PR #1, opened 2015-04-13
Machinery for building the documentation

drocco007:


takluyver:
The test failures are not related to this - it's trying to import the notebook, and not finding it. Adding a dependency on the notebook seems like the wrong thing to do given that the notebook already depends on nbconvert. I'm not sure why it wasn't failing before.

jupyter_nbconvert/_version.py
```diff
@@ -1,2 +1,41 @@
-version_info = (4, 0, 0, 'dev')
-__version__ = '.'.join(map(str, version_info))
```
takluyver: I think I'd rather keep this file small and simple, and just add `_version_extra` to check for the 'dev' tag.

drocco007:
@takluyver Done! :)

takluyver:
Thanks!

jupyter_nbconvert/_version.py
minrk: I would have preferred to leave it how it was, actually.

jupyter_nbconvert/_version.py
```diff
@@ -1,2 +1,3 @@
-version_info = (4, 0, 0, 'dev')
+_version_extra = 'dev'
+version_info = (4, 0, 0, _version_extra)
```
minrk: This will do the wrong thing at release time when _version_extra is empty. My plan for this was to have a release replace this with `version_info = (4, 0, 0)`, having no extra at all.

If version_extra is to be treated as a variable, it needs to be conditionally added to version_info only if it's non-empty.

If this is prompted by the _version_extra logic in conf.py, I would prefer to remove that logic, rather than change this here.

docs/source/conf.py
```diff
@@ -0,0 +1,292 @@
...
+copyright = u'The Project Jupyter Development Team'
+
+# The version info for the project you're documenting, acts as replacement for
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+#
+# The full version, including alpha/beta/rc tags.
+release = "%s" % iprelease['__version__']
+# Just the X.Y.Z part, no '-dev'
+version = iprelease['__version__'].split('-', 1)[0]
```
minrk: '-' will never be in `__version__`

docs/source/conf.py
```diff
@@ -0,0 +1,292 @@
...
+# serve to show the default.
+
+import sys
+import os
+
+
+# If extensions (or modules to document with autodoc) are in another directory,
+# add these directories to sys.path here. If the directory is relative to the
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+sys.path.insert(0, os.path.abspath('../sphinxext'))
```
minrk: there is no sphinxext here.

docs/source/index.rst
```diff
@@ -3,24 +3,24 @@
...
-Newly added in the 1.0 release of IPython is the ``nbconvert`` tool, which 
-allows you to convert an ``.ipynb`` notebook document file into various static 
-formats. 
+Newly added in the 1.0 release of IPython is the ``nbconvert`` tool, which
+allows you to convert an ``.ipynb`` notebook document file into various static
+formats.
 
-Currently, ``nbconvert`` is provided as a command line tool, run as a script 
-using IPython. A direct export capability from within the 
-IPython Notebook web app is planned.
```
minrk: I don't care here, but for future reference, it's a good idea to disable stripping whitespace on save in editors. It's a horrible misfeature.

minrk:
It looks like the sphinx conf.py was copied from IPython rather than started fresh, and I think perhaps we should avoid that. There are a lot of things we've accumulated in IPython over the years that we don't necessarily want to inherit in the new projects.

takluyver:
Sorry, I was a bit too eager with the merge button.

minrk:
No problem, it's good to get these off the ground. I just want to make sure everything we copy over from IPython we actually want to inherit. Lots of the docs stuff is *super* old.

takluyver:
The people copying over the docs stuff did go through conf.py and remove several bits that we don't want to inherit - e.g. there shouldn't be a need to special-case Readthedocs. I guess there were always going to be some bits we missed, though.

PR #2, opened 2015-04-14
init sphinx conf.py with sphinx-quickstart

minrk:
removes a few things inherited from IPython

@takluyver can you identify anything that this removes that you think we should keep?


docs/source/conf.py
```diff
@@ -88,15 +71,15 @@
...
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
-exclude_patterns = ['whatsnew/pr']
+exclude_patterns = []
 
 # The reST default role (used for this markup: `text`) to use for all
 # documents.
-default_role = 'literal'
+#default_role = None
```
takluyver: You probably want to leave this one as 'literal' - that's what lets us use single backticks for inline monospace text.

PR #5, opened 2015-04-19
Remove pdf from display order for HTML export

takluyver:
The HTML template has no means to display pdf, so this was hiding outputs with pdf representations.

The live notebook also appears to prefer png to svg, whereas the HTML export prefers svg. Should this be harmonised?

Closes jupyter/jupyter_notebook#31

### DIFF
--- a/jupyter_nbconvert/exporters/html.py
+++ b/jupyter_nbconvert/exporters/html.py
@@ -37,7 +37,15 @@ class HTMLExporter(TemplateExporter):
     def default_config(self):
         c = Config({
             'NbConvertBase': {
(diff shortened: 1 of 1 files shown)


Carreau:
+1, 

We should  fix travis that try to get things from jupyter_notebook.

JanSchulz:
Just a question: what version of the notebook (if there is any "all together" releases? Or the 3.x stream?) and what version of nbviewer will get this fix? The R kernel currently defaults to pdf+png+svg output, so if we keep it that way we will probably get more than a few bugreports of nbviewer is kept at the current version...

takluyver:
This should be in the notebook version 4.0. We could backport it to 3.x, but I think we're aiming to have 4.0 out quite soon - it will certainly be quicker than our usual development cycle.

PR #6, opened 2015-04-20
add jupyter_notebook to list of deps to get CI working

Carreau:
try to fix travis...


takluyver:
This fixes that problem, but there's still a couple of other errors in the tests. I imagine that will be a separate fix, so I'll merge this and we can work out the other issues tomorrow.

PR #14, opened 2015-04-30
reorder preprocessors

minrk:
- ensure execute runs first
- ensure extract output runs last

Any operations on outputs are incorrect if done before execute. Similarly, any modifications should be done before extraction.

Thanks @olgabot for revealing the problem.

This should be backported for IPython 3.x.

### DIFF
--- a/jupyter_nbconvert/exporters/exporter.py
+++ b/jupyter_nbconvert/exporters/exporter.py
@@ -65,15 +65,17 @@ class Exporter(LoggingConfigurable):
 
     _preprocessors = List()
 
-    default_preprocessors = List(['jupyter_nbconvert.preprocessors.coalesce_streams',
+    default_preprocessors = List([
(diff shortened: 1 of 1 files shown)


jupyter_nbconvert/exporters/exporter.py
```diff
@@ -227,6 +229,7 @@ def _init_preprocessors(self):
         # Load user-specified preprocessors.  Enable by default.
         for preprocessor in self.preprocessors:
             self.register_preprocessor(preprocessor, enabled=True)
+        print([ p for p in self._preprocessors ])
```
takluyver: Leftover debugging code.

minrk: fixed

takluyver:
Merged - I'll have a go at the backport.

PR #18, opened 2015-05-07
s/jupyter_nbconvert/nbconvert/

minrk:
The 4.0 nbconvert package is just called nbconvert.


damianavila:
Changes makes sense, but it's failing on TravisCI, maybe because some issue in the circular dependency??

PR #20, opened 2015-05-14
get notebook css from cdn.jupyter.org

minrk:
removes dependency on the notebook package

This will only be 4.0.0-dev while notebook is unreleased, after which point it will be pinned to a stable release, even for dev versions of nbconvert.

### DIFF
--- a/nbconvert/preprocessors/csshtmlheader.py
+++ b/nbconvert/preprocessors/csshtmlheader.py
@@ -50,10 +50,12 @@ class CSSHTMLHeaderPreprocessor(Preprocessor):
         from pygments.formatters import HtmlFormatter
         header = []
         
-        # Construct path to IPy CSS
-        from jupyter_notebook import DEFAULT_STATIC_FILES_PATH
-        sheet_filename = os.path.join(DEFAULT_STATIC_FILES_PATH,
-            'style', 'style.min.css')
+        # Construct path to Jupyter CSS
+        import nbconvert.resources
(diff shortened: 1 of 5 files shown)


rgbkrk:
A whole CDN for one file.

Excited to see a green light here.

rgbkrk:
Welp.

``` 
    Failed to download css from https://cdn.jupyter.org/notebook/4.0.0-dev/style/style.min.css: addinfourl instance has no attribute '__exit__'
    
    error: Need Notebook CSS to proceed: nbconvert/resources/style.min.css
```

minrk:
Oops. That's me forgetting about Python 2. urlopen isn't a context manager on py2.

damianavila:
There is still an error but I am :+1: on the idea...

minrk:
After a bit more fiddling, Builds passing now.

nbconvert/preprocessors/csshtmlheader.py
```diff
@@ -67,10 +69,15 @@ def _generate_header(self, resources):
         # Load the user's custom CSS and IPython's default custom CSS.  If they
         # differ, assume the user has made modifications to his/her custom CSS
         # and that we should inline it in the nbconvert output.
+        try:
+            from notebook import DEFAULT_STATIC_FILES_PATH
+        except ImportError:
+            DEFAULT_STATIC_FILES_PATH = None
```
rgbkrk: Why keep the fallback if notebook is no longer a dependency?

PR #28, opened 2015-06-01
add tightlist command

minrk:
provided by pandoc 1.14

backported to 3.x as ipython/ipython#8499


Carreau:
Kicked the tests, though the 3 were all raising the same OS Error... weird.

Carreau:
Faillure seem to be real... I don't see why.

minrk:
Poked travis after merging jupyter/jupyter_core#18. All green.

PR #30, opened 2015-06-04
install ipython kernelspec with -m

minrk:


takluyver:
The test failure on this is something unrelated that @Carreau and I tracked down the other day - jupyter/jupyter_core#18 should resolve it.

minrk:
Poked travis after merging jupyter/jupyter_core#18. All green.

PR #31, opened 2015-06-09
Initial docs updates

jhamrick:
This is just an initial PR to start updating the documentation. This:

1. Renames "ipython" to "jupyter"
2. Moves the example notebooks into the docs folder and converts them to rst (though this is not currently automated in anyway -- I just ran `jupyter nbconvert --to rst`)
3. Splits apart `index.rst` into a few separate files and then makes `index.rst` mostly just be a TOC

Things that are not addressed here:

1. Fleshing out the documentation or making it up-to-date
2. API docs

I figured I would make an initial PR to make sure I am heading in the right direction with this, and then address the second two things in separate PRs.

One question I had about this -- there is the `ipython/nbconvert-examples` repository. Do we want to move the examples in there to the nbconvert docs in this repository?


docs/source/customizing.rst
```diff
@@ -0,0 +1,90 @@
...
+Customizing nbconvert
+=====================
+
+Look at the first 20 lines of the ``python`` exporter
+
+.. code:: python
+
+    pyfile = !jupyter nbconvert --to python 'Index.ipynb' --stdout
+    for l in pyfile[20:40]:
+        print l
```
takluyver: Things like this will really want the output in the converted result as well as the code, I guess. Might need a custom template!

docs/source/index.rst
```diff
@@ -1,168 +1,21 @@
...
-Converting notebooks to other formats
+=====================================
+Converting Notebooks to Other Formats
 =====================================
 
-Newly added in the 1.0 release of IPython is the ``nbconvert`` tool, which
-allows you to convert an ``.ipynb`` notebook document file into various static
-formats.
+Originally added to IPython in version 1.0 (and moved to Jupyter in the 4.0
+release), the ``nbconvert`` tool allows you to convert an ``.ipynb`` notebook
```
takluyver: We can probably drop the description of its history now.

docs/source/customizing.rst
jhamrick: :-P Yes, definitely. Probably whatever build system we have for the notebook docs will need to run the notebooks first, so that the output gets included when converting to rst.

docs/source/nbconvert_library.ipynb
```diff
@@ -86,7 +86,7 @@
    },
    "outputs": [],
    "source": [
-    "from IPython import nbformat\n",
+    "from jupyter import nbformat\n",
```
takluyver: It's just `import nbformat` now ;-)

docs/source/nbconvert_library.ipynb
```diff
@@ -323,7 +323,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "You can use `Preprocessors` to accomplish the task at hand.  IPython has preprocessors built in which you can use.  One of them, the `ExtractOutputPreprocessor` is responsible for crawling the notebook, finding all of the figures, and putting them into the resources directory, as well as choosing the key (i.e. `filename_xx_y.extension`) that can replace the figure inside the template.\n",
+    "You can use `Preprocessors` to accomplish the task at hand.  Jupyter has preprocessors built in which you can use.  One of them, the `ExtractOutputPreprocessor` is responsible for crawling the notebook, finding all of the figures, and putting them into the resources directory, as well as choosing the key (i.e. `filename_xx_y.extension`) that can replace the figure inside the template.\n",
```
takluyver: probably 'Nbconvert' here instead of 'Jupyter' 

(how are we capitalising nbconvert now that it's a project name? No capitals? Nbconvert? NBconvert? NbConvert?)

takluyver:
I think this looks like a good start. We can steal the machinery to document config options from e.g. [jupyter_console](https://github.com/jupyter/jupyter_console/tree/master/docs), where I've just set it up, but that can be in a later PR.

docs/source/nbconvert_library.ipynb
jhamrick: I would say either no caps or `Nbconvert`, but neither is really perfect. For nbgrader, I typically use no caps and avoid putting it in places where it needs to be capitalized, but you can't *always* avoid that :-/

docs/source/nbconvert_library.ipynb
jhamrick: Ah, ok. Meh, I can't actually easily test these examples at the moment because I have my global install of IPython still at 3.x and Jupyter 4.0 installed in a virtualenv. Let me see if I can set it up so I can do some better testing here...

jhamrick:
Ok, updated this to make the documentation be more up-to-date with the current version of nbconvert. I think it can still use more fleshing out and organizing a bit, but that can come later. I'll work next on getting the API docs working and building this on RTD.

takluyver:
Awesome, merging.

docs/source/nbconvert_library.rst
```diff
@@ -0,0 +1,364 @@
...
+
+Real World Uses
+---------------
+
+@jakevdp uses Pelican and Jupyter Notebook to blog. Pelican `will
+use <https://github.com/getpelican/pelican-plugins/pull/21>`__ nbconvert
+programatically to generate blog post. Have a look a `Pythonic
+Preambulations <http://jakevdp.github.io/>`__ for Jake's blog post.
+
+@damianavila wrote the Nicholas Plugin to `write blog post as
```
damianavila: Nikola

jhamrick:
Thanks!

docs/source/nbconvert_library.rst
jhamrick: Whoops, sorry Damien! I'll fix it in the next PR...

damianavila: No problem... I just see it and I did not want to forget about the typo...

PR #32, opened 2015-06-10
Add API docs

jhamrick:
This adds some very barebones API docs, with some minor changes to the actual source here and there to make the docs build nicely.


docs/source/conf.py
```diff
@@ -48,24 +57,25 @@
...
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
 # built documents.
 #
-_version_py = '../../nbconvert/_version.py'
-version_ns = {}
-exec(compile(open(_version_py).read(), _version_py, 'exec'), version_ns)
 # The short X.Y version.
-version = '%i.%i' % version_ns['version_info'][:2]
+version = '4.0'
```
jhamrick: Hmm, probably this should use the code that was there before so the version gets automatically updated.

docs/source/api/filters.rst
```diff
@@ -0,0 +1,54 @@
+Filters
+=======
```
takluyver: This should probably have a little intro paragraph explaining that these are for use in templates.

docs/source/conf.py
```diff
@@ -101,12 +114,15 @@
...
 
+# If true, `todo` and `todoList` produce output, else they produce nothing.
+todo_include_todos = False
+
 
 # -- Options for HTML output ----------------------------------------------
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-html_theme = 'default'
```
takluyver: I generally comment this out when we're using readthedocs, then it will use the RTD theme.

takluyver:
Thanks!

jhamrick:
Ok, updated as per your comments.

docs/autogen_config.py
```diff
@@ -0,0 +1,22 @@
...
+=====================
+
+These options can be set in ``~/.jupyter/nbconvert_config.py``, or
+at the command line when you start it.
+"""
+
+try:
+    indir = os.path.dirname(__file__)
+except NameError:
+    indir = os.getcwd()
```
takluyver: I think on RTD, the cwd will be where conf.py is, so this should be `os.path.dirname(os.getcwd())`. This is definitely something of a hack, but it's very useful to have docs on Readthedocs.

jhamrick:
@takluyver Fixed!

docs/source/api/exporters.rst
```diff
@@ -0,0 +1,69 @@
...
+    .. automethod:: register_preprocessor
+
+Specialized exporter classes
+----------------------------
+
+.. autoclass:: HTMLExporter
+
+.. autoclass:: SlidesExporter
+
+.. autoclass:: TemplateExporter
```
takluyver: I think this should move up to just under Exporter - TemplateExporter is the main API point that people should inherit from to define custom exporters; it's not specialised in the same sense as the others in this section.

takluyver:
Thanks Jess, and sorry that I keep thinking of things one at a time rather than in one big group!

docs/source/api/exporters.rst
jhamrick: Good point, fixed

jhamrick:
No worries!

takluyver:
OK, merging this before I come up with another thing. Let's see what it looks like on RTD :-)

PR #33, opened 2015-06-11
Add command line switch --abort-on-error to abort conversion if execution fails

maxalbert:
This PR adds a command line switch `--abort-on-error`. When used in conjunction with `--execute` this will make `nbconvert` abort in case any of the cells raises an exception during execution.

I have added a custom exception which captures the output of the notebook execution until the failure because we have a use case in our group where we need to be able to inspect this. The downside is that it makes the output somewhat noisy when `nbconvert` is used from the command line. Let me know your thoughts and whether you have any suggestions how this can be handled differently in case you think it is a problem. Thanks!

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -89,11 +89,10 @@ The currently supported export formats are:
       jupyter nbconvert --to notebook --execute mynotebook.ipynb
 
   will open the notebook, execute it, capture new output, and save the result in
-  :file:`mynotebook.nbconvert.ipynb`. Any exceptions occurring during execution
-  of a cell will not result in an error, but the output from the exception will
-  simply be included in the cell output. If you specify ``--abort-on-error``
-  in addition to the ``--execute`` flag then ``nbconvert`` will abort if an
-  exception occurs during execution.
+  :file:`mynotebook.nbconvert.ipynb`. By default, ``nbconvert`` will abort conversion
+  if any exceptions occur during execution of a cell. If you specify ``--allow-errors``
+  (in addition to the ``--execute`` flag) then conversion will continue and the output
+  from any exception will be included in the cell output.
 
   ::
 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -89,7 +89,11 @@ The currently supported export formats are:
       jupyter nbconvert --to notebook --execute mynotebook.ipynb
 
   will open the notebook, execute it, capture new output, and save the result in
-  :file:`mynotebook.nbconvert.ipynb`.
+  :file:`mynotebook.nbconvert.ipynb`. Any exceptions occurring during execution
+  of a cell will not result in an error, but the output from the exception will
+  simply be included in the cell output. If you specify ``--abort-on-error``
+  in addition to the ``--execute`` flag then ``nbconvert`` will abort if an
+  exception occurs during execution.
 
   ::
 
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -50,7 +50,7 @@ class ExecutePreprocessor(Preprocessor):
         )
     )
 
-    abort_on_error = Bool(
+    allow_errors = Bool(
         False, config=True,
         help=dedent(
             """
@@ -170,7 +170,7 @@ class ExecutePreprocessor(Preprocessor):
             else:
                 outs.append(out)
 
-            if out.output_type == 'error' and self.abort_on_error:
+            if out.output_type == 'error' and not self.allow_errors:
                 raise CellExecutionError(outs)
 
         return outs
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -26,8 +26,8 @@ class CellExecutionError(ConversionException):
     using nbconvert as a library, since it allows to deal with
     failures gracefully.
     """
-    def __init__(self, outputs):
-        self.outputs = outputs
+    def __init__(self, traceback):
+        self.traceback = traceback
 
 
 class ExecutePreprocessor(Preprocessor):
@@ -124,6 +124,9 @@ class ExecutePreprocessor(Preprocessor):
                 else:
                     raise
 
+            if msg['msg_type'] == 'execute_reply' and msg['metadata']['status'] == 'error' and not self.allow_errors:
+                raise CellExecutionError(msg['content']['traceback'])
+
             if msg['parent_header'].get('msg_id') == msg_id:
                 break
             else:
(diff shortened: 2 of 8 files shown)


minrk:
I think this is the right way to go, but here's one question: Should the default behavior match the live notebook (halt on error) or the initial release of `nbconvert --execute`, which captures errors and moves on?

cc @jhamrick and @ellisonbg

nbconvert/preprocessors/execute.py
```diff
@@ -146,4 +170,7 @@ def run_cell(self, cell):
             else:
                 outs.append(out)
 
+            if out.output_type == 'error' and self.abort_on_error:
```
takluyver: Should this be listening for 'error' output messages, or should it be checking the for an error status in `execute_reply` messages? The protocol doesn't force the two to be linked. I'd say that the execute_reply status is what really determines whether execution should proceed or not.

jhamrick:
Hmmm, good question. I'd say from a consistency standpoint with the rest of the notebook ecosystem it should be the default to throw the error, and there should be a flag to ignore errors -- but this would definitely break existing usages of `nbconvert` where it is assumed the error will be ignored. It's definitely something I can easily change in nbgrader, but there might be others who are relying on that default.

takluyver:
I think I'd lean towards errors-are-fatal by default, but that does make it a bit awkward if you need to support different versions of nbconvert.

If we switch the default, the flag could be `--on-error-continue-next` ;-)

Carreau:
The problem with error are fatal is batch convert.

minrk:
@Carreau I don't think that's true. If I were batch processing my notebooks and one failed, I would want it to die unless I explicitly told it that exceptions were okay.

Carreau:
Process can return 1. when you unit-test a bunch of test, you don't want 1 failling test to prevent others from running.

minrk:
That's true, but nbconvert's not a unittest runner.

takluyver:
I suspect we might end up with additional options for things like running all notebooks even if one fails, but I think we can get a long way with just a boolean for abort vs. carry on.

maxalbert:
Thanks for the feedback so far! @Carreau: The unit-testing application is indeed one of the use cases I had in mind when adding the custom exception. In our group we wrote a plugin for `py.test` to validate notebooks: https://github.com/computationalmodelling/pytest_validate_nb This is largely based on [this gist](https://gist.github.com/minrk/2620735) by @minrk from a while ago, but now that I've familiarised myself better with the `nbconvert` internals it feels like it could be almost trivially extended from `nbconvert`'s `ExecutePreprocessor`. I agree with @minrk though that testing should probably best be handled by a separate tool and not nbconvert itself (although they should of course share the underlying execution machinery).

maxalbert:
So, is the consensus to make "abort-on-error" the default behaviour? I'm happy to change the PR if that's the case. What should the name of the flag be? If we go with something along the line of @takluyver's suggestion, would it be ok to change it to `--on-error-continue-next-cell` to make it clearer what "next" means in this context? Any other suggestions, in case this is too verbose?

takluyver:
That wasn't a serious suggesting, I was making a joke about Basic. Although now that I go and look it up it's actually 'resume' instead of continue.

I'd probably just call it something like `--ignore-errors` or `--allow-errors`.

minrk:
I think either `--ignore-errors` or `--allow-errors` is fine. Maybe a slight preference for `allow`, since they aren't exactly ignored.

damianavila:
>Maybe a slight preference for allow, since they aren't exactly ignored.

+1

maxalbert:
Apologies for the somewhat sluggish response on this, it's been a slighly crazy week. The flag is now called `--allow-errors` (which I like for the same reason that Min mentioned) and I have changed the default behaviour so that conversion is aborted if a cell raises an error (I had to supply the `allow_errors` flag to a few test notebooks which raise exceptions). Also, the code now checks for messages of type "error" -- I hope this is what @takluyver meant with his comment. (I tried to check for "execute_reply" messages but there were actually none of that type at all; and if I interpret the messaging spec correctly then the "error" message type should be the relevant one.)

docs/source/usage.rst
```diff
@@ -89,7 +89,11 @@ The currently supported export formats are:
       jupyter nbconvert --to notebook --execute mynotebook.ipynb
 
   will open the notebook, execute it, capture new output, and save the result in
-  :file:`mynotebook.nbconvert.ipynb`.
+  :file:`mynotebook.nbconvert.ipynb`. Any exceptions occurring during execution
+  of a cell will not result in an error, but the output from the exception will
+  simply be included in the cell output. If you specify ``--abort-on-error``
```
takluyver: This doc needs updating now

docs/source/usage.rst
```diff
@@ -89,7 +89,11 @@ The currently supported export formats are:
       jupyter nbconvert --to notebook --execute mynotebook.ipynb
 
   will open the notebook, execute it, capture new output, and save the result in
-  :file:`mynotebook.nbconvert.ipynb`.
+  :file:`mynotebook.nbconvert.ipynb`. Any exceptions occurring during execution
```
minrk: This description needs to be updated to reflect `--allow-errors`

takluyver:
There is an execute_reply message, but it's separate from the output message - it's being fetched in the block immediately above the iopub messages. I think it's probably OK to do it this way for now - we haven't really defined any expected case where an `error` message can be sent without the execute_reply status being error, or vice versa. If we do, we can always change this later.

maxalbert:
Blimey, I must have been tired last night... Thanks for picking up on the missing doc update. The code now also checks the status of `execute_reply` messages to determine whether an error occurred (as suggested by @takluyver). Let me know if there is anything else that needs changing or fixing.

nbconvert/preprocessors/execute.py
```diff
@@ -100,6 +124,9 @@ def run_cell(self, cell):
                 else:
                     raise
 
+            if msg['msg_type'] == 'execute_reply' and msg['metadata']['status'] == 'error' and not self.allow_errors:
+                raise CellExecutionError(msg['content']['traceback'])
```
takluyver: You can put this inside the `if msg['parent_header'].get('msg_id') == msg_id:` check just below, to ensure that we're looking at the message which is a response to this execute request (it shouldn't be anything else, but it's worth checking). And then I think we can lose the msg_type check, because the only message on the shell channel in response to an `execute_request` should be `execute_reply`.

maxalbert: Thanks for these suggestions! That makes sense. (Implemented in b17eac4)

takluyver:
Thanks, I think this is looking good.

PR #36, opened 2015-06-14
two passes through cells

bollwyvl:
This loops through the cells twice to ensure the appropriate metadata exists for all cells before trying to use them.


bollwyvl:
https://github.com/jupyter/nbviewer/issues/466

damianavila:
I miss this when I implemented #8415, thanks for the catch, this PR LGTM... please somebody press the green button...
Also, this should be backported to be cached by nbviewer

bollwyvl:
Should I propose the backport, or is someone already on it?

takluyver:
Go for it, thanks!

Issue #41, opened 2015-06-17
Totally broken by recent change to traitlets


takluyver:
The default value for `Type` when no default specified changed from `None` to `object`, to make validation more consistent. However, nbconvert relies on the default for `NbconvertApp.postprocessor_factory` being None (or Falsy, at least).

This is simple to fix, I'm just contemplating the best way to do it.

minrk:
Not sure, but `Type(None, allow_none=True)`?

Carreau:
@parleur was looking for some easy to get started issues.

PR #42, opened 2015-06-17
Use explicit None default for postprocessor_factory trait

takluyver:
Closes gh-41

Depends on ipython/traitlets#39


takluyver:
Failure is due to a similar problem in jupyter_client, PR coming up

takluyver:
OK, tests are now passing again :-)

PR #44, opened 2015-06-17
Simplify TemplateExporter

takluyver:
I was finding the code in TemplateExporter hard to follow because it had quite a bit of state, particularly around loading templates.

Because the class stores all the information needed to load the template at any time, storing the loaded template is really a cache. These changes make this explicit.

This has a few functional consequences:

1. Previously you could load a template separately and pass it in to TemplateExporter, at least in principle (I'm not sure if it would actually have worked). This is not possible at present, because the template attribute is considered a cache. However, since most of the logic in TemplateExporter is to set up the environment and load the template, code that wants to load a template itself may be better off not using TemplateExporter.
2. If there's an error in a template (e.g. invalid syntax), with this PR it will escape to the caller. I was confused while trying to make custom templates for Fernando, because it was logging the template error, then continuing and raising a "template file ... could not be found" error instead, which is not really accurate.
3. The template syntax is no longer configurable, instead the Latex overrides are hardcoded. As we're encouraging people to inherit from our templates, it doesn't really make sense to reconfigure the template syntax. If people really need custom syntax, they can still make that work by subclassing `TemplateExporter` in their own code and overriding `_create_environment()`, like `LatexExporter` does here.

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -2,17 +2,8 @@
 that uses Jinja2 to export notebook files into different formats.
 """
 
-#-----------------------------------------------------------------------------
-# Copyright (c) 2013, the IPython Development Team.
-#
+# Copyright (c) IPython Development Team.
 # Distributed under the terms of the Modified BSD License.
-#
-# The full license is in the file COPYING.txt, distributed with this software.
-#-----------------------------------------------------------------------------
-
-#-----------------------------------------------------------------------------
-# Imports
-#-----------------------------------------------------------------------------
 
 from __future__ import print_function, absolute_import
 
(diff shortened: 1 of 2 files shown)


minrk:
I think this makes a lot of sense. I don't think configurable template syntax is something that would ever make sense, so making it part of the class is sensible.

nbconvert/exporters/templateexporter.py
```diff
@@ -150,27 +148,19 @@ def __init__(self, config=None, extra_loaders=None, **kw):
...
         super(TemplateExporter, self).__init__(config=config, **kw)
 
-        #Init
-        self._init_template()
-        self._init_environment(extra_loaders=extra_loaders)
-        self._init_filters()
+        self.on_trait_change(self._invalidate_environment_cache,
+                     list(self.traits(affects_environment=True)))
+        self.on_trait_change(self._invalidate_template_cache,
+                     list(self.traits(affects_template=True)))
```
minrk: I like this use of trait metadata. Clever, but easy to follow.

takluyver: Thank you, I was pleased with the idea.

PR #46, opened 2015-06-18
Document template structure

takluyver:
The documentat is hand-written HTML, because HTML is actually just what I wanted to illustrate the nested blocks that comprise a template. I wrote it while working on custom templates for Fernando's tutorial.


Carreau:
+1. we can probably include it in the middle of a sphing ..raw:html

takluyver:
It's a complete page, with styles defined in the header, so it's probably not trivial to embed it into another page, unfortunately.

minrk:
Would there be a benefit to putting it in a `.. html::` block?

takluyver:
Well, it would be nice for it to be a more integrated part of the docs, so it showed up on a page with the regular docs sidebar and so on. But I don't see a neat way of achieving that.

minrk:
Doesn't the html directive provide that? The CSS can be dropped into the site css, or maybe in a style tag in an html directive on the page.

takluyver:
You can only define CSS rules inside `<head>`, AFAIK (maybe I'm wrong about this?). So if it's contained within the page, all the styles have to be inlined.

It could go in site CSS, but it's referenced from a doc written in the notebook, so it would then look all wrong in the notebook (or not show up at all, depending on whether we need to do a raw rst cell with the `.. html::` directive, or whether pandoc will do the right thing with a bunch of HTML-in-Markdown.

minrk:
You can have style tags anywhere on the page. It's used in the notebook pretty often.

damianavila:
:+1:

takluyver:
OK, with that knowledge, I've got it embedded in the page in a way that shows up in both notebook and rst output. We have to be careful to save the notebook with that code cell executed, though.

PR #48, opened 2015-06-23
Add html_head block

### LINKED ISSUE #39: Blocks to add to HTML header
The HTML full.tpl template has a large (~50 line) header block. Templates inheriting from this will likely want to add to the HTML headers, but they can't use `super()` to add to it, because it closes the `<head></head>` tag inside the block, so additions would fall outside that.

There should be one or more Jinja blocks inside the `<head></head>` tag which other templates can override.

takluyver:
By overriding this, you can add things in the `<head>` tag without having to maintain a copy of the default content that goes there.

Closes #39

I've also documented the extra blocks that are overridable in the HTML and Latex templates. There are quite a few for Latex - do we want to publicly document all of these, or leave some out so we have more flexibility to rearrange them later?

### DIFF
--- a/docs/source/customizing.ipynb
+++ b/docs/source/customizing.ipynb
@@ -95,7 +95,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 4,
+   "execution_count": 1,
    "metadata": {
     "collapsed": false
    },
@@ -118,6 +118,11 @@
        "    padding: 5px;\n",
        "    border: 1px solid hsla(120, 60%, 70%, 0.5);\n",
        "    border-left: 2px solid black;\n",
+       "}\n",
+       "\n",
+       ".template_block pre {\n",
+       "    background: transparent;\n",
+       "    padding: 0;\n",
        "}\n",
        "\n",
        ".big_vertical_ellipsis {\n",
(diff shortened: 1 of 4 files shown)


docs/source/customizing.ipynb
```diff
@@ -279,7 +316,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.4.3+"
+   "version": "3.4.3"
```
takluyver: The `+` shouldn't really have been here. Possibly we should be more careful about how we get the Python version number from `sys`.

nbconvert/templates/html/full.tpl
```diff
@@ -52,7 +52,7 @@ div#notebook {
 
 <!-- Loading mathjax macro -->
 {{ mathjax() }}
-
+{%- endblock html_head %-}
```
minrk: end with `-%`, not `%-`. Travis is reporting syntax errors.

takluyver: Gah, every time I make a template, I make this mistake. Fixed.

damianavila:
LGTM

PR #49, opened 2015-06-23
ENH: input cells now rendered in fenced code blocks

spencerlyon2:


minrk:
Nice. :+1:, assuming tests pass.

jdfreder:
This looks great, glad that was the right key. : )

spencerlyon2:
Thanks for the help!

nbconvert/templates/markdown.tpl
```diff
@@ -8,7 +8,9 @@
 {%- endblock output_prompt %}
 
 {% block input %}
-{{ cell.source | indent(4)}}
+```{% if nb.metadata.kernelspec %}{{ nb.metadata.kernelspec.language }}{% endif %}
```
takluyver: I think we should be using `metadata.language_info.name` instead of `metadata.kernelspec.language`. The latter is really just an ID to find a sensible substitute kernel when we can't find one matching `kernelspec.name`.

damianavila:
Nice, I like it...

spencerlyon2:
@takluyver  thanks for the tip -- changed.

jdfreder:
Thanks @spencerlyon2 !

PR #53, opened 2015-06-26
Make sure outputs are cleared immediately before execution rather than immediately after

nitind:
The current ordering or preprocessors clears the output fields after executing cells rather than before, so --to notebook no longer leaves results. This affects 3.2 as well.

### DIFF
--- a/nbconvert/exporters/exporter.py
+++ b/nbconvert/exporters/exporter.py
@@ -66,8 +66,8 @@ class Exporter(LoggingConfigurable):
     _preprocessors = List()
 
     default_preprocessors = List([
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks, this looks good to me.

In the longer run, we may need to think more carefully about how preprocessors are ordered: at present, no third-party preprocessor can run before one of the default preprocessors. But we can work that out another day.

nitind:
Thanks for the quick response! Should I open a separate one for 3.x?

rgbkrk:
Oh! Maybe that's why I was having weird issues with https://github.com/sigmavirus24/github3.py/pull/394

minrk:
@nitind yes, please do open a copy against 3.x

PR #57, opened 2015-07-11
prepare for mistune changes

minrk:
- blacklist mistune 0.6, which has some buggy behavior escaping *some* raw html.
- mistune 0.7 will start escaping html by default, so prepare for this with explicit `escape=False`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -171,7 +171,7 @@ if 'develop' in sys.argv or any(a.startswith('bdist') for a in sys.argv):
 
 setuptools_args = {}
 install_requires = setuptools_args['install_requires'] = [
-    'mistune',
+    'mistune!=0.6',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -171,7 +171,7 @@ def run(self):
 
 setuptools_args = {}
 install_requires = setuptools_args['install_requires'] = [
-    'mistune',
+    'mistune!=0.6',
```
damianavila: should not be `>0.6`? Or passing `escape=False` to previous version will work too?

takluyver: Just checked with 0.5.1: passing `escape=False` should work, I think just the default changed from False to True.

damianavila: Then LGTM...

PR #58, opened 2015-07-12
Fix missing slideshow key.

damianavila:
This should do the trick fixing https://github.com/jupyter/nbviewer/issues/466

### DIFF
--- a/nbconvert/preprocessors/tests/test_revealhelp.py
+++ b/nbconvert/preprocessors/tests/test_revealhelp.py
@@ -20,12 +20,15 @@ class Testrevealhelp(PreprocessorTestsBase):
 
         slide_metadata = {'slideshow' : {'slide_type': 'slide'}}
         subslide_metadata = {'slideshow' : {'slide_type': 'subslide'}}
+        fragment_metadata = {'slideshow' : {'slide_type': 'fragment'}}  
 
         cells=[nbformat.new_code_cell(source="", execution_count=1, outputs=outputs),
                nbformat.new_markdown_cell(source="", metadata=slide_metadata),
                nbformat.new_code_cell(source="", execution_count=2, outputs=outputs),
                nbformat.new_markdown_cell(source="", metadata=slide_metadata),
-               nbformat.new_markdown_cell(source="", metadata=subslide_metadata)]
+               nbformat.new_markdown_cell(source="", metadata=subslide_metadata),
+               nbformat.new_markdown_cell(source="", metadata=fragment_metadata),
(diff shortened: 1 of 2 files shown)


damianavila:
pinging @bollwyvl who was interested in a solution...

Carreau:
+1

bollwyvl:
+1 if it fixes... Test coverage?

nbconvert/preprocessors/tests/test_revealhelp.py
```diff
@@ -20,12 +20,15 @@ def build_notebook(self):
 
         slide_metadata = {'slideshow' : {'slide_type': 'slide'}}
         subslide_metadata = {'slideshow' : {'slide_type': 'subslide'}}
+        fragment_metadata = {'slideshow' : {'slide_type': 'fragment'}}
```
minrk: Why is the metadata here `md.slideshow.slide_type` but `md.slide_type` above?

damianavila:
Done :wink:

minrk:
Merged since it fixes the issue, but I'd still like a note about why the metadata seems to get promoted one level.

bollwyvl:
Yeah, that looks very suspicious. seems that all the checks in the code, templates, and tests use `metadata.*`, not `metadata.slideshow.*`. I didn't look too closely when I added the second loop through, but it's pretty glaring now that you mention it. is this just to save some keystrokes?

Additionally, using `enumerate` in this way may have some unintended consequences: when `i` is 0, `slide_helper` will be set on `cells[-1]` for certain values. Perhaps a stack-based approach would be more robust? or having a `prev_cell` and `next_cell` each iteration that could be checked.

takluyver:
I did some cleanup of similar code in RISE, so I might have a go at cleaning up this logic too at some point.

damianavila:
>Yeah, that looks very suspicious. seems that all the checks in the code, templates, and tests use metadata.*, not metadata.slideshow.*. I didn't look too closely when I added the second loop through, but it's pretty glaring now that you mention it. is this just to save some keystrokes?

Probably saving keystrokes... but probably we need to keep consistency and respect the levels

>Additionally, using enumerate in this way may have some unintended consequences: when i is 0, slide_helper will be set on cells[-1] for certain values. Perhaps a stack-based approach would be more robust? or having a prev_cell and next_cell each iteration that could be checked.

It does not have consequences if you set up the first cell as slide, becase the last cell should have the slide_end helper, but this is not the case for subslides... and the reality is that the logic is a little bit convoluted. We should simplify this... I agree.

PR #59, opened 2015-07-12
remove released dependencies from requirements.txt

minrk:


takluyver:
nbformat needs to be added as a dependency in setup.py

Issue #34, opened 2015-06-12
slides display problem: reveal.js version conflict?


cranmer:
I'm using IPython 3.0.0 and having various issues with slide conversion.
(I had to manually install mistune (pip installed  mistune.__version__ '0.5.1')

Scenario 0:
`ipython nbconvert --to slides TestSlides.ipynb --post serve`
This works, but you see it is getting an old reveal 2.6.2
`[NbConvertApp] Redirecting reveal.js requests to https://cdn.jsdelivr.net/reveal.js/2.6.2`

Note, if you use nbviewer it uses reveal 3.1.0

The problem is that you can't put these on a live server unless it is has reveal.js installed in that local directory.

Scenario 1:
```
ipython nbconvert --to slides TestSlides.ipynb
python -m SimpleHTTPServer 8000
````
This browser sees the html, but they don't act as slides b/c reveal.js isn't found. 

Scenario 2:
```
ipython nbconvert --to slides TestSlides.ipynb
wget https://github.com/hakimel/reveal.js/archive/3.0.0.tar.gz
tar -xvf 3.0.0.tar.gz
mv reveal.js-3.0.0 reveal.js
python -m SimpleHTTPServer 8000
```
Now the browser sees the reveal.js (js and css) resources, but for some reason it renders as a blank screen. It also still has problem that you can't deploy without making a copy of reveal.js in the corresponding directory on the live server.

Scenario 3:
```
ipython nbconvert --to slides TestSlides.ipynb  --reveal-prefix https://cdn.jsdelivr.net/reveal.js/2.6.2
python -m SimpleHTTPServer 8000
```

This works now locally and on live server, but with old reveal.js

Scenario 4:
```
ipython nbconvert --to slides TestSlides.ipynb  --reveal-prefix http://cdn.bootcss.com/reveal.js/3.1.0
python -m SimpleHTTPServer 8000
```
This gives similar behavior as Scenario 2, the resources are there but the browser renders an empty screen. In both Scenario 2 & 4 I see these errors in the console:
```
[Error] ReferenceError: Can't find variable: Reveal
	global code (TestSlides.slides.html, line 333)
[Error] Error: Mismatched anonymous define() module: function () {
			root.Reveal = factory();
			return root.Reveal;
		}
http://requirejs.org/docs/errors.html#mismatch
	da (require.min.js, line 8)
	w (require.min.js, line 12)
	M (require.min.js, line 15)
	(anonymous function) (require.min.js, line 26)
```

[...]

PR #62, opened 2015-07-14
[Fixes using reveal.js>3] Wait for reveal.js to load before using it.

cjrh:
This issue is described in these places:

http://stackoverflow.com/a/30128179/170656

The solution is described here:

https://github.com/hakimel/reveal.js/issues/389

It turns out that since requirejs is already in the template,
all we have to do is defer the use of Reveal till after
the reveal.js script has loaded.

### DIFF
--- a/nbconvert/templates/html/slides_reveal.tpl
+++ b/nbconvert/templates/html/slides_reveal.tpl
@@ -164,24 +164,31 @@ div.output_prompt {
 
 <script src="{{resources.reveal.url_prefix}}/lib/js/head.min.js"></script>
 
-<script src="{{resources.reveal.url_prefix}}/js/reveal.js"></script>
-
 <script>
+requirejs(["{{resources.reveal.url_prefix}}/js/reveal.js"], function() {
 
-// Full list of configuration options available here: https://github.com/hakimel/reveal.js#configuration
-Reveal.initialize({
-controls: true,
-progress: true,
-history: true,
+    // Full list of configuration options available here: https://github.com/hakimel/reveal.js#configuration
+    Reveal.initialize({
+    controls: true,
+    progress: true,
+    history: true,
 
-theme: Reveal.getQueryHash().theme, // available themes are in /css/theme
-transition: Reveal.getQueryHash().transition || 'linear', // default/cube/page/concave/zoom/linear/none
+    theme: Reveal.getQueryHash().theme, // available themes are in /css/theme
+    transition: Reveal.getQueryHash().transition || 'linear', // default/cube/page/concave/zoom/linear/none
 
-// Optional libraries used to extend on reveal.js
-dependencies: [
-{ src: "{{resources.reveal.url_prefix}}/lib/js/classList.js", condition: function() { return !document.body.classList; } },
-{ src: "{{resources.reveal.url_prefix}}/plugin/notes/notes.js", async: true, condition: function() { return !!document.body.classList; } }
-]
+    // Optional libraries used to extend on reveal.js
+    dependencies: [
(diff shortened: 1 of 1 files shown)


cjrh:
I see there is already an issue for this #34.

takluyver:
Thanks! I'll let @damianavila have a look at this (when he's recovered from a very long flight).

damianavila:
I have a branch for this I did on the plane :wink: 
It covers some other things as well, but I guess we could eventually merge this if this is working OK and then I can PR my branch... so this contribution get merged.
In any case, I have to test it yet... (still no recovered... just landed a few hours ago and a local flight to take in a little bit... I will check this tomorrow)

cjrh:
Hi @damianavila :)

By all means, if your patch is a superset I'm happy to go with yours.

There is one other thing I've noticed.  I have played around with creating my own template that extends `slides_reveal.tpl` in the following way:

```jinja
{%- extends 'slides_reveal.tpl' -%}

{% block body %}

{{ super() }}

<script>
requirejs(["reveal.js/js/reveal.js"], function(Reveal) {
    console.log("Running configure.");
    Reveal.configure({

        // Display controls in the bottom right corner
        //controls: true,

        // Display a presentation progress bar
        //progress: true,

        // Push each slide change to the browser history
        //history: false,

        // Enable keyboard shortcuts for navigation
        //keyboard: true,

        // Enable touch events for navigation
        //touch: true,

        // Enable the slide overview mode
        //overview: true,

        // Vertical centering of slides
        //center: true,

        // Loop the presentation
        //loop: false,

        // Change the presentation direction to be RTL
        //rtl: false,

        // Number of milliseconds between automatically proceeding to the
        // next slide, disabled when set to 0, this value can be overwritten
        // by using a data-autoslide attribute on your slides
        //autoSlide: 0,

        // Enable slide navigation via mouse wheel
        //mouseWheel: false,

        // Transition style
        transition: 'fade', // default/cube/page/concave/zoom/linear/fade/none

        // Transition speed
        transitionSpeed: 'slow', // default/fast/slow

        // Transition style for full page backgrounds
        //backgroundTransition: 'default', // default/linear/none

    });
    // Change the reveal js theme like this
    document.getElementById("theme")
         .setAttribute('href', "{{resources.reveal.url_prefix}}/css/theme/night.css");
});

</script>

{% endblock body %}
```

I _suspect_ that it is poor form to use `requirejs` twice to load revealjs (the first time happens in my patched `slides_reveal.tpl`).  It seems quite nice though to allow someone to run `.configure()` on the revealjs instance to change the defaults, as per above.

What I've been thinking is that perhaps one could add a 

```jinja
{% block revealconf %}
{% endblock revealconf %}
```

to the bottom part of the requirejs code, say underneath where the `Reveal.addEventListener(...` line appears (but before the `requirejs` callback function closes).  Then, in the extending template one could simply add

```jinja
{% block revealconf %}
    Reveal.configure({
        // Display controls in the bottom right corner
        //controls: true,

        // etc.
    });
    // Change the reveal js theme like this
    document.getElementById("theme")
         .setAttribute('href', "{{resources.reveal.url_prefix}}/css/theme/night.css");
{% endblock revealconf %}
```

and not bother about knowing that requirejs is being used. Thoughts?

damianavila:
>By all means, if your patch is a superset I'm happy to go with yours.

Great, I will push it later today...

>Thoughts?

That's really interesting and I think we could provide that point so people can customize their slides easily... can you open a new issue to discuss this even further? No need to rewrite anything, just link the your post comment. Thanks.

cjrh:
Since I've added the issue as per @damianavila, I'm closing this PR.

PR #65, opened 2015-07-20
Make slides to work with latest Reveal

### LINKED ISSUE #34: slides display problem: reveal.js version conflict?

damianavila:
Closes #34 

* [x] Change how to load latest reveal
* [x] Update CDN to get the latest `reveal.js`
* [x] Load the mathjax macro in a proper place at `head`

### DIFF
--- a/nbconvert/templates/html/slides_reveal.tpl
+++ b/nbconvert/templates/html/slides_reveal.tpl
@@ -144,9 +144,11 @@ div.output_prompt {
   margin: 5px 5px 0 0;
 }
 div.text_cell.rendered .rendered_html {
+  /* The H1 height seems miscalculated, we are just hidding the scrollbar */
   overflow-y: hidden;
 }
 a.anchor-link {
+  /* There is still an anchor, we are only hidding it */
   display: none;
 }
 .rendered_html p {
--- a/nbconvert/templates/html/slides_reveal.tpl
+++ b/nbconvert/templates/html/slides_reveal.tpl
@@ -74,6 +74,9 @@ if( window.location.search.match( /print-pdf/gi ) ) {
 <script src="{{resources.reveal.url_prefix}}/lib/js/html5shiv.js"></script>
 <![endif]-->
 
+<!-- Loading the mathjax macro -->
+{{ mathjax() }}
+
 <!-- Get Font-awesome from cdn -->
 <link rel="stylesheet" href="//netdna.bootstrapcdn.com/font-awesome/4.1.0/css/font-awesome.css">
 
@@ -155,9 +158,6 @@ a.anchor-link {
 
 <!-- Custom stylesheet, it must be in the same directory as the html file -->
 <link rel="stylesheet" href="custom.css">
-
-<!-- Loading the mathjax macro -->
-{{ mathjax() }}
 
 </head>
 {% endblock header%}
--- a/nbconvert/templates/html/slides_reveal.tpl
+++ b/nbconvert/templates/html/slides_reveal.tpl
@@ -156,6 +156,9 @@ a.anchor-link {
 <!-- Custom stylesheet, it must be in the same directory as the html file -->
 <link rel="stylesheet" href="custom.css">
 
+<!-- Loading the mathjax macro -->
+{{ mathjax() }}
+
 </head>
 {% endblock header%}
 
@@ -168,37 +171,50 @@ a.anchor-link {
 </div>
 </div>
 
-<script src="{{resources.reveal.url_prefix}}/lib/js/head.min.js"></script>
-
-<script src="{{resources.reveal.url_prefix}}/js/reveal.js"></script>
-
 <script>
 
-// Full list of configuration options available here: https://github.com/hakimel/reveal.js#configuration
-Reveal.initialize({
-controls: true,
-progress: true,
-history: true,
+require(
+    {
+      // for really large notebook this probably makes sense
+      waitSeconds: 30
+    },
+    [
+      "{{resources.reveal.url_prefix}}/lib/js/head.min.js",
+      "{{resources.reveal.url_prefix}}/js/reveal.js"
+    ],
 
-theme: Reveal.getQueryHash().theme, // available themes are in /css/theme
-transition: Reveal.getQueryHash().transition || 'linear', // default/cube/page/concave/zoom/linear/none
+    function(head, Reveal){
 
-// Optional libraries used to extend on reveal.js
-dependencies: [
-{ src: "{{resources.reveal.url_prefix}}/lib/js/classList.js", condition: function() { return !document.body.classList; } },
-{ src: "{{resources.reveal.url_prefix}}/plugin/notes/notes.js", async: true, condition: function() { return !!document.body.classList; } }
-]
-});
-</script>
+        // Full list of configuration options available here: https://github.com/hakimel/reveal.js#configuration
+        Reveal.initialize({
+            controls: true,
+            progress: true,
+            history: true,
 
-<!-- Loading mathjax macro -->
-{{ mathjax() }}
+            theme: Reveal.getQueryHash().theme, // available themes are in /css/theme
+            transition: Reveal.getQueryHash().transition || 'linear', // default/cube/page/concave/zoom/linear/none
 
-<script>
-Reveal.addEventListener( 'slidechanged', function( event ) {
-  window.scrollTo(0,0);
-  MathJax.Hub.Rerender(event.currentSlide);
-});
+            // Optional libraries used to extend on reveal.js
+            dependencies: [
+                { src: "{{resources.reveal.url_prefix}}/lib/js/classList.js",
+                  condition: function() { return !document.body.classList; } },
+                { src: "{{resources.reveal.url_prefix}}/plugin/notes/notes.js",
+                  async: true,
+                  condition: function() { return !!document.body.classList; } }
+            ]
+        });
+
+        var update = function(event){
+          window.scrollTo(0,0);
+          if(MathJax.Hub.getAllJax(Reveal.getCurrentSlide())){
+            MathJax.Hub.Rerender(Reveal.getCurrentSlide());
+          }
+        };
+
+        Reveal.addEventListener('slidechanged', update);
+    }
+);
+
 </script>
 
 </body>
--- a/nbconvert/templates/html/slides_reveal.tpl
+++ b/nbconvert/templates/html/slides_reveal.tpl
@@ -175,8 +175,9 @@ a.anchor-link {
 
 require(
     {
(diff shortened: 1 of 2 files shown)


damianavila:
WIP yet:

* [ ] Make scroll work again
* [ ] Render math on fragments too

damianavila:
@bollwyvl probably this is interesting for you. I migrate the `overflow-y`into the `reveal` class. In this way we only have a functional scrolling bar when there is content exceeding the slides view. I chose to use `scroll` instead of `auto`, because `auto` generates an horizontal movement when you switch from a "non scrolling" slide to a "scrolling" one. We discussed this in the past in the nbviewer repo and I think this solution, even it is not ideal, it solves some of the problem on mobile and actually represent better the "scroll" status of a slide...

bollwyvl:
Very cool, looks much cleaner!

damianavila:
ping @jupyter/nbconvert

nbconvert/templates/html/slides_reveal.tpl
```diff
@@ -168,37 +171,52 @@ a.anchor-link {
...
-]
-});
-</script>
-
-<!-- Loading mathjax macro -->
-{{ mathjax() }}
+require(
+    {
+      // for really large notebook this probably makes sense
+      waitSeconds: 30
```
takluyver: What is this timeout for? The size of the notebook shouldn't affect how long it takes to load JS libraries, right?

nbconvert/templates/html/slides_reveal.tpl
```diff
@@ -168,37 +171,52 @@ a.anchor-link {
...
+                { src: "{{resources.reveal.url_prefix}}/lib/js/classList.js",
+                  condition: function() { return !document.body.classList; } },
+                { src: "{{resources.reveal.url_prefix}}/plugin/notes/notes.js",
+                  async: true,
+                  condition: function() { return !!document.body.classList; } }
+            ]
+        });
+
+        var update = function(event){
+          window.scrollTo(0,0);
```
takluyver: Is the scrollTo needed here? It seems unrelated to rendering mathjax.

nbconvert/templates/html/slides_reveal.tpl
```diff
@@ -74,6 +74,9 @@ if( window.location.search.match( /print-pdf/gi ) ) {
 <script src="{{resources.reveal.url_prefix}}/lib/js/html5shiv.js"></script>
 <![endif]-->
 
+<!-- Loading the mathjax macro -->
+{{ mathjax() }}
```
takluyver: Where is this macro defined? I can't immediately find it in our code.

takluyver:
Overall I think this looks good, there are a few specific bits I've asked for clarification on.

nbconvert/templates/html/slides_reveal.tpl
damianavila: You are right... I was working on RISE at the same time, where the time to load the live notebook could affect the reveal.js load and arise a timeout...
Certainly, I don't think this setting is needed here, unless I am missing something else... so I will delete it for now...

nbconvert/templates/html/slides_reveal.tpl
damianavila: It is not related with mathjax, I should add a comment here... also I have to check if the scroll is still working after change the overflow into the reveal class... essentially you want to scroll to the top on every `slidechange` event (otherwise when you come back to a previous slide, you probably miss content because the slide keep scrolled to the bottom), so probably I have to rewrite this to only scroll on the `slidechange` event and not in the fragments events.

nbconvert/templates/html/slides_reveal.tpl
damianavila: At `nbconvert/nbconvert/templates/html/`, next to others html templates.

takluyver: Aha, thanks.

nbconvert/templates/html/slides_reveal.tpl
takluyver: OK, if you want to delete this and add the comment below, I think this is good to go.

damianavila:
@takluyver, made some additional changes:

1. I would like to use the waitSecond settting because it makes sense when you are in slow connections loading reveal from a CDN. I set that t 15 seconds.
2. I fix the scroll to top on the slidechange event. Now the code is self-explanatory and decupled from the mathjax update, so there is no need for a comment, I think...

@bollwyvl, regarding my last comment about the scrolling, regrettably I still see the overflow bar in some no-scroll slides because it seems to take next largest slide to set up the bars... but still is better than before because people not using any "scrolling" slides will have a no-scrolling slideshow (with the bar but deactivated).

bollwyvl:
Bummer about the scrolling. I hate to be defending slides as a medium, but
I just have trouble with the "user has so much crap to say it doesn't fit
on a weirdly-margined-slide use case" :P

Apropos of https://github.com/jupyter/nbviewer/issues/484 (bokeh table
css), is there a reason we don't have a #notebook somewhere in slides?

Also, might it be possible to break up the reveal stuff into even more
blocks to make nbviewer easier? I had to strip most of the page chrome to
get all the paths right, which lead to the real bug in that issue. If I
could sub the template and pull the no-fooling `<div class="reveal">` that
might help avoid these kinds of problems.

damianavila:
>Bummer about the scrolling. I hate to be defending slides as a medium, but
I just have trouble with the "user has so much crap to say it doesn't fit
on a weirdly-margined-slide use case" :P

Can you elaborate a little more on this...

>Apropos of https://github.com/jupyter/nbviewer/issues/484 (bokeh table
css), is there a reason we don't have a #notebook somewhere in slides?

No reason I am aware of... or I remember...

>Also, might it be possible to break up the reveal stuff into even more
blocks to make nbviewer easier? I had to strip most of the page chrome to
get all the paths right, which lead to the real bug in that issue. If I
could sub the template and pull the no-fooling `<div class="reveal">` that
might help avoid these kinds of problems.

Yes, please, open an issue for that... there is another asking for that to inject reveal configure in a custom template, I think it makes sense to decouple things with more block to help people (and us) to use it in custom settings (like nbviewer).

bollwyvl:
> Can you elaborate a little more on this...

I think the phrase I used before was that they "break the medium"... slides "unintentionally" bigger (i.e. not infinite like prezi, etc.) than the screen...
- are not supported on any other slide platform
- magnify the "i'm going to read you this slide" presentation style... most people can read well north of 300 words per minute in their mother tongue, and showing someone a dense slide, they'll likely tune you out until they've read it
- break URLs: no way to link to a particular scrolled location. not that the reveal ones are great to begin with
- are not liked by reveal, hence all our troubles
  - will likely never printable or exportable, one of the use cases i value a lot but haven't been able to handle succinctly... even without scrolling slides... but at least it is theoretically doable!
- which require a scroll wheel/bar during a talk is distracting to the speaker and the listeners, and content will be in a half-revealed state, which again breaks focus for both parties
  - if the content really is big, like a big diagram or visualization, then temporary full screen/scroll/zoom on that content is a better fit... with text and margins so big, it's hard to even get 80 character code on a screen
- break touch gestures (unless you've fixed that!)

perhaps an explicit `longslide` slidetype that, if found, would enable the scrolling, while `slide` would conform to the reveal model would be less surprising.

> No reason I am aware of... or I remember...

wasn't sure if there was some un-reset stuff from notebook that would mess us up... in browser testing suggests it looks okay...

> Yes, please, open an issue for that...

Will do!

> inject reveal configure in a custom template

Would love to see this in the notebook metadata as the "last word"... but i could see how having nbviewer options, config options (as per rise) etc.

damianavila:
>perhaps an explicit longslide slidetype that, if found, would enable the scrolling, while slide would conform to the reveal model would be less surprising.

Well, lot of good reasons, we should discuss this point again... I like the idea of a long slide, but no sure how complicated could be... I passed those critics to new issue to discuss further: #78

>wasn't sure if there was some un-reset stuff from notebook that would mess us up... in browser testing suggests it looks okay...

Great, should be an easy fix...

>Will do!

Thanks

>Would love to see this in the notebook metadata as the "last word"...

Yep... I agree, we should probably re-discuss that point, or provide a plugin (extension) to configure this from the notebook metadata... maybe you can open a new issue to discuss that :wink:

damianavila:
btw, @takluyver I think this is ready... all the other discussion are in parallel and to be solved later.

takluyver:
Thanks, I think this looks good.

PR #66, opened 2015-07-20
Fix 37 and 38 overriding css at the slides template.

### LINKED ISSUE #37: Anchors in reveal
Markdown heading anchors carry over to slides, where they have no effect due to being overridden. Their presence leads to visual junk on mouseover, and the slides jump to start when an anchor is accidentally clicked.

I guess removing anchor generation in `markdown2html` would be a solution.

### LINKED ISSUE #38: h1 scrollbars in reveal
A markdown cell containing only a single first level heading acquires an ugly scrollbar.

caption_scrollbar_chrome

Adding an extra markdown line, or changing to heading 2 removes the scrollbar.

damianavila:
closes #37 
closes #38

### DIFF
--- a/nbconvert/templates/html/slides_reveal.tpl
+++ b/nbconvert/templates/html/slides_reveal.tpl
@@ -142,6 +142,12 @@ div.output_prompt {
   /* 5px right shift to account for margin in parent container */
   margin: 5px 5px 0 0;
 }
+div.text_cell.rendered .rendered_html {
+  overflow-y: hidden;
+}
+a.anchor-link {
+  display: none;
+}
 .rendered_html p {
   text-align: inherit;
 }


akhmerov:
For future reference, these two issues are mostly hidden now. The height of `<h1>` or of the cell is still miscalculated, and there's still an anchor appended by nbconvert to each header. Of

damianavila:
We can probably add a comment in the source code about these issues until we come up with a better implementation. I will add that in some other PR I am working on right now, so people from the future can expend some time to deal with this in a better way. Right now, I think the cost-benefit of the these fixes is enough to live with them... Thanks @akhmerov for making this comment.

damianavila:
Cooments added here: https://github.com/damianavila/jupyter_nbconvert/commit/bca0a73f7a409b69885f60267be5c4c2e7cd5534

PR #69, opened 2015-07-20
Rehash and simplify slide export

takluyver:
This is a medium size reworking of the slide export code. The major changes are:

1. Got rid of the RevealHelpPreprocessor, merging its functionality into SlideExporter. Neither is useful without the other, so it doesn't make sense to have a preprocessor that can be turned on and off separately. This slightly breaks config: `RevealHelpPreprocessor.url_prefix` becomes `SlidesExporter.reveal_url_prefix`. I can do a compatibility shim if needed, but I suspect it's not worth the added complexity.

2. The template and the cell metadata added to slides to facilitate grouping them into slides/subslides/fragments are (I hope) simplified. It now has six keys: `{slide,subslide,fragment}_{start,end}` (plus slide_type). In the template, we just prefix & suffix the cell with appropriate opening and closing tags according to the start/end attributes. I suspect there may be one or two corner cases fixed by this, but I haven't confirmed any.

3. The extra metadata is now added to a copy of the notebook, rather than modifying the notebook object. This has no effect when using nbconvert at the command line, but if you use it in a Python script, this means the notebook object you pass in is not changed.

Ping @damianavila


damianavila:
Nice!! 
I started to work on this in the afternoon, but patching the previous logic and this PR really seems to simplify the logic a **lot**, so I would like to go with your implementation :stuck_out_tongue_winking_eye:
I will comment/review tomorrow... a little bit late here right now.

nbconvert/exporters/slides.py
```diff
@@ -1,28 +1,78 @@
...
 
 from .html import HTMLExporter
 
-#-----------------------------------------------------------------------------
-# Classes
-#-----------------------------------------------------------------------------
+def prepare(nb):
+    """Add some convenenience metadata on cells for the slide template.
+    """
+    nb = deepcopy(nb)
```
damianavila: perfect, this simplify a lot of things and keep things clean...

nbconvert/exporters/slides.py
```diff
@@ -1,28 +1,78 @@
...
+
+    for cell in nb.cells:
+        # Make sure every cell has a slide_type
+        cell.metadata.slide_type = cell.metadata.get('slideshow', {}).get('slide_type', '-')
+
+    # Find the first visible cell
+    for index, cell in enumerate(nb.cells):
+        if cell.metadata.slide_type not in {'notes', 'skip'}:
+            cell.metadata.slide_type = 'slide'
```
damianavila: We are forcing the `slide_type=slide` in a cell maybe no labelled as is... but I am OK with that because we are working in a copy of the nb, not the nb itself, so changing the property from what it was selected thought the user interface should not be a problem anymore...

Additionally, what about detecting the slide_type before this so we can warn the user about how properly start an slideshow with the `slyde_type=slide` instead of `-` or `subslide`? I guess we should tell people that we are using this "not slide" first visible cell as slide beginning because they never start the slideshow with the proper slide_type.
I think a warning in the console (but proceed with the conversion, obviously) should be enough in this case...

nbconvert/exporters/slides.py
```diff
@@ -1,28 +1,78 @@
...
+            cell.metadata.slide_start = True
+            cell.metadata.subslide_start = True
+            first_slide_ix = index
+            break
+    else:
+        raise ValueError("All cells are hidden, cannot create slideshow")
+
+    in_fragment = False
+
+    for index, cell in enumerate(nb.cells[first_slide_ix+1:],
```
damianavila: jeje... at overall, I made the same in my branch :wink:, detecting the first slide and enumerating from that point (well the point+1 in this case)

nbconvert/exporters/slides.py
```diff
@@ -1,28 +1,78 @@
...
+                                 start=(first_slide_ix+1)):
+
+        previous_cell = nb.cells[index - 1]
+
+        # Get the slide type. If type is start, subslide, or slide,
+        # end the last subslide/slide.
+        if cell.metadata.slide_type == 'slide':
+            previous_cell.metadata.slide_end = True
+            cell.metadata.slide_start = True
+        if cell.metadata.slide_type in {'subslide', 'slide'}:
```
damianavila: we should make a comment here to help understand people that we defined our slides as 
```
<section>
  <section>
  ...
  </section>
<section>```

otherwise is really confusing why we check the slide_type of this set (`{'subslide', 'slide'}`).

File: nbconvert/exporters/slides.py [hunk 5]
```diff
@@ -1,28 +1,78 @@
...
+            previous_cell.metadata.subslide_end = True
+            cell.metadata.subslide_start = True
+            in_fragment = False
+
+        elif cell.metadata.slide_type == 'fragment':
+            cell.metadata.fragment_start = True
+            if in_fragment:
+                previous_cell.metadata.fragment_end = True
+            else:
+                in_fragment  = True
```
damianavila: In general, nice way to solve the fragment problem without using data-fragment-index :tada:

damianavila:
First of all, sorry for the delay and thanks for this :wink:

I left specific comments above, now some more general comments:

>Got rid of the RevealHelpPreprocessor, merging its functionality into SlideExporter. Neither is useful without the other, so it doesn't make sense to have a preprocessor that can be turned on and off separately. 

Sound OK to me to merge it.

>This slightly breaks config: RevealHelpPreprocessor.url_prefix becomes SlidesExporter.reveal_url_prefix. I can do a compatibility shim if needed, but I suspect it's not worth the added complexity.

I don't think we need it...

>The template and the cell metadata added to slides to facilitate grouping them into slides/subslides/fragments are (I hope) simplified.

Yep, certainly simplified.

>I suspect there may be one or two corner cases fixed by this, but I haven't confirmed any.

All the cases I tested seems to work OK.

>The extra metadata is now added to a copy of the notebook, rather than modifying the notebook object. This has no effect when using nbconvert at the command line, but if you use it in a Python script, this means the notebook object you pass in is not changed.

Perfect as I mentioned above in one of my comments.

To summarize, this is great! I only left some minor specific comments. After fixing them or answer them, this LGTM to go...

nbconvert/exporters/slides.py
takluyver: I think conversion is the wrong time to warn people - it would be better to give them information like that in the interface when they're using the slide toolbar. But that's obviously much trickier to design.

Additionally, it doesn't really matter whether they start the slideshow 'properly' with a slide cell - we don't lose anything by accepting any visible cell as the start of the first slide (and RISE [does the same](https://github.com/damianavila/RISE/blob/56eff79aac0a0c49ad212c6944ff2fd16c774b65/livereveal/main.js#L71)).

nbconvert/exporters/slides.py
takluyver: I've added a comment - hopefully it makes this a bit clearer.

takluyver:
Thanks for the review :-)

damianavila:
Are we having a 4.0.1 or we are progressing to 4.1? Because I think this could be merged now...

takluyver:
I think we're heading for 4.1 - there are 3 PRs merged already tagged 4.1, and a fix for #74 would also be appropriate. Thanks!

takluyver:
@minrk I have added the config shim.

PR #71, opened 2015-07-21
Make 'element' available for JS output to add to document

takluyver:
From discussion with Min. JS output in the notebook can add to 'element' to put things on the page in that output area. This adds the same variable for use by JS output in nbconverted HTML.

I'm playing with this in altwidgets-prototype, and it seems to work. I haven't thought through all the possible scoping issues, though.

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -9,6 +9,7 @@ from __future__ import print_function, absolute_import
 
 # Stdlib imports
 import os
+import uuid
 
 # other libs/dependencies are imported at runtime
 # to move ImportErrors to runtime when the requirement is actually needed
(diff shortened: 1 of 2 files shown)


nbconvert/exporters/templateexporter.py
```diff
@@ -283,6 +284,8 @@ def _create_environment(self):
             extensions=JINJA_EXTENSIONS
             )
 
+        environment.globals['uuid4'] = uuid.uuid4
```
takluyver: I tried putting this in filters, but the `set` statement in the template didn't appear to look in that namespace. I'm happy to make a more general apparatus in a separate PR for adding things to the `globals` namespace if we think that's useful.

minrk: Shouldn't this be `lambda : uuid.uuid4().hex` returning a string instead of the instance?

takluyver: It appears to work as is - I guess that Jinja does `str()` on objects, which gives you the same as `.hex`. This way feels neater, but I can also see the 'explicit is better than implicit' argument.

minrk:
@takluyver sorry, needs rebase. If you don't want to do that while you are traveling, I can finish it up.

minrk:
Rebased and merged as #98

PR #83, opened 2015-08-16
setuptools entrypoint

minrk:
because windows can't handle vanilla scripts

cf jupyter/jupyter#15


Carreau:
ok for me.

minrk:
We should get 4.1 out soon with this. @damianavila, @takluyver anything more you guys want in 4.1?

damianavila:
#71 is milestoned for 4.1. I also have some more slides-related fixes but we can wait for another release, not rush (I still have to test it more in depth).

PR #84, opened 2015-08-19
Update docs a bit.

Carreau:
use a v4 notebook


nicktimko:
I might mention in passing how to configure the Exporter to extract embedded files (lines 86-90), i.e. by adding `'ExtractOutputPreprocessor': {'enabled': True}` to the `config` dict (leave the rest to the reader to explore), but that's just me because I'll probably be looking at that in a while. Saying "you can do X" without providing a clue is tantalizing.

rgbkrk:
Looking good. The Exporter configuration @nicktimko mentioned can probably go in another PR to not let this languish.

PR #87, opened 2015-08-21
Update slides_reveal.tpl

zabeloliver:
added missing "http:"

### DIFF
--- a/nbconvert/templates/html/slides_reveal.tpl
+++ b/nbconvert/templates/html/slides_reveal.tpl
@@ -73,7 +73,7 @@ if( window.location.search.match( /print-pdf/gi ) ) {
 {{ mathjax() }}
 
 <!-- Get Font-awesome from cdn -->
-<link rel="stylesheet" href="//netdna.bootstrapcdn.com/font-awesome/4.1.0/css/font-awesome.css">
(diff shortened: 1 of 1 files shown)


Carreau:
Did you had a bug ? A URL that start with double slash make sens. It is called a protocol relative URL, that is to say it is https if the page is https, and http if the page is http.

zabeloliver:
Oh, i wasn't aware of this. because the actual version of jupyter/nbconvert doesn't work with reveal 3.x i thought that this could may be the problem .. but in the meantime i changed the the local reveal to 2.6 and i worked then, so i tought that my code change did it... so sorry for thi PR and for wasting your time..

Carreau:
No problem, protocol relative URL might trigger issues, and are not a widely known. 
Issuing a PR was the right things to do, as it might allow other people to find this information later. 

No need to apologize, you are not wasting our time, we are here to direct and answer to new contributors question. So feel free to open issues or ask if you are not sure of something or need help. 

We will gladly spend an afternoon helping you in doing a contribution even if it would have taken us half an hour. We'll have went through the noob phase where we would indent with tab, finish line with semicolon and were guided by more experienced developper and we keep doing that for new ones. 

Thanks, and looking forward to your next PR.

PR #101, opened 2015-09-14
render markdown output in rst

juhasch:
Currently `text/markdown` output is ignored and only `text/plain` is used in conversion to rst.

### DIFF
--- a/nbconvert/utils/base.py
+++ b/nbconvert/utils/base.py
@@ -13,7 +13,7 @@ class NbConvertBase(LoggingConfigurable):
     Useful for display data priority that might be use by many transformers
     """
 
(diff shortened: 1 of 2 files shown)


nbconvert/utils/base.py
```diff
@@ -13,7 +13,7 @@ class NbConvertBase(LoggingConfigurable):
     Useful for display data priority that might be use by many transformers
     """
 
-    display_data_priority = List(['text/html', 'application/pdf', 'text/latex', 'image/svg+xml', 'image/png', 'image/jpeg', 'text/plain'],
+    display_data_priority = List(['text/html', 'application/pdf', 'text/latex', 'text/markdown', 'image/svg+xml', 'image/png', 'image/jpeg', 'text/plain'],
```
minrk: Can you move text/markdown to second-lowest priority (just above text/plain)?

juhasch:
Sure.

minrk:
Thanks!

Issue #43, opened 2015-06-17
Reveal odd latex formatting


akhmerov:
This is going to be super-fun to debug:

`* $A \parallel B_{CD}$`

renders as

![reveal_bug](https://cloud.githubusercontent.com/assets/2069677/8215956/c7ddaa30-1533-11e5-9c80-8b16f454cd57.png)

in `--to slides` (a line break is inserted or so). It renders correctly in `--to html`. Removing any one of `*`, `\parallel`, or replacing `{CD}` with `{C}` fixes the rendering.

A workaround is to split a single math environment into several: `$A$ $\parallel$ $B_{CD}$`. Not sure what to do with `$$ $$`.

takluyver:
@damianavila , another slides related thing. Maybe some mathjax CSS (does mathjax use CSS?) is clashing with reveal stuff.

damianavila:
Interesting...  it could be some clashing, I have to investigate (IIRC, reveal have a mathjax plugin, so maybe they are using some other configuration we can research into if we don't figured out at first sight what is the issue here).

btw @takluyver can you assign these issues to me, so I can filter them easily... or add me to the repo and I assign the slides-related thing to myself... thx

akhmerov:
It might help to know that whatever it is, it doesn't happen in RISE.

damianavila:
Thanks for the RISE note... then seems to be nbconvert related... we'll need to figured out.

takluyver:
@damianavila I've invited you to collaborate on this repo so we can assign issues to you. As you've probably already seen, we all try to collaborate via pull requests rather than pushing directly. You'll be able to review and merge other people's pull requests, but if you're in any doubt about a PR, feel free to ask for another pair of eyes on it. :-)

takluyver:
I also made a slides tag to group issues related to the slides output.

damianavila:
Thanks.
The slides tag is very useful... and about the practices, I practice (and encourage to others) the same ones in several of the open source project I am involved with :wink:

damianavila:
@akhmerov, I can not reproduce the issue...

I tried `* $A \parallel B_{CD}$` and it renders OK, without the final `$` it does not render even in the notebook, so I guess it is the correct expression (I don't know too much about latex).

I tried with several versions of mistune (there was an issue recently with 0.6) but even in that case I could not reproduce the issue... do you have a notebook where you see it? Cheers.

akhmerov:
Here:

[...]

PR #104, opened 2015-09-15
[slides] updating uninitialized resources in from_notebook_node

bollwyvl:
Was having a look at options for #43... not sure how we haven't been bitten by this on nbviewer, as we call `from_notebook_node` without `resources`. In some local testing, however, I am seeing `The error was: argument of type 'NoneType' is not iterable`. This fixes it!

### DIFF
--- a/nbconvert/exporters/slides.py
+++ b/nbconvert/exporters/slides.py
@@ -102,7 +102,7 @@ class SlidesExporter(HTMLExporter):
     output_mimetype = 'text/html'
 
     def from_notebook_node(self, nb, resources=None, **kw):
-        self._init_resources(resources)
+        resources = self._init_resources(resources)
         if 'reveal' not in resources:
             resources['reveal'] = {}
         resources['reveal']['url_prefix'] = self.reveal_url_prefix


damianavila:
upps... how we missed this... LGTM (keeping it opening for another view, but I will merge tomorrow if anyone beats me first).

bollwyvl:
Hooray!

I realize now that were the current master to ship, slides would be really
broken on nbviewer. Thinking about adding a template_path over there, and
overriding slides_reveal... Or, perhaps we consider going to an iframe for
all rendered notebooks? It's kind of crazy to expect all of these things to
coexist. Since we don't care about SEO, there no real disadvantage to
having our chrome and the "meat" in different contexts. Also the "embed
this notebook" user story is strong... Especially for slides.

damianavila:
>Thinking about adding a template_path over there, and
overriding slides_reveal... 

We talked about this in another thread, but we should provide a nbconvert all the abstractions in the template so you can use it in nbviewer without pain...

>Or, perhaps we consider going to an iframe for
all rendered notebooks? It's kind of crazy to expect all of these things to
coexist. 

It could be another option...

I would like to start with the first option because it will useful for you and other as well...

PR #106, opened 2015-09-19
fix broken href

akhmerov:
Trivial cruft fix.


Carreau:
No this is not a mistake, `//something` is a [protocol-relative url](http://billpatrianakos.me/blog/2013/04/18/protocol-relative-urls/) it behave as `http://` on http page and `https://` on https pages.

akhmerov:
My apologies; didn't know. But it does break for a locally opened presentation, since that one uses file://.

Carreau:
`use --serve` when converting to a slideshow it will startup a http server. 
Other things will break if you don't sever through an http server.

PR #107, opened 2015-09-19
Use shell=True on Windows when calling latex

jschueller:
on windows the env is ignored with shell=False
the notebook.tex file is generated in that module, so shell can be safely enabled
I did the same as in the svg2pdf preprocessor

### DIFF
--- a/nbconvert/exporters/pdf.py
+++ b/nbconvert/exporters/pdf.py
@@ -77,7 +77,8 @@ class PDFExporter(LatexExporter):
         with open(os.devnull, 'rb') as null:
             stdout = subprocess.PIPE if not self.verbose else None
             for index in range(count):
-                p = subprocess.Popen(command, stdout=stdout, stdin=null)
+                shell = (sys.platform == 'win32')
+                p = subprocess.Popen(subprocess.list2cmdline(command) if shell else command, stdout=stdout, stdin=null, shell=shell)
                 out, err = p.communicate()
                 if p.returncode:
                     if self.verbose:


nbconvert/exporters/pdf.py
```diff
@@ -77,7 +77,7 @@ def run_command(self, command_list, filename, count, log_function):
         with open(os.devnull, 'rb') as null:
             stdout = subprocess.PIPE if not self.verbose else None
             for index in range(count):
-                p = subprocess.Popen(command, stdout=stdout, stdin=null)
+                p = subprocess.Popen(command, stdout=stdout, stdin=null, shell=True) #Shell=True okay since input is trusted.
```
minrk: command is a list, and Popen(shell=True) only accepts when command is a string. If we do this, we will have to properly collapse the command-list into a command-string (with proper quoting, etc.)

jschueller:
@minrk I read that Popen takes care of converting the command list:
"On Windows, if args is a sequence, it will be converted to a string in a manner described in Converting an argument sequence to a string on Windows. This is because the underlying CreateProcess() operates on strings."

However I also read this: "On Windows [...] You do not need shell=True to run a batch file or console-based executable."

In fact I created a pdflatex.bat provided by my miktex conda package that calls pdflatex.exe in the right place. The pdflatex.bat lies in the conda path (in /Scripts) as the pdflatex.exe cannot just be moved there, as I guess it needs the relative layout wrt the other tex files of the distribution.
Setting shell=True makes this work, so I guessed PATH was not taken into account.

minrk:
> I read that Popen takes care of converting the command list:

This is true *unless* `shell=True`, in which case you need to make it a single string.

> I guessed PATH was not taken into account

PATH is definitely taken into account without `shell=True`. What can be different is if the current process has a different PATH than what you would get from instantiating a shell. But the current process' PATH is certainly used.

jschueller:
Hi,

I finally understood what went wrong: whith shell=False, Popen uses CreateProcess which only looks for .exe.

I also looked at subprocess.py, Popen does the jobs of converting it for us:
```
if not issinstance(args, str):
    args = list2cmdline(args)
```

I think this PR just needs renaming.

minrk:
> I also looked at subprocess.py, Popen does the jobs of converting it for us:

This is true *unless* `shell=True`. The command must be a string when `shell=True`:

```python
In [1]: from subprocess import check_call
In [2]: check_call(['sleep', '1'])
Out[2]: 0

In [3]: check_call(['sleep', '1'], shell=True)
usage: sleep seconds
CalledProcessError...
```

minrk:
If you add the conversion from list to string, then this is okay.

minrk:
Alternately, you can restrict it to `shell=(sys.platform == 'win32')`

jschueller:
Ok, you were right.

PR #108, opened 2015-09-20
nbconvert --inplace implies --to notebook

minrk:
Since --inplace only makes sense if using `--to notebook`, it should be implied.


damianavila:
Seems reasonable, LGTM.

PR #110, opened 2015-09-21
Define missing tokens.

Carreau:
Need a test.

Closes jupyter/notebook#466

(and maybe a few other PRs)


minrk:
@Carreau I'm going to assume that since you made another Issue for the tests, this should be merged now. Merging away...

michaelpacer:
I don't actually know how this passed — when I run nosetests locally on using ipython I get three errors running `$ nosetests nbconvert/tests/test_nbconvertapp.py`. Is the tests directory intentionally not included in the travis build?

Carreau:
Do you want to make a PR  in nbconvert/tests/test_nbconvertapp.py that make the test fails to see if travis fail ? Cause the test should be included in travis testing. 

There might be a skipif when tex not included.

Carreau:
I get 185 test locally 1 fail.

Carreau:
Reverting something wrong with my system, this is not actually needed, and travis does not test PDF.

PR #113, opened 2015-09-22
Revert "Define missing tokens."

Carreau:
Reverts jupyter/nbconvert#110

investigating with @michaelpacer and @jhamrick to test PDF building on travis.


Carreau:
(and exceptionally merging immediately)

PR #114, opened 2015-09-22
test to test pdf generation

Carreau:


.travis.yml
```diff
@@ -3,8 +3,9 @@ python:
     - 3.4
     - 3.3
     - 2.7
-sudo: false
+sudo: true
```
jhamrick: Not sure if this will work, but you might be able to do:

```
addons:
  apt:
    packages:
    - pdflatex
    - texlive-base
    - pandoc
```

And then you wouldn't need to use sudo.

jhamrick:
Blergh, looks like the version of pandoc is too old, so maybe we will have to use sudo to install a more recent version e.g. from here: https://github.com/jgm/pandoc/releases/tag/1.15.0.6

minrk:
We were fetching a pandoc binary [in IPython 3.x](https://github.com/ipython/ipython/blob/3.x/.travis.yml#L18). Sorry, I should have ported that over for Travis. But I suspect we are going to need sudo: true for latex

minrk:
Merged as #126 with pandoc binary we had been using. We can update pandoc to 1.15 at a later point.

PR #115, opened 2015-09-23
add import line to test notebook

### LINKED ISSUE #111: Write a test for missing tokens
Missing tokens definitions when converting to Pdf

michaelpacer:
Adds a line to the notebook2 test notebook to see import and base class token types. The test itself is already incorporated in the `test_pdf` test in `test_nbconvertapp.py` (i.e., it will fail to build if it cannot find token command names).

Closes #111

### DIFF
--- a/nbconvert/tests/files/notebook2.ipynb
+++ b/nbconvert/tests/files/notebook2.ipynb
@@ -1569,6 +1569,8 @@
    "metadata": {},
    "source": [
     "```python\n",
+    "import math, numpy\n",
+    "def matrix_calc(L):\n",
     "def foo(bar=1):\n",
     "    \"\"\"docstring\"\"\"\n",
     "    raise Exception(\"message\")\n",
@@ -1579,7 +1581,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "This string contains a ~~strikethrough~~, struckthrough. "
+    "This string contains a ~~strikethrough~~, struckthrough. ",
    ]
   }
  ],
--- a/nbconvert/tests/files/notebook2.ipynb
+++ b/nbconvert/tests/files/notebook2.ipynb
@@ -1581,7 +1581,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "This string contains a ~~strikethrough~~, struckthrough. ",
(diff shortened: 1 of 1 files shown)


nbconvert/tests/files/notebook2.ipynb
```diff
@@ -1569,6 +1569,8 @@
    "metadata": {},
    "source": [
     "```python\n",
+    "import math, numpy\n",
+    "def matrix_calc(L):\n",
```
jhamrick: Why is there also a function definition? There's a function defined immediately below this line, so technically this is a syntax error... unless that's what you want to test for?

michaelpacer: not intended. I'll fix it.

nbconvert/tests/files/notebook2.ipynb
```diff
@@ -1569,7 +1569,8 @@
    "metadata": {},
    "source": [
     "```python\n",
-    "def foo(bar=1):\n",
+    "import math, numpy\n",
+    "def matrix_calc(L):\n",
```
jhamrick: Sorry to be nitpicky about this, but I have a suspicion that the `bar=1` keyword argument in the original function definition was intentional. Could you either make this function definition also have a keyword argument or just use the old one?

minrk:
@michaelpacer do you have time to address @jhamrick's comment? If not, I can finish this up in a new PR.

michaelpacer:
Hi @minrk I'd missed this — will fix right now.

michaelpacer:
Not sure what happened when i tried to rebase…I'll try to clean this up

nbconvert/tests/files/notebook2.ipynb
```diff
@@ -1569,7 +1569,9 @@
    "metadata": {},
    "source": [
     "```python\n",
-    "def foo(bar=1):\n",
+    "import math \n",
+    "import numpy as np\n",",
```
jhamrick: You've introduced a JSON syntax error here, which is causing the tests to fail.

michaelpacer:
ok I think that should address @jhamrick's comment and it builds successfully, huzzah!

jhamrick:
:+1:

minrk:
Thanks, @michaelpacer and @jhamrick!

PR #120, opened 2015-09-24
Handle cell execution error in preprocess_cell() not in run_cell()

tritemio:
Delay raising CellExecutionError until the cell output is saved.
In this way, when an execution error occurs, the stack trace is stored in 
the cell output. Also, CellExecutionError is raised with a message showing
the source of the cell causing the error.

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -93,14 +93,19 @@ class ExecutePreprocessor(Preprocessor):
         """
         if cell.cell_type != 'code':
             return cell, resources
-        try:
-            outputs = self.run_cell(cell)
-        except Exception as e:
-            self.log.error("failed to run cell: " + repr(e))
-            self.log.error(str(cell.source))
-            raise
+        
+        outputs = self.run_cell(cell)
         cell.outputs = outputs
+    
+        if not self.allow_errors:
+            for out in outputs:
+                if out.output_type == 'error':
+                    msg = 'Error executing the following the notebook cell:\n'
+                    msg += str(cell.source)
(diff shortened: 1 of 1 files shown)


minrk:
Makes sense.

minrk:
Can you add a test that verifies the changed behavior? The existing tests are in `nbconvert/preprocessors/tests/test_execute.py`

tritemio:
Sure.

Thinking about it, I didn't change any behaviour. `CellExecutionError` is still raised in the exact same case. Only the error message is better.

How can I retrieve the content of the error message in the test? 

I'm thinking expanding [this test](https://github.com/jupyter/nbconvert/blob/master/nbconvert/preprocessors/tests/test_execute.py#L160).

tritemio:
I reformatted the test and added a white space removal commit. I can remove if it creates too much noise.

minrk:
We do generally prefer to avoid things like commits that do nothing but cosmetic changes, so please avoid whitespace commits in the future, but this is fine for now. Thanks!

tritemio:
Thanks for the quick merge!

malliwi88:
Hello boys @tritemio @minrk 

I have a question, could I use this functionality to prevent an error in a cell interrupt execution of my notebook, from the command line?

I am using the following statement:
``` bash
jupyter nbconvert --to notebook --ExecutePreprocessor.kernel_name=python2 --ExecutePreprocessor.enabled=True --ExecutePreprocessor.timeout=3000 --execute --inplace in_notebook.ipynb --output=out_notebook.ipynb
```
What should I add to prevent execution stop for an error in some cell?

Thanks, any help is appreciated.

PR #123, opened 2015-09-25
warn about missing IPython in ipython2python

### LINKED ISSUE #122: dependency on IPython in ipython2python
There is a dependence on `IPython.core` in [filters.strings](https://github.com/jupyter/nbconvert/blob/2ffe7cb2661532db40a8f463a61252e89d5ed973/nbconvert/filters/strings.py#L20]. This isn't listed in the requirements though. 

This means that I need to require ipython as well as nbconvert if my package only has a dependence on nbconvert.

minrk:
and leave as a no-op in its absence

closes #122

### DIFF
--- a/nbconvert/filters/strings.py
+++ b/nbconvert/filters/strings.py
@@ -11,13 +11,14 @@ templates.
 import os
 import re
 import textwrap
+import warnings
+
 try:
     from urllib.parse import quote  # Py 3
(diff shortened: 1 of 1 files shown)


aaren:
Looks good to me.

damianavila:
LGTM as well

PR #131, opened 2015-10-08
Get Inkscape path from Windows registry

juhasch:
Otherwise Inkscape needs to be manually added to the path variable.

### DIFF
--- a/nbconvert/preprocessors/svg2pdf.py
+++ b/nbconvert/preprocessors/svg2pdf.py
@@ -20,6 +20,12 @@ from .convertfigures import ConvertFiguresPreprocessor
 
 INKSCAPE_APP = '/Applications/Inkscape.app/Contents/Resources/bin/inkscape'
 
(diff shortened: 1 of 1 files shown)


nbconvert/preprocessors/svg2pdf.py
```diff
@@ -20,6 +20,12 @@
 
 INKSCAPE_APP = '/Applications/Inkscape.app/Contents/Resources/bin/inkscape'
 
+if sys.platform == "win32":
+    if sys.version_info < (2, 99):
```
minrk: Probably best to use `try import/except import` instead of targeting a specific version

minrk:
Minor note in how to handle version checking, but otherwise +1

juhasch:
Now using try/except to import winreg module

meeseeksmachine:

https://discourse.jupyter.org/t/inkscape-not-linking-for-pdf-export/4292/1

PR #132, opened 2015-10-08
Run latex in current directory

### LINKED ISSUE #136: Using relative image paths breaks PDF generation
When using a recent version of Jupyter with the Python2 kernel, I am running into an issue savings PDFs for notebooks using relative image filepaths.

Using the relative path causes the image to render fine in the notebook, but when going to export to PDF, the following error is encountered:

```
nbconvert failed: PDF creating failed
```

And on the shell:

```
    ...
    LaTeX Warning: File `image/matplotlib_gallery_small.png' not found on input lin
    e 1116.


    ! Package pdftex.def Error: File `image/matplotlib_gallery_small.png' not found
    .

    See the pdftex.def package documentation for explanation.
     ...                                              

    l.1116 ...hics{image/matplotlib_gallery_small.png}

    ? 
    ! Emergency stop.
     ...                                              

    l.1116 ...hics{image/matplotlib_gallery_small.png}

    !  ==> Fatal error occurred, no output PDF file produced!
```

Changing to absolute filepaths on the other hand breaks the image rendering in the notebook, but causes the PDF to work again.

**System info**
- Python 3.5.0 / Python 2.7.10
- Jupyter 4.0.6
- IPython 4.0.0
-  pdfTeX, Version 3.14159265-2.6-1.40.16

juhasch:
Currently, `pdflatex` is executed in some random temporary directory. Including local files (images) doesn't work then.

A test for this would be good, but I have no idea how to check for the correct PDF result.
Would checking for an error (image found/not found) be sufficient ?

closes #136


minrk:
@juhasch I think that would be sufficient

juhasch:
Added test for pdf generation with locally linked image in markdown cell.

minrk:
@juhasch this one doesn't seem to be working. I believe I've found an alternate fix in #140, can you test that?

PR #137, opened 2015-10-12
add get_metadata filter

### LINKED ISSUE #75: Top-level shape metadata ignored for image output
Try viewing [this notebook](https://github.com/IRkernel/IRkernel/blob/master/Display.ipynb) on github/nbconvert vs as running notebook.

The last cell output should be 50×50 pixel big, but only is when viewed in a running notebook. The nbconvert (and therefore GitHub’s) rendering doesn’t respect that.

minrk:
for extracting output metadata properly

output may be in the top-level, or specified for a particular mimetype

closes #75


damianavila:
:+1:

PR #140, opened 2015-10-14
PDF: Look for tex inputs in notebook directory

minrk:
Restores running latex in a temp directory. This is an alternate fix for #132, which doesn't appear to work.

### DIFF
--- a/nbconvert/exporters/pdf.py
+++ b/nbconvert/exporters/pdf.py
@@ -8,8 +8,8 @@ import os
 import sys
 
 from ipython_genutils.py3compat import which
-from traitlets import Integer, List, Bool, Instance
-from ipython_genutils.tempdir import TemporaryDirectory
+from traitlets import Integer, List, Bool, Instance, Unicode
+from ipython_genutils.tempdir import TemporaryWorkingDirectory
 from .latex import LatexExporter
 
 
@@ -36,6 +36,7 @@ class PDFExporter(LatexExporter):
         help="File extensions of temp files to remove after running."
     )
     
+    texinputs = Unicode(help="texinputs dir. A notebook's directory is added")
     writer = Instance("nbconvert.writers.FilesWriter", args=())
 
     def run_command(self, command_list, filename, count, log_function):
@@ -74,11 +75,16 @@ class PDFExporter(LatexExporter):
         
         times = 'time' if count == 1 else 'times'
         self.log.info("Running %s %i %s: %s", command_list[0], count, times, command)
+        
+        shell = (sys.platform == 'win32')
+        if shell:
+            command = subprocess.list2cmdline(command)
+        env = os.environ.copy()
+        env['TEXINPUTS'] = os.pathsep.join([self.texinputs, env.get('TEXINPUTS', '')])
         with open(os.devnull, 'rb') as null:
             stdout = subprocess.PIPE if not self.verbose else None
             for index in range(count):
-                shell = (sys.platform == 'win32')
-                p = subprocess.Popen(subprocess.list2cmdline(command) if shell else command, stdout=stdout, stdin=null, shell=shell)
+                p = subprocess.Popen(command, stdout=stdout, stdin=null, shell=shell, env=env)
                 out, err = p.communicate()
                 if p.returncode:
                     if self.verbose:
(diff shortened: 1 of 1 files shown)


juhasch:
Sorry, can't test this. `p.communicate()` crashes on my Windows 7 installation.
Using the `TEXINPUTS` environment variable sounds reasonable.

nbconvert/exporters/pdf.py
```diff
@@ -74,11 +75,16 @@ def run_command(self, command_list, filename, count, log_function):
         
         times = 'time' if count == 1 else 'times'
         self.log.info("Running %s %i %s: %s", command_list[0], count, times, command)
+        
+        shell = (sys.platform == 'win32')
+        if shell:
+            command = subprocess.list2cmdline(command)
+        env = os.environ.copy()
+        env['TEXINPUTS'] = ':'.join([self.texinputs, env.get('TEXINPUTS', '')])
```
juhasch: This gives me `i:\\notebook:` instead of `i:\\notebook` as `TEXINPUTS` env. The trailing `:` makes the compilation fail.

minrk: Ah, of course it should be os.pathsep. Hopefully that works, now.

tsailiming:
This fix doesn't seem to work for me. Running on Mac and using `nbconvert==4.1.0`

````
l.298 \includegraphics{images/grMHW4F.png}
                                          
? 
! Emergency stop.
````

**Comment by seltzered** [comment]
I have a similar error. (environment: osx el capitan, nbconvert 4.1.0, notebook 4.1.0 , mactex installed homebrew cask package)

I have a python notebook and a directory called report_images, with the line`` which renders the image in the python notebook but has the following error when attempting to export to pdf.

    ! Package pdftex.def Error: File `./report_images/myImage.png' not fou
    nd.
    
    See the pdftex.def package documentation for explanation.
     ...                                              
                                                      
    l.293 ...ics{./report_images/myImage.png}
                                                      
    ? 
    ! Emergency stop.

Attempted workarounds: 
- Made sure file extension actually had .png at the end, removed any whitespace characters
- Tried using an absolute path but this wouldn't even render in the notebook - seems the notebook assumes a path based on either the directory notebook was launched or the ipynb's path

**Comment by seltzered** [comment]
I spent some a little time debugging this today - It seems like in my scenario the path key isn't in my resources dictionary, which causes pdf.py to default to the cwd of where I launched the notebook server.

I tried added some print statements pdf.py to print out the path info / resource dict:

```
    def from_notebook_node(self, nb, resources=None, **kw):
        latex, resources = super(PDFExporter, self).from_notebook_node(
            nb, resources=resources, **kw
        )
        print("texinputs: resources: ", str(resources))
        # set texinputs directory, so that local files will be found
        if resources and resources.get('metadata', {}).get('path'):
            self.texinputs = resources['metadata']['path']
            print("texinputs: using resources path")
        else:
            self.texinputs = os.getcwd()
            print("texinputs: using cwd")

        print("texinputs: " + self.texinputs)
```

This gave me the following: 
```
...
texinputs: resources:  defaultdict(None, {'outputs': {}, 'latex': {'....... lots of things I'm not going to share here, but ctrl+f'ing for "path" shows that it's not here}
texinputs: using cwd
texinputs: /Users/vivek
[I 18:01:10.556 NotebookApp] Writing 38978 bytes to notebook.tex
[I 18:01:10.557 NotebookApp] Building PDF
...

```

So, there's two temporary workarounds for those needing images exported right now - you can either launch `jupyter notebook` from the same directory of the ipynb document you're launching, or temporarily copy your images to wherever you usually launch `jupyter notebook` at. For either case, you may need to play with removing the relative path period too.

newhobby2016:
nbconvert should be able to translate the relative path of image into an absolute path when generating  markdown file. Especially for large document project with notebook in different folders, it is natural to generate markdown in differet folders as well. But at end, all these markdown files in different folders  will be concatenated in the parent fold, and further transformed to latex for generating PDF. For this scenario, only absolute path works.

bidhya:
I have the same problem with the "Download as PDF" command. The workaround for me was to start Jupyter notebook from the folder where the images are located. It would be nice to have a feature that allows embedding images from anywhere in the computer.

I would say this is problem only with Markdown of Jupyter notebook. Using pandoc for example, I was able to convert the markdown files to PDF with no issue.

PR #141, opened 2015-10-14
[WIP] allow nbconvert template loading from other libraries

gonzolabs:
The goal here is to make it easier to use nbconvert with custom html/css templates.


Carreau:
I'm guiding @gonzolabs into contributing to nbconvert. 

This is working locally but WIP.

minrk:
Thanks, looking forward to it.

Carreau:
Would like to ping @damianavila and @bollwyvl on this.

bollwyvl:
Cool! A supported, well-documented way to do this would be great. :+1: 

`templatelib` is kind of a weird name. It makes me think that i can use some other template engine, which i don't think is the intent.

We're (going to soon be) doing something similar on the commit i just linked over on nbviewer... I do think `templatepath` is Good Enough for our purposes... but the package loader would have some benefits in terms of portability... but not as portable as entry_points, as was adopted over on [ipymd](https://github.com/rossant/ipymd/blob/master/setup.py#L79). Anyhow, for us, we would have to pop `__init__.py`s into all of our template dirs, and the resulting path would be `nbviewer.templates.nbconvert.slide_reveal` which is pretty nice.

Carreau:
Yeah, I need to dig a bit more into how Jinja could load template.
My goal would be to have a way to refer to as template as a variable in a Python package instead of an actual file. 

I'm teaching John how to use git and how nbconvert work at the same time, so i did not dived too much into the actual details. 

I honestly haven't touched nbconvert in a while, so am re-discovering all the internal at the same time, so any advice is greatly appreciated.

damianavila:
> Cool! A supported, well-documented way to do this would be great. :+1:

:+1:

bollwyvl:
Ah, that's not what packageloader does... It still expects template files,
but in directories with __init__.py files, you just avoid having to know
the physical path.

http://jinja.pocoo.org/docs/dev/api/#jinja2.FunctionLoader on the other
hand might do the trick. You'd have to do an __import__ of the callable,
which takes a template name and returns a template string (or none) and
create the loader with that.

Then you'd want to adopt some convention to do the naming, though if it's
installed after the base loader(s), those should always execute first...
This means you wouldn't be able to overload stock templates, don't know if
that is a requirement... Ordering of config is hard!

Carreau:
> http://jinja.pocoo.org/docs/dev/api/#jinja2.FunctionLoader on the other
> hand might do the trick. You'd have to do an __import__ of the callable,
> which takes a template name and returns a template string (or none) and
> create the loader with that.

Yeah, that's what I figured out. 



> Then you'd want to adopt some convention to do the naming, though if it's
> installed after the base loader(s), those should always execute first...
> This means you wouldn't be able to overload stock templates, don't know if
> that is a requirement... Ordering of config is hard!

I'm also thinking of having a DottedObjectName for exporter_format, that would allow to not only being able to select template, but potentially Pre/Post Processor. 

I'll play with @gonzolabs to see what his needs are, and potentially make this change as "Provisional" without guaranties of being supported. 

Note that the end goal is to be able to export in a format that would suit, science, nature, or other science Journal editors.

nbconvert/nbconvertapp.py
```diff
@@ -30,6 +31,36 @@
...
+        self.values = values
+        Enum.__init__(self, values=values, default_value=default_value, **metadata)
+
+
+    def validate(self, obj, value):
+        if '.' not in value:
+            CaselessStrEnum.validate(self, obj, value)
+            pass
+        else: 
+            DottedObjectName.validate(self, obj, value)
```
minrk: Should we be using Type instead of DottedObjectName? In most cases where we have DottedObjectName, Type is actually the right thing to use.

nbconvert/nbconvertapp.py
```diff
@@ -17,8 +17,9 @@
 from jupyter_core.application import JupyterApp, base_aliases, base_flags
 from traitlets.config import catch_config_error, Configurable
 from traitlets import (
-    Unicode, List, Instance, DottedObjectName, Type, CaselessStrEnum, Bool,
+    Unicode, List, Instance, DottedObjectName, Type, CaselessStrEnum, Bool, Undefined, Enum
```
Carreau: @gonzolabs Type is already imported here :-P

gonzolabs: Ah right of course! 



--  
John Bohannon, PhD
Contributing Correspondent, Science Magazine
johnbohannon.org <http://www.johnbohannon.org/>

nbconvert/nbconvertapp.py
Carreau: IIRC we did not use `Type` because `Type` imports the module, and that would make us have to deal with 'isInstance' everywhere, in particular validate will imediately complain that `module` object is not iterable.  Dotted are always strings, and for sanity, I would prefer the trait to stay a string (at it was in 4.0), that to deal with an object which type can vary. 

Though, I agree that a subtype of `Type` that alias a few enums to Type would likely be more appropriate in the long run.

Carreau:
Also you committed a `Untilted1.ipynb`, can you remove it.

Carreau:
> Also you committed a Untilted1.ipynb, can you remove it.

Thanks !

nbconvert/nbconvertapp.py
```diff
@@ -30,6 +31,36 @@
 #Classes and functions
 #-----------------------------------------------------------------------------
 
+class CaselessStrEnumOrDottedObjectName(CaselessStrEnum, DottedObjectName):
```
takluyver: I think rather than defining a fiddly new traitlet type to hold this, I'd just drop it down to a `Unicode` trait, and leave checking whether it refers to a real exporter until we try to use it.

docs/source/external_exporters.rst
```diff
@@ -0,0 +1,156 @@
...
+A few built-in formats are available by default: `html`, `pdf`,
+`script`, `latex`. Each of these has its own _exporter_ with many configuration
+options that can be extended. Having the option to point to a different _exporter_ 
+allows authors to create their own fully customized templates. 
+
+A custom _exporter_ must be a globally importable python object. We recommend that
+these be distributed as python libraries.
+
+To export using an exporter from an external library, use the `fully qualified`
+name of this exporter on the command line as the argument of the ``--to`` flag::
```
takluyver: Do we want to do this, or use [entry points](http://entrypoints.readthedocs.org/en/latest/), with a group name like `nbconvert.exporters`? I think I'd lean towards entry points, because then it's trivial to list all the available exporters you have installed.

ellisonbg: I like the idea of entry points, but in the past it seemed like it was too
setuptools centric. That is why the new nbextension stuff didn't use them.

On Wed, Jan 6, 2016 at 9:22 AM, Thomas Kluyver <notifications@github.com>
wrote:

> In docs/source/external_exporters.rst
> <https://github.com/jupyter/nbconvert/pull/141#discussion_r48983354>:
>
> > +
> > +  $ jupyter nbconvert --to FORMAT notebook.ipynb
> > +
> > +This will convert the Jupyter document file ``notebook.ipynb`` into the output
> > +format designated by the ``FORMAT`` string as explained below.
> > +
> > +A few built-in formats are available by default: `html`, `pdf`,
> > +`script`, `latex`. Each of these has its own _exporter_ with many configuration
> > +options that can be extended. Having the option to point to a different _exporter_
> > +allows authors to create their own fully customized templates.
> > +
> > +A custom _exporter_ must be a globally importable python object. We recommend that
> > +these be distributed as python libraries.
> > +
> > +To export using an exporter from an external library, use the `fully qualified`
> > +name of this exporter on the command line as the argument of the ``--to`` flag::
>
> Do we want to do this, or use entry points
> <http://entrypoints.readthedocs.org/en/latest/>, with a group name like
> nbconvert.exporters? I think I'd lean towards entry points, because then
> it's trivial to list all the available exporters you have installed.
>
> —
> <https://github.com/jupyter/nbconvert/pull/141/files#r48983354>.
>



-- 
Brian E. Granger

takluyver: The [entrypoints](https://pypi.python.org/pypi/entrypoints) module is my attempt to free the technology from setuptools - it's a completely standalone module that can find and load entry points, so we wouldn't need  nbconvert to depend on setuptools. It's not battle-tested yet, but I think it's simple enough that we can iron out any bugs that come up.

Third parties who wanted to offer nbconvert exporters would need to either use setuptools (which almost all Python packages now do anyway), or package with something like flit.

ellisonbg: Ohh, cool. Some questions:

* Does it find/load regular setuptools entry points?
* How does one declare the entrypoints?
* Who needs to install it? Consumers of entrypoints? Providers of
entrypoints? Both?
* For us to rely on something like this in our core stack, I think it would
need to be pretty stable and battle tested. How far aware is it from that?

Because this type of thing is starting to come up in many different parts
of our projects, I think we should figure out a robust consistent approach
that all our projects can use. Having different projects use different
approaches will only confuse users. Maybe we want to bring this up on the
list or as an enchancemnt proposal?

Just to clarify, I have "re-invented" entry points in this PR as well:

https://github.com/jupyter/notebook/pull/879

which is targeted for a quick 4.2 release of jupyter/notebook so it would
be good to figure this stuff out on a pretty short time frame.




On Wed, Jan 6, 2016 at 9:57 AM, Thomas Kluyver <notifications@github.com>
wrote:

> In docs/source/external_exporters.rst
> <https://github.com/jupyter/nbconvert/pull/141#discussion_r48987153>:
>
> > +
> > +  $ jupyter nbconvert --to FORMAT notebook.ipynb
> > +
> > +This will convert the Jupyter document file ``notebook.ipynb`` into the output
> > +format designated by the ``FORMAT`` string as explained below.
> > +
> > +A few built-in formats are available by default: `html`, `pdf`,
> > +`script`, `latex`. Each of these has its own _exporter_ with many configuration
> > +options that can be extended. Having the option to point to a different _exporter_
> > +allows authors to create their own fully customized templates.
> > +
> > +A custom _exporter_ must be a globally importable python object. We recommend that
> > +these be distributed as python libraries.
> > +
> > +To export using an exporter from an external library, use the `fully qualified`
> > +name of this exporter on the command line as the argument of the ``--to`` flag::
>
> The entrypoints <https://pypi.python.org/pypi/entrypoints> module is my
> attempt to free the technology from setuptools - it's a completely
> standalone module that can find and load entry points, so we wouldn't need
> nbconvert to depend on setuptools. It's not battle-tested yet, but I think
> it's simple enough that we can iron out any bugs that come up.
>
> Third parties who wanted to offer nbconvert exporters would need to either
> use setuptools (which almost all Python packages now do anyway), or package
> with something like flit.
>
> —
> <https://github.com/jupyter/nbconvert/pull/141/files#r48987153>.
>



-- 
Brian E. Granger

takluyver: > Does it find/load regular setuptools entry points?

Yes.

> How does one declare the entrypoints?

Providers of entry points declare them in their packaging, using setuptools or flit (or other future packaging tools).

> Who needs to install it? Consumers of entrypoints? Providers of entrypoints? Both?

The entrypoints module is for consumers of entrypoints. If we used it like this, nbconvert would add a dependency on entrypoints.

> For us to rely on something like this in our core stack, I think it would need to be pretty stable and battle tested. How far aware is it from that?

I don't know if anyone has used it in earnest yet, but it's [documented](http://entrypoints.readthedocs.org/en/latest/) and [tested](https://travis-ci.org/takluyver/entrypoints). It's also only about 200 lines of code, so I'm pretty confident that we can quickly fix any problems we come across.

Carreau: I would be fine using entry points, though the approach with DottedObjectNames has IMHO a few advantages:
 - you don't require to install a package you just need the thing to be importable, so it can be a local file to CWD.
 - and understand how to use entrypoint, which mean you don't have to understand how to package something.
 - it forces namespaces. Otherwise N people will try to grab the `--to docx` or others.
 - it does not allow dynamic generation of exporters. 

So in general I would lean toward keep this as is (full qualified named as argument) **for now**, and add ability to register entrypoints on top of that.

So basically provide shorthands as entrypoints (docx, rst, md, py, script, html, slides...) and also allow fully qualified names. Which would be easy to distinguish because full qualified names have dots in their names.

nbconvert/nbconvertapp.py
Carreau: -1 on that because right now the `Enum` thing  for the error message if you give a wrong value:

```
...must be any of ['custom', 'html', 'latex', 'markdown', 'notebook', 'pdf', 'python', 'rst', 'script', 'slides'], but a value of 'tex' <class 'str'> was specified.
```

which goes away for unicode.

takluyver: You can provide a nicer error message in the code below where we're setting up `self.exporter`, rather than hammering traitlets into giving a roughly right but rather ugly error.

Carreau: I'm not 100% sure of the simplicity/ugliness tradeof. Doing so also remove the automatic listing of known format from out `--help` (yes we can still do it manually) and move the check of wether an exporter format is right much later in the code. Plus that would duplicate quite a bit of logic already present. 

If you feel like `CaselessStrEnumOrDottedObjectName` is a too generic name, we can rename to `ExporterTrait` or something shorter. 

We can try though.

Carreau: See #185 not as horrible as I thought, but not a huge fan either.

Carreau:
Merged as #185 !

PR #150, opened 2015-10-30
Make sure env value is always str

scottfitzenrider:
subprocess.Popen will fail if env contains anything other than str, including unicode


Carreau:
i'm not sure the fix is `str()` in this case, it might be `.encode()`

Carreau:
And I'm curious how you got bytes here, ant calling `str()` and bytes here will definitively do the wrong thing in some case:

```
>>> s = b'bytes'
>>> str(s)
"b'bytes'"
```

what was your value of `TEXINPUTS` ? Python2 or Python3 ?

minrk:
I think the fix is `py3compat.cast_bytes_py2`.

PR #151, opened 2015-10-31
test on python 3.5

Carreau:


minrk:
Merged identical #153 instead

Carreau:
Rha did it on too many repo at the same time from github. ight have had nbconvert tab openend twice.

PR #154, opened 2015-11-04
--inplace should actually run in-place

minrk:
empty build_directory is meant to indicate that exported files are in the same directory as their source.

Use path resource if build_directory is unspecified.

### DIFF
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -72,6 +72,7 @@ nbconvert_aliases.update({
     'writer' : 'NbConvertApp.writer_class',
     'post': 'NbConvertApp.postprocessor_class',
     'output': 'NbConvertApp.output_base',
+    'output-dir': 'FilesWriter.build_directory',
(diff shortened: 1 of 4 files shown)


nbconvert/nbconvertapp.py
```diff
@@ -51,6 +51,7 @@ def validate(self, obj, value):
     'writer' : 'NbConvertApp.writer_class',
     'post': 'NbConvertApp.postprocessor_class',
     'output': 'NbConvertApp.output_base',
+    'output-dir': 'FilesWriter.output_directory',
```
jdfreder: I think this line is what is causing Travis to fail

jdfreder:
:+1:

dhimmel:
Until this fix arrives, is the suggested workaround to specify `--FilesWriter.build_directory`?

For example I would use:

```
jupyter nbconvert --execute path/to/file.ipynb --inplace --FilesWriter.build_directory path/to
```

stonefly:
i think it breaks the TemporaryWorkingDirectory in from_notebook_node in the pdf.py.

PR #162, opened 2015-11-26
run tests with stable ipython/ipykernel

minrk:


minrk:
Merging so other tests can pass

Issue #164, opened 2015-11-29
Allow nbconvert to read from stdin


akhmerov:
This would allow using it directly as a git filter. Specifically, the `ClearOutputPreprocessor` and `ExecutePreprocessor` are very useful as filters, but one can think of many more examples.

Carreau:
Marking as Sprint friendly.

It might require some refactor though. 

Ping @parleur

minrk:
closed by #173

PR #166, opened 2015-11-30
Allow --stdin for nbconvert

parleur:
allow to use nbconvert from stdin with option '--stdin' (issue #164)
cc @akhmerov

### DIFF
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -13,7 +13,6 @@ import logging
 import sys
 import os
 import glob
-import uuid
 
 from jupyter_core.application import JupyterApp, base_aliases, base_flags
 from traitlets.config import catch_config_error, Configurable
@@ -317,7 +316,7 @@ class NbConvertApp(JupyterApp):
 
         # Get a unique key for the notebook and set it in the resources object.
         if not notebook_filename:
-            notebook_name = '%s' % uuid.uuid4()
+            notebook_name = 'notebook'
         else:
             basename = os.path.basename(notebook_filename)
             notebook_name = basename[:basename.rfind('.')]
@@ -354,8 +353,8 @@ class NbConvertApp(JupyterApp):
             else:
                 output, resources = self.exporter.from_filename(shell_input, resources=resources)
         except ConversionException:
-            self.log.error("Error while converting '%s'", notebook_filename, exc_info=True)
-            self.exit(1)
+            self.log.error("Error while converting notebook", exc_info=True)
+            self.exit('Could not convert notebook.')
 
         return output, resources
 
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -71,7 +71,11 @@ nbconvert_flags.update({
          "is only relevant if '--execute' was specified, too.")
         ),
     'stdin' : (
-        {'NbConvertApp' : {'from_stdin' : True}},
+        {'NbConvertApp' : {
+            'from_stdin' : True,
+            'writer_class':'StdoutWriter'
+            }
+        },
         "read a single notebook file from stdin, and write the converted result to stdout. Implies `--stdout`"
         ),
     'stdout' : (
@@ -192,9 +196,6 @@ class NbConvertApp(JupyterApp):
             new = self.writer_aliases[new.lower()]
         self.writer_factory = import_item(new)
 
-    def _from_stdin_changed(self, name, old, new):
-        if new == True:
-            self.writer_class = "StdoutWriter"
 
     # Post-processor specific variables
     postprocessor = Instance('nbconvert.postprocessors.base.PostProcessorBase',
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -13,6 +13,7 @@ import logging
 import sys
 import os
 import glob
+import uuid
 
 from jupyter_core.application import JupyterApp, base_aliases, base_flags
 from traitlets.config import catch_config_error, Configurable
@@ -68,6 +69,10 @@ nbconvert_flags.update({
          "an error and include the error message in the cell output "
          "(the default behaviour is to abort conversion). This flag "
          "is only relevant if '--execute' was specified, too.")
+        ),
+    'stdin' : (
+        {'NbConvertApp' : {'from_stdin' : True}},
+        "read a single notebook file from stdin, and write the converted result to stdout. Implies `--stdout`"
         ),
     'stdout' : (
         {'NbConvertApp' : {'writer_class' : "StdoutWriter"}},
@@ -176,6 +181,7 @@ class NbConvertApp(JupyterApp):
     writer_class = DottedObjectName('FilesWriter', config=True,
                                     help="""Writer class used to write the
                                     results of the conversion""")
+    from_stdin   = Bool(False, config=True, help="read a single notebook from stdin.")
     writer_aliases = {'fileswriter': 'nbconvert.writers.files.FilesWriter',
                       'debugwriter': 'nbconvert.writers.debug.DebugWriter',
                       'stdoutwriter': 'nbconvert.writers.stdout.StdoutWriter'}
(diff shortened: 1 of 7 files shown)


nbconvert/nbconvertapp.py
```diff
@@ -398,18 +422,21 @@ def convert_notebooks(self):
...
             )
             self.exit(1)
-        
+
         # initialize the exporter
         self.exporter = exporter_map[self.export_format](config=self.config)
 
         # no notebooks to convert!
-        if len(self.notebooks) == 0:
+        if len(self.notebooks) and not self.from_stdin == 0:
```
Carreau: I think you ment to keep the `==0` earlier.

nbconvert/nbconvertapp.py
```diff
@@ -300,18 +310,24 @@ def init_single_notebook_resources(self, notebook_filename):
...
-        if self.output_base:
-            # strip duplicate extension from output_base, to avoid Basname.ext.ext
-            if getattr(self.exporter, 'file_extension', False):
-                base, ext = os.path.splitext(self.output_base)
-                if ext == self.exporter.file_extension:
-                    self.output_base = base
-            notebook_name = self.output_base
+        if not notebook_filename:
+            notebook_name = '%s' % uuid.uuid4()
+
```
Carreau: You can remove this blank line.

nbconvert/nbconvertapp.py
```diff
@@ -300,18 +310,24 @@ def init_single_notebook_resources(self, notebook_filename):
             - output_files_dir: a directory where output files (not including
               the notebook itself) should be saved
 
+        If filename is none (converting notebook from in memory), fill in the required
```
Carreau: `None` Upper case.

nbconvert/nbconvertapp.py
```diff
@@ -378,9 +398,13 @@ def convert_single_notebook(self, notebook_filename):
             4. (Maybe) postprocess the written file
 
         """
-        self.log.info("Converting notebook %s to %s", notebook_filename, self.export_format)
-        resources = self.init_single_notebook_resources(notebook_filename)
-        output, resources = self.export_single_notebook(notebook_filename, resources)
+        if self.from_stdin:
+            self.log.info("Converting notebook from stdin to stdout", shell_input, self.export_format)
```
Carreau: the `shell_input` here is not needed, and the `self.export_format` here is not used, you might want something like:

`"Converting notebook read from stdin to %s on  stdout", self._export_format"` , A native english Speaker, might find a better sentence.

nbconvert/nbconvertapp.py
```diff
@@ -323,16 +339,20 @@ def init_single_notebook_resources(self, notebook_filename):
...
 
-    def export_single_notebook(self, notebook_filename, resources):
+    def export_single_notebook(self, shell_input, resources):
         """Step 2: Export the notebook
 
         Exports the notebook to a particular format according to the specified
         exporter. This function returns the output and (possibly modified)
         resources from the exporter.
 
+        shell_input: a filename or buffer from stdin depending on the '--stdin' option
```
Carreau: the butter does not need to be from stdin, so you can say "for example from stdin when the the '--stdin' is used", 
but as a API, you do not **need** to be stdin.

nbconvert/nbconvertapp.py
```diff
@@ -390,7 +414,7 @@ def convert_notebooks(self):
         """
         # check that the output base isn't specified if there is more than
         # one notebook to convert
-        if self.output_base != '' and len(self.notebooks) > 1:
```
Carreau: I don't think you should remove this :-P

akhmerov:
Hi @parleur, awesome, thanks! One remark: you don't seem to have tests, and I found testing for unicode input particularly important.

nbconvert/nbconvertapp.py
```diff
@@ -69,6 +70,10 @@ def validate(self, obj, value):
          "(the default behaviour is to abort conversion). This flag "
          "is only relevant if '--execute' was specified, too.")
         ),
+    'stdin' : (
+        {'NbConvertApp' : {'from_stdin' : True}},
```
minrk: This can include `'writer_class': 'StdoutWriter'`, removing the need for `_from_stdin_changed` below.

nbconvert/nbconvertapp.py
```diff
@@ -300,18 +310,23 @@ def init_single_notebook_resources(self, notebook_filename):
...
-        notebook_name = basename[:basename.rfind('.')]
-        if self.output_base:
-            # strip duplicate extension from output_base, to avoid Basname.ext.ext
-            if getattr(self.exporter, 'file_extension', False):
-                base, ext = os.path.splitext(self.output_base)
-                if ext == self.exporter.file_extension:
-                    self.output_base = base
-            notebook_name = self.output_base
+        if not notebook_filename:
+            notebook_name = '%s' % uuid.uuid4()
```
minrk: I think this can be 'notebook' instead of a UUID.

nbconvert/nbconvertapp.py
parleur: I agree.

nbconvert/nbconvertapp.py
Carreau: Done.

parleur:
Thanks for yours comments @akhmerov. Test have been added (with help of @Carreau).

nbconvert/nbconvertapp.py
```diff
@@ -300,18 +310,23 @@ def init_single_notebook_resources(self, notebook_filename):
             - output_files_dir: a directory where output files (not including
               the notebook itself) should be saved
 
+        If filename is None (converting notebook from in memory), fill in the required
+        parameters with UUID.
```
minrk: Comment about UUID is no longer accurate

akhmerov:
OK, my knowledge of codebase is nearing its limits, but a few remarks:
* I'm not sure what are the dev's traditions, but you introduce several whitespace changes outside of where you made the edits; it might be better to revert those edits (with `git checkout -p` or so).
* While it's probably not very relevant to convert from `stdin` to a file, you're tying two unrelated options together — source of the notebook and output of the notebook. It might be cleaner to not do so (otoh it makes the interface more complex).
* Another option to consider, probably also a wrong one, is to treat lack of notebooks to convert as a signal to read from stdin. This would mimic behavior of other terminal tools, like `cat`. The downside is, of course, that the change from current behavior may be confusing.
* I couldn't help noticing that most of the changes consist of adding the same switch: `from_file` vs `from_filename`. It would be cool if it was possible to reduce duplication by first converting any `filename` into a `file`, but it seems to break too much of API. OTOH, you're changing the call signature already now by renaming an argument.
* Finally, I'm not sure what should be done with the changelog entry; the contribution guide is still written with a single ipython repo in mind.

parleur:
Hi Anton, thanks for your remarks. Trying to respond step by step.

- I recently switched to atom ... so the remaining whitespase ... My editor is fixed now, I agree with you and will remove the whitespace changes
- A use case for your second point would be ``` nbsomething untitled.ipynb --stdout | nbconvert --stdin my_shiny_document.ipynb ```. I don't think nbconvert have been design in this spirit but this kind of feature is just great ( see pandoc interface ...) I 'll try to do do that.
- I 'll not do that as it is completely not compatible with previous feature and I prefers it.
- In the principle you are right ... I ll have a eye on it.

Carreau:
> While it's probably not very relevant to convert from stdin to a file, you're tying two unrelated options together — source of the notebook and output of the notebook. It might be cleaner to not do so (otoh it makes the interface more complex).

I asked pierre to do that, there are a few assumption when you from a file about a filename and where to put the resources files (like extracted images). These assumption break down when using stdin, and --stdout is probably the only thing right now that can handle that. Though we **could** handle that.


> Another option to consider, probably also a wrong one, is to treat lack of notebooks to convert as a signal to read from stdin.

I think we tend to print help in this case, which seem to be friendlier than hanging on the command line for new comer.  

> OTOH, you're changing the call signature already now by renaming an argument.

Good point, I missed that I tend to treat args as positional only. 

> Finally, I'm not sure what should be done with the changelog entry; the contribution guide is still written with a single ipython repo in mind.

We're bad a change log in general, but true.

nbconvert/nbconvertapp.py
```diff
@@ -323,19 +338,23 @@ def init_single_notebook_resources(self, notebook_filename):
 
         return resources
 
-    def export_single_notebook(self, notebook_filename, resources):
+    def export_single_notebook(self, shell_input, resources):
```
Carreau: can you rename `shell_input` here back to `notebook_filename` (also in docstring)

You can assign after but the comment than changing the signature break api is right.

takluyver:
Closing in favour of #173, where development has continued.

PR #171, opened 2015-12-08
allow custom exporters to be importable classes

minrk:
rather than only aliases


takluyver:
This came up in discussions at CERN. I think the fix looks good, just waiting for Travis to finish.

PR #56, opened 2015-07-11
ENH: Add SanitizeHTML preprocessor.

ssanderson:
Uses [bleach](https://github.com/jsocol/bleach) as the underlying engine for sanitization.

PR #172, opened 2015-12-09
ENH: Add SanitizeHTML preprocessor. 

yuvipanda:
Uses bleach as the underlying sanitizing engine.

Based off of https://github.com/jupyter/nbconvert/pull/56

### DIFF
--- a/nbconvert/preprocessors/sanitize.py
+++ b/nbconvert/preprocessors/sanitize.py
@@ -29,7 +29,7 @@ class SanitizeHTML(Preprocessor):
     tags = List(
         Unicode,
         config=True,
-        default=ALLOWED_TAGS,
+        default_value=ALLOWED_TAGS,
         help="List of HTML tags to allow",
     )
     styles = List(
--- a/nbconvert/preprocessors/sanitize.py
+++ b/nbconvert/preprocessors/sanitize.py
@@ -77,13 +77,17 @@ class SanitizeHTML(Preprocessor):
 
         Cell Types:
           raw:
-            Return unchanged
+            Sanitize literal HTML
           markdown:
             Sanitize literal HTML
           code:
             Sanitize outputs that could result in code execution
         """
         if cell.cell_type == 'raw':
+            # Sanitize all raw cells anyway.
+            # Only ones with the text/html mimetype should be emitted
+            # but erring on the side of safety maybe.
+            cell.source = self.sanitize_html_tags(cell.source)
             return cell, resources
         elif cell.cell_type == 'markdown':
             cell.source = self.sanitize_html_tags(cell.source)
--- a/nbconvert/preprocessors/sanitize.py
+++ b/nbconvert/preprocessors/sanitize.py
@@ -8,14 +8,14 @@ from bleach import (
     ALLOWED_TAGS,
     clean,
 )
-from IPython.utils.traitlets import (
+from traitlets import (
     Any,
     Bool,
     List,
     Set,
     Unicode,
 )
-from IPython.nbconvert.preprocessors.base import Preprocessor
+from .base import Preprocessor
 
 
 class SanitizeHTML(Preprocessor):
--- a/nbconvert/preprocessors/sanitize.py
+++ b/nbconvert/preprocessors/sanitize.py
@@ -1,0 +1,134 @@
+"""
+NBConvert Preprocessor for sanitizing HTML rendering of notebooks.
+"""
+
+from bleach import (
+    ALLOWED_ATTRIBUTES,
+    ALLOWED_STYLES,
+    ALLOWED_TAGS,
+    clean,
+)
+from IPython.utils.traitlets import (
+    Any,
+    Bool,
+    List,
+    Set,
+    Unicode,
+)
+from IPython.nbconvert.preprocessors.base import Preprocessor
+
+
+class SanitizeHTML(Preprocessor):
+
+    # Bleach config.
+    attributes = Any(
+        config=True,
+        default_value=ALLOWED_ATTRIBUTES,
+        help="Allowed HTML tag attributes",
+    )
+    tags = List(
+        Unicode,
+        config=True,
+        default=ALLOWED_TAGS,
+        help="List of HTML tags to allow",
+    )
+    styles = List(
+        Unicode,
+        config=True,
+        default_value=ALLOWED_STYLES,
+        help="Allowed CSS styles if <style> tag is whitelisted"
+    )
+    strip = Bool(
+        config=True,
+        default_value=False,
+        help="If True, remove unsafe markup entirely instead of escaping"
+    )
+    strip_comments = Bool(
+        config=True,
+        default_value=True,
+        help="If True, strip comments from escaped HTML",
+    )
+
+    # Display data config.
+    safe_output_keys = Set(
+        config=True,
+        default_value={
+            'metadata',  # Not a mimetype per-se, but expected and safe.
+            'text/plain',
+            'text/latex',
+            'application/json',
+            'image/png',
+            'image/jpeg',
+        },
+        help="Cell output mimetypes to render without modification",
+    )
+    sanitized_output_types = Set(
+        config=True,
+        default_value={
+            'text/html',
+            'text/markdown',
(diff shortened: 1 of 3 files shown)


yuvipanda:
I'll probably add more tests tomorrow. Need to add tests to make sure that:

1. Cells that should just pass through, pass through
2. Test that all cells that need preprocessing are preprocessed (rather than just one as is right now)
3. Test that output cells are sanitized properly.

yuvipanda:
@minrk I think this is complete-ish now. I have a few questions:

# Should 'raw' cells be sanitized?
# Should 'stdout/stderr' cells be sanitized?

My gut feel is to try to break the sanitization by attempting to inject JS in different ways (aka when the attacker controls the input nb file), and make sure that all of those fail. If someone has disallowed `text/javascript` and `text/css`, all other cells shouldn't let anything through if sanitizer preprocessor is turned on.

Does that sound right or am I missing something fundamental? (/me wears newbie hat)

minrk:
> Should 'raw' cells be sanitized?

I think so, yes. Since raw cells are meant to produce 'output' (i.e. HTML) unmodified, then they should be treated the same as an HTML output.

> Should 'stdout/stderr' cells be sanitized?

We should rely on the existing mechanisms for stdout/stderr to be safe. If they aren't, there's a problem.

ssanderson:
> Should 'raw' cells be sanitized?

Yep.  Not sanitizing these was a misreading of the purpose of `raw` cells in my original design (thanks @yuvipanda for picking up the slack here btw).  I fixed this downstream by just calling `cgi.escape`/`html.escape` (that function moved in the py2 -> py3 transition) on the contents of raw cells, which has the effect of making them render as plain text.  That may or may not be the desired behavior here, but some escaping is definitely necessary.

minrk:
@ssanderson yup, escaping raw cells is definitely the wrong thing to do. They should be treated as literal HTML, and sanitized (or not) the same as HTML output.

ssanderson:
It's also worth noting that the escaping strategy here doesn't prevent users from passing markdown links to external SVGs, which are potentially dangerous because they can include arbitrary javascript.  

I'm not sure that there's a good solution to that problem that can be handled in the preprocessing layer.

takluyver:
Raw cells have a mimetype attached, IIRC? So we should be sanitising those marked as text/html (& js + css?), escaping text/plain, and skipping over anything else.

minrk:
@takluyver this is only for HTML output, so only HTML raw cells are considered (no mime-type specified is treated as HTML). If the mime-type is other than HTML, the cell is excluded altogether, since it's by definition for inclusion only in other nbconvert output types.

takluyver:
Great. Just checking since the discussion of raw cells so far didn't seem to have mentioned that.

yuvipanda:
I've updated this too.

yuvipanda:
@minrk can you point me to where the other stdout / stderr sanitization occurs?

yuvipanda:
Hmm, I'm sanitizing all raw cells - maybe I should just do the text/html ones and rely on other mechanisms for stripping everything else?

minrk:
Sorry, I might have given the wrong idea above - stdout/stderr shouldn't be sanitized. They should already be handled in a safe way. If they aren't, it's a bug.

yuvipanda:
Alright :) I shall await more scrunity on the rest of the patch.

minrk:
@ssanderson can I ask for your help on review, or are you busy?

takluyver:
To elaborate a bit (@yuvipanda asked for clarification on gitter):

In the HTML template, [stream outputs](https://github.com/jupyter/nbconvert/blob/41c232dbaf57d5b76afa0bc0d9e096d480c3e826/nbconvert/templates/html/basic.tpl#L94-L108) are passed through the [ansi2html filter](https://github.com/jupyter/nbconvert/blob/560a1fc8ead384e1953349f2300185ee7041882c/nbconvert/filters/ansi.py#L40), which escapes significant characters in the text using HTML entities, e.g. `< -> &lt;`. So it shouldn't be possible for arbitrary HTML to get on the page from a stream output.

ssanderson:
@minrk @yuvipanda the changes here look reasonable, but after thinking about this some more, I'm a bit nervous about including this in the main repository because of the SVG link sanitizing issue I outlined above.  The issue is that I can include a markdown image link that references an external image, and we can't know whether or not we should escape that link without fetching and examining the linked resource.  External images are a potential attack vector because SVGs can embed arbitrary Javascript.  

On Quantopian, where we have a version of this deployed, we take the extra precaution of serving rendered notebooks in an iframe on a domain separate from the main page.  Without that precaution I'd be nervous about presenting this as a strong security layer for sharing notebooks.

I suppose a conservative option for dealing with this would be to always remove markdown image links?

yuvipanda:
@ssanderson that's a good point. I'm not sure if we can do that easily. Perhaps one way is to make this a postprocessor maybe, so we can just deal with the final markdown output? Or I guess we can find a way to properly strip image links from markdown, and making that an option.

yuvipanda:
This should definitely be configurable, so people can do other rewrites if they want (we'd probably pass it through https://github.com/wikimedia/mediawiki/blob/master/includes/upload/UploadBase.php#L1288 and serve it off a different domain, for example)

yuvipanda:
I'm now not actually sure if the postprocessors as they currently exist can fit that use case. Would we have to parse the markdown, strip out images and spit it back?

Carreau:
What's the status of this ?

yuvipanda:
I basically agree with @ssanderson - we can't actually call this secure because of JS in SVG issues. And JS in SVG is a hardish problem to solve statically.

Our options now are:

1. Ship it as is, but with a big warning about SVG + JS
2. Add ability to postprocess the markdown produced HTML, and strip out images from that by default.

I'm not sure which one to do. I'm also not sure if other things can 'leak in' svg + js...

brion:
Can you clarify the SVG+JS issue? As I recall the gamut runs roughly:
* SVG loaded in an  element is strictly sandboxed and doesn't run scripts.
* SVG loaded in an <iframe> or <object> generally can run scripts, but in the scripting context of the site it comes from. This means cross-site embedding is protected just like an HTML document in an iframe.
* _Inline_ SVG embedded within an HTML document can run scripts just like HTML can.

michaelpacer:
Any further thoughts on this? I'm going to put it in wishlist for lack of consensus for a few months on next steps

michaelpacer:
So @yuvipanda and I were looking at this today, and what we found matches what @brion was saying. If we were to only allow  elements and disallow <embed> and <object> elements we shouldn't have to worry about any js injection issues. It also avoids needing to include any <iframe> elements, which avoid some of the rendering issues that have cropped up around that.

Any images defined using the `` syntax will already be interpreted as  elements so the challenge is going to arise in supporting arbitrary html in the markdown cells (which already is an issue when converting…see other issues that I'm having difficulty finding right now).

yuvipanda:
I think the things to do now are:

1. Resolve the merge conflict
2. Add more tests to make sure we don't allow object or embed tags
3. Merge?

@michaelpacer Arbitrary HTML in markdown should be taken care of by the markdown parser rather than at this layer, no?

yuvipanda:
@michaelpacer ok, I've fixed the merge conflict now!

yuvipanda:
If tests pass, I'll try to add a few more to catch the img / object / embed stuff later this week. Feel free to poke at it before that too if you wish :)

yuvipanda:
Added extra test! Ready for review and hopefully merge again :)

michaelpacer:
This looks really good! If you look at the code coverage you'll see that a couple of places aren't being hit by the tests.

The one I'd be worried about is the removing tags if they aren't safe or sanitised. People might end up confused if we don't pass back a message saying that that those are being removed, so if possible, making sure that that is hit by the test would be good. 

But other than that I think it should be good to go.

michaelpacer:
@Carreau @takluyver @minrk 

I just committed to this branch, so I'm no longer comfortable merging on my own. But if you want to merge, I think this is good to go.

minrk:
Thanks, @michaelpacer and @yuvipanda!

PR #173, opened 2015-12-09
Allow --stdin for nbconvert (alternative command line interface)

parleur:
Follow #166 and #164, @akhmerov @Carreau @minrk 
Here the command line interface is:

| nbconvert --stdin --to html 
write in "notebook.html" by default
| nbconvert --stdin --to html --stdout
write to stdout

The corner in this API is default basename I am forced to introduce. Public APIs are changed but not broken, just extended. Right now help, and docstring havn't been double cheked (coming ...).

### DIFF
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -335,7 +335,7 @@ class NbConvertApp(JupyterApp):
 
         return resources
 
-    def export_single_notebook(self, notebook_filename, resources):
+    def export_single_notebook(self, notebook_filename, resources, input_buffer=None):
         """Step 2: Export the notebook
 
         Exports the notebook to a particular format according to the specified
@@ -345,8 +345,8 @@ class NbConvertApp(JupyterApp):
         notebook_filename: a filename or buffer (for instance from stdin using '--stdin' option)
         """
         try:
-            if self.from_stdin:
-                output, resources = self.exporter.from_file(notebook_filename, resources=resources)
+            if input_buffer:
+                output, resources = self.exporter.from_file(input_buffer, resources=resources)
             else:
                 output, resources = self.exporter.from_filename(notebook_filename, resources=resources)
         except ConversionException:
@@ -385,22 +385,27 @@ class NbConvertApp(JupyterApp):
         if hasattr(self, 'postprocessor') and self.postprocessor:
             self.postprocessor(write_results)
 
-    def convert_single_notebook(self, notebook_filename):
+    def convert_single_notebook(self, notebook_filename, input_buffer=None):
         """Convert a single notebook. Performs the following steps:
 
             1. Initialize notebook resources
             2. Export the notebook to a particular format
             3. Write the exported notebook to file
             4. (Maybe) postprocess the written file
-
+            
+            If input_buffer is not None, convertion is done using buffer as source ignoring
+            the notebook_filename argument.
         """
-        if self.from_stdin:
-            self.log.info("Converting notebook from stdin into %s", self.export_format)
-            resources = self.init_single_notebook_resources(None)
-        else:
+        if not input_buffer:
             self.log.info("Converting notebook %s to %s", notebook_filename, self.export_format)
             resources = self.init_single_notebook_resources(notebook_filename)
-        output, resources = self.export_single_notebook(notebook_filename, resources)
+            output, resources = self.export_single_notebook(notebook_filename, resources)
+        else:
+            notebook_filename = "notebook"
+            self.log.info("Converting notebook into %s", self.export_format)
+            resources = self.init_single_notebook_resources(None) #init_single_notebook_resources need to know there is no file
+            output, resources = self.export_single_notebook(notebook_filename, resources, input_buffer=input_buffer)
+
         write_results = self.write_single_notebook(output, resources)
         self.postprocess_single_notebook(write_results)
 
@@ -428,11 +433,11 @@ class NbConvertApp(JupyterApp):
             sys.exit(-1)
 
         # convert each notebook
-        if not self.writer_class == 'StdoutWriter': #if stdout there is no file
+        if not self.from_stdin:
             for notebook_filename in self.notebooks:
                 self.convert_single_notebook(notebook_filename)
         else:
-            self.convert_single_notebook(sys.stdin)
+            self.convert_single_notebook(None, input_buffer=sys.stdin)
             
 #-----------------------------------------------------------------------------
 # Main entry point
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -71,7 +71,11 @@ nbconvert_flags.update({
          "is only relevant if '--execute' was specified, too.")
         ),
     'stdin' : (
-        {'NbConvertApp' : {'from_stdin' : True}},
+        {'NbConvertApp' : {
+            'from_stdin' : True,
+            'writer_class':'StdoutWriter'
+            }
+        },
         "read a single notebook file from stdin, and write the converted result to stdout. Implies `--stdout`"
         ),
     'stdout' : (
@@ -192,9 +196,6 @@ class NbConvertApp(JupyterApp):
             new = self.writer_aliases[new.lower()]
         self.writer_factory = import_item(new)
 
-    def _from_stdin_changed(self, name, old, new):
-        if new == True:
-            self.writer_class = "StdoutWriter"
 
     # Post-processor specific variables
     postprocessor = Instance('nbconvert.postprocessors.base.PostProcessorBase',
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -25,6 +25,7 @@ from .exporters.export import get_export_names, exporter_map
 from nbconvert import exporters, preprocessors, writers, postprocessors, __version__
 from .utils.base import NbConvertBase
 from .utils.exceptions import ConversionException
+from .utils.io import unicode_stdin_stream
 
 #-----------------------------------------------------------------------------
 #Classes and functions
@@ -430,8 +431,9 @@ class NbConvertApp(JupyterApp):
             for notebook_filename in self.notebooks:
                 self.convert_single_notebook(notebook_filename)
         else:
+            input_buffer = unicode_stdin_stream()
             # default name when conversion from stdin
-            self.convert_single_notebook("notebook.ipynb", input_buffer=sys.stdin)
+            self.convert_single_notebook("notebook.ipynb", input_buffer=input_buffer)
             
 #-----------------------------------------------------------------------------
 # Main entry point
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -72,7 +72,6 @@ nbconvert_flags.update({
     'stdin' : (
         {'NbConvertApp' : {
             'from_stdin' : True,
-            'writer_class':'StdoutWriter'
             }
         },
         "read a single notebook file from stdin, and write the converted result to stdout. Implies `--stdout`"
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -360,7 +360,7 @@ class NbConvertApp(JupyterApp):
         resources from the exporter.
 
         notebook_filename: a filename
-        input_buffer: a containing notebook buffer, if not None notebook_filename is ignored
+        input_buffer: a readable file like object returning unicode, if not None notebook_filename is ignored
         """
         try:
             if input_buffer is not None:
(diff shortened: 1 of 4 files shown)


parleur:
Maybe random basename could be an option ...

akhmerov:
This replaces #166, right?

Carreau:
No I think it is an alternative where --stdin does not imply --stdout

And the API change where you can pass a buffer as a kwarg instead of the first arg eing (filename or buffer)

parleur:
Sorry to be unclear ... It's a more more invasive alternative for #166

It responds to your comments about cleaning the code (introduce API extention `input_buffer` which is clearer than previously.

>While it's probably not very relevant to convert from stdin to a file, you're tying two unrelated options together — source of the notebook and output of the notebook. It might be cleaner to not do so (otoh it makes the interface more complex)

 And modify the command line interface as you pointed.

parleur:
As @Carreau said it modify the API (API just extended). But could be balanced by making the code a bit more clearer than having `notebook_filename` being a buffer or a filename and testing external flags to kwow what we are dealing with.

nbconvert/nbconvertapp.py
```diff
@@ -69,6 +69,13 @@ def validate(self, obj, value):
          "(the default behaviour is to abort conversion). This flag "
          "is only relevant if '--execute' was specified, too.")
         ),
+    'stdin' : (
+        {'NbConvertApp' : {
+            'from_stdin' : True,
+            }
+        },
+        "read a single notebook file from stdin, and write the converted result to stdout. Implies `--stdout`"
```
akhmerov: The description seems obsolete.

nbconvert/nbconvertapp.py
```diff
@@ -323,19 +335,23 @@ def init_single_notebook_resources(self, notebook_filename):
...
 
         Exports the notebook to a particular format according to the specified
         exporter. This function returns the output and (possibly modified)
         resources from the exporter.
 
+        notebook_filename: a filename or buffer (for instance from stdin using '--stdin' option)
         """
         try:
-            output, resources = self.exporter.from_filename(notebook_filename, resources=resources)
+            if input_buffer:
```
akhmerov: `if input_buffer is not None`
(same in analogous places)

nbconvert/nbconvertapp.py
```diff
@@ -170,12 +177,13 @@ def _classes_default(self):
...
-                      help="""Instance of the writer class used to write the 
+    writer = Instance('nbconvert.writers.base.WriterBase',
+                      help="""Instance of the writer class used to write the
                       results of the conversion.""", allow_none=True)
-    writer_class = DottedObjectName('FilesWriter', config=True, 
-                                    help="""Writer class used to write the 
+    writer_class = DottedObjectName('FilesWriter', config=True,
+                                    help="""Writer class used to write the
                                     results of the conversion""")
+    from_stdin   = Bool(False, config=True, help="read a single notebook from stdin.")
```
akhmerov: This is located in a confusing place in the file: it's definitely not a writer-specific variable. I suggest to move to the group of `# Other configurable variables`.

nbconvert/nbconvertapp.py
```diff
@@ -323,19 +335,23 @@ def init_single_notebook_resources(self, notebook_filename):
...
 
-    def export_single_notebook(self, notebook_filename, resources):
+    def export_single_notebook(self, notebook_filename, resources, input_buffer=None):
         """Step 2: Export the notebook
 
         Exports the notebook to a particular format according to the specified
         exporter. This function returns the output and (possibly modified)
         resources from the exporter.
 
+        notebook_filename: a filename or buffer (for instance from stdin using '--stdin' option)
```
akhmerov: Obsolete, cannot be a buffer.

nbconvert/nbconvertapp.py
```diff
@@ -323,19 +335,23 @@ def init_single_notebook_resources(self, notebook_filename):
...
-            output, resources = self.exporter.from_filename(notebook_filename, resources=resources)
+            if input_buffer is not None:
+                output, resources = self.exporter.from_file(input_buffer, resources=resources)
+            else:
+                output, resources = self.exporter.from_filename(notebook_filename, resources=resources)
         except ConversionException:
-            self.log.error("Error while converting '%s'", notebook_filename, exc_info=True)
-            self.exit(1)
+            self.log.error("Error while converting notebook", exc_info=True)
+            self.exit('Could not convert notebook.')
```
akhmerov: This change seems unrelated to the PR; is it an improvement?

nbconvert/nbconvertapp.py
```diff
@@ -369,18 +385,27 @@ def postprocess_single_notebook(self, write_results):
...
             4. (Maybe) postprocess the written file
-
+            
+            If input_buffer is not None, convertion is done using buffer as source ignoring
+            the notebook_filename argument.
         """
-        self.log.info("Converting notebook %s to %s", notebook_filename, self.export_format)
-        resources = self.init_single_notebook_resources(notebook_filename)
-        output, resources = self.export_single_notebook(notebook_filename, resources)
+        if not input_buffer:
```
akhmerov: `if input_buffer is None`

nbconvert/nbconvertapp.py
```diff
@@ -369,18 +385,27 @@ def postprocess_single_notebook(self, write_results):
...
+            self.log.info("Converting notebook %s to %s", notebook_filename, self.export_format)
+            resources = self.init_single_notebook_resources(notebook_filename)
+            output, resources = self.export_single_notebook(notebook_filename, resources)
```
akhmerov: This line can be taken out of `if` statement (together with the matching line in `else`.

nbconvert/nbconvertapp.py
```diff
@@ -369,18 +385,27 @@ def postprocess_single_notebook(self, write_results):
...
+        else:
+            notebook_filename = "notebook"
```
akhmerov: This seems unnecessary. You can assume it's a legitimate name even if `input_buffer` is set. Then you can just provide the value `notebook` in line 440.

nbconvert/nbconvertapp.py
```diff
@@ -398,18 +423,21 @@ def convert_notebooks(self):
...
             sys.exit(-1)
 
         # convert each notebook
-        for notebook_filename in self.notebooks:
-            self.convert_single_notebook(notebook_filename)
+        if not self.from_stdin:
+            for notebook_filename in self.notebooks:
+                self.convert_single_notebook(notebook_filename)
+        else:
+            self.convert_single_notebook(None, input_buffer=sys.stdin)
```
akhmerov: Related to comment about l404: 

`self.convert_single_notebook('notebook', input_buffer=sys.stdin)`

akhmerov:
Other things still to consider potentially:
* fixing the whitespace changes by `git checkout -i`.
* whatsnew entry.

Other than that and above comments, everything seems perfect to me.

parleur:
Thanks a lot for all your comments. I am keep trying to suppress my coding glitches. 

What do mean by 'whatsnew entry' ? Precisely where it is supposed to be written ?

akhmerov:
> What do mean by 'whatsnew entry' ? Precisely where it is supposed to be written ?

Just a description for changelog, @Carreau should know where it needs to go.

Carreau:
> @Carreau should know where it needs to go.

`docs/source/changelog.rst` I guess. 

Thanks !

akhmerov:
Happy new year everyone! Should this be merged? I don't see any further problems.

Carreau:
let's get notebook 4.1, and then pass on all those things that have waited too long.

But in principle,+1.

nbconvert/nbconvertapp.py
```diff
@@ -323,16 +330,21 @@ def init_single_notebook_resources(self, notebook_filename):
...
-    def export_single_notebook(self, notebook_filename, resources):
+    def export_single_notebook(self, notebook_filename, resources, input_buffer=None):
         """Step 2: Export the notebook
 
         Exports the notebook to a particular format according to the specified
         exporter. This function returns the output and (possibly modified)
         resources from the exporter.
 
+        notebook_filename: a filename
+        input_buffer: a containing notebook buffer, if not None notebook_filename is ignored
```
takluyver: This description wasn't completely clear to me. The convention I'm used to in Python is to call these 'file like objects'. Annoyingly, there's a lot of 'file like objects' with slightly different properties, so it's a good idea to clarify, e.g.

> a readable file like object returning unicode

nbconvert/nbconvertapp.py
```diff
@@ -403,13 +421,17 @@ def convert_notebooks(self):
...
 
         # convert each notebook
-        for notebook_filename in self.notebooks:
-            self.convert_single_notebook(notebook_filename)
+        if not self.from_stdin:
+            for notebook_filename in self.notebooks:
+                self.convert_single_notebook(notebook_filename)
+        else:
+            # default name when conversion from stdin
+            self.convert_single_notebook("notebook.ipynb", input_buffer=sys.stdin)
```
takluyver: Unfortunately, I think this will run into Unicode problems on Python 3 in non-UTF-8 systems, primarily Windows. Python will use the system's default encoding to try to decode stdin, and if a notebook is piped in containing a non-ascii character, and it tries to decode it using cp1252 or something, it will blow up.

Have a look at `nbconvert.utils.io.unicode_std_stream` for what we need to do here. That's for output, but basically the same steps should work for stdin.

parleur:
Happy new year, and sorry for the latency.

nbconvert/tests/test_nbconvertapp.py
```diff
@@ -315,8 +316,33 @@ def test_fenced_code_blocks_markdown(self):
             output2, _ = self.nbconvert('--to markdown --stdout notebook_jl.ipynb')
             assert '```julia' in output2  # shouldn't have language
             assert "```" in output2  # but should also plain ``` to close cell
-
-        pass
+    
+    def test_convert_from_stdin_to_stdout(self):
+        """
+        Verify that conversion can be done via stdin, and that 
+        --stdin  implies --stdout.
```
takluyver: This docstring is misleading now - `--stdin` no longer implies `--stdout`.

nbconvert/tests/test_nbconvertapp.py
```diff
@@ -315,8 +316,33 @@ def test_fenced_code_blocks_markdown(self):
...
+            with io.open('notebook1.ipynb') as f:
+                notebook = f.read().encode()
+                output1, _ = self.nbconvert('--to markdown --stdin --stdout', stdin=notebook)
+            assert_not_in('```python', output1) # shouldn't have language
+            assert_in("```", output1) # but should have fenced blocks
+
+    def test_convert_from_stdin(self):
+        """
+        Verify that conversion can be done via stdin, and that 
+        --stdin  implies --stdout.
```
takluyver: And again

nbconvert/utils/io.py
```diff
@@ -31,3 +31,20 @@ def unicode_std_stream(stream='stdout'):
...
+
+    This ignores environment variables and default encodings, to reliably read unicode from stdin.
+
+    ::
+
+        totreat = unicode_stdin_stream().read()
+    """
+    stream  = sys.stdin
+    if PY3:
+        stream_b = stream.buffer
```
takluyver: I think this should do the same try/except AttributeError that we do in `unicode_std_stream()` above. This is to deal with the case where someone has set `sys.stdin` to another object, like a `StringIO` instance. That's less common for stdin than for stdout, but it's still a valid thing to do.

takluyver:
Apart from a few minor comments, I think this looks good. Hopefully we can finally get this merged soon!

Carreau:
+1, looking for ward to get back on "dev and break all the things".

nbconvert/utils/io.py
parleur: I think it is done in tests ... that's why I dealt with in a proper way.

Carreau:
+1, thanks for you patience. 

5.0 of 4.2 ? Is does not break API.

takluyver:
I don't think we have any really major breaking changes planned for nbconvert, so I'm not sure how we want to handle the version numbering. Are 4.x releases roll-ups of bugfixes, as we've traditionally done things, and 5.0 will be a feature release, or do we want to do a longer series of 4.x feature releases and switch to 5.0 when we want to break things. The latter is more semver style, but I don't think semver should be the only way to do things.

Carreau:
I would do 4.2 with a few new features, as we already have new things, like `--to Qualified.names`.

I just asked because the current version shows 4.2.dev, and you tagged 5.0. 

I'm find starting a 5.0.dev and having a 4.2 branch also. 

If you are fine with 4.2 with new features, I woudl just merge that now.

minrk:
I think in general, across all repos, we don't bump major until we start merging breaking changes. And in general, we should be requiring stronger justifications for any breaking changes we make, since any breaking change may force a coordinated release. We only need to start backporting once we've started down the path of breaking changes, but want to keep releasing bugfixes for prior versions.

I would expect most repos other than notebook/ipython to spend much of their time in minor releases adding features and not breaking stuff.

takluyver:
Great, then I think this can be landed. Thanks @parleur for your patience in working on this!

PR #175, opened 2015-12-20
FIX for bad ".jpe" extension which crashes latex compilation.

bkasel:
Hi, I encountered trouble using nbconvert for conversion to latex and pdf because jpeg images were converted to ".jpe" files which are not recognised by latex. The issue has been confirmed on several systems on several OS (see ipython repo: https://github.com/ipython/ipython/issues/8340), but since nbconvert has migrated here, I think this is the place to apply the fix.

### DIFF
--- a/nbconvert/preprocessors/extractoutput.py
+++ b/nbconvert/preprocessors/extractoutput.py
@@ -14,6 +14,17 @@ from traitlets import Unicode, Set
 from .base import Preprocessor
 from ipython_genutils import py3compat
 
+def guess_extension_without_jpe(mimetype):
+    """
+    This function fixes a problem with '.jpe' extensions
+    of jpeg images which are then not recognised by latex.
+    For any other case, the function works in the same way
+    as mimetypes.guess_extension
+    """
+    ext = guess_extension(mimetype)
+    if ext==".jpe":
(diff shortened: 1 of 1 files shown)


Carreau:
Seems good to me,  Thanks a lot !

Do you feel comfortable trying to write a test ? (Otherwise we can do that ourself)

bkasel:
I am not entirely sure. The main problem is that the issue seems to arise erratically over different systems. How  would you test for that? Maybe just check if guess_extension_without_jpe("image/jpeg") returns either .jpg or .jpeg

juhasch:
Here is the Python issue for this behavior: https://bugs.python.org/issue4963
It might make sense to reference this.

Carreau:
We can likely monkeypatch `mimetype` to explicitly return `jpe` on `.jpeg`, and check that the conversion still happen corectly. 

But we can just open an issue to do that later.

bkasel:
@juhasch yes, that one is the root of all evil ;-) Where should I add that?

@Carreau I had a look at the way the tests are done and I do not really feel comfortable messing around in there.

minrk:
@bkasel no problem. I've opened #176 as a reminder that we should add a test.

PR #177, opened 2015-12-23
ENH: add option to specify (and possibly override) the kernel name.

tritemio:
Add trailet `kernel_name` to `ExecutePreprocessor` to allow specifying a kernel for the notebook execution. When no `kernel_name` is set, the kernel name is taken from the notebook metadata.

When running notebooks on python 2/3 projects, it is convenient to run the same notebook under both python 2 and python 3. Before this patch, it was not possible to directly run a notebook with a different kernel form the one saved in the notebook metadata. With this patch is possible, for example, to use a python2 kernel to run a notebook that specifies python3.

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -65,7 +65,7 @@ class ExecutePreprocessor(Preprocessor):
     extra_arguments = List(Unicode())
 
     kernel_name = Unicode(
-        "python", config=True,
+        '', config=True,
         help=dedent(
             """
             Name of kernel to use to execute the cells.
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -81,9 +81,8 @@ class ExecutePreprocessor(Preprocessor):
             path = None
 
         from jupyter_client.manager import start_new_kernel
-        if not self.kernel_name:
-            kernel_name = nb.metadata.get('kernelspec', {}).get('name', 'python')
-        else:
+        kernel_name = nb.metadata.get('kernelspec', {}).get('name', 'python')
+        if self.kernel_name:
             kernel_name = self.kernel_name
         self.log.info("Executing notebook with kernel: %s" % kernel_name)
         self.km, self.kc = start_new_kernel(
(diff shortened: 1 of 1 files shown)


nbconvert/preprocessors/execute.py
```diff
@@ -64,13 +64,27 @@ class ExecutePreprocessor(Preprocessor):
 
     extra_arguments = List(Unicode())
 
+    kernel_name = Unicode(
+        "python", config=True,
```
yuvipanda: I think it should default to being unset, rather than 'python'?

tritemio: Absolutely, it needs to be changed to None (otherwise it always overrides the kernel_spec kernel).

Thinking about defaults, note that even before this PR, the kernel name is set to "python" when no name is specified in the kernel_spec (see [this line](https://github.com/jupyter/nbconvert/blob/4.1.0/nbconvert/preprocessors/execute.py#L73)). Apparently on both python2 and 3 the is always a kernel called `python` which points to the current python version. This is why I did catch the wrong default in the `kernel_name` traitlet.

In principle no kernel in kernel_spec and no explicit `kernel_name` passed should raise an error. However, I'm a bit reluctant adding this exception because it will break backward compatibility if somebody is relying on automatic selection of "python" kernel when no kernel is specified in the notebook metadata. On the other hand, it will make more explicit which kernel the user wants to use.

So, what do you guys prefer?

yuvipanda: Yeah I think retaining current behavior and just making the override an
additional option is the right thing to do.

tritemio:
I changed the default value `kernel_name` value to `''` (instead of `'python'`) and slightly simplified the logic.

takluyver:
Yep, I think that's fine. The fallback to 'python' if no kernel is specified is because older versions of IPython (pre 3.0) weren't aware of different kernels, so saved notebooks without didn't embed information about their kernel. In that case, Python should be a sensible default.

tritemio:
@takluyver, ok makes sense. Let me know if you need something else for merging.

takluyver:
Thanks

PR #178, opened 2015-12-27
Pdflatex figurefixes

fperez:
Mostly improvements to figure handling, but a few extra fixes as well, to improve the quality of PDF output out of the box.

With these fixes, I was able to generate the final Sloan report with the command `jupyter nbconvert --to pdf <notebook-filename>`, without any further manual modification.

Note to whoever does final merge: merge commit message should list credit to https://gist.github.com/rwst/1366514 by @rwst, which is the main logic of this PR, the rest are small tweaks.

### DIFF
--- a/nbconvert/templates/latex/base.tplx
+++ b/nbconvert/templates/latex/base.tplx
@@ -16,7 +16,26 @@ This template does not define a docclass, the inheriting class must define this.
     % Nicer default font than Computer Modern for most use cases
     \usepackage{palatino}
 
-    \usepackage{graphicx} % Used to insert images
+    % Basic figure setup, for now with no caption control since it's done
+    % automatically by Pandoc (which extracts ![](path) syntax from Markdown).
+    \usepackage{graphicx}
+    % We will generate all images so they have a width \maxwidth. This means
+    % that they will get their normal width if they fit onto the page, but
+    % are scaled down if they would overflow the margins.
+    \makeatletter
+    \def\maxwidth{\ifdim\Gin@nat@width>\linewidth\linewidth
+    \else\Gin@nat@width\fi}
+    \makeatother
+    \let\Oldincludegraphics\includegraphics
+    % Set max figure width to be 80% of text width, for now hardcoded.
+    \renewcommand{\includegraphics}[1]{\Oldincludegraphics[width=.8\maxwidth]{#1}}
+    % Ensure that by default, figures have no caption (until we provide a
+    % proper Figure object with a Caption API and a way to capture that
+    % in the conversion process - todo).
+    \usepackage{caption}
+    \DeclareCaptionLabelFormat{nolabel}{}
+    \captionsetup{labelformat=nolabel}
+
     \usepackage{adjustbox} % Used to constrain images to a maximum size 
     \usepackage{color} % Allow colors to be defined
     \usepackage{enumerate} % Needed for markdown enumerations to work
--- a/nbconvert/templates/latex/base.tplx
+++ b/nbconvert/templates/latex/base.tplx
@@ -13,6 +13,9 @@ This template does not define a docclass, the inheriting class must define this.
     ((* block docclass *))((* endblock docclass *))
     
     ((* block packages *))
+    % Nicer default font than Computer Modern for most use cases
+    \usepackage{palatino}
+
     \usepackage{graphicx} % Used to insert images
     \usepackage{adjustbox} % Used to constrain images to a maximum size 
     \usepackage{color} % Allow colors to be defined
--- a/nbconvert/templates/latex/article.tplx
+++ b/nbconvert/templates/latex/article.tplx
@@ -13,5 +13,5 @@
 %===============================================================================
 
 ((* block docclass *))
-\documentclass{article}
+\documentclass[11pt]{article}
 ((* endblock docclass *))


rwst:
Thanks for the note. I have no problem with this, please go ahead.

fperez:
Great, many thanks @rwst! I've updated the PR description accordingly. Credit in the corresponding commit linking to your original gist, please let us know if you'd like any changes made, happy to do so.

nbconvert/templates/latex/base.tplx
```diff
@@ -13,7 +13,29 @@ This template does not define a docclass, the inheriting class must define this.
...
-    \usepackage{graphicx} % Used to insert images
+    % Nicer default font than Computer Modern for most use cases
+    \usepackage{palatino}
+
+    % Basic figure setup, for now with no caption control since it's done
+    % automatically by Pandoc (which extracts ![](path) syntax from Markdown).
+    \usepackage{graphicx}
+    % We will generate all images so they have a width \maxwidth. This means
+    % that they will get their normal width if they fit onto the page, but
+    % are scaled down if they would overflow the margins.
```
Carreau: This should already be handled line 204 and works correctly for me with pandoc 1.15.2 (with a default of min witdth 80% pagewidth, height 80% pageheight,to ensure extra-long image still fit.) So This would be a regression.

nbconvert/templates/latex/base.tplx
```diff
@@ -13,7 +13,29 @@ This template does not define a docclass, the inheriting class must define this.
...
+    \makeatletter
+    \def\maxwidth{\ifdim\Gin@nat@width>\linewidth\linewidth
+    \else\Gin@nat@width\fi}
+    \makeatother
+    \let\Oldincludegraphics\includegraphics
+    % Set max figure width to be 80% of text width, for now hardcoded.
+    \renewcommand{\includegraphics}[1]{\Oldincludegraphics[width=.8\maxwidth]{#1}}
+    % Ensure that by default, figures have no caption (until we provide a
+    % proper Figure object with a Caption API and a way to capture that
+    % in the conversion process - todo).
```
Carreau: If you are speaking on markdown figure, this part is untrue. The way to provide a caption (or alt text in html) is the following:

```
![ This is the alt text or caption ](/this/is/the/path/to/the/image.ext)
```

Pandoc also extract that correctly, and It works locally. So this would introduce a regression.

Carreau:
I guess what you want is to set no-caption (not to get the `Figure X:`) when there is actually no text in the markdown alt-text. That make sens. But as is this would introduce regression to nbconvert. 

For the width/height things if it does not work for you we can try to debug it.

nbconvert/templates/latex/base.tplx
fperez: Mmh, it wasn't working for me... The reason I introduced this logic was precisely because when I started creating the Sloan report, I was getting all my figures to be absolutely huge, blowing way past the page borders.  I'm on OS X, pandoc 1.15.2.1.

So, evidently, something wasn't quite right until I put this logic in, but we should sort it out first...

nbconvert/templates/latex/base.tplx
fperez: The problem is that often people don't actually use that, since that only produces alt text, not an actual caption for the figure. So the result would be an empty "Figure:" label, which is ugly.

fperez:
Yes, we need a solution for the case where there's no caption in the alt-text, to avoid the ugly empty `Figure X:` markers.

As far as the figure size, I'm happy to drop that commit, if it worked, but in my case, it didn't at all. Can you try to create the Sloan report without this PR at all, and see what you get? Just run `nbconver --to pdf` with master? I got *huge* figures that oveflowed the page very badly.

Carreau:
> As far as the figure size, I'm happy to drop that commit, if it worked, but in my case, it didn't at all. Can you try to create the Sloan report without this PR at all, and see what you get? Just run nbconvert --to pdf with master? I got huge figures that oveflowed the page very badly.

Can you send me the images you use ? I just have the ipynb. (see if I get a pixel for pixel PDF)

fperez:
On Tue, Dec 29, 2015 at 10:36 AM, Matthias Bussonnier <
notifications@github.com> wrote:

> Can you send me the images you use ? I just have the ipynb. (see if I get
> a pixel for pixel PDF)


sent!


-- 
Fernando Perez (@fperez_org; http://fperez.org)
fperez.net-at-gmail: mailing lists only (I ignore this when swamped!)
fernando.perez-at-berkeley: contact me here for any direct mail

Carreau:
> sent!

Ah indeed, my bad. The current handling of width/height apply only in outputs,not pandokized markdown.

I think we can just update that to `max size={0.9\linewidth}{0.9\paperheight}` for uniformity with outputs (or equivalent syntax). 

@fperez dont' worry about that, we'll take care of it !

fperez:
Ok, thanks! I played a lot with sizes, and I think a bit smaller than 90% would be a little better, perhaps 75% or 80%.  Visually it makes them somewhat more inset from the text. 90% barely shows as an inset, think of a block quote, and it's typically more inset than 10% (keep in mind that a 10% inset has to be split into 5% on each side, so it's almost nothing visually on each side to differentiate it).  So I'd argue for at least 80%...

ellisonbg:
+1 on 80%

On Wed, Dec 30, 2015 at 10:23 AM, Fernando Perez <notifications@github.com>
wrote:

> Ok, thanks! I played a lot with sizes, and I think a bit smaller than 90%
> would be a little better, perhaps 75% or 80%. Visually it makes them
> somewhat more inset from the text. 90% barely shows as an inset, think of a
> block quote, and it's typically more inset than 10% (keep in mind that a
> 10% inset has to be split into 5% on each side, so it's almost nothing
> visually on each side to differentiate it). So I'd argue for at least 80%...
>
> —
> <https://github.com/jupyter/nbconvert/pull/178#issuecomment-168049093>.
>



-- 
Brian E. Granger

minrk:
80% it is.

minrk:
Do we want to merge this and refine in master, or is there anything left to do in this PR?

Carreau:
:-( Need rebase now.

nbconvert/templates/latex/base.tplx
minrk: I just tested, and with this PR pandoc still generates captions from markdown. It only excludes the `Figure 1:` prefix.

minrk:
@fperez I think the conflict is only on the last commit, because the same fix had been found and merged in another PR. If you pop that, I think we can merge this.

fperez:
thanks!

Carreau:
:cake:

Issue #156, opened 2015-11-05
customizing and nbconvert_library doc pages are missing 


Carreau:
Reported by @takluyver [here](https://github.com/jupyter/nbconvert/commit/183ac9b95f085ab8ef4e74499764e6c3c777da57#commitcomment-14215343)

Readthedocs should really make a build **fail**.

takluyver:
We either need to:

- Keep the rst files checked in, remembering to only change the corresponding notebooks and regenerate them, or
- Do some more conf.py trickery to get them converted on RTD.

PR #180, opened 2015-12-28
Update documentation and add a readme for building docs

willingc:
- Add a `README_DOCS.md` to aid developers in docs development and local builds
- Expanded existing comments in `conf.py` to be more understandable to less frequent users of Sphinx configuration settings and RTD settings
- Added code to enable local builds of documentation with `sphinx_rtd_theme` instead of the default alabaster theme
- Added a `_static` directory with an empty text file to suppress warning during `make html`. This could have been achieved by editing the `conf.py` file to comment out the static file directory setting. Since we will likely have some screenshots in the near future, it seemed prudent to add it now. I'm happy to remove it if that is preferred.
- Edited notebook markdown to correct bad link to JVP's blog post

### DIFF
--- a/docs/README-DOCS.md
+++ b/docs/README-DOCS.md
@@ -1,0 +1,42 @@
+# Documenting nbconvert
+
+[Documentation for `nbconvert`](https://nbconvert.readthedocs.org/en/latest/)
+is hosted on ReadTheDocs.
+
+## Build Documentation locally
+
+0. Change directory to documentation root:
+
+    $ cd docs
+
+1. Install requirements:
+
+    $ pip install -r requirements.txt
+
+2. Build documentation using Makefile for Linux and OS X:
+
+    $ make html
+
+or for Windows:
+
+    $ make.bat html
+
+3. Display the documentation locally:
+
+    $ open build/html/index.html
+
+Or alternatively you may run a local server to display the docs. In Python 3:
+
+    $ python -m http.server 8000
+
+In your browser, go to `http://localhost:8000`
+
+## Developing Documentation
+
+### Helpful files and directories
+
+* conf.py - Sphinx build configuration file
+* source directory - source for documentation
+* source/api - source files for generated API documentation
+* autogen_config.py - Generates configuration of ipynb source files to rst
+* index.rst - Main landing page of the Sphinx documentation
--- a/docs/README-DOCS.md
+++ b/docs/README-DOCS.md
@@ -5,38 +5,40 @@ is hosted on ReadTheDocs.
 
 ## Build Documentation locally
 
-0. Change directory to documentation root:
+1. Change directory to documentation root:
 
-    $ cd docs
+       $ cd docs
 
-1. Install requirements:
+2. Install requirements:
 
-    $ pip install -r requirements.txt
+       $ pip install -r requirements.txt
 
-2. Build documentation using Makefile for Linux and OS X:
+3. Build documentation using Makefile for Linux and OS X:
 
-    $ make html
+       $ make html
 
-or for Windows:
+   or for Windows:
 
-    $ make.bat html
+       $ make.bat html
 
-3. Display the documentation locally:
+4. Display the documentation locally:
 
-    $ open build/html/index.html
+       $ open build/html/index.html
 
-Or alternatively you may run a local server to display the docs. In Python 3:
+   This command will open a window in your browser.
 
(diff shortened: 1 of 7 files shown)


docs/README-DOCS.md
```diff
@@ -0,0 +1,44 @@
+# Documenting nbconvert
```
takluyver: I'd be inclined to call this file just `README.md` instead of `README-DOCS.md`. Github will display `README.md` underneath the files list when you view the docs folder, but it doesn't do anything with `README-DOCS.md`.

docs/README-DOCS.md
```diff
@@ -0,0 +1,44 @@
...
+
+[Documentation for `nbconvert`](https://nbconvert.readthedocs.org/en/latest/)
+is hosted on ReadTheDocs.
+
+## Build Documentation locally
+
+1. Change directory to documentation root:
+
+       $ cd docs
```
takluyver: These aren't showing up as monospace blocks on Github ([view file here](https://github.com/willingc/nbconvert/blob/doc-sphinxone/docs/README-DOCS.md)). It looks like you need one more space of indentation for each:

```
1. Change directory to documentation root:

        $ cd docs
```

becomes:


1. Change directory to documentation root:

        $ cd docs

docs/README-DOCS.md
```diff
@@ -0,0 +1,44 @@
...
+
+       $ make html
+
+   or for Windows:
+
+       $ make.bat html
+
+4. Display the documentation locally:
+
+       $ open build/html/index.html
```
takluyver: `open` doesn't do this on Linux (though I have a personal alias, because typing `xdg-open` is annoying), and I'm not sure about Windows either. I don't think there is a reliable cross platform command for this, unfortunately. Maybe it's best to give the example with e.g. `firefox`, and rely on people knowing enough to substitute that with their preferred browser if they want.

docs/requirements.txt
```diff
@@ -2,5 +2,6 @@ traitlets
 jupyter_core
 jupyter_client
 nbformat
+sphinx_rtd_theme
 -e git+https://github.com/ipython/ipython.git#egg=ipython
 -e git+https://github.com/ipython/ipykernel.git#egg=ipykernel
```
takluyver: These lines are probably a hangover from when we were doing the 4.0 releases, I think we can just list `ipykernel` now.

docs/source/conf.py
```diff
@@ -23,7 +23,8 @@
 #sys.path.insert(0, os.path.abspath('.'))
 
 if os.environ.get('READTHEDOCS', ''):
-    # RTD doesn't use the Makefile, so re-run autogen_config.py here.
+    # RTD doesn't use the repo's Makefile to build docs. Run autogen_config.py
+    # here to generate rst files from ipynb files.
```
takluyver: `autogen_config.py` is actually generating the [config docs](http://nbconvert.readthedocs.org/en/latest/config_options.html), rather than converting ipynb to rst.

takluyver: Ideally we should also run the ipynb -> rst conversion here - see #156 - but I think that needs pandoc installed on RTD. Maybe now that RTD supports conda, we can [get that](http://anaconda.org/asmeurer/pandoc) - but that's definitely a matter for another PR

willingc:
@takluyver Thanks for the review and the helpful suggestions. I'll incorporate the suggestions and resubmit. And...Happy New Year :)

takluyver:
Happy new year!

There should be no need to resubmit - if you just add the changes to the same branch, they'll show up on the pull request, and you can leave a comment to say that you've finished addressing comments, so I get an email and can look at it again.

docs/source/conf.py
willingc: Agreed. There's a bunch of activity on projects for conversion now over at RTD, as well as other Sphinx extensions, ipymd, etc. Between all of us, we are aware of a bunch of different tools. I'm going to create a section in my calpoly repo to create both a matrix of conversion tools for the different options md->ipymd, ipymd->md, md in sphinx, md -> rst, etc. I'll ping you, min and matthias for comments/additions/deletions of the options. After that we can run some conversion examples to see what we like best. I really would like to get some consistency across the many repos on tools/packages.

willingc:
@takluyver I've added the suggested changes.

takluyver:
Thanks!

PR #184, opened 2016-01-06
Use conda on readthedocs, install pandoc and convert notebooks to rst

takluyver:
Alternative to #183 for addressing #156. I'll set up RTD on my fork to test if this works.

### DIFF
--- a/docs/conda-env.yml
+++ b/docs/conda-env.yml
@@ -1,0 +1,8 @@
+name: nbconvert_docs
+channels:
+- asmeurer
+dependencies:
+- pandoc
+- nbformat
+- jupyter_client
+- ipykernel
--- a/docs/conda-env.yml
+++ b/docs/conda-env.yml
@@ -6,3 +6,4 @@ dependencies:
 - nbformat
 - jupyter_client
 - ipykernel
+- numpydoc
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,6 +1,0 @@
-traitlets
-jupyter_core
-jupyter_client
-nbformat
--e git+https://github.com/ipython/ipython.git#egg=ipython
--e git+https://github.com/ipython/ipykernel.git#egg=ipykernel
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -28,6 +28,10 @@ if os.environ.get('READTHEDOCS', ''):
     with open('../autogen_config.py') as f:
         exec(compile(f.read(), 'autogen_config.py', 'exec'), {})
 
+        # Convert notebook files to rst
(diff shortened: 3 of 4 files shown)


Carreau:
Need rebase. Sorry.

docs/conda-env.yml
```diff
@@ -0,0 +1,9 @@
+name: nbconvert_docs
+channels:
+- asmeurer
+dependencies:
+- pandoc
+- nbformat
+- jupyter_client
+- ipykernel
+- numpydoc
```
minrk: we don't use numpydoc anymore, do we?

takluyver: I got a failure without this.

takluyver:
Unfortunately, while this is conceptually nicer than #183, it is not currently working - see rtfd/readthedocs.org#1902. As we don't know when that will be resolved, I think we should merge #183 as a stopgap so that our docs are complete.

Carreau:
> Unfortunately, while this is conceptually nicer than #183, it is not currently working - see rtfd/readthedocs.org#1902. As we don't know when that will be resolved, I think we should merge #183 as a stopgap so that our docs are complete.

then we need a travis test that the ipynb->rst conversion is stable, to make sure people don't change the `.rst` file without changing the ipynb.

minrk:
@takluyver makes sense. I merged #183 while we wait for conda to work properly.

mgeier:
There's a possible alternative approach: #212.

mgeier:
I guess this is now obsolete since #212 was merged, right?

takluyver:
I think switching to conda will still make sense, but we no longer need pandoc. I'll open a new PR for it.

takluyver:
Ah, no, we can't easily switch to conda until there's a conda package of nbsphinx. Never mind, requirements.txt works for now.

mgeier:
I have clue about conda packages, but if you tell me how, I can probably make one for `nbsphinx`.
Or feel free to create one yourself ...

damianavila:
If nbsphinx in on pypi it should be easy:

```
conda skeleton pypi nbsphinx
conda builg nbsphinx
anaconda upload -u org path/to/your/conda_package
```

mgeier:
I don't have an anaconda account and I'm not planning to become a `conda` user in the near future. Can somebody else please take care of the package?

And yes, it is on PyPI: https://pypi.python.org/pypi/nbsphinx/

minrk:
nbsphinx doesn't need to be on conda for us to use it. The environment can specify pip dependencies as well (See #222):

```yaml
dependencies:
- pandoc
- ...
- pip:
  - nbsphinx
```

PR #185, opened 2016-01-07
allow external exporter for nbconvert.

Carreau:
Would that better suit @takluyver (alternate to #141)


nbconvert/nbconvertapp.py
```diff
@@ -204,13 +238,23 @@ def _postprocessor_class_changed(self, name, old, new):
...
     )
 
+    def _export_format_changed(self, old, new):
+        try:
+            _validate_exporter(new)
+        except ValueError as e:
+            sys.exit(e.args)
+        except ImportError as e:
+            self.log.debug(e, exc_info=True)
+            sys.exit("Cannot import exporter '%s'" % new)
```
takluyver: Yuck, no, traitlet validation should never call `sys.exit()`.

I don't think it's really necessary to check the name as part of setting the traitlet at all. Just let traitlets check that it's a string, and throw an error when we try to use it if it's the wrong string. That should make the code simpler. Do you want me to open a mini-PR against this or #141 to show you what I mean?

I'm sympathetic to the argument that it's useful in help to see the values it can have, but since with external exporters that can't be a complete list, we can just do it in the help string.

Carreau: > Yuck, no, traitlet validation should never call sys.exit()

Yeah, I need to find a better way, but them that was to avoid traceback on bad exporter.



> Just let traitlets check that it's a string, and throw an error when we try to use it if it's the wrong string

I would prefer to check earlier, otherwise you end up with an exception relatively late in the process. But if it's really cleaner why not.

takluyver: PR Carreau/nbconvert#1 illustrates what I mean about simplifying this.

nbconvert/nbconvertapp.py
```diff
@@ -400,7 +419,7 @@ def convert_notebooks(self):
             self.exit(1)
         
         # initialize the exporter
-        self.exporter = exporter_map[self.export_format](config=self.config)
+        self.exporter = get_exporter(self.export_format)(config=self.config)
```
Carreau: @takluyver I think the double call here can be confusing and decrease readability. Thoughts ?

takluyver: It doesn't bother me at all, but it's not unreasonable to break it up a bit.

```python
cls = get_exporter(self.export_format)
self.exporter = cls(config=self.config)
```

Carreau: > It doesn't bother me at all

but you are part of the 1% of the 1% !

docs/source/external_exporters.rst
```diff
@@ -0,0 +1,156 @@
...
+        
+
+And the template file, that inherits from the html `full` template and prepend/append text to each markdown cell (see Jinja2 docs for template syntax)::
+
+    {% extends "full.tpl" %}
+
+    {% block markdowncell -%}
+
+
+    ## this is a markdown cells
```
willingc: Typo cells. Strike out the 's'

docs/source/external_exporters.rst
```diff
@@ -0,0 +1,156 @@
...
+    {super()}
+    ## THIS IS THE END
+
+
+    {% endblock markdowncell %}
+
+
+Assuming you install this pacakge locally, or from PyPI, you can now use::
```
willingc: typo package

@Carreau This is a really helpful doc. I spent the morning reading the nbconvert source. Love this addition.

docs/source/external_exporters.rst
```diff
@@ -0,0 +1,156 @@
...
+
+    juyter nbconvert --to mypackage.MyEporter notebook.ipynb
```
willingc: typo jupyter

docs/source/external_exporters.rst
Carreau: Typo fixed.

bollwyvl:
Is this the better PR to review now? I'm definitely interested in helping push this forward... this would be great for, say, giving us a way forward for getting additional exporters into nbviewer without having to control how closely the actual exporter instance gets created. Even better if we actually started [using ConfigManager](https://github.com/jupyter/nbviewer/issues/436) over there.

To that end, on nbpresent, I have to [wrap a lot of stuff myself](https://github.com/Anaconda-Server/nbpresent/blob/master/nbpresent/export.py), and have even gone to having my own cmd entry_point... which I may still do if there end up being lots of switches needed for the "simple" case. But from an ecosystem point of view, it will be really nice for the doc to just `nbconvert --to nbpresent.HTMLExporter` or `nbconvert.PDFExporter`. And obviously much easier to on-board new contributors.

Exporters I'd love to see:
- [TufteCSS](https://edwardtufte.github.io/tufte-css/) for pretty static sites
- [Phonon](http://phonon.quarkdev.com/) or [ionic](http://ionicframework.com/) mobile apps
- [baker](http://www.bakerframework.com/) for serial items
- [ebooks](https://github.com/aerkalov/ebooklib)
- khan academy lessons
- runnable docker images aka local binder :)

Obviously, some of these would have pretty heavy-duty non-python dependencies, but making these standalone, installable things, with in-browser GUI configuration as opposed to complex scripted pipelines will amplify the impact a notebook author can have.

Carreau:
> Is this the better PR to review now?

I guess so, there does not seem to be a huge interest in it. 

I would like to even mark that as provisional API (ie merge it, potentially release it with a "not stable" warning) and iterate on it.

nbconvert/nbconvertapp.py
```diff
@@ -30,6 +31,24 @@
...
+    Raises ValueError if exporter is not found
+    """
+    if name in exporter_map:
+        return exporter_map[name]
+
+    if '.' in name:
+        try:
+            return import_item(name)
+        except ImportError:
+            pass
```
bollwyvl: I'd like a little more detail on the error, perhaps:
```py
        try:
            return import_item(name)
        except ImportError as err:
            self.log.error("Error importing %s\n%s" % (name, err))
```

or combine it with the re-raise below....

minrk:
I think this proposal is good as-is (though it should get some exercise in tests before merge). If we go the route of entrypoints, which I think we should, there's no need for that to be exclusive of this one.

So +1 to merge as-is after some tests.

bollwyvl:
I was just able to successfully use this with nbpresent with only minor modifications (i needed to do some `encode()`ing to keep py3 happy).

Also, using the import notation works just fine with the shipped exporters:
```bash
jupyter nbconvert --to=nbconvert.HTMLExporter Some.ipynb
```

Worth adding that as a test for coverage sake?
```python
    def test_dotted_exporterself):
        """Do dotted exporters work?"""
        with self.create_temp_cwd(['notebook1.ipynb']):
            out, err = self.nbconvert('--log-level 0 --to nbconvert.HTMLExporter notebook1')
            assert os.path.isfile('notebook1.html')
```

Probably don't need to do all of the other formats.

Otherwise, LGTM!

nbconvert/nbconvertapp.py
Carreau: Good point. I assumed the module would just not exist.

Carreau:
> I think this proposal is good as-is (though it should get some exercise in tests before merge). If we go the route of entrypoints, which I think we should, there's no need for that to be exclusive of this one.
> 
> So +1 to merge as-is after some tests.

test added.

Carreau:
> Worth adding that as a test for coverage sake?

I did that but with a custom exporter that change the file extension, for the sake of making sure it's imported and use.

willingc:
@Carreau I'm cool with the docs. We can merge and iterate as needed.

Carreau:
Yeah ! Awesome ! Thanks @gonzolabs who did a lot of the work too !

PR #186, opened 2016-01-08
Added development install instructions to README

captainsafia:
Noticed there were no dev install instructions while working on some nbconvert stuff so here they are.

cc: @willingc

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,15 @@ Convert Juptyer notebook file, `mynotebook.ipynb`, to HTML using:
 
 This command creates an HTML output file named `mynotebook.html`.
 
+## Dev Install
+Install nbconvert for development using:
+
+```
+git clone https://github.com/jupyter/nbconvert.git
+cd nbconvert
+python setup.py develop
+```
+
 
 ## Resources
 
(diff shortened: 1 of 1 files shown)


willingc:
@captainsafia I like it! Do you think it would be helpful to mention here that pandoc should be installed too (http://nbconvert.readthedocs.org/en/latest/install.html#pandoc)?

README.md
```diff
@@ -36,6 +36,15 @@ Convert Juptyer notebook file, `mynotebook.ipynb`, to HTML using:
 
 This command creates an HTML output file named `mynotebook.html`.
 
+## Dev Install
+Install nbconvert for development using:
+
+```
+git clone https://github.com/jupyter/nbconvert.git
+cd nbconvert
+python setup.py develop
```
Carreau: `pip install -e .`  is preferred IIRC, (maybe even `python -m pip install -e .` but that's another story). Because if you don't use pip, pip might not be able to correctly uninstall later. In general you should never invoke setup.py directly.

captainsafia:
Updated to reflect both your PR comments. Thanks!

Carreau:
Heck you can even `conda install pandoc` now.

PR #190, opened 2016-01-11
get css from notebook-4.1

minrk:


willingc:
@minrk +1

damianavila:
If I go to 4.1.0 cdn, I don't see it minified... could that be the case?

minrk:
I think we stopped minifying because the minifier regularly introduced bugs, so it's just concatenated.

damianavila:
>I think we stopped minifying because the minifier regularly introduced bugs, so it's just concatenated.

OK, then LGTM

Carreau:
Should we upload a 4.2 version ?

minrk:
@Carreau when we release notebook-4.2, yes.

Carreau:
> @Carreau when we release notebook-4.2, yes.

So should we add that to the release things to do ?

willingc:
@carreau @minrk We should add to the release process that the docs should be up to date for the release (esp. conf.py and index.rst) before the final release is tagged in GitHub. ReadtheDocs unfortunately builds its releases off the commit in the GitHub tag. If the up to date docs are not in the release tagged, we can't easily tie the docs to the specific release for the stable build.

minrk:
There should be nothing to do. It ought to show up automatically.

willingc:
@minrk The stickler was the sidebar box that I was using for the new features, link to the announcement blog post (from a user perspective, it's good to include this), and install instructions. On most repos, it likely will not matter much. Not a big deal, I'm happy with a ping right before pushing the final release since changes to docs (if any) will be minimal. Alternatively, I can just make sure the docs have no mention of the RC once we enter RC period. Either is cool with me.

minrk:
@willingc we might be able to get away with some `if dev` logic, so that it does the right thing if it's a release version without needing manual changes. We used to have a few of those in the IPython docs.

willingc:
@minrk Sounds reasonable. We can try it out for 4.2.

Issue #125, opened 2015-10-02
Add a top-level function to execute notebooks


tritemio:
With all the machinery to run notebooks in-place in nbconvert, would be nice to offer a top-level function to simply execute a notebook programmatically.

I use something like this, if others agree I can send a PR:

```
def run_notebook(notebook_name):
    """Runs the notebook `notebook_name` (file name with no extension).

    This function executes notebook with name `notebook_name` (no extension)
    and saves the fully executed notebook in a new file appending "-out"
    to the original file name.

    It also displays links to the original and executed notebooks.
    """
    nb_name_full  = notebook_name + '.ipynb'
    display(FileLink(nb_name_full))

    out_path = 'out_notebooks/'
    out_nb_name = out_path + notebook_name + '-out.ipynb'

    nb = nbformat.read(nb_name_full, as_version=4)
    ep = ExecutePreprocessor(timeout = 3600)

    try:
        out = ep.preprocess(nb, {'metadata': {'path': './'}})
    except Exception:
        msg = 'Error executing the notebook "%s".\n\n' % notebook_name
        msg += 'See notebook "%s" for the traceback.' % out_nb_name
        print(msg)
        raise
    finally:
        nbformat.write(nb, out_nb_name)
        display(FileLink(out_nb_name))
```

Having a function like this would save some gimmick to users wanting to run a series of notebooks in one shot.

michaelpacer:
@minrk Is there anything that needs to be done for this issue? 

With @tritemio's https://github.com/tritemio/nbrun he seems to have covered the use case, is the idea that we should attempt to integrate that functionality into the core library? What are the barriers to doing that?

One thought is that the approach to distributing nbrun separately is going to discourage people from using it. 
Still, it's not clear where it would fit into the nbconvert codebase.

tritemio:
nbrun also executes notebooks passing parameters, but this is not the topic here. What I was proposing was to add a simple function that given a notebook (i.e. filename) it executes it and writes the executed version to disk.

A use case is executing a set of notebooks on a CI server (for both testing and rendering). Right now, I'm adding nbrun.py to the notebook repository as it is only a small file (and one function). Having such a function directly in nbconvert would avoid the need of vendorizing nbrun.

API-wise, the function should be in the top-level module and have only a few options to specify the output file (including not writing the output, maybe just returning the `nb` object) and expose a few `ExecutePreprocessor` traitlets such as `kernel_name`,  `timeout` and `allow_errors`.

[...]

PR #191, opened 2016-01-11
DOC: Add section on executing notebooks

tritemio:
This section shows how to execute notebooks using nbconvert API. The executed notebook is saved in notebook format. There is no discussion on exporting notebooks to other format (topic already addressed in another section of the docs).

This is heavily based on this blog post: [Batch Execution of Jupyter Notebooks](http://tritemio.github.io/smbits/2016/01/02/execute-notebooks/).

Except for links/formatting fixes it should be ready for review. I can rebase to clean history before merge.

### DIFF
--- a/docs/source/execute_api.rst
+++ b/docs/source/execute_api.rst
@@ -1,17 +1,19 @@
 Executing notebooks
 ===================
 
+.. module:: nbconvert.preprocessors
+
 In this section we show how to execute a ``.ipynb`` notebook
 document saving the result in notebook format.
 To export the notebook to other formats see section
-:ref:`Using nbconvert as a library <nbconvert_library>`.
+:doc:`nbconvert_library`.
 
 Executing notebooks programmatically is useful, for example, as a test layer
 in python libraries that include example notebooks, or as a way to
 automate the data analysis in projects involving more than one notebook.
 
 The same functionality of executing notebooks is exposed through a
-:ref:`command line interface <commandline_usage>` or a python API interface.
+:doc:`command line interface <usage>` or a python API interface.
 In this section we will (mostly) illustrate how to use the python API interface.
 
 A Quick example
@@ -20,7 +22,7 @@ A Quick example
 Let's start with a complete quick example, leaving detailed explanations
 to the following sections.
 
-First we import nbconvert and the ``ExecutePreprocessor`` class:
+First we import nbconvert and the :class:`ExecutePreprocessor` class:
 
 .. code-block:: python
 
--- a/docs/source/execute_api.rst
+++ b/docs/source/execute_api.rst
@@ -9,16 +9,16 @@ To export the notebook to other formats see section
 :doc:`nbconvert_library`.
 
 Executing notebooks programmatically is useful, for example, as a test layer
-in python libraries that include example notebooks, or as a way to
+in Python libraries that include example notebooks, or as a way to
 automate the data analysis in projects involving more than one notebook.
 
 The same functionality of executing notebooks is exposed through a
-:doc:`command line interface <usage>` or a python API interface.
+:doc:`command line interface <usage>` or a Python API interface.
 As an example, a notebook can be executed from the command line with::
 
     jupyter nbconvert --to notebook --execute mynotebook.ipynb
 
-In this section we will (mostly) illustrate how to use the python API interface.
+In this section we will (mostly) illustrate how to use the Python API interface.
 
 A Quick example
 ---------------
@@ -46,7 +46,7 @@ define respectively the execution timeout and the execution kernel.
 
     The option to specify **kernel_name** is new in nbconvert 4.2.
     When not specified or when using nbconvert <4.2,
-    the default python kernel is chosen.
+    the default Python kernel is chosen.
 
 To actually run the notebook we call the method ``preprocess``::
 
@@ -88,11 +88,11 @@ The second traitlet, ``kernel_name``, allows specifying the name of the kernel
 to be used for the execution. By default, the kernel name is obtained from the
 notebook metadata. The traitlet ``kernel_name`` allows to specify a user-defined
 kernel, overriding the value in the notebook metadata. A common use case
-is that of a python 2/3 library which includes documentation/testing
+is that of a Python 2/3 library which includes documentation/testing
 notebooks. These notebooks will specify either a python2 or python3 kernel
 in their metadata
 (depending on the kernel used the last time the notebook was saved).
-In reality, these notebooks will work on both python 2/3 and, for testing,
+In reality, these notebooks will work on both Python 2/3 and, for testing,
 it is important to be able to execute them programmatically on both
 versions. Here the traitlet ``kernel_name`` comes to help:
 we can just run each notebook twice, specifying first "python2" and then
--- a/docs/source/execute_api.rst
+++ b/docs/source/execute_api.rst
@@ -37,12 +37,12 @@ we can load it with::
 
 Next, we configure the notebook execution mode::
 
-    ep = ExecutePreprocessor(timeout=3600, kernel_name='python3')
+    ep = ExecutePreprocessor(timeout=600, kernel_name='python3')
 
 We specified two (optional) arguments ``timeout`` and ``kernel_name``, which
 define respectively the execution timeout and the execution kernel.
 
-    The option to specify **kernel_name** it's new in nbconvert 4.2.
+    The option to specify **kernel_name** is new in nbconvert 4.2.
     When not specified or when using nbconvert <4.2,
     the default python kernel is chosen.
 
--- a/docs/source/execute_api.rst
+++ b/docs/source/execute_api.rst
@@ -32,7 +32,8 @@ we can load it with:
 
 .. code-block:: python
 
-    nb = nbformat.read(open(notebook_filename), as_version=4)
+    with open(notebook_filename) as f:
+        nb = nbformat.read(f, as_version=4)
 
 Next, we configure the notebook execution mode:
 
--- a/docs/source/execute_api.rst
+++ b/docs/source/execute_api.rst
@@ -3,8 +3,8 @@ Executing notebooks
 
 In this section we show how to execute a ``.ipynb`` notebook
 document saving the result in notebook format.
-To export the notebook to other formats see the
-:ref:`next section <nbconvert_library>`.
+To export the notebook to other formats see section
+:ref:`Using nbconvert as a library <nbconvert_library>`.
 
 Executing notebooks programmatically is useful, for example, as a test layer
 in python libraries that include example notebooks, or as a way to
@@ -43,8 +43,9 @@ Next, we configure the notebook execution mode:
 We specified two (optional) arguments ``timeout`` and ``kernel_name``, which
 define respectively the execution timeout and the execution kernel.
 
-    The **kernel_name** keyword `requires <https://github.com/jupyter/nbconvert/pull/177>`__
-    nbconvert 4.2 (unreleased, use master branch from github).
+    The option to specify **kernel_name** it's new in nbconvert 4.2.
+    When not specified or when using nbconvert <4.2,
+    the default python kernel is chosen.
 
 To actually run the notebook we call the method ``preprocess``:
 
@@ -59,15 +60,16 @@ Finally, to save the resulting notebook:
 
 .. code-block:: python
 
-    nbformat.write(nb, open('executed_notebook.ipynb', mode='wt'))
+    with open('executed_notebook.ipynb', 'wt') as f:
+        nbformat.write(nb, f)
 
 That's all. Your executed notebook will be saved in the current folder
-in the file *executed_notebook.ipynb*.
+in the file ``executed_notebook.ipynb``.
 
 Execution arguments
 -------------------
 
-The arguments passed to ``ExecutePreprocessor`` are configuration options
+The arguments passed to :class:`ExecutePreprocessor` are configuration options
 called `traitlets <http://traitlets.readthedocs.org/>`_.
 There are many cool things about traitlets, for example
 they enforce the type of the input and they can be accessed/modified as
@@ -75,13 +77,14 @@ class attributes. Moreover, each traitlet is automatically exposed
 as command-line options. For example, we can pass the timeout from the
 command-line like this::
 
-    jupyter nbconvert --ExecutePreprocessor.timeout=3600 --to notebook --execute mynotebook.ipynb
+    jupyter nbconvert --ExecutePreprocessor.timeout=600 --to notebook --execute mynotebook.ipynb
 
 Let's now discuss in more detail the two traitlets we used.
 
-The ``timeout`` traitlet defines the maximum time (in seconds) the notebook is
+The ``timeout`` traitlet defines the maximum time (in seconds) each notebook
+cell is
 allowed to run, if the execution takes longer an exception will be raised.
-The default is only 30 s, so in many cases you may want to specify
+The default is 30 s, so in cases of long-running cells you may want to specify
 an higher value.
 
 The second traitlet, ``kernel_name``, allows specifying the name of the kernel
@@ -115,7 +118,7 @@ After an error, we can still save the notebook as before:
 
 The saved notebook contains the output up until the failing cell,
 and includes a full stack-trace and error (which can help debugging).
-A pattern I use to execute notebooks while handling errors is the following:
+A useful pattern to execute notebooks while handling errors is the following:
 
 .. code-block:: python
 
@@ -134,12 +137,13 @@ In case of errors, however, an additional message is printed and the
 ``CellExecutionError`` is raised. The messages directs the user to
 the saved notebook for further inspection.
 
-As a last scenario, sometimes notebooks contains independent computations
-in each code cell.
-In this case it can be useful to run the notebook until the end,
-in order to get a complete picture of all cells that are failing.
-Luckily enough, the ``allow_errors`` traitlet (default False) allows to do that.
+As a last scenario, it is sometimes useful to execute notebooks which
+raise exceptions, for example to show an error conditions.
+In this case, instead of stopping the execution on the first error,
+we can keep executing the notebook using the traitlet ``allow_errors``
+(default False).
 With ``allow_errors=True``,
-the notebook is executed until the end, and a ``CellExecutionError`` is raised
-if one or more cells threw an error. In this case, the output notebook
-will contain the stack-traces and error messages for all the failing cells.
+the notebook is executed until the end, regardless of any error encountered
+during the execution. The output notebook,
+will contain the stack-traces and error messages for **all** the cells
+raising exceptions.
--- a/docs/source/execute_api.rst
+++ b/docs/source/execute_api.rst
@@ -5,12 +5,12 @@ Executing notebooks
 
 Jupyter notebooks are often saved with cleared output cells. nbconvert
 provides a helpful way to execute the input cells of an .ipynb notebook
-file and save the results, both input and output cells, as an .ipynb file.
+file and save the results, both input and output cells, as a .ipynb file.
 
 In this section we show how to execute a ``.ipynb`` notebook
 document saving the result in notebook format.
-If you need to export notebooks to other formats (optionally executing them
-during conversion) see section :doc:`nbconvert_library`.
+If you need to export notebooks to other formats, such as reStructured Text
+or Markdown (optionally executing them) see section :doc:`nbconvert_library`.
 
 Executing notebooks can be very helpful, for example, to run all notebooks
 in Python library in one step, or as a way to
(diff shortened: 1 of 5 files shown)


docs/source/execute_api.rst
```diff
@@ -0,0 +1,145 @@
+Executing notebooks
+===================
+
+In this section we show how to execute a ``.ipynb`` notebook
+document saving the result in notebook format.
+To export the notebook to other formats see the
+:ref:`next section <nbconvert_library>`.
```
takluyver: It's probably better to refer to this by name, because if we rearrange the docs or put another page in between, we won't think about this.

docs/source/execute_api.rst
```diff
@@ -0,0 +1,145 @@
...
+
+Executing notebooks programmatically is useful, for example, as a test layer
+in python libraries that include example notebooks, or as a way to
+automate the data analysis in projects involving more than one notebook.
+
+The same functionality of executing notebooks is exposed through a
+:ref:`command line interface <commandline_usage>` or a python API interface.
+In this section we will (mostly) illustrate how to use the python API interface.
```
takluyver: Since the title of the page is just 'executing notebooks', I think it should at least briefly mention the `--execute` command line flag.

docs/source/execute_api.rst
```diff
@@ -0,0 +1,145 @@
...
+
+.. code-block:: python
+
+    ep = ExecutePreprocessor(timeout=3600, kernel_name='python3')
+
+We specified two (optional) arguments ``timeout`` and ``kernel_name``, which
+define respectively the execution timeout and the execution kernel.
+
+    The **kernel_name** keyword `requires <https://github.com/jupyter/nbconvert/pull/177>`__
+    nbconvert 4.2 (unreleased, use master branch from github).
```
takluyver: It's better not to scatter things like this ("unreleased, use master") through the docs, because we're definitely going to forget to update them. Just say that it's new in 4.2.

Also, I wouldn't show the `kernel_name` parameter in the first example at all. If it's not there, nbconvert will take the kernel name as saved in the notebook, which should be a sensible default in most cases.

docs/source/execute_api.rst
```diff
@@ -0,0 +1,145 @@
...
+    ep.preprocess(nb, {'metadata': {'path': 'notebooks/'}})
+
+Hopefully, we will not get any errors during the notebook execution
+(see the last section for error handling). Note that ``path`` specifies
+in which folder to execute the notebook.
+Finally, to save the resulting notebook:
+
+.. code-block:: python
+
+    nbformat.write(nb, open('executed_notebook.ipynb', mode='wt'))
```
takluyver: Let's use a context manager:

```python
with open('executed_notebook.ipynb', 'wt') as f:
    nbformat.write(nb, f)
```

I like to follow best practices in code examples for docs much more rigorously than I do in actual code files, because code examples will end up being copied and pasted everywhere.

docs/source/execute_api.rst
```diff
@@ -0,0 +1,145 @@
...
+
+That's all. Your executed notebook will be saved in the current folder
+in the file *executed_notebook.ipynb*.
```
takluyver: I think the filename should be `monospaced` rather than *italicised*

docs/source/execute_api.rst
```diff
@@ -0,0 +1,145 @@
...
+
+Execution arguments
+-------------------
+
+The arguments passed to ``ExecutePreprocessor`` are configuration options
```
takluyver: You can create cross links to the description of this class with syntax like this:

```
:class:`ExecutePreprocessor`
```

docs/source/execute_api.rst
```diff
@@ -0,0 +1,145 @@
...
+class attributes. Moreover, each traitlet is automatically exposed
+as command-line options. For example, we can pass the timeout from the
+command-line like this::
+
+    jupyter nbconvert --ExecutePreprocessor.timeout=3600 --to notebook --execute mynotebook.ipynb
+
+Let's now discuss in more detail the two traitlets we used.
+
+The ``timeout`` traitlet defines the maximum time (in seconds) the notebook is
+allowed to run, if the execution takes longer an exception will be raised.
```
takluyver: The timeout is per cell, not for the whole notebook

docs/source/execute_api.rst
```diff
@@ -0,0 +1,145 @@
...
+the error and the original error name and message are also printed.
+After an error, we can still save the notebook as before:
+
+.. code-block:: python
+
+    nbformat.write(nb, open('executed_notebook.ipynb', mode='wt'))
+
+The saved notebook contains the output up until the failing cell,
+and includes a full stack-trace and error (which can help debugging).
+A pattern I use to execute notebooks while handling errors is the following:
```
takluyver: I think we should reword this to avoid using "I" - it's fine for blog posts, but a bit confusing for docs.

docs/source/execute_api.rst
```diff
@@ -0,0 +1,145 @@
...
+the saved notebook for further inspection.
+
+As a last scenario, sometimes notebooks contains independent computations
+in each code cell.
+In this case it can be useful to run the notebook until the end,
+in order to get a complete picture of all cells that are failing.
+Luckily enough, the ``allow_errors`` traitlet (default False) allows to do that.
+With ``allow_errors=True``,
+the notebook is executed until the end, and a ``CellExecutionError`` is raised
+if one or more cells threw an error. In this case, the output notebook
```
takluyver: I don't think this is right. If allow_errors is True,  it runs through to the end, without throwing CellExecutionError at all. It's intended for notebooks that demonstrate an error for some reason, so an error is an expected part of their output.

I've just looked at the help string for `allow_errors`, and I think it has a confusing mistake - it says `True` when it should say `False`. I think it was initially designed as an `ignore_errors` flag and we forgot to update the help when we switched it.

tritemio: Ops, yes, in fact due to the switched True/False I mis-interpreted it as raising an exception (but at the end) when `allow_errors=True`. I'm including a docstring fix in the updated PR.

tritemio:
@takluyver, thanks for the comments. I fixed the the points you raised. 

Still, I left a note on `kernel_name` and regarding the command-line example (`--execute` flag), it is already there.

tritemio:
Wait for a few fixes left.

tritemio:
Done with fixes. 

- Locally, it builds with no warnings and the internal links work.
- Added example of command-line usage in the section intro
- Improved docstring for `timeout` and `allow_errors` traitlets
- Document `ExecutePreprocessors.process` and add it to API docs 
- Reworded docs index intro

Would be nice to add the `ExecutePreprocessor` traitlet docstrings to the API docs
but I don't know if it's possible using autodoc.

tritemio:
I also I rebased to clean the history.

willingc:
@tritemio Thanks for this PR. It's very helpful. I have some grammar and style edits to make the section more user friendly. :sunglasses:

docs/source/execute_api.rst
```diff
@@ -0,0 +1,143 @@
+Executing notebooks
+===================
+
+.. module:: nbconvert.preprocessors
+
+In this section we show how to execute a ``.ipynb`` notebook
```
willingc: Introduction paragraph:

Jupyter notebooks are often saved with cleared output cells. `nbconvert` provides a helpful way to execute the input cells of an `.ipynb` notebook file and save the results, both input and output cells, as an `.ipynb` file. This conversion can be very helpful when teaching using notebooks or exploring notebooks created by other researchers.

willingc: Typo ...execute an ``.ipynb`` ...

docs/source/execute_api.rst
```diff
@@ -0,0 +1,143 @@
...
+document saving the result in notebook format.
+To export the notebook to other formats see section
```
willingc: ...notebook file to other formats, such as reStructured Text or Markdown, see ...

docs/source/execute_api.rst
```diff
@@ -0,0 +1,143 @@
...
+in python libraries that include example notebooks, or as a way to
+automate the data analysis in projects involving more than one notebook.
+
+The same functionality of executing notebooks is exposed through a
+:doc:`command line interface <usage>` or a python API interface.
+As an example, a notebook can be executed from the command line with::
+
+    jupyter nbconvert --to notebook --execute mynotebook.ipynb
+
+In this section we will (mostly) illustrate how to use the python API interface.
```
willingc: This section will illustrate ... the Python API interface.

docs/source/execute_api.rst
```diff
@@ -0,0 +1,143 @@
...
+Next, we configure the notebook execution mode::
+
+    ep = ExecutePreprocessor(timeout=600, kernel_name='python3')
+
+We specified two (optional) arguments ``timeout`` and ``kernel_name``, which
+define respectively the execution timeout and the execution kernel.
+
+    The option to specify **kernel_name** is new in nbconvert 4.2.
+    When not specified or when using nbconvert <4.2,
+    the default python kernel is chosen.
```
willingc: Python

docs/source/execute_api.rst
```diff
@@ -0,0 +1,143 @@
...
+
+To actually run the notebook we call the method ``preprocess``::
+
+    ep.preprocess(nb, {'metadata': {'path': 'notebooks/'}})
+
+Hopefully, we will not get any errors during the notebook execution
+(see the last section for error handling). Note that ``path`` specifies
+in which folder to execute the notebook.
+Finally, to save the resulting notebook::
```
willingc: Finally, ~~to~~ save the resulting notebook

docs/source/execute_api.rst
```diff
@@ -0,0 +1,143 @@
...
+to be used for the execution. By default, the kernel name is obtained from the
+notebook metadata. The traitlet ``kernel_name`` allows to specify a user-defined
+kernel, overriding the value in the notebook metadata. A common use case
+is that of a python 2/3 library which includes documentation/testing
+notebooks. These notebooks will specify either a python2 or python3 kernel
+in their metadata
+(depending on the kernel used the last time the notebook was saved).
+In reality, these notebooks will work on both python 2/3 and, for testing,
+it is important to be able to execute them programmatically on both
+versions. Here the traitlet ``kernel_name`` comes to help:
```
willingc: ~~comes to help~~ is helpful:

docs/source/execute_api.rst
```diff
@@ -0,0 +1,143 @@
...
+we can just run each notebook twice, specifying first "python2" and then
+"python3" as kernel name.
+
+Error Handling
+--------------
+
+In the previous sections we saw how to save an executed notebook, assuming
+there are no execution error. But, what if there are errors?
```
willingc: ... execution errors.

docs/source/execute_api.rst
```diff
@@ -0,0 +1,143 @@
...
+        msg = 'Error executing the notebook "%s".\n\n' % notebook_filename
+        msg += 'See notebook "%s" for the traceback.' % notebook_filename_out
+        print(msg)
+        raise
+    finally:
+        nbformat.write(nb, open(notebook_filename_out, mode='wt'))
+
+This will save the executed notebook regardless of execution errors.
+In case of errors, however, an additional message is printed and the
+``CellExecutionError`` is raised. The messages directs the user to
```
willingc: The messages direct the

docs/source/execute_api.rst
```diff
@@ -0,0 +1,143 @@
...
+the saved notebook for further inspection.
+
+As a last scenario, it is sometimes useful to execute notebooks which
+raise exceptions, for example to show an error conditions.
```
willingc: ... an error condition.

docs/source/index.rst
```diff
@@ -3,10 +3,12 @@ Converting Notebooks to Other Formats
 =====================================
 
 The ``nbconvert`` tool allows you to convert an ``.ipynb`` notebook
-document file into various static formats.
+document file into various static formats. It can also execute notebooks
+programmatically.
 
-Currently, ``nbconvert`` is provided as a command line tool, run as a script
-using Jupyter. It also powers the 'Download as' feature within the
+``nbconvert`` is both a python library and a command line tool (invoked as
```
willingc: Python

docs/source/execute_api.rst
takluyver: Interesting, I'd say "a dot-ipynb file", so using "an" looks odd to me

willingc: @takluyver Good point. It's been done both ways throughout the docs. Since we do use ``.ipynb`` throughout the docs instead of ``ipynb`` file, I agree with you. Let's leave this as "a" ``.ipynb`` file.

docs/source/execute_api.rst
```diff
@@ -0,0 +1,143 @@
+Executing notebooks
+===================
```
Carreau: I think that this file conveys that manually creating an execute preprocessor is **the way** to execute notebook, which is not false, but not technically the canonical way. We have a special Exporter :`from nbconvert.exporters import NotebookExporter` which is made for notebook to notebook conversion (it does extra things). 

The canonical way would be

```python
from nbconvert.exporters import NotebookExporter, HTMLExporter
from traitlets.config import Config

c = Config({'ExecutePreprocessor':{'enable':True}})
E = NotebookExporter(config=c) # or E = HTMLExporter(config=c)

notebook, resources = E.from_filename('0 - notebook.ipynb')
# TODO save notebook
```

One of the big advantage of this approach is that it's identical wether you convert to HTML, rst, Latex, you "just" have to change the exporter you use. 

The other advantage is that it allows to imedialelty now how to do the same things for others and multiple preprocessors, in particular the following that all exporters inherit from : 

```python
List([
                                  'nbconvert.preprocessors.ClearOutputPreprocessor',
                                  'nbconvert.preprocessors.ExecutePreprocessor',
                                  'nbconvert.preprocessors.coalesce_streams',
                                  'nbconvert.preprocessors.SVG2PDFPreprocessor',
                                  'nbconvert.preprocessors.CSSHTMLHeaderPreprocessor',
                                  'nbconvert.preprocessors.LatexPreprocessor',
                                  'nbconvert.preprocessors.HighlightMagicsPreprocessor',
                                  'nbconvert.preprocessors.ExtractOutputPreprocessor',
                              ],
```

You "just" have to enable the preprocessors you use

The exporter should also take care of having the export process stateless, which make the conversion less prone to error. 

While the "how to use preprocessor" is nice, I think it should mention that unless you know what you are doing, using preprocessor directly is likely not what you want to do.

takluyver: Hmm, I would disagree with you there. I think executing and converting are two separate steps, and if all you want to do is execute a notebook, you shouldn't need an exporter object involved. So I would consider using `ExecutePreprocessor` the more canonical approach.

Carreau: Well the example give here is explicitly reading from disk, executing and writing back. So it is re-writing a basic exporter. Using the exporter also remove from many underlying details, like reading the notebook correctly (with nbformat) and setting the kernel to the right kerne, and then writing it back. 

Plus we know from experience that the order of preprocessors is tricky. So we likely do not want people to use preprocessor in the wrong way/wrong order and complain.

takluyver: The notebook exporter only really exists to fit execution+saving into the framework of nbconvert. It's not really necessary, and I think it's unwanted complexity for this example. Even if it can result in slightly fewer lines of code, it obscures what's going on.

 ExecutePreprocessor already handles getting the kernel name from the notebook metadata, so Exporter is no help there. Reading a notebook file with nbformat is a) trivial, and b) definitely a public API, so it's good to show people how to use it.

I think preprocessor order has been problematic precisely because nbconvert separates enabling the preprocessors from their priority. I think using them explicitly as Python objects rather than enabling them through config makes it easier to get them organised correctly.

tritemio: I agree with @takluyver here. In most of my use cases I need to execute notebooks and save the results as notebook (using `nbformat`, saving is not even handled by nbconvert).

Saving as notebook has 2 advantages for me. One is that the file can be distributed and is "executable". Compare this to a static representation in a sphinx documentation, that is harder to re-run (I should add a link to the original unexecuted notebook, but it's nice to have the output). Second, for debugging, it's easier to start from a notebook with all the output and re-execute it step by step, instead of starting from a  clean (no output) notebook. In this latter case, I use nbconvert for batch execution of a set of notebooks and when something fails I dig the problem by opening the failing notebook (containing output).

tritemio:
@willingc I applied your fixes. Let me know if you want to further tweak the intro.

tritemio:
@Carreau, I think your example on exporting notebooks should go on a different section, for example here:

- [Python API for working with nbconvert](http://nbconvert.readthedocs.org/en/latest/api/index.html)

However this section is currently a mess and should be heavily re-edited IMHO.

@willingc, @Carreau, @takluyver is there a way to make traitlets descriptions appear in sphinx API documentation?

takluyver:
> is there a way to make traitlets descriptions appear in sphinx API documentation?

I don't think so, unfortunately. Maybe it could be done with a Sphinx extension, if somebody is interested in writing one.

tritemio:
@takluyver, regarding the `with` statement, on a second though, I think in this case it may be ok not to use it. Take this example (reading):

```
nb = nbformat.read(open(notebook_filename), as_version=4)
```

No reference to the opened file is saved and the file is closed when `nbformat.read()` returns. The same goes for writing:

```
nbformat.write(nb, open('executed_notebook.ipynb', 'wt'))
```

And, in the final example on error handling, adding the `with` inside the `finally` becomes a little heavier on the eyes:

```
    try:
        out = ep.preprocess(nb, {'metadata': {'path': run_path}})
    except CellExecutionError:
        msg = 'Error executing the notebook "%s".\n\n' % notebook_filename
        msg += 'See notebook "%s" for the traceback.' % notebook_filename_out
        print(msg)
        raise
    finally:
        with open(notebook_filename_out, mode='wt') as f
            nbformat.write(nb, f)
```

What's your take on this?

takluyver:
On Python 3, a file that isn't explicitly closed and gets cleaned up when all references to it are gone generates a ResourceWarning. Admittedly these aren't visible by default, and perhaps it's overly strict, but they do show up during tests, so I still think it's better to point people to `with open(...`.

tritemio:
Ok, I didn't know that, I'll leave the PR as is then, using `with open(...`.

tritemio:
@takluyver 

> ResourceWarning is ignored unless Python was built in debug mode.

https://docs.python.org/3.5/library/warnings.html#default-warning-filters

takluyver:
Yep, like I said, they're not visible by default. You can enable them without rebuilding Python, though, and doing that when testing is good practice.

docs/source/execute_api.rst
Carreau: Fair enough.

Carreau:
seeing that this is an improvement on our docs I would be +1 to merge and iterate.

tritemio:
Regarding the verbosity of loading, executing and saving the notebook, I proposed in the past a top-level function https://github.com/jupyter/nbconvert/issues/125 to run notebooks in one step.

Nowadays I use a slightly improved version of that function which allows to optionally pass parameters to the notebook to be executed (~like a function). See [nbrun](https://github.com/tritemio/nbrun), rudimentary but incredibly useful in my workflow.

willingc:
@Carreau +1 on merge and iterate. Will leave to you or @takluyver to merge. 

@tritemio Thank you! :cookie:

tritemio:
@willingc, thanks, glad I earn a :cookie:!

willingc:
@tritemio :+1: Also, :surfer: :palm_tree: :cake: is a virtual vacation ;)

tritemio:
Just add one last commit, I'm done.

nbconvert/preprocessors/execute.py
```diff
@@ -36,7 +36,13 @@ class ExecutePreprocessor(Preprocessor):
     """
 
     timeout = Integer(30, config=True,
-        help="The time to wait (in seconds) for output from executions."
+        help=dedent(
+            """
+            The time to wait (in seconds) for output from executions.
+            If a cell execution takes longer, a `CellExecutionError`
```
takluyver: Looking at the code, it actually raises `TimeoutError` on Python 3 and `RuntimeError` on Python 2 (because TimeoutError is new in Python 3)

docs/source/index.rst
takluyver: @willingc is right, it should be a capital P here.

nbconvert/preprocessors/execute.py
tritemio: @takluyver fixed in the last commit.

takluyver:
Thanks for all your work on this!

willingc:
Yes! Thanks @tritemio for the good work and @takluyver for the helpful reviews.

Carreau:
:cake:

tritemio:
:+1: @willingc @Carreau @takluyver

PR #195, opened 2016-01-18
Re-allow capitalised --to options

takluyver:
e.g. `--to HTML`

Fixes a failing test

### DIFF
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -36,8 +36,8 @@ def get_exporter(name):
     
     Raises ValueError if exporter is not found
     """
-    if name in exporter_map:
-        return exporter_map[name]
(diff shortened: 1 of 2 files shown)


takluyver:
This was accidentally broken in #185 (by my changes, I think). We should ensure it gets into any release those changes are in.

minrk:
Want to capitalize one of our tests to catch this in the future?

takluyver:
There already is one - you may have seen a Jenkins failure email this morning. But it seemingly doesn't run on Travis.

takluyver:
Added a non-pandoc test so we can catch it on Travis.

minrk:
Gotcha. I hadn't gotten to that notification in my pile of emails from the weekend, yet.

PR #197, opened 2016-01-19
Expose external exporter section in TOC.

Carreau:


docs/source/external_exporters.rst
```diff
@@ -1,4 +1,4 @@
-.. _external_exporter
+.. _external_exporter:
```
minrk: s/exporter/exporters/

takluyver: The TOC is based on file names, so this doesn't break anything, but it's probably clearer if it's consistent.

There is a `:doc:` role, so in principle we shouldn't need labels at the top of each file. However, IIRC it's still useful for intersphinx, because I don't think `:doc:` references work with intersphinx.

Carreau: +s

PR #198, opened 2016-01-19
Update copyright date in docs.

Carreau:
Unlike source code, we can make it dynamic here. 
And I put the 2 dates just to avoid dispute of wether we want `initialYear` or `currentYear` in copyright notice.


minrk:
InitialYear would presumably be 2013 or something.

Carreau:
> InitialYear would presumably be 2013 or something.

It was part of IPython at that date.... SO no sure.

PR #205, opened 2016-01-21
DOC: fix rst newline, add syntax highlighting directive

ebolyen:
Noticed this while reading the docs [here](http://nbconvert.readthedocs.org/en/latest/external_exporters.html#writing-a-custom-exporter).


takluyver:
Thanks!

Issue #208, opened 2016-01-21
[docs] Bad link


Carreau:
http://nbconvert.readthedocs.org/en/latest/nbconvert_library.html literally shows `One of [@jakevdp](https://github.com/jakevdp)’s great blog posts will be used`  Pandoc might be confusd by the `@` in the link, or the `'s`

PR #210, opened 2016-01-21
Fix bad link in converted notebook docs

takluyver:
It was the @ sign causing the issue. Converted to Jake's full name.

Closes gh-208


takluyver:
Already fixed by #180. Looks like docs aren't being built; investigating now.

PR #213, opened 2016-01-22
Don't remove the newline when converting notebooks to notebooks (#207)

### LINKED ISSUE #207: --to notebook removes newline at end of file
When I run the following on a notebook:
```
jupyter-nbconvert --ClearOutputPreprocessor.enabled=True --to notebook --inplace *.ipynb
```
it removes the newline after the last ``}``.

That is a bit of an inconvenience, since that changes all notebook files, even if they were cleared already.
I'm running version 4.1.0

amueller:
Fixes #207.
I checked, and the regression test without the ``strip`` didn't pass before, but now passes.

### DIFF
--- a/nbconvert/exporters/notebook.py
+++ b/nbconvert/exporters/notebook.py
@@ -29,4 +29,6 @@ class NotebookExporter(Exporter):
         else:
             resources['output_suffix'] = '.nbconvert'
         output = nbformat.writes(nb_copy, version=self.nbformat_version)
+        if not output.endswith("\n"):
+            output = output + "\n"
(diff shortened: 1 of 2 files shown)


amueller:
Is it ok to leave the test like this, or should I make sure that ``file_contents`` actually ends with a newline?

takluyver:
That looks fine to me. I'll give others a chance to look at it, but it should be merged within a day or so.

amueller:
thanks for your help :)

Carreau:
Staled build. Relaunched tests.

PR #215, opened 2016-01-23
Added a comment on Jinja expression syntax

karldw:
Support for Jinja expressions syntax seems to be present, but undocumented (as far as I could tell).

To do something simple, adding one LaTeX package, it's appealing to create a new template that inherits from 'article.tplx' and uses Jinja's super() expression in the packages block.  The super() expression will inherit the previous contents of the block, which makes things much easier.
Example:
```latex
((* extends 'article.tplx' *))
((* block packages *))
% Use a different font:
\usepackage{libertine}
((( super() )))
((* endblock packages *))
```

### DIFF
--- a/docs/source/customizing.rst
+++ b/docs/source/customizing.rst
@@ -208,6 +208,10 @@ A few gotchas
 Jinja blocks use ``{% %}`` by default which does not play nicely with
 LaTeX, so those are replaced by ``((* *))`` in LaTeX templates.
 
+Jinja expressions use ``{{ }}`` by default which does not play nicely with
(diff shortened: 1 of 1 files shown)


takluyver:
+1 to documenting this, but can we make the two sentences read a bit better together? It's really two parts of the same idea, so we needn't repeat the exact same wording twice.

Cue a 'trouble with programmers writing docs' joke ;-)

karldw:
Sorry, I wasn't sure how to continue editing this pull request.  See #216 instead.

takluyver:
No problem. It looks like you made this PR by editing the file directly in Github, and I'm not sure that there is an easy way to update it after that.

PR #221, opened 2016-01-29
Add textcomp to latex template for textquotesingle (single tick)

### LINKED ISSUE #220: textcomp issue when using --to pdf
I have trouble rendering this in a markup:
```
Much `tex` such `'markup'`
```
It will result in this tex code:
```
    Much \texttt{tex} such
\texttt{\textquotesingle{}markup\textquotesingle{}}
```
which will lead to the latex error
```
Untitled.tex:251: Undefined control sequence \textquotesingle.
Untitled.tex:251: leading text: ...extquotesingle{}markup\textquotesingle{}}
```
It seems the ``textcomp`` package is not included.
Adding ``\usepackage{textcomp}`` at the start of the file solves the problem.

amueller:
Fixes #220.
Is that an ok place to add a regression test?


minrk:
Yup, AOK. Thanks!

PR #227, opened 2016-02-03
Add support for discovering separately installed exporters using entry points

takluyver:
The group name to advertise additional exporters is `nbconvert.exporter`.

This adds a dependency on my (tiny) [entrypoints](http://entrypoints.readthedocs.org/en/latest/index.html) module.

### DIFF
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -14,6 +14,7 @@ import sys
 import os
 import glob
 
+import entrypoints
 from jupyter_core.application import JupyterApp, base_aliases, base_flags
 from traitlets.config import catch_config_error, Configurable
 from traitlets import (
@@ -47,9 +48,16 @@ def get_exporter(name):
             log = logging.getLogger()
             log.error("Error importing %s" % name, exc_info=True)
             pass
+    else:
+        try:
+            return entrypoints.get_single('nbconvert.exporter', name).load()
+        except entrypoints.NoSuchEntryPoint:
+            pass
 
+    valid_names = sorted(get_export_names() +
+                     list(entrypoints.get_group_named('nbconvert.exporter')))
     raise ValueError('Unknown exporter "%s", did you mean one of: %s?'
-                     % (name, ', '.join(sorted(get_export_names()))))
+                     % (name, ', '.join(valid_names)))
 
 
 class DottedOrNone(DottedObjectName):
--- a/nbconvert/tests/exporter_entrypoint/eptest-0.1.dist-info/entry_points.txt
+++ b/nbconvert/tests/exporter_entrypoint/eptest-0.1.dist-info/entry_points.txt
@@ -1,0 +1,2 @@
+[nbconvert.exporter]
+entrypoint_test = eptest:DummyExporter
--- a/nbconvert/tests/exporter_entrypoint/eptest.py
+++ b/nbconvert/tests/exporter_entrypoint/eptest.py
@@ -1,0 +1,4 @@
+from nbconvert.exporters import Exporter
+
+class DummyExporter(Exporter):
+    pass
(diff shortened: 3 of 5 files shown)


ellisonbg:
With the nbextention stuff we decided to not use entry points. @minrk what are your thought on this usage ?

minrk:
Unlike nbextensions, I think entrypoints do make sense for nbconvert, where listing options makes more sense to me. It also allows exporters to pick their names, enabling things like `jupyter nbconvert --to asciidoc`, which I don't think is as relevant to nbextensions.

I view nbconvert exporters as being similar to pygments formatters, which uses entrypoints, whereas nbextensions are more like regular packages.

ellisonbg:
k, great!

On Wed, Feb 3, 2016 at 7:34 AM, Min RK <notifications@github.com> wrote:

> Unlike nbextensions, I think entrypoints do make sense for nbconvert,
> where listing options makes more sense to me. It also allows exporters to
> pick their names, enabling things like jupyter nbconvert --to asciidoc,
> which I don't think is as relevant to nbextensions.
>
> I view nbconvert exporters as being similar to pygments formatters, which
> uses entrypoints, whereas nbextensions are more like regular packages.
>
> —
> <https://github.com/jupyter/nbconvert/pull/227#issuecomment-179296029>.
>



-- 
Brian E. Granger

Carreau:
+1

we are slowly eating setuptools marketshare and features !

damianavila:
+1 as well

Carreau:
:-( unfortunately need rebase.

takluyver:
Rebased.

takluyver:
Ooh, we now get a 'Project Jupyter member' badge on our comments. That's new.

Carreau:
> Ooh, we now get a 'Project Jupyter member' badge on our comments. That's new.

That replace the "owner" thing.

Merging, thanks.

jezdez:
FWIW, since `entrypoints` isn't available as a source distribution on PyPI this change breaks using setuptools' `test_require` parameter and its `python setup.py test` command line tool for any Jupyter extension.

When it tries to fulfill the test requirements setuptools looks for suitable releases and is not available as a source distribution but only as a wheel distribution, it'll give up. I'm not sure what the best course of action here is, just wanted to let you know that.

Personally I'd recommend using setuptools' own entrypoints, but uploading a source distribution to PyPI for the entrypoints library would also work. Any opinion on that @takluyver?

takluyver:
There are PEPs in the works for a new sdist format that won't require setup.py. The official position now seems to be that `setup.py` should only ever be invoked through pip, never directly.

jezdez:
@takluyver I'm aware of the PEPs but with all due respect it seems as if you're jumping the gun a bit with only supporting whl files.

FWIW I fully understand and support your design decisions for flit, I've "fixed" way too many setup.py files in my own time and still recover from working on pip for a while. But I'm worried that basing a dependency for Jupyter -- especially as it was mentioned as "tiny" in this PR -- adds unnecessary risk for installation issues.

In other words, I'm not sure if Jupyter should be trying solve packaging problems on its own but wait for the upstream tools like pip adopt the PEPs. Until then I'd recommend reducing complexity and go with what currently works, e.g. via setuptools' entrypoints or creating a sdist for the entrypoints library.

takluyver:
I think the PEPs can formalise a solution that people are already experimenting with, but if we all sit back and wait for the PEPs, there won't be enough pressure to make those PEPs happen.

In the meantime, setuptools should AIUI see a package once it's been installed from a wheel, so it should be possible to work around it by running `pip install .` before testing.

jezdez:
Totally agreed about PEP and building pressure to finally getting rid of the hacks on top of hacks!

That's a great work-around, I hadn't thought about that, it was even enough to add `entrypoints` as a requirement to the tox config (that the project in particular has as well and which calls `python setup.py test`), so setuptools won't try to fulfill that dependency on its own. Thanks and cheers!

takluyver:
Excellent, glad it's working out :-)

PR #230, opened 2016-02-04
Bump version number to 5.0

takluyver:
Branch 4.x from the commit before the merge of this.


minrk:
Have we already merged breaking changes?

Carreau:
> Have we already merged breaking changes?

As far as I can tell, no , we did our best to actually keep API.

takluyver:
OK, this is deferred indefinitely until there are some breaking changes. Closing so it doesn't clutter up the PR list.

minrk:
I'm fine with bumping to 5.0 if we've started making these changes, I just didn't think we had done so yet.

PR #233, opened 2016-02-08
capture latex stderr

minrk:
allows forwarding latex errors to users (e.g. notebook save-as PDF)

closes jupyter/notebook#715

### DIFF
--- a/nbconvert/exporters/pdf.py
+++ b/nbconvert/exporters/pdf.py
@@ -12,6 +12,21 @@ from traitlets import Integer, List, Bool, Instance, Unicode
 from ipython_genutils.tempdir import TemporaryWorkingDirectory
 from .latex import LatexExporter
 
+class LatexFailed(IOError):
+    """Exception for failed latex run
+    
+    Captured latex output is in error.output.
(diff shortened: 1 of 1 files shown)


takluyver:
Does it make sense to try to do anything to try to cut down the Latex output to the relevant part? I suspect the answer is that no, any part of the output may be informative, though I know Latex output tends to be extremely verbose.

What about an accompanying UI change in the notebook so that the starting view shows the last lines of the error message, as the terminal does, since this is most likely to indicate what the problem is?

minrk:
I wouldn't know how to trim down latex output to what's relevant. Do you just mean to scroll to the bottom of the error page by default?

takluyver:
Yep, or rather display the output in its own scrolling box which is scrolled to the bottom by default, rather than scrolling the whole page down.

minrk:
Makes sense. Done in jupyter/notebook#1064

takluyver:
:+1: I'll give others a chance to look at this, but it looks fine to me.

Carreau:
Awesome !

Issue #234, opened 2016-02-09
ExecutePreprocessor.timeout does not work as expected


jlamcanopy:
In the documentation, it says ExecutePreprocessor.timeout is "The time to wait (in seconds) for output from executions"

My observation is that even the cell takes 1 second, it will wait until the timeout expires before moving to the next cell. It does not seem to be right because most of the cells take seconds only a few take longer. Is that something I'm missing here?

takluyver:
If that's what you're seeing, that's a bug. If a cell finishes in 1 second, it should notice that immediately and execute the next cell.

Is there anything unusual about your setup that might relate to this? E.g. are you using a different kernel?

jlamcanopy:
@takluyver Thank you for the quick reply. You are right! I'm not using a python kernel. I'm using a kernel called ispark. Are you suspecting it is the kernel issue?

jlamcanopy:
@takluyver You are right. If I tested it with a python kernel, it behaves as described. 
I used ispark kernel for long time interactively and I never encountered issues before. Do you have a clue why it behaves strangely with nbconvert?

takluyver:
I'd guess that it's not sending the kernel status messages:
http://jupyter-client.readthedocs.org/en/latest/messaging.html#kernel-status

The execute code in nbconvert looks for a *kernel idle* message to tell it that there's no more output. If it doesn't see that, it will wait for the full timeout again. We can probably make that timeout shorter, because it's collecting the outputs after it's received `execute_reply`, which should already mean that execution has finished.

minrk:
Yeah, it's probably a good idea to have a long-ish (~30 second) timeout for the execute_reply, and then a relatively short timeout for the output (~5s, conservatively), since it's just waiting for network funny business.

But this behavior does mean that the ispark kernel is not following the messaging protocol correctly.

jlamcanopy:
Thanks guys! I'm going to make some changes to ispark to see if I could make it to conform to the messaging protocol.

jlamcanopy:
@takluyver Sorry about bothering you but my knowledge in the protocol is quite limited but I want to fix ispark so I can use nbconvert with it. I found the code that might cause the issue you mentioned:

```
def busy[T](block: => T): T = {
    send_status(ExecutionStates.busy)

    try {
      block
    } finally {
      send_status(ExecutionStates.idle)
    }
  }
```

[...]

PR #235, opened 2016-02-10
Shorter timeout for collecting iopub messages after executing

takluyver:
When this runs, we have already waited up to self.timeout seconds for execute_reply, so all the outputs should already be queued.

This should make execution much faster for kernels that don't send the kernel status messages.

Closes gh-234


minrk:
One second is *probably* okay, but it might be safer to use 3-5s for slow networks (e.g. CI VMs, which can have super weird, crazy-slow network behavior).

takluyver:
There you go, 4.

minrk:
I like how you think.

Issue #239, opened 2016-02-12
Can't easily define a Latex template to render nb for embedding in other Latex file


takluyver:
For HTML, we have the `full` template with the document structure, and `basic` which produces HTML intended to be embedded in another document. However, there's no equivalent to the latter for producing Latex, and we couldn't see a convenient way to do it by inheriting from `base.tplx` - in particular, `\begin{document}` and `\end{document}` can't easily be disabled from that. Starting from the parent of `base`, one has to copy a number of blocks from base to get useful output.

We probably just need to adjust the template inheritance a bit so there's a template with the support blocks but without the main document structure.

Ping @maxalbert

PR #240, opened 2016-02-12
Restructure LaTeX templates

maxalbert:
This is a first attempt to slightly restructure the LaTeX templates with the aim to make it easier to embed nbconvert's LaTeX output in another LaTeX document (see #239). The only thing this PR currently does is move the support blocks into a separate template called `latex/document_contents.tplx` as suggested by @takluyver.

With this change, when I run nbconvert using `--template document_contents.tplx` this creates a .tex file which contains only the parts between `\begin{document}` and `\end{document}` so that I can include it in another LaTeX document. However, the conversion also removes the "input" parts of any code cells because the relevant template blocks are defined in `style_ipython.tplx`, which is further down the template hierarchy. For my use case, stripping code input happens to be precisely what I need, but it should of course be supported to keep the input.

In an attempt to support this, I tried to move `style_ipython.tplx` further up in the hierarchy but ran into issues due to the way some template blocks currently inherit from / depend on each other. I don't currently have time to debug this but wanted to open a PR for discussion anyway.

One additional complication is that in order to correctly compile the resulting LaTeX code a whole lot of preamble definitions are needed (at least if the input part of code cells is to be retained; otherwise only a handful of packages need to be included). These LaTeX definitions live in `base.tplx`, but if nbconvert is called with the `document_contents.tplx` template then the user needs to copy these manually as they won't be part of the converted .tex file. I wonder if it would be feasible to provide a .sty file to make this less painful (but I don't know how to easily make this visible to the user's LaTeX installation).

Any comments / input / suggestions on any of this are welcome.

### DIFF
--- a/nbconvert/templates/latex/base.tplx
+++ b/nbconvert/templates/latex/base.tplx
@@ -3,7 +3,7 @@ This template builds upon the abstract template, adding common latex output
 functions.  Figures, data_text,
 This template does not define a docclass, the inheriting class must define this.=))
 
-((*- extends 'display_priority.tplx' -*))
+((*- extends 'document_contents.tplx' -*))
 
 %===============================================================================
 % Abstract overrides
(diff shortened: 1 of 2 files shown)


Carreau:
Alway happy to take refactoring/improvement like that.
I'm going to mark as 5.0, as IIUC this is a change of API as out template structure change. Am I right ?

I have no objection in stating that master branch of nbconvert is 5.0.dev if that's the case.

takluyver:
I initially thought it would be possible to do this without breaking anything inheriting from the existing templates - by inserting a new template in the inheritance order as the parent of `base.tplx`. But based on @maxalbert's comments, it sounds like it might not be that easy.

minrk:
We could probably do the styles a little differently with macro/variable definitions, rather than relying on template inheritance.

As I see this now, this PR does indeed only add to the stack, without breaking anything, since it doesn't include the style reworking that might be hard to do without a major revision. All that together, it's fine to call this 5.0 if we've got a breaking change. There's no reason that major releases have to break *lots* of stuff, they can still be quite small, just indicating that at least one thing changed in an incompatible way.

maxalbert:
Thanks for merging this! I haven't actually used these templates in a little while but seem to remember that at the time it felt like they needed a bigger restructuring to provide proper flexibility for including LaTeX output in external documents (e.g. including a converted notebook as a section in a report). So while this PR won't hurt there is likely more work to do, but it would be good to have a small example demonstrating a limitation before starting more restructuring (unfortunately I don't have one at hand).

PR #242, opened 2016-02-12
Fixes #199 : make emph be rendered as italics in pdf

### LINKED ISSUE #199: markdown italics being translated as underline in pdf
When I go to convert a simple notebook with some italics markdown in it, like *some text*, the resulting converted pdf renders as underline, not italics. I'm using IPython 4.0.0 and nbconvert 4.1.0

amueller:
Fixes #199.
Makes sure ``\emph`` are rendered as italics and not as underlines.
I tested and it works. As this is a rendering issue (I think), I'm not sure how to add a test.


takluyver:
I think we can really test things like this, so it should be fine without a test.

From #199 this looks OK to me, but I'll let someone who knows about Latex OK it.

amueller:
I imagine you meant "can't" ;)

takluyver:
You imagine quite correctly. And that's what I said all along, honest. ;-)

PR #246, opened 2016-02-17
Make Exporter.from_filename() use self.from_file()

mgeier:
In the spirit of polymorphism, `nbconvert.Exporter.from_filename()` should delegate to `self.from_file()`, or am I missing something?

BTW, what's with the ignored keyword args?
In my experience, unchecked keyword arguments are often a source of subtle bugs and major frustration.

### DIFF
--- a/nbconvert/exporters/exporter.py
+++ b/nbconvert/exporters/exporter.py
@@ -162,7 +162,7 @@ class Exporter(LoggingConfigurable):
         resources['metadata']['modified_date'] = modified_date.strftime(text.date_format)
 
         with io.open(filename, encoding='utf-8') as f:
-            return self.from_notebook_node(nbformat.read(f, as_version=4), resources=resources, **kw)
+            return self.from_file(f, resources=resources, **kw)
 
 
     def from_file(self, file_stream, resources=None, **kw):


nbconvert/exporters/exporter.py
```diff
@@ -95,7 +95,7 @@ def __init__(self, config=None, **kw):
         with_default_config = self.default_config
         if config:
             with_default_config.merge(config)
-
```
takluyver: Please configure your editor not to strip trailing whitespace, it produces unnecessary diffs.

takluyver:
Other than the whitespace changes, I think this is fine.

nbconvert/exporters/exporter.py
mgeier: I'm not a fan of trailing spaces, but if you like them, you shall have them!

takluyver:
Thanks! I'm not particularly attached to trailing spaces, but I am in favour of keeping diffs minimal - it reduces the likelihood of introducing merge conflicts, and it makes tools like `git blame` more useful. We have a general policy that we don't do stylistic cleanups beyond the code that we're actually working on.

mgeier:
OK, got it. Thanks for merging!

Issue #176, opened 2015-12-21
test for .jpe issue in mimetypes


minrk:
We should add a test for the issue fixed by #175

takluyver:
The function added is so simple I'm not sure it's really worth testing, so I'm inclined to close this unless we want to leave it as a simple task for a new contributor to work on?

WillEngler:
New contributor here looking for a simple task. I'm going to claim this issue for JupyterDay Chicago.

I hear what @takluyver is saying with
> The function added is so simple I'm not sure it's really worth testing

So I'm thinking of writing a test for `ExtractOutputPreprocessor.preprocess_cell` that confirms it succeeds on a simple cell with a .jpeg in it.

WillEngler:
Looking closer at the existing test code, it seems like the preferred style is making a fixture notebook and doing a more integration-level test than what I'd suggested, so I'll go that route instead and make a notebook with a .jpeg. (By the way, lovely test organization! I love all the helpers.)

minrk:
closed by #251

PR #251, opened 2016-02-20
Add test for linked .jpeg

WillEngler:
In response to #176

I added a new fixture .ipynb to tests/files along with a .jpeg that the notebook links to. I figured you'd rather have a separate file to test .jpegs rather than adding another cell to one of the existing notebooks with a .png.

I'm happy to iterate on this if you have any critiques.


takluyver:
Technical matter: awwww!

takluyver:
Real technical matter: unfortunately I don't think this test is exercising the problem fixed by #175 (which is what #176 was opened for). The problem was in the extract output preprocessor, which turns embedded data in outputs into separate files. So to test it, we need a notebook with an embedded jpeg in the output of a code cell, rather than just linked from a markdown cell.

WillEngler:
Ahhhh. Pretty fundamental misunderstanding on my part. Thanks for setting me straight.

WillEngler:
Double checking: _this_ is embedded output, right? If so, I'll fix up the PR and send it back. Thanks for bearing with me!

screen shot 2016-02-20 at 1 07 47 pm

takluyver:
IIRC you need an `embed=True` parameter in there, otherwise it just links to the existing file. Check that there's a blob of base64 encoded data in the ipynb file.

WillEngler:
(If this looks good to you, I could squash that second commit if you like)

takluyver:
Looks good to me, thanks. I don't think there's a particular need to squash here, so if you want to fix the docstring, this can be merged.

WillEngler:
Good catch on the docstring. Fixed!

takluyver:
:cake:

Issue #228, opened 2016-02-04
PDF conversion choke on some bytes  (ansi sequences.??)


Carreau:
cf [notebook](https://github.com/mjbright/jupyter_notebooks/blob/bc9663c748ae14822c298036798e23c31261a6d9/2016-Feb_Docker_Build_Lab/2016-Feb_Docker_Build_Lab.ipynb) reference in [this comment](https://github.com/jupyter/notebook/issues/840#issuecomment-179646131).

```
 Package inputenc Error: Keyboard character used is undefined
(inputenc)                in inputencoding `utf8x'.

See the inputenc package documentation for explanation.
Type  H <return>  for immediate help.
 ...

l.1444 ^^[[0B

?
! Emergency stop.
 ...

l.1444 ^^[[0B
```

takluyver:
The offending bit is the output of a docker command:

```
    \begin{Verbatim}[commandchars=\\\{\}]
{\color{incolor}In [{\color{incolor}35}]:} docker push dockerlabs/toolset:userN
\end{Verbatim}

    \begin{Verbatim}[commandchars=\\\{\}]
The push refers to a repository [docker.io/dockerlabs/toolset]

[0B
[0B
[3BuserN: digest: sha256:8a2e3f3fae154e60e95adbbe95d4e67ada66c64adfbdbde4f5e56e8a56ac8a41 size: 3658
    \end{Verbatim}
```

It's not rendered correctly in [the notebook on Github](https://github.com/mjbright/jupyter_notebooks/blob/bc9663c748ae14822c298036798e23c31261a6d9/2016-Feb_Docker_Build_Lab/2016-Feb_Docker_Build_Lab.ipynb), either, but it doesn't break the whole document.

minrk:
Would #225 fix this?

mgeier:
I believe it would, since it removes all sequences starting with `'\x1b['` and ending with an ASCII letter in the range `[@-~]`.

According to [Wikipedia](https://en.wikipedia.org/wiki/ANSI_escape_code#Sequence_elements), there are also two-character sequences consisting of `'\x1b'` followed by one character in the range `[@-_]`. Those are still not handled.

And there seems to be an alternative starting character `'\x9b'` (used instead of `'\x1b['`), which isn't handled either.

I don't understand the details and I don't know if any of this is used in practice, but if needed, I can add support for those cases, too.

mgeier:
[...]

PR #252, opened 2016-02-22
Move \r handling from coalescestreams to ANSI filters

mgeier:
See also https://github.com/jupyter/nbconvert/issues/228#issuecomment-183977301.

### DIFF
--- a/nbconvert/filters/ansi.py
+++ b/nbconvert/filters/ansi.py
@@ -13,6 +13,7 @@ __all__ = [
 ]
 
 _ANSI_RE = re.compile('\x1b\\[(.*?)([@-~])')
+_CARRIAGE_RETURN_RE = re.compile(r'.*\r(?=[^\n])')
 
 _FG_HTML = (
     'ansiblack',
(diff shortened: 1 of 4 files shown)


minrk:
I'm not sure this is right because it means this now happens in the jinja stage, and won't happen for non-jinja outputs (e.g. `nbconvert --to notebook --execute`), right?

mgeier:
Good point, I wasn't thinking about that!

Ideally, the "execute" preprocessor would use the same code as the live notebook to store the output streams back to the notebook data structure. But I guess it doesn't, right?

minrk:
It doesn't because that's what the coalesce_streams preprocessor already accomplishes. Older versions of the javascript also didn't apply the coalescing, or only applied some, so it should probably still be performed at the preprocessing stage, whether handling is *added* to the ANSI filters or not.

mgeier:
OK, thanks for the explanation!
I see now why the `\r` handling happens in `coalesce_streams`.

I'm closing this ...

PR #261, opened 2016-03-03
Raise error on iopub timeout in execute preprocessor

jhamrick:
For nbgrader, it is possible that long-running, memory consuming cells will produce an iopub timeout. This causes the nbconvert preprocessor to just break out of the loop and continue executing other cells, which isn't what we'd want for nbgrader (see https://github.com/jupyter/nbgrader/issues/427). This PR makes it configurable whether to break out of the loop or throw an error, with the current behavior remaining the default option.

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -85,6 +85,20 @@ class ExecutePreprocessor(Preprocessor):
         )
     )
 
+    raise_on_iopub_timeout = Bool(
+        False, config=True,
+        help=dedent(
+            """
+            If `False` (default), then the kernel will continue waiting for
+            iopub messages until it receives all of them. If `True`, then an
+            error will be thrown after the first timeout. This option generally
+            does not need to be used, but may be useful in contexts where there
+            is the possibility of executing notebooks with memory-consuming
+            infinite loops.
+            """
+            )
+        )
+
 
     def preprocess(self, nb, resources):
         """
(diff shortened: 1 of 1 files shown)


nbconvert/preprocessors/execute.py
```diff
@@ -85,6 +85,20 @@ class ExecutePreprocessor(Preprocessor):
         )
     )
 
+    raise_on_iopub_timeout = Bool(
+        False, config=True,
+        help=dedent(
+            """
+            If `False` (default), then the kernel will continue waiting for
+            iopub messages until it receives all of them. If `True`, then an
```
takluyver: This doesn't seem quite right - if False, it will stop waiting when it hits the fixed timeout.

(Also, there's not really an 'all of them' for iopub messages - the client can't know that there are no more to come, it just makes an intelligent guess that a kernel status idle message means it can stop waiting. If the notebook does things with threads, there could be more messages after that.)

takluyver:
In what situation is that coming up? It should only be collecting iopub messages after it's got execute_reply, so any output should already be queued in zmq. Is the memory use just bogging things down so much that there can actually be a 4 second delay between two consecutive messages being delivered? If so, that seems like a bigger problem; I wonder if we can do something to limit the memory available to the kernel and kill it if it tries to exceed that.

jhamrick:
See the issue I linked to (https://github.com/jupyter/nbgrader/issues/427):

> However, it seems that if the infinite loop also memory consuming (think while True: some_list.append(1)), kernel sometimes freezes and does not respond on KeyboardInterrupt (actually, we see this behaviour from time to time using Jupyter Notebook interactively, when Interrupt Kernel doesn't work and we have to restart kernel). Then we have a warning: Timeout waiting for IOPub output. No exception is recorded into the notebook output and the corresponding test is not marked as failed. After that, we are trying to execute the following cells. But the kernel doesn't respond (it seems it is actually dead), and after timeout we again does not record the exception. Finally, we see that no cells gave us exceptions and mark all test as passed, which is definitely wrong.

jhamrick:
Updated the help string as per your comment.

minrk:
:+1: I'll let @takluyver merge in case he has another comment.

PR #260, opened 2016-03-02
Allowed timeout to be none

sandyshao:
Hi all, I am an undergraduate student at Berkeley working with Matthias. This is to fix the timeout problem. I set the allow_none to truw to make sure the notebook cell will continue to work. Thanks in advance for all the suggestions and advices!

PR #264, opened 2016-03-08
Cleanup and document that `timeout` can be None, -1

### LINKED ISSUE #256: Option to disable ExecutePreprocessor.timeout
Is there a value for `--ExecutePreprocessor.timeout` that will disable the per-cell timer? I want my notebooks cells to run until completion, no matter how long that takes.

I'd venture further and propose that the default should be to have no timeout. As the notebook ecosystem progresses, I expect people to want to automate their analysis pipelines, which will include notebooks. When you're using nbconvert to run all inplace, the timeout is unwanted. It's as if `python script.py` had a timeout enabled by default.

Carreau:
Cf #260 for implementation and closes #256 

Also add a preventive header in a non-git-tracked file, that it is auto generated. 
@sandyshao worked on #260 so adding here to the issue so that she can
see how documentation is added to a project. 

@sandyshao once this get merged, you will be able to see the docs automatically updating [here](http://nbconvert.readthedocs.org/en/latest/).

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -45,7 +45,7 @@ class ExecutePreprocessor(Preprocessor):
             If a cell execution takes longer, an exception (TimeoutError
             on python 3+, RuntimeError on python 2) is raised.
 
-            `None`, or `0` will disable the timeout.
+            `None` or `-1` will disable the timeout.
             """
         )
     )
@@ -184,7 +184,7 @@ class ExecutePreprocessor(Preprocessor):
         while True:
             try:
                 timeout = self.timeout
-                if timeout <= 0:
+                if timeout < 0:
                     timeout = None
                 msg = self.kc.shell_channel.get_msg(timeout=timeout)
             except Empty:
--- a/docs/autogen_config.py
+++ b/docs/autogen_config.py
@@ -4,6 +4,10 @@ import os.path
 from nbconvert.nbconvertapp import NbConvertApp
 
 header = """\
+
+.. This is an automatically generated file.
+.. do not modify by hand.
+
 Configuration options
 =====================
 
--- a/docs/source/execute_api.rst
+++ b/docs/source/execute_api.rst
@@ -86,7 +86,8 @@ The ``timeout`` traitlet defines the maximum time (in seconds) each notebook
 cell is
 allowed to run, if the execution takes longer an exception will be raised.
 The default is 30 s, so in cases of long-running cells you may want to specify
-an higher value.
+an higher value. The ``timeout`` option can also be set to ``None`` or ``-1``
+to remove any restriction on execution time.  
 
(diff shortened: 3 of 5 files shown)


minrk:
From the discussion in #256, the conclusion was that we should document that `-1` is an infinite timeout, and `0` should *not* be an infinite timeout. I also wouldn't document "any negative integer", because it should only be -1, even if other values work.

Carreau:
> 0 should not be an infinite timeout

Ok, I misunderstood this part. I thought that 0 was making things fail immediately so would not make sens according to last `dhimmel` behavior. Happy to revert 0 to failing immediately behavior.

> I also wouldn't document "any negative integer"

Ok, documented only `None` and `-1`

Carreau:
Ping @takluyver who was also in the other PR.

nbconvert/preprocessors/execute.py
```diff
@@ -45,7 +45,7 @@ class ExecutePreprocessor(Preprocessor):
             If a cell execution takes longer, an exception (TimeoutError
             on python 3+, RuntimeError on python 2) is raised.
 
-            `None`, or `0` will disable the timeout.
+            `None` or `-1` will disable the timeout.
```
takluyver: Do we have single-backticks-are-double-backticks configured for nbconvert's docs? If not, I'd be fine with just leaving these un-decorated. Or configure that option.

nbconvert/preprocessors/execute.py
```diff
@@ -184,7 +184,7 @@ def run_cell(self, cell):
         while True:
             try:
                 timeout = self.timeout
-                if timeout <= 0:
```
takluyver: I was suggesting that we document what was already happening, rather than change behaviour. If `0` does already work to suspend the timeout, that's fine. I thought it didn't, but I may have been mistaken. I don't see any need for 0 to be instant-timeout, because execution always takes some time, even if it's very small.

nbconvert/preprocessors/execute.py
Carreau: I'm matching the style of other helps in this file:

```

    allow_errors = Bool(
        False, config=True,
        help=dedent(
            """
            If `False` (default), when a cell raises an error the
            execution is stoppped and a `CellExecutionError`
            is raised.
            If `True`, execution errors are ignored and the execution
            is continued until the end of the notebook. Output from
            exceptions is included in the cell output in both cases.
            """
        )
    )

```

I dont 'care if we decorate or not. I can double them all.

nbconvert/preprocessors/execute.py
Carreau: > If 0 does already work to suspend the timeout, that's fine

It does not,  it times out immediately so the `<=` introduce by previous PR did actually change the behavior. 
I'm just reverting what it did.

I personally think that `0` , `-1`, and `None` should have the same effect, but that the opposite of what was decide in previosu PR.

PR #270, opened 2016-03-09
Update docs to be consistent with other Jupyter docs

willingc:
Here's a first pass at reorganizing the table of contents.

### DIFF
--- a/docs/conda_env.yml
+++ b/docs/conda_env.yml
@@ -1,11 +1,10 @@
 name: nbconvert_docs
-channels:
-- asmeurer
 dependencies:
-- pandoc
+- nbconvert
 - nbformat
+- jupyter
+- notebook
 - jupyter_client
 - sphinx
-- pip:
-    - nbsphinx
-    - entrypoints
+- nbsphinx
+- entrypoints
--- a/docs/conda_env.yml
+++ b/docs/conda_env.yml
@@ -2,3 +2,5 @@ name: nbconvert_docs
 dependencies:
 - sphinx
 - nbconvert
+- pip:
+    - nbsphinx
--- a/docs/conda_env.yml
+++ b/docs/conda_env.yml
@@ -1,3 +1,4 @@
 name: nbconvert_docs
 dependencies:
 - sphinx
+- nbconvert
--- a/docs/conda_env.yml
+++ b/docs/conda_env.yml
@@ -1,10 +1,6 @@
 name: nbconvert_docs
 dependencies:
-- nbconvert
-- nbformat
-- jupyter
-- notebook
-- jupyter_client
-- sphinx
-- nbsphinx
-- entrypoints
+- sphinx>=1.3
+- sphinx_rtd_theme
+- pip:
+    - nbsphinx
--- a/docs/conda_env.yml
+++ b/docs/conda_env.yml
@@ -1,7 +1,3 @@
 name: nbconvert_docs
 dependencies:
-- sphinx>=1.3
-- sphinx_rtd_theme
-- pygments==2.1.1
-- pip:
-    - nbsphinx
+- sphinx
--- a/docs/conda_env.yml
+++ b/docs/conda_env.yml
@@ -2,5 +2,6 @@ name: nbconvert_docs
 dependencies:
 - sphinx>=1.3
 - sphinx_rtd_theme
+- pygments==2.1.1
 - pip:
     - nbsphinx
--- a/docs/conda_env.yml
+++ b/docs/conda_env.yml
@@ -2,5 +2,7 @@ name: nbconvert_docs
 dependencies:
 - sphinx
 - nbconvert
+- jupyter_client
 - pip:
     - nbsphinx
+    - entrypoints
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -1,6 +1,11 @@
 Changes in nbconvert
 ====================
 
+4.2 (in development)
+--------------------
+
+``4.2 on GitHub - Add link once release is tagged``
+
 - allow nbconvert reading from stdin with "--stdin" option (write into "notebook" basename)
 
 4.1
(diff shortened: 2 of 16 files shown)


willingc:
I'm sorry for the size of this PR. The PR contains:
- updates the table of contents to be consistent with the other Jupyter docs
- make the conda_env.yml similar to the other Jupyter projects and use Python 3 
- lots of edits to the user guide docs including adding subsections, adding more user friendly example names, and improving (I hope!) contents

minrk:
I seem to recall that the docs didn't build on RTD when using Python 3, and that's why py2 was used. Can you double check that it works with py3?

docs/conda_env.yml
```diff
@@ -1,11 +1,8 @@
 name: nbconvert_docs
-channels:
-- asmeurer
 dependencies:
-- pandoc
-- nbformat
-- jupyter_client
 - sphinx
+- nbconvert
```
minrk: nbconvert shouldn't be included here, because it's installed as part of readthedocs.yml. If we install nbconvert to build the docs, it should get the latest stable version, not the current one.

docs/source/changelog.rst
```diff
@@ -1,6 +1,11 @@
 Changes in nbconvert
 ====================
 
+4.2 (in development)
+--------------------
+
+``4.2 on GitHub - Add link once release is tagged``
```
minrk: The milestone link always exists: https://github.com/jupyter/nbconvert/milestones/4.2

readthedocs.yml
```diff
@@ -1,5 +1,4 @@
 conda:
-    file: docs/environment.yml
+    file: docs/conda_env.yml
```
minrk: I think this should stay as `environment.yml`. That is the standard name for conda environment files.

readthedocs.yml
```diff
@@ -1,5 +1,4 @@
...
 python:
-    version: 2 # pandoc fails with Python 3 (missing libgmp??)
-    setup_py_install: true
```
minrk: setup_py_install is needed, I think.

docs/conda_env.yml
willingc: I will double check this.

docs/source/changelog.rst
willingc: Excellent. I didn't know that. Thanks for pointing it out :sunny:

readthedocs.yml
willingc: No problem. We've used both conda_env.yml in most of the other repos but I think going with the standard environment.yml makes more sense.

willingc:
@minrk So I built a pandoc 1.15.2 package based off of the pandoc in [conda-recipes](https://github.com/conda/conda-recipes) and uploaded to my Anaconda channel. Although RTD seems to have some issue with wiping the cached environment from prior builds completely, the docs will now build cleanly with Python 3 and setup.py install may be set to false.

minrk:
By setting `setup_py_install` to False, it's actually building the API docs for the last release of nbconvert, rather than the current version in the repo at build time. I think we still need `setup_py_install` to be True.

takluyver:
The alternative to `setup_py_install` is that we can add the source directory of nbconvert to `sys.path` in `conf.py`.

minrk:
@takluyver true. In either case, we shouldn't be adding `nbconvert` as a dependency in the docs environment, because that allows this kind of mismatch. We want to get an ImportError trying to make the API docs if the current version isn't on sys.path (whether by installation or conf.py)

willingc:
@minrk I made the change you requested. I don't believe nbconvert is a dependency any longer in the docs environment.

I'm still getting one exception when Sphinx tries to autodoc of postprocessors.ServePostProcessor while building the docs (using the conda environment in the PR or the current environment found in master). The built docs seem to have the class in the API docs. 

```
(nbconvert_docs) ~pj/nbconvert/docs   doc-outline ●  make html
sphinx-build -b html -d build/doctrees   source build/html
Running Sphinx v1.3.6
making output directory...
loading pickled environment... not yet created
loading intersphinx inventory from https://docs.python.org/objects.inv...
loading intersphinx inventory from http://jinja.pocoo.org/docs/dev/objects.inv...
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 18 source files that are out of date
updating environment: 18 added, 0 changed, 0 removed
reading sources... [100%] usage
/Users/carol/ProjectJup/nbconvert/docs/source/api/postprocessors.rst:18: WARNING: autodoc: failed to import class 'ServePostProcessor' from module 'postprocessors'; the following exception was raised:
  File "/Users/carol/anaconda/envs/nbconvert_docs/lib/python3.5/site-packages/sphinx/ext/autodoc.py", line 386, in import_object
    __import__(self.modname)
ImportError: No module named 'postprocessors'
```

willingc:
@minrk I fixed the build warning caused by an out of date autodoc command. Fixed all the links (but 2) that threw errors when running ``make linkcheck``. Also added a docstring for the autogen_config.py.

This is ready for re-review.

docs/source/usage.rst
```diff
@@ -58,26 +78,43 @@ The currently supported export formats are:
...
-  If there were any magics in an Jupyter notebook, this may only be executable from
-  an Jupyter session.
-
+  This is the simplest way to get a Python (or other language, depending on
+  the kernel) script out of a notebook. If there were any magics in an
+  Jupyter notebook, this may only be executable from a Jupyter session.
+  
+  For example, to convert a Julia notebook to a Julia executable script::
+      
+      jupyter nbconvert --to script my_julia_notebook.ipynb my_julia_script.jl
```
takluyver: We shouldn't need to pass the name of the output file (and I'm not sure this is the syntax if we do). It should pick up the file extension automatically from the kernel.

willingc: I forgot the --output. But likely simpler to just leave it the output file out and just use ``jupyter nbconvert --to script my_julia_notebook.ipynb``. Good catch.

minrk:
Thanks!

PR #274, opened 2016-03-26
Use upright quotes for code in LaTeX

mgeier:
I'm not quite sure if that's the right fix, but at least it seems to fix the problem.

### DIFF
--- a/nbconvert/templates/latex/base.tplx
+++ b/nbconvert/templates/latex/base.tplx
@@ -13,6 +13,7 @@ This template does not define a docclass, the inheriting class must define this.
     ((* block docclass *))((* endblock docclass *))
     
     ((* block packages *))
+    \usepackage[T1]{fontenc}
     % Nicer default font than Computer Modern for most use cases
     \usepackage{palatino}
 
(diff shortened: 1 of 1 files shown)


takluyver:
I've noticed before that it uses smart quotes in code when converted to PDF. With this, if you copy and paste code to somewhere else, is it valid? I.e. do quotes show up as the regular ascii `'` and `"` quotes?

mgeier:
Good point!

I just tried it and the correct quotes show up after copy/paste, which wasn't the case before this PR.

But there is a different problem now: When copy/pasting from Pygmentized code, each single quote is surrounded by spaces! In the PDF it looks OK, this only appears after pasting.

I don't know what causes this ... any ideas?

For code without highlighting (e.g. `stdout`), everything is fine, i.e. there are no additional spaces.
I also tried it with Sphinx (using `nbsphinx`), which uses very similar LaTeX code, but there this error does *not* appear. I haven't figured out yet what's the actual difference between the two ...

mgeier:
I tried a few things from Sphinx's LaTeX output, and I found out that the unwanted spaces disappear when using this:

     \usepackage[T1]{fontenc}

AFAIK, this is a quite common thing to do in LaTeX, is there a reason why this should not be used in `nbconvert`'s LaTeX templates?

minrk:
@mgeier probably an oversight. If that fixes it , let's go with that.

mgeier:
OK, I rebased my first commit and I added `\usepackage[T1]{fontenc}`.

minrk:
Thanks!

PR #276, opened 2016-03-29
Incorrectly transforms multiple markdown email

bollwyvl:
This was uncovered in the wild:
https://github.com/jupyter/nbviewer/issues/582

From the [spec](https://daringfireball.net/projects/markdown/syntax):
```
Markdown supports a shortcut style for creating “automatic” links for URLs and email
addresses: simply surround the URL or email address with angle brackets. What this
means is that if you want to show the actual text of a URL or email address, and also
have it be a clickable link, you can do this:

<http://example.com/>

Markdown will turn this into:

<a href="http://example.com/">http://example.com/</a>
Automatic links for email addresses work similarly, except that Markdown will also
perform a bit of randomized decimal and hex entity-encoding to help obscure your
address from address-harvesting spambots. For example, Markdown will turn this:

<address@example.com>
```

This adds a minimum failing test, looking at solutions... probably regex-fu...

### DIFF
--- a/nbconvert/filters/tests/test_markdown.py
+++ b/nbconvert/filters/tests/test_markdown.py
@@ -178,9 +178,8 @@ i.e. the $i^{th}$"""
             self._try_markdown(markdown2rst, test, tokens[index])
 
     def test_multiple_email_links(self):
-        cases = {'<d.e@f.g>': '<a href="mailto:a@b.c">a@b.c</a>. '}
-        for md, html in cases.items():
-            self.assertIn(html, markdown2html(md))
+        # https://github.com/jupyter/nbviewer/issues/582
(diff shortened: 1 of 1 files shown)


takluyver:
I don't think Github does this either - at least in the preview:

<address@example.com>

takluyver:
I thought @bollwyvl was referring to the obfuscation in the second paragraph above, which seemingly the major markdown implementations don't do. The issue is actually with having a second email address link on a single line.

michaelpacer:
When I run the originally linked error locally using the current verison of nbconvert (5.0.0dev), and older versions (4.0, 4.1, 4.2) and process --to html, I don't see this.

I think this is because it was fixed in mistune. 

Do we want to include a test for something that mistune no longer errs on?

michaelpacer:
I'm going to close this, since it's fixed in the dependency that was causing the original error.

PR #284, opened 2016-04-06
Add/revise docstrings for utils directory and nbconvertapp.py

willingc:
As both a follow-up to the dev meeting and an opportunity to really get to know the nbconvert codebase, I have begun editing and/or adding docstrings. In this PR, the `utils` directory files and `nbconvertapp.py` have been done.

Also a minor clarification in the README on pandoc installation.

### DIFF
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -33,10 +33,9 @@ from .utils.io import unicode_stdin_stream
 #-----------------------------------------------------------------------------
 
 class DottedOrNone(DottedObjectName):
+    """A string holding a valid dotted object name in Python, such as A.b3._c
+    Also allows for None type.
     """
-    A string holding a valid dotted object name in Python, such as A.b3._c
-    Also allows for None type."""
-    
     default_value = u''
 
     def validate(self, obj, value):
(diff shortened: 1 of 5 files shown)


nbconvert/nbconvertapp.py
```diff
@@ -238,21 +238,20 @@ def initialize(self, argv=None):
 
 
     def init_syspath(self):
-        """
-        Add the cwd to the sys.path ($PYTHONPATH)
-        """
+        """Add the cwd to the sys.path ($PYTHONPATH)"""
```
takluyver: In general, I would consider changes like this one purely stylistic cleanup, which we prefer to avoid.

nbconvert/nbconvertapp.py
```diff
@@ -360,6 +367,17 @@ def write_single_notebook(self, output, resources):
...
+        ----------
+        output :
+        resources : dict
+            resources for a single notebook including name, config directory
+            and directory to save output
+
+        Returns
+        -------
+        file
+            results from the specified writer output of exporter
```
takluyver: Changes like this which improve the docstrings are welcome, however :-)

nbconvert/nbconvertapp.py
willingc: I understand that approach with executable code and, in general, I would leave style as it is within a document if I wasn't editing the document already. I'm happy to put it back as it was. Though, I could make the case that beautiful docstrings encourage more docstrings ;-)

Carreau: One thing is that reformatting for style screw up git blame, but I thins it's ok if there is just small change like that for this time.

willingc: Fair enough. Thanks.

PR #286, opened 2016-04-07
Improve docs navigation to supported output types

willingc:
This PR contains:

- Improved links and navigation to the different output types and their commands
- Minor formatting on change log
- Updated introduction to document


Carreau:
Thanks !

Issue #281, opened 2016-04-05
pdflatex compilation error with emojis and other utf8


steveyang95:
If you try to have emojis or other utf8 characters in the notebook, then this will result in an error because certain utf8 characters are unrecognizable to the pdflatex compiler. For example, try to export a notebook into pdf with this emoji (😁). Because pdflatex is not very comprehensive for utf8, what if instead of pdflatex, XeLatex is used? XeLatex "natively supports Unicode and the input file is assumed to be in UTF-8 encoding by default".

takluyver:
The latex command is configurable (`PDFExporter.latex_command`). I think pdflatex is the default as the lowest common denominator, rather than because it's the best.

minrk:
I, too, always use xelatex. I don't know enough about latex installations to say if we should switch nbconver to xelatex-by-default, or xelatex-if-available. It would make more sense in my own environments, at least.

takluyver:
I am wary of if-available, because that means more potential bugs and
another variable for users to think about when stuff goes wrong.

minrk:
Very true, though it's perhaps not the best to stick with the worst implementation (pdflatex) just because it's the oldest and most likely to be present. I'm not sure how prevalent xelatex installs are, so I don't know how often people are likely to have pdflatex and not xelatex. If that number is small, I'd be inclined to switch the default to xelatex with no check.

ryanlovett:
Debian has some data on [popularity](http://popcon.debian.org). texlive-xetex has rank 4199/154308. Of 8753 installs, 3075 use it regularly (during apt-get trigger?). It has [four open bugs](https://bugs.debian.org/cgi-bin/pkgreport.cgi?pkg=texlive-xetex;dist=stable) in Debian. pdflatex is provided by texlive-latex-base so they're not really comparable, but it has rank 2292/154308; of 29796 installs, 8971 use it regularly.

+1 for it being the default, though that is for our own needs and not based on the above.

naicode:
Given that `xetex` (xelatex) and `pdflatex` are not drop-in replacements for each other and using `pdflatex` for Notebooks with e.g. Chinese character (or any Unicode character not working with the current template witch seem to be  alot) is completely *broken* I would highly recommend against using pdflatex.

[...]

PR #303, opened 2016-05-02
use xelatex by default

minrk:
xelatex seems to be strictly more capable than pdflatex, and a more sensible default in 2016.

Just like with pdflatex -> xelatex, people can use config to pick another implementation if the default doesn't suit them for whatever reason.

cf #281

### DIFF
--- a/nbconvert/exporters/pdf.py
+++ b/nbconvert/exporters/pdf.py
@@ -35,7 +35,7 @@ class PDFExporter(LatexExporter):
         help="How many times latex will be called."
     )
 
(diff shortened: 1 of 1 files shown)


takluyver:
Expect an influx of bug reports that conversion to PDF has stopped working. ;-)

janschulz:
Just checked: at least in the [upcoming windows miktex package in conda-forge](https://github.com/conda-forge/staged-recipes/pull/469), it is included :-)

After that it will be ...

```
conda install -c conda-forge nbconvert pandoc miktex-portable
```

...in the root environment (which is always in PATH) and you can run a nbconvert to pdf :-)

PR #308, opened 2016-05-07
Adding an index to the pdf

damontallen:
This modification adds an index to the pdf viewer when a pdf is created from LaTeX.


minrk:
@damontallen thanks! It looks like the merge commit has caused this to not apply anymore. Can you rebase on master?

 @fperez what do you think about PDFs having an index by default?

damontallen:
@minrk Rebased.

damontallen:
This appears to only be an issue with larger notebooks.

PR #321, opened 2016-05-31
traitlets 4.2

minrk:
I caught one or two other deprecations in the process (e.g. use str/unicode-stable `binascii.a2b_base64` instead of `base64.decodestring`)


nbconvert/preprocessors/extractoutput.py
```diff
@@ -76,8 +78,7 @@ def preprocess_cell(self, cell, resources, cell_index):
 
                         # data is b64-encoded as text (str, unicode)
                         # decodestring only accepts bytes
```
takluyver: This comment can go now.

takluyver:
:+1:

nbconvert/preprocessors/extractoutput.py
minrk: Right, fixed.

PR #336, opened 2016-07-14
Add kernel-specific templates

flying-sheep:
this adds kernel-specific templates and removes caching. several runs of the test suite showed that caching had no effect on its runtime that is above noise level.

### DIFF
--- a/nbconvert/exporters/tests/test_kernel_templates.py
+++ b/nbconvert/exporters/tests/test_kernel_templates.py
@@ -1,0 +1,26 @@
+"""Tests for Kernel-specific templates"""
+
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+from testpath import modified_env
+
+from .base import ExportersTestsBase
+from ..markdown import MarkdownExporter
+
+
+class TestKernelTemplates(ExportersTestsBase):
+    """Tests for Kernel-specific templates"""
+
+    exporter_class = MarkdownExporter
+    should_include_raw = ['markdown', 'html']
+
+    def setUp(self):
+        super(TestKernelTemplates, self).setUp()
+        self.env = modified_env({'JUPYTER_PATH': self._get_files_path()})
+
+    def test_export(self):
+        """Will the kernel-specific template be used?"""
+        with self.env:
+            output, resources = MarkdownExporter().from_filename(self._get_notebook('kernel_template.ipynb'))
+        self.assertIn('Test md code: some *markdown*', output)
--- a/nbconvert/exporters/exporter.py
+++ b/nbconvert/exporters/exporter.py
@@ -13,6 +13,8 @@ import copy
 import collections
 import datetime
 
+from ipython_genutils.py3compat import PY3
+from jupyter_client.kernelspec import find_kernel_specs
 from traitlets.config.configurable import LoggingConfigurable
 from traitlets.config import Config
 import nbformat
@@ -123,6 +125,9 @@ class Exporter(LoggingConfigurable):
         """
         nb_copy = copy.deepcopy(nb)
         resources = self._init_resources(resources)
+
+        # if found, add a kernel name
+        resources['kernel_name'] = nb['metadata'].get('kernelspec', {}).get('name', None)
         
         if 'language' in nb['metadata']:
             resources['language'] = nb['metadata']['language'].lower()
--- a/nbconvert/exporters/latex.py
+++ b/nbconvert/exporters/latex.py
@@ -76,8 +76,8 @@ class LatexExporter(TemplateExporter):
                              Highlight2Latex(pygments_lexer=lexer, parent=self))
         return super(LatexExporter, self).from_notebook_node(nb, resources, **kw)
 
-    def _create_environment(self):
-        environment = super(LatexExporter, self)._create_environment()
+    def get_environment(self, kernel_name):
+        environment = super(LatexExporter, self).get_environment(kernel_name)
 
         # Set special Jinja2 syntax that will not conflict with latex.
         environment.block_start_string = "((*"
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -19,6 +19,8 @@ from traitlets import HasTraits, Unicode, List, Dict, default, observe
 from ipython_genutils.importstring import import_item
 from ipython_genutils import py3compat
 
+from jupyter_client.kernelspec import get_kernel_spec, find_kernel_specs, NoSuchKernel
+
 from nbconvert import filters
 from .exporter import Exporter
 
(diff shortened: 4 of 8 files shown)


takluyver:
Kernel specific templates is something that probably requires more discussion, and I'd be happier if that was separate from the purely mechanistic change of removing caching.

It's probably fine to get rid of caching; the caching design itself was a rationalisation of a buggy stateful API. However, our test suite is unlikely to be a good performance test. In particular, if you're always initialising a new Exporter, there's self-evidently no benefit to caching. So I'd like to see some comparisons with an exporter initialised once and then reused many times. If we have to read the template file again each time, it may be meaningful to keep the cache.

Jinja appears to have its own cache of templates, but that's on the environment, so if we recreate the environment each time as well, we can't use that.

flying-sheep:
we could just cache the environments.

and tbh, i mainly removed the caching because i couldn’t get it to work with per-kernel templates. something really weird goes on there

flying-sheep:
uuh, this test failure is really weird, can anyone shed light on it?

takluyver:
I've prodded it to rerun in case it was a sporadic failure.

takluyver:
And it seems it was.

flying-sheep:
oo, how to rerun without closing and reopening again?

takluyver:
Travis has a button to do it, but it's only available if you have push access to the repo.

flying-sheep:
OK, wtf?

[it says it uses the correct template, but the wrong stuff is put out?](https://travis-ci.org/jupyter/nbconvert/jobs/145007659#L389-L395)

or it uses the wrong template, but for some reason `get_source` finds the correct one.

in any case i can’t debug it since it works locally.

takluyver:
That appears to me to be saying it's loading the default template, not your custom one:

```
Loaded template markdown.tpl from .../nbconvert/exporters/../templates/markdown.tpl
```

nbconvert/exporters/tests/test_kernel_templates.py
```diff
@@ -0,0 +1,27 @@
+"""Tests for Kernel-specific templates"""
+
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+from test.support import EnvironmentVarGuard
```
takluyver: https://docs.python.org/3/library/test.html

> The test package is meant for internal use by Python only. It is documented for the benefit of the core developers of Python. Any use of this package outside of Python’s standard library is discouraged as code mentioned here can change or be removed without notice between releases of Python.

We already require testpath for tests, so you can use [testpath.modified_env](http://testpath.readthedocs.io/en/latest/env.html#testpath.modified_env) to achieve this.

flying-sheep:
oh, true. but why? [here](https://github.com/jupyter/nbconvert/pull/336/files#diff-43c2e238630461c0a2c422a0b09fae55R263) i add the kernel path right after the custom path, so the loader should try (and find) stuff there first.

nbconvert/exporters/tests/test_kernel_templates.py
flying-sheep: thanks, will do!

takluyver:
Nothing springs to mind, sorry.

flying-sheep:
OK, this should help. i’m logging the shit out of this and duplicating the code used to search for template paths

flying-sheep:
uuh, my search path [isn’t added](https://travis-ci.org/jupyter/nbconvert/jobs/145066017#L395):

> ```py
['.',
 '.../site-packages/nbconvert/exporters/../templates',
 '.../site-packages/nbconvert/exporters/../templates/skeleton']

flying-sheep:
wtf? why is only the python3 kernel found [within those kernel dirs](https://travis-ci.org/jupyter/nbconvert/jobs/145073722#L394-L396)

`def find_kernel_specs()` [contains only](https://github.com/jupyter/jupyter_client/blob/112095fe0e85d5857c9150893b88b2437124db13/jupyter_client/kernelspec.py#L281-L283) `return KernelSpecManager().find_kernel_specs()`

---

# wait

before i can understand this, please explain how `kernel_dirs` gets set at all: [i see no `@default`](https://github.com/jupyter/jupyter_client/blob/112095fe0e85d5857c9150893b88b2437124db13/jupyter_client/kernelspec.py#L109) here…

takluyver:
The @default decorator is a change in the latest version of traitlets, and that code hasn't been updated yet. It's relying on the older API where a traitlet named `foo` looks for a method `_foo_default()`.

takluyver:
I suspect the issue is that you need to add the test files to package_data here so that they get installed:

https://github.com/jupyter/nbconvert/blob/a0a4a5ff3f364666640fea7916e32fbc73bc55c2/setup.py#L55

flying-sheep:
aaah, makes sense. but may i ask why you install tests at all?

takluyver:
Different approaches - some people like to be able to install a package and run its tests, and Jupyter code is generally structured that way, with tests inside the package.

flying-sheep:
OK. what do you say about this PR?

takluyver:
Much the same as I said before:

* You're making two quite different changes, I'd be happier if they were two separate PRs we could discuss individually.
* I'd like to see some more relevant performance measurements before we rip out all the caching, because the tests are not designed for that. It might be fine, or we may want to keep the environment caching, and let Jinja handle template caching.
* I'd like to have a bit more discussion with the other devs about whether and how to implement kernel specific templates. E.g. is it better to allow kernels to override exporters, rather than templates? Could we handle the case you want this for better by adding more flexibility in the current template, plus something in notebook metadata? What are the merits of kernels overriding existing export formats versus adding new ones?

flying-sheep:
i’m happy to split this, but as said: i didn’t get it to work with caching and i suspect this is at least partly due to bugs in caching.

so yeah, we should probably decide first if it’s worth to cache, before i adapt this all to caching only to have caching ripped out later for buggyness and uselessness..

> Could we handle the case you want this for better by adding more flexibility in the current template, plus something in notebook metadata?

i though about this as well and think it’s also a good idea. maybe we should extend the kernel.json spec to have a `"resources": { ... }` or `"template_data": { ... }` object that’s merged into the resources?

janschulz:
Just to chime in: I'm also helping out in the development of a environmental kernel manager (register all conda/virtualenv environments as kernels, if they have irkernel or the ipykernel installed: https://github.com/Cadair/jupyter_environment_kernels). Part of that is that we currently (or better after https://github.com/Cadair/jupyter_environment_kernels/pull/24)  "set" our own kernel spec dir to get logos to the UI... 

I'm not sure how hard it would be to find the right dir in the env itself, but this felt easier (and is also the thing which the current conda implementation of such an kernel env manager does).

Carreau:

flying-sheep:
There was no progress as I was waiting for an answer to

> so yeah, we should probably decide first if it’s worth to cache, before i adapt this all to caching only to have caching ripped out later for buggyness and uselessness.

So what about it?

PR #343, opened 2016-07-22
Graceful kernel shutdown by ExecutePreprocessor

netj:
A new `ExecutePreprocessor.kill_kernel_at_end` option controls whether it should kill the kernel immediately (as before) or allow it to shutdown properly (new default), e.g., executing all its `atexit` hooks.  The previous behavior of immediately terminating kernel causes many carefully prepared libraries in notebooks to result in a unexpected state after executing them with nbconvert.

Fixes Calysto/matlab_kernel#11

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -109,9 +109,9 @@ class ExecutePreprocessor(Preprocessor):
         default_value='graceful',
         help=dedent(
             """
-            If `graceful` (default), then the kernel will given time to clean
+            If `graceful` (default), then the kernel is given time to clean
             up after executing all cells, e.g., to execute its `atexit` hooks.
-            If `immediate` or `now, then the kernel is signaled to immediately
+            If `immediate`, then the kernel is signaled to immediately
             terminate.
             """
             )
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -103,7 +103,16 @@ class ExecutePreprocessor(Preprocessor):
             """
             )
     ).tag(config=True)
-
+    
+    kill_kernel_at_end = Bool(False,
+        help=dedent(
+            """
+            If `False` (default), then the kernel will given time to clean up
+            after executing all cells, e.g., to execute its `atexit` hooks.
+            If `True`, then the kernel is signaled to immediately terminate.
+            """
+            )
+    ).tag(config=True)
 
     def preprocess(self, nb, resources):
         """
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks, this looks straightforward and logical.

My only question is whether we can make the naming clearer. The boolean `kill_kernel_at_end` makes me think that if it's False, it will just leave the kernel running.

Maybe `shutdown_option='kill'` or `shutdown_option='shutdown'`? I'm not precisely happy with that either, though. Maybe @willingc can help think of a good design.

willingc:
Thanks for the PR. The approach makes good sense. 

I agree that it's worth a bit of extra time to work on naming. Here's my thought process:

1. Currently, shutdown of the kernel will happen at the completion of ExecutePreprocessor (run all cells in the notebook).
2. The type of shutdown will be either 'graceful' or 'immediate or now'.

Perhaps:

``shutdown_execution='graceful'`` or ``shutdown_execution='immediate'``. 'graceful' would be the default.

takluyver:
:+1: to that suggestion, except that I'd call it `shutdown_kernel` instead of `shutdown_execution`.

netj:
@takluyver @willingc Thanks for taking a look!  Renamed according to your suggestion.  I agree the original name was misleading.

nbconvert/preprocessors/execute.py
```diff
@@ -103,7 +103,18 @@ class ExecutePreprocessor(Preprocessor):
             """
             )
     ).tag(config=True)
-
+    
+    shutdown_kernel = Enum(['graceful', 'immediate', 'now'],
```
takluyver: Sorry to quibble, but I don't like having two values that mean the same thing. I'd keep 'immediate' and get rid of 'now'.

willingc: I agree that it would be best to remove 'now'. Thanks so much for helping with this PR. It should be a helpful improvement.

netj: Appreciate the careful look!  Updated. PTAL

nbconvert/preprocessors/execute.py
```diff
@@ -103,7 +104,18 @@ class ExecutePreprocessor(Preprocessor):
...
     ).tag(config=True)
-
+    
+    shutdown_kernel = Enum(['graceful', 'immediate'],
+        default_value='graceful',
+        help=dedent(
+            """
+            If `graceful` (default), then the kernel will given time to clean
+            up after executing all cells, e.g., to execute its `atexit` hooks.
+            If `immediate` or `now, then the kernel is signaled to immediately
```
damianavila: missing quote after `now`

damianavila: now reading the other comments, I see `now` is a leftover... so, just need to remove it from the docstring as well...

netj: Thanks for catching this!  cleaned now.

willingc:
@netj Thanks so much. The latest changes look really good. :cake:

@takluyver This looks ready to merge 👍

takluyver:
Thanks @netj for your patience as we've worked out how to do this. In it goes!

PR #345, opened 2016-07-27
allow absolute path templates to be pass in

benjamincaldwell:
Currently, when you pass an absolute path in for the template nbconvert fails to find the template. This change accepts both relative and absolute paths.

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -107,7 +107,7 @@ class TemplateExporter(Exporter):
 
     default_template = Unicode(u'').tag(affects_template=True)
 
-    template_path = List(['.']).tag(config=True, affects_environment=True)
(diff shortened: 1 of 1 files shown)


nbconvert/exporters/templateexporter.py
```diff
@@ -107,7 +107,7 @@ def _template_file_default(self):
 
     default_template = Unicode(u'').tag(affects_template=True)
 
-    template_path = List(['.']).tag(config=True, affects_environment=True)
+    template_path = List(['.', '/']).tag(config=True, affects_environment=True)
```
takluyver: I'd like to find a way to handle absolute paths that's not also searching relative paths starting from the root dir. Also, we probably need to think about Windows.

minrk:
#398 allows specifying templates by absolute path without adding the full filesystem to the template path.

Issue #346, opened 2016-07-27
Timeout that depends on cell content


j2kun:
I would like to set the ExecutePreprocessor timeout to a callable that computes the timeout based on the contents of the cell.

I see the place in the code that would make this change very easy:

```
    def run_cell(self, cell):
        msg_id = self.kc.execute(cell.source)
        self.log.debug("Executing cell:\n%s", cell.source)
        # wait for finish, with timeout
        while True:
            try:
                timeout = self.timeout
                if timeout < 0:
                    timeout = None
                msg = self.kc.shell_channel.get_msg(timeout=timeout)
```

Instead using something like

```
    def run_cell(self, cell):
            ...
            try:
                if callable(self.timeout):
                    timeout = self.timeout(cell.source)
                else:
                    timeout = self.timeout
                ...
```

I'm not submitting a PR because I also noticed that the attributes have some [strange type-checking](https://github.com/jupyter/nbconvert/blob/master/nbconvert/preprocessors/execute.py#L48) that I don't understand. If someone showed me how to do the proper type checking with a callable, I could submit a PR.

minrk:
That sounds interesting. I wouldn't add further complex cases to the `timeout` trait, but a separate `timeout_func` that returns an integer or None as an `Any` trait should be doable.

j2kun:
This feature was merged into master. Closing.

PR #347, opened 2016-07-28
preprocessors/execute: added timeout_func for variable cell timeouts

j2kun:
Implements `timeout_func` from https://github.com/jupyter/nbconvert/issues/346

Note I couldn't figure out how to get fully passing tests, as I was getting a bunch of errors like `FileNotFoundError: '<snip>/nbconvert/resources/style.min.css'` when I ran `py.test`. Do you use a different testing framework? If so, could you add a short description in the readme of the process of setting up a local dev version to run the tests specifically?

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -12,7 +12,7 @@ try:
 except ImportError:
     from Queue import Empty  # Py 2
 
-from traitlets import List, Unicode, Bool, Enum
+from traitlets import List, Unicode, Bool, Enum, Any
 
 from nbformat.v4 import output_from_msg
 from .base import Preprocessor
@@ -35,7 +35,7 @@ class CellExecutionError(ConversionException):
         if not isinstance(s, str):
             s = s.encode('utf8', 'replace')
         return s
-    
+
     def __unicode__(self):
         return self.traceback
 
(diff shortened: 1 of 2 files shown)


nbconvert/preprocessors/execute.py
```diff
@@ -197,8 +217,12 @@ def run_cell(self, cell):
         # wait for finish, with timeout
         while True:
             try:
-                timeout = self.timeout
-                if timeout < 0:
+                if self.timeout_func is not None:
+                    timeout = self.timeout_func(cell.source)
```
minrk: Can you pass the cell itself, rather than cell.source? It's conceivable that one could write a timeout_func that looks at metadata, etc.

minrk:
Tests are run with nose, as seen in .travis.yml, though py.test does work fine as well. If it's missing style.min.css, that suggests that maybe it hasn't been installed, or was installed with `pip install .` (no `-e`), and run from the repo, in which case the files generated during the install process would be in the installed package, but the not-installed source repo would be imported for the tests.

j2kun:
Updated the PR.

As an aside: in order to get it working with both `py.test` and `nosetests` from a fresh virtualenv, I had to install the following packages which were not included from `pip install -e .`

```
nose
pytest
jupyter   # for some reason it needs the client, not just core
testpath
```

(this still skipped a handful of tests)

Would be nice to include these in a requirements-dev.txt or something

takluyver:
Doing this should work:

```
pip install -e .[test]
```

j2kun:
It might be my shell? I get 

```
$ pip install -e .[test]
zsh: no matches found: .[test]
```

j2kun:
I see the test is failing, but I can't reproduce it locally:

```
$ py.test -v --pyargs nbconvert --ignore venv 
...
nbconvert/preprocessors/tests/test_execute.py::TestExecute::test_allow_errors PASSED
nbconvert/preprocessors/tests/test_execute.py::TestExecute::test_constructor PASSED
nbconvert/preprocessors/tests/test_execute.py::TestExecute::test_disable_stdin PASSED
nbconvert/preprocessors/tests/test_execute.py::TestExecute::test_empty_path PASSED
nbconvert/preprocessors/tests/test_execute.py::TestExecute::test_run_notebooks PASSED
nbconvert/preprocessors/tests/test_execute.py::TestExecute::test_timeout PASSED
nbconvert/preprocessors/tests/test_execute.py::TestExecute::test_timeout_func PASSED
...
```

takluyver:
I think in zsh you may have to quote `".[test]"`

takluyver:
Test failure might be a random thing; I've restarted that build.

takluyver:
Thanks, I think this looks good.

j2kun:
Thanks! Do you have a ballpark estimate on when 5.0 will be released?

minrk:
I think we can do 5.0 pretty soon.

PR #348, opened 2016-07-29
run tests with pytest

minrk:
and add a note for running the tests in the dev-install section of the README.

all tests pass without modification, as long as nose is present, since we still use nose.tools in a few places.

We can remove the nose dependency in a later patch.


takluyver:
```
ERROR: file not found: nbconvert
```

PR #351, opened 2016-08-08
Markdown display data priority

### LINKED ISSUE #342: Markdown image extraction issues
I'm trying to convert to markdown, and there seems to be some issues with the image extraction.
It looks like they are always extracted properly, but sometimes no reference is added to the text.
I think this is the case when both PNG and PDF are embedded.

For example for this file:
https://github.com/amueller/introduction_to_ml_with_python/blob/master/02-supervised-learning.ipynb

Or maybe pdf is never embedded in markdown? Should it be converted to SVG?
Or maybe rather: if pdf and png are present, and pdf is not supported, shouldn't it pick up png?

amueller:
Fixes #342.

This adds the same ``display_data_priority`` to ``MarkdownExporter`` that is used in ``HTMLExporter``.
I also added a regression test that checks that an image is included in the output if both png and pdf are present. This fails on master.

### DIFF
--- a/nbconvert/exporters/markdown.py
+++ b/nbconvert/exporters/markdown.py
@@ -30,6 +30,20 @@ class MarkdownExporter(TemplateExporter):
 
     @property
     def default_config(self):
-        c = Config({'ExtractOutputPreprocessor':{'enabled':True}})
+        c = Config({
+            'ExtractOutputPreprocessor':{'enabled':True},
+            'NbConvertBase': {
+                'display_data_priority' : ['application/javascript',
+                                           'text/html',
+                                           'text/markdown',
+                                           'image/svg+xml',
+                                           'text/latex',
+                                           'image/png',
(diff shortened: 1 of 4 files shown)


nbconvert/exporters/markdown.py
```diff
@@ -30,6 +30,20 @@ def _raw_mimetypes_default(self):
 
     @property
     def default_config(self):
-        c = Config({'ExtractOutputPreprocessor':{'enabled':True}})
+        c = Config({
+            'ExtractOutputPreprocessor':{'enabled':True},
+            'NbConvertBase': {
+                'display_data_priority' : ['application/javascript',
```
minrk: Probably want to exclude javascript here, as that won't usually work in markdown. Otherwise, 👍

minrk:
Great! one comment inline, but 👍 after that.

amueller:
done

PR #353, opened 2016-08-10
add -shell-escape flag to xelatex to fix test errors

### LINKED ISSUE #350: Test failure in jack-of-none builds
Problem running Latex:

```
Package ifplatform Warning: 
    shell escape is disabled, so I can only detect \ifwindows.

))

! Package minted Error: You must invoke LaTeX with the -shell-escape flag.

See the minted package documentation for explanation.
 ...                                              
                                                  
l.67     \usepackage
                    [normalem]{ulem} % ulem is needed to support strikethrou...

? 
! Emergency stop.
 ...                                              
                                                  
l.67     \usepackage
                    [normalem]{ulem} % ulem is needed to support strikethrou...

No pages of output.
```

[Test log](https://jenkins.jupyter.org/job/jack-of-none/372/console)

amueller:
Fixes #350.
I'm not entirely sure if this is the right fix. I'm also not sure why travis didn't error on this.

### DIFF
--- a/nbconvert/exporters/pdf.py
+++ b/nbconvert/exporters/pdf.py
@@ -12,6 +12,7 @@ from traitlets import Integer, List, Bool, Instance, Unicode
 from ipython_genutils.tempdir import TemporaryWorkingDirectory
 from .latex import LatexExporter
 
+
 class LatexFailed(IOError):
     """Exception for failed latex run
     
@@ -35,7 +36,7 @@ class PDFExporter(LatexExporter):
         help="How many times latex will be called."
     ).tag(config=True)
 
-    latex_command = List([u"xelatex", u"{filename}"],
+    latex_command = List([u"xelatex", u"-shell-escape", u"{filename}"],
         help="Shell command used to compile latex."
     ).tag(config=True)
 


takluyver:
I suspect Latex isn't installed on Travis, so those tests are skipped. I don't know enough about latex to know if this is the right fix. @michaelpacer, your input would be welcome :-)

janschulz:
I cherry-picked this commit into #356

michaelpacer:
So everything I can find on this would suggest this as fine as anything else (in that what we're doing is allowing LaTeX to have access to the shell directly through the use of the \write18{} command). If we don't trust one of the third party libraries that we're using by default, it's probably better we run into that problem in a Travis build than on a user build.

takluyver:
I don't know much about the way Latex works, but the concern is that the flag would allow shell commands not just by libraries we're using, but by the notebook being converted, which may be untrusted. Like our security model in the browser, it's a matter of expectations: if you explicitly ask for code to be executed, we trust that you understand the implications, but if you are opening/converting a document, you probably don't expect that action to execute arbitrary code, so you may not have checked it beforehand.

As @flying-sheep suggested on #356, if someone wants to do something extra with latex, they can always nbconvert `--to latex` and then run Latex on the result themselves. We also offer a config option for the Latex command to run. But `-shell-escape` does not sound like something that we should enable by default.

takluyver:
Merged #367 instead, so this should no longer be needed.

amueller:
Thanks for looking into it!

PR #355, opened 2016-08-10
asciidoc export

amueller:
I've started on writing an asciidoc exporter.
This is helpful particularly for people that want to write books with O'Reilly, which are a few folks these days, I think.
This PR is not ready for reviews yet, but I wanted to ask

1. Is this something you'd like to have in nbconvert?
2. What kinds of tests would you like to see for adding a new output format?


nbconvert/exporters/__init__.py
```diff
@@ -4,6 +4,7 @@
 from .templateexporter import TemplateExporter
 from .latex import LatexExporter
 from .markdown import MarkdownExporter
+from .asciidoc import ASCIIdocExporter
```
takluyver: I think you forgot to add a new file when you committed ;-)

takluyver:
I think it would be OK to add an asciidoc exporter to nbconvert, assuming it doesn't add much complexity. However, if we decide not to, it should now be easy to distribute it as a separate module, using entry points to register the exporter:

http://nbconvert.readthedocs.io/en/latest/external_exporters.html#registering-a-custom-exporter-as-an-entry-point

nbconvert/exporters/__init__.py
amueller: indeed, fixed now, I thin ;)

amueller:
Either way is fine for me. Adding in here would provide more exposure. I don't see a lot of complications. It's supported by pandoc, so things should be fairly simple. It's mostly tweaking the template.

nbconvert/exporters/asciidoc.py
```diff
@@ -0,0 +1,35 @@
+"""Markdown Exporter class"""
```
amueller: whoops I totally didn't copy & paste this

amueller:
This should be good now. Suggestions for tests welcome.

minrk:
I think asciidoc is appropriate to include, especially if @amueller promises to support it going forward.

amueller:
Yeah I'm happy to help.

michaelpacer:
Added to 5.0, I'll think more about the tests question. 

@amueller what sorts of issues were you running into while debugging this? That might be a good place to start? 

I'm fine shipping it by default for exposure and ease of use sake, but could you still convert this to entry points as per https://github.com/jupyter/nbconvert/pull/352. Right now you're set up to conflict with that once that pr is merged.

amueller:
@michaelpacer can you elaborate on the entry points?
That PR doesn't seem to touch any of the exporters.

One thing that I noticed when debugging was that latex was dropped (as it is when converting to html,  think? or RST?)

Things were pretty straight-forward. There is a work-around for a pandoc bug in there, the rest was pretty straight-forward. I'm wondering if we want better support for internal links. Markdown has (afaik) internal links via ``#tags`` like ``[reference name](#reference-tag)``. That is currently converted to a standard link, not an internal reference in asciidoc. arguably a pandoc bug, too. I don't think these are converted into proper internal links in any of the other conversions, though (apart from html, where there is no difference between internal and external references).

takluyver:
@michaelpacer - I think we'll just have to deal with some conflicts. If @amueller converts this to use entry points, then it's definitely going to conflict with #352. Sorry, I haven't had much chance to review either of these PRs recently, what with travel.

amueller:
I think I need to clean this up a bit. It seems to me instead of writing many ``x2x`` functions, there should be a ``generic_pandoc`` filter that can do all pandoc conversions. Otherwise I'll also need to add a ``html2asciidoc`` one and that all seems to blow up.

michaelpacer:
My intuition is that while the generic pandoc converter is a good idea, that should be a separate PR as it will require a more broad refactoring of the code base at which point new bugs are likely to be introduced.

amueller:
So should I first create a PR for the generic pandoc filter? Or I could add a separate ``html2asciidoc`` filter here (not sure if that's the last one).

michaelpacer:
The only thing I thought would be good to have in the previous version, would be to use entrypoints to access it in addition to using the hardcoded `export_map` as we were trying to govern all the exporters via entrypoints. 

In fact  my worry is that your asciidoc exporter won't be seen by the `get_exporter` function given the removal at https://github.com/jupyter/nbconvert/pull/352/files#diff-592b46ef5ddfed7d0b228485401ea2a7L183

Could you write a test that takes one of the notebooks and exports it to asciidoc? I'm pretty sure if you add that it'll fail right now.

michaelpacer:
As for the order of the PRs, if you think you can put it together relatively quickly it shouldn't be an issue.

The changes you'll be making for that will likely be covered by existing tests, since it's modding current infrastructure rather than adding a new exporter. So I imagine it'll be faster to debug and faster to merge, and it'll probably save you the time of needing to rewrite some of this PR if you do that first.

michaelpacer:
Also apologies for not elaborating on the entrypoints stuff before, it shouldn't be hard, just add a line in the `setup.py` `setup_args['entrypoints']` near https://github.com/jupyter/nbconvert/pull/352/files#diff-2eeaed663bd0d25b7e608891384b7298R217 instead of using `export_map`.

amueller:
Ok, thanks. Don't expect anything this week, but hopefully next.

michaelpacer:
I'm going to move this to 5.1, which I anticipate won't need too many things to be released shortly after 5.0. 

If we don't release 5.0 before you can cover those cases then it should definitely be in `5.0`. A new default template for working directly with a publishing house's format is a nice feature to have on a `x.0`  version, so that would be my preference (ceteris paribus).

michaelpacer:
Yay, #436 was merged, so this can proceed. Feel free to define a convenience method, but I ask that for consistency inside the template itself that you use the `generic_pandoc` function.

@amueller Will you have a chance to work on this soon?

amueller:
I'll try over the weekend, ok?
Do you have a deadline for 5.0?

Carreau:
> Do you have a deadline for 5.0?

I think @mpacer would like to have a beta soonish (if possible before end of october).

If it does not make it into 5.0 we can make it into 5.1 which should be sooninsh after 5.0 cause we'll probably get complaint as we often break things unexpectedly.

michaelpacer:
Hi @amueller, I was hoping to have this ready to go (at least as a beta release) by the end of this week. 

And yes I support what @Carreau said, we can put this in 5.1, if it doesn't make it into 5.0.  But given that the generic pandoc filter is included it'd be nice to have with it together, illustrating its use.

michaelpacer:
Also, I think I'm going to not touch generic pandoc and solve the rst filtering in a slightly different way.

amueller:
Ok, trying to do this today.
Somehow I get 100 scikit-learn notifications a day right now, that sort of keeps me busy :-/

amueller:
Just ran this over my book and got no errors... I think that's a relatively decent test. I'll add something to the test file after dinner.

michaelpacer:
This is looking good. Restarted the build on travis, since it looked like it failed for bad reasons. 

Also, if the `latex2asciidoc` filter is just using the pandoc command with no intermediate layers, could you use `convert_pandoc(from_format='latex', to_format='asciidoc')` in the actual template? 

I'm ok if you want to leave the helper there for other people in the LaTeX filter module itself.

michaelpacer:
Also, thank's for turning this around so quickly!

amueller:
I removed the ``latex2asciidoc`` (as that was kind of the point of the other PR), but left in the ``markdown2asciidoc`` for now. The bug has been fixed in pandoc master, but hasn't been released yet.

amueller:
I added a test to the exporters. Was that what you had in mind? I just realized that before I kept cramming stuff into ``test_nbconvertapp.py`` which I guess is more of an integration test? Should I test there, too?

amueller:
I don't understand the test failure. That's unrelated, right? Can someone retrigger on 2.7?

takluyver:
It did indeed look unrelated; I've rerun it and it passed. :-)

michaelpacer:
@amueller Apologies for the lack of quick response. I wasn't able to spend much of any time on GitHub today. 

So this looks good except that the test case that you are making the exception for in the `markdown2asciidoc` is not currently being hit by the testing materials. Could you modify the testing file to include whatever the error condition is for that pandoc bug?

Other than that this looks good to merge.

michaelpacer:
Made a PR at https://github.com/amueller/nbconvert/pull/2 to fix the code coverage issue. 

As soon as that's added, I'll merge. 

@amueller Sorry for not contributing directly… I would have just pushed an update here, but then I would have felt uncomfortable also merging, and I wanted to be able to do that as soon as I could.

michaelpacer:
Thanks @amueller for all the work you put into this!

Merging

amueller:
thank you for your help in getting this in :)

PR #356, opened 2016-08-10
Make travis run all tests (not SKIPPED tests)

janschulz:
This is a test to see if thats the problem which lead the pdf test not beeing run https://github.com/jupyter/nbconvert/issues/350

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,8 @@ sudo: false
 addons:
   apt:
     packages:
-    - texlive-latex-recommended
-    - texlive-xetex # both for pdf output tests
+    - texlive-latex-extra # we need this for all the latex package we use, recommended is not enough
+    - texlive-xetex # latex to pdf converter
     - inkscape # for svgs in pdf output
 before_install:
     - git clone --quiet --depth 1 https://github.com/minrk/travis-wheels travis-wheels
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,8 @@ sudo: false
 addons:
   apt:
     packages:
-    - texlive-base
+    - texlive-latex-recommended
+    - texlive-xetex
 before_install:
     - git clone --quiet --depth 1 https://github.com/minrk/travis-wheels travis-wheels
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ addons:
   apt:
     packages:
     - texlive-latex-extra # we need this for all the latex package we use, recommended is not enough
+    - latex-xcolor # ... and we even need another latex package
     - texlive-fonts-recommended # fonts...
     - cm-super # more fonts
     - texlive-xetex # latex to pdf converter
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,8 @@ addons:
   apt:
     packages:
     - texlive-latex-extra # we need this for all the latex package we use, recommended is not enough
-    - latex-xcolor # ... and we even need another latex package
+    - texlive-generic-recommended # .. and more ...
+    - latex-xcolor # ... and more latex packages
     - texlive-fonts-recommended # fonts...
     - cm-super # more fonts
     - texlive-xetex # latex to pdf converter
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ addons:
     packages:
     - texlive-latex-extra # we need this for all the latex package we use, recommended is not enough
     - texlive-fonts-recommended # fonts...
+    - cm-super # more fonts
     - texlive-xetex # latex to pdf converter
     - inkscape # for svgs in pdf output
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ addons:
   apt:
     packages:
     - texlive-latex-extra # we need this for all the latex package we use, recommended is not enough
+    - texlive-fonts-recommended # fonts...
     - texlive-xetex # latex to pdf converter
     - inkscape # for svgs in pdf output
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,8 @@ addons:
   apt:
     packages:
     - texlive-latex-recommended
-    - texlive-xetex
+    - texlive-xetex # both for pdf output tests
+    - inkscape # for svgs in pdf output
 before_install:
     - git clone --quiet --depth 1 https://github.com/minrk/travis-wheels travis-wheels
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+sudo: required
+dist: trusty
+
 language: python
 python:
     - 3.5
@@ -7,7 +10,6 @@ python:
 env:
   global:
     - PATH=$TRAVIS_BUILD_DIR/bin:$PATH
-sudo: false
 addons:
   apt:
     packages:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ python:
     - 2.7
 env:
   global:
-    - PATH=$TRAVIS_BUILD_DIR/pandoc:$PATH
+    - PATH=$TRAVIS_BUILD_DIR/bin:$PATH
 sudo: false
 addons:
   apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+# Use a newer travis environment:
+# https://docs.travis-ci.com/user/trusty-ci-environment/
+# needs these two lines:
 sudo: required
 dist: trusty
 
--- a/nbconvert/exporters/tests/test_pdf.py
+++ b/nbconvert/exporters/tests/test_pdf.py
@@ -26,7 +26,7 @@ class TestPDF(ExportersTestsBase):
         self.exporter_class()
 
 
-    @dec.onlyif_cmds_exist('pdflatex')
+    @dec.onlyif_cmds_exist('xelatex')
     @dec.onlyif_cmds_exist('pandoc')
     def test_export(self):
         """Smoke test PDFExporter"""
--- a/nbconvert/tests/test_nbconvertapp.py
+++ b/nbconvert/tests/test_nbconvertapp.py
@@ -91,7 +91,7 @@ class TestNbConvertApp(TestsBase):
             assert os.path.isfile('notebook2.py')
 
 
-    @dec.onlyif_cmds_exist('pdflatex')
+    @dec.onlyif_cmds_exist('xelatex')
     @dec.onlyif_cmds_exist('pandoc')
     def test_filename_spaces(self):
         """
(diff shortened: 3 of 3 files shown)


flying-sheep:
PDF stuff seems still to be skipped:

* https://travis-ci.org/jupyter/nbconvert/jobs/151307788#L445
* https://travis-ci.org/jupyter/nbconvert/jobs/151307788#L550
* https://travis-ci.org/jupyter/nbconvert/jobs/151307788#L564
* https://travis-ci.org/jupyter/nbconvert/jobs/151307788#L576

janschulz:
I have the feeling that this might have somthign to do with https://github.com/jupyter/nbconvert/commit/1db4edc503269dd2e42ce07abb1b20aa35d18029 which changed the command but not the test?

janschulz:
Looks like no latex is installed on travis? 

```
$ which pdflatex
The command "which pdflatex" exited with 1.
$ which xelatex
The command "which xelatex" exited with 1.
```

janschulz:
And now it fails: https://travis-ci.org/jupyter/nbconvert/jobs/151315097#L485

janschulz:
This now runs all tests in travis (no more SKIPPED tests -> the "purpose" of this PR is therefore accomplished), but that exposed quite a lot of failures: https://travis-ci.org/jupyter/nbconvert/jobs/151317439

janschulz:
And now we are in trouble: 

```
E           ! LaTeX Error: File `adjustbox.sty' not found.
```

But according to [this SO answer](http://tex.stackexchange.com/a/116668/12733) this file isn't in the version of `texlive-latex-extra` in ubuntu precise which is used on travis.

takluyver:
Travis does have a beta option to use Trusty (14.04) for builds. I don't know what the queue time is like for those.

https://docs.travis-ci.com/user/trusty-ci-environment/

takluyver:
Looks like the queue is not too bad, at least at the moment - those builds had started within a couple of minutes of your commit.

takluyver:
The builds seems to run for a similar amount of time to those on master, which is a good sign if we can get them working.

takluyver:
OK, now we have reached the same test failure described in #350, which #353 aims to fix.

janschulz:
This is interesting: the test actually sets `' --PDFExporter.verbose=True'` but the test doesn't capture the latex error :-(

```
/home/travis/virtualenv/python3.5.0/lib/python3.5/site-packages/nbconvert/tests/test_nbconvertapp.py:117: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <nbconvert.tests.test_nbconvertapp.TestNbConvertApp testMethod=test_pdf>
parameters = ['--log-level', '0', '--to', 'pdf', 'notebook2', '--PDFExporter.latex_count=1', ...]
ignore_return_code = False, stdin = None
[...]
E               raise LatexFailed('\n'.join(self._captured_output))
E           nbconvert.exporters.pdf.LatexFailed: PDF creating failed, captured latex output:
```

janschulz:
I cherrypicked e1f1e1111ea9e9bfa65dad3b0d4a406cee39fc6b into this PR, lets see what happens...

janschulz:
Now it runs into a kernel startup timeout on py3.3: https://travis-ci.org/jupyter/nbconvert/jobs/151344622

???

takluyver:
That may be a random failure, I've restarted it to see if it passes.

janschulz:
seems to be, now it's green :-)

flying-sheep:
seems like my weird choices have uncovered yet another bug :wink: 

great work fixing it!

takluyver:
So now, as on #353, we just have to work out if adding the `-shell-escape` flag is actually the right thing to do.

@michaelpacer @fperez, other latex users - does adding that flag make sense? Is it going to break other things?

flying-sheep:
it’s not breaking anything, but it disables a sandbox: malicious LaTeX code could do bad things with it.

so the question is if we trust the LaTeX, or if adding `-shell-escape` only for trusted notebooks would be feasible

janschulz:
My vote (still :-)) goes to removing the minted package: in IRkernel there are two ways to highlight source code of a displayed function (NOT the code in a cell, but things like `display(function)`): one with a package in R, one with latex (=minted). IMO it would be enough to only use the R stuff. If someone wants special latex function, it's only an overwrite of a repr funciton away. IMO thats enough for a mostly unused feature (not a lot of use cases to display source code for already defined functions).

flying-sheep:
maybe catch the error in the LaTeX output and give a user friendly warning that this specific notebook can only be converted to PDF if it’s trusted?

takluyver:
Maybe, but I'm wary of anything that fails on untrusted notebooks, because it's very easy to end up with users going "whatever, I'll trust this so I can do what I want". That's the 'enable macros' curse - you can't rely on users reading a security warning if it's blocking the path to what they want to do.

If `minted` only works with this sandbox disabled, I'd rather explore @janschulz's ideas for how we can do things without it.

flying-sheep:
yeah, minted calls into pygments and therefore needs to access external commands. `\write18` is TeX’ way to call commands and is disabled by default.

without it, TeX can only create files, not execute things. (which can also be used to cause mayhem of course)

we just have to think about some things before deciding this: do we already have output formats that can have arbitrary code execution on creation?

if not i’d agree to removing the minted line again and having users install `highr` which they’ll hopefully do since we have a nice error message if it isn’t installed

fperez:
I agree that arbitrary external code execution on format conversion sounds like something we should avoid...  @takluyver's comment about the "curse of macros" is on-point.

If users pass `--execute` they *know* they are asking the code in the notebook to run, but at least they can assume that it's only that code that runs, so they have a sense of the security perimeter.  But having simple PDF conversion trigger third-party code calls would be very surprising. My vote is to avoid it, even at the cost of some loss of functionality.

Given my admittedly limited understanding of the issue, I lean towards @janschulz's suggestion of dropping minted altogether.

takluyver:
And I don't think any of our other export formats allow arbitrary code execution from the notebook (though we haven't closely audited that).

flying-sheep:
OK. if people use any `\write18` code, they can still convert to `.tex` and then run `pdflatex -shell-escape file.tex`

fperez:
On Fri, Aug 12, 2016 at 1:39 AM, Philipp A. <notifications@github.com>
wrote:

> OK. if people use any \write18 code, they can still convert to .tex and
> then run pdflatex -shell-escape file.tex
>

That sounds like an excellent solution.  Perhaps someone is willing to add
this as a tip to the `nbconvert` docs?  That would sufficiently take care
if the issue as far as I'm concerned.

Thanks!

-- 
Fernando Perez (@fperez_org; http://fperez.org)
fperez.net-at-gmail: mailing lists only (I ignore this when swamped!)
fernando.perez-at-berkeley: contact me here for any direct mail

takluyver:
If the aim is to add syntax highlighting of displayed code, I wonder if we can do something smarter with the fact that `pygments` is already a dependency for nbconvert. How about, if a mimebundle is displayed with a `text/plain` component, it could have metadata like:

```json
{'text/plain': {'highlight_as': 'text/x-r'}}
```

In principle this could be used both by pygments in nbconvert, and codemirror in the live notebook. In practice, codemirror currently has a different mimetype for R (`text/x-rsrc`) than any of the ones [pygments recognises](http://pygments.org/docs/lexers/#lexers-for-the-r-s-languages) :angry:, but maybe we could get that fixed.

takluyver:
Or we could do what we already do in kernelspecs, and allow metadata to specify names for a codemirror mode and a pygments lexer to override a more general specifier.

flying-sheep:
why not just serve it as `text/x-r` and have notebook and nbconvert handle the displaying/conversion to HTML?

takluyver:
That's also possible, but it requires them to handle outputs with a much larger range of mimetypes. It's also a breaking change: frontends and tools not aware of the change processing output/notebooks from things that are will show no output (or they'll need a second copy included as `text/plain`). With my proposal, frontends which aren't aware of the highlighting metadata will still display it as unhighlighted text, without the kernel needing to provide a fallback.

My idea also allows a frontend to distinguish between e.g. an HTML output (`text/html`) and highlighted HTML source code (`text/plain` with highlight_as metadata).

flying-sheep:
> or they'll need a second copy included as `text/plain`

that was the idea

takluyver:
I think sending as `text/plain` to display source code is a pragmatic approach. Github does this for the 'raw' view of text-based files, including SVG and HTML.

Also, I don't think it's a lie to call source code `text/plain`; there may be a more specific mime type for it, but  the larger category is not inaccurate. The XDG mime database includes a notion of inheritance, and sure enough, `text/x-python` and `text/html` both inherit from `text/plain`.

flying-sheep:
i’m convinced. i think that `highlight_as` is a bit too specific though. maybe other consumers might do something else with it.

what that metadata is is the “most specific programming/markup language mimetype”, and we should base its name on that. maybe `lang-mime`/ `language-mime`/ `lang-mimetype`/ `language-mimetype`? or is there precedent for using underscores? i think `language-mime` is my favourite.

takluyver:
I don't have a strong opinion on the name; `language-mime` sounds fine to me. I'll open a new issue for this, so we can get back to the immediate issue here (removing `minted`).

takluyver:
#363 is the issue for that feature.

flying-sheep:
can @janschulz just remove it in this PR? then we’ll also see if this runs without `-shell-escape` and without minted.

i sent a repr PR: IRkernel/repr#77

michaelpacer:
I'm curious, why don't we enable a flag when invoking nbconvert that allows it to pass `--shell-escape` to the XeLaTeX interpreter? That way we could write tests that run on trusted notebooks and allow users to use `\write18{}` commands which they're likely only going to do if they know what `--shell-escape` is anyway. I'm on board for by default getting rid of minted which would require this, but it seems like we can keep options open for users who know what including `--shell-escape` is going to do.

takluyver:
The command used to invoke Latex is already [configurable](https://github.com/jupyter/nbconvert/blob/39b86e995ccfccd9d2c367b21187c61349f6189f/nbconvert/exporters/pdf.py#L38). I'd be wary of:

1. making it too easy to disable a security feature, especially where the name makes it not obvious that that's what it is.
2. starting down the road of exposing lots of latex options through nbconvert. If you want to use latex in all its glory, you can get a .tex file from nbconvert and run latex separately.

janschulz:
rebased on top of master: IMO this is ready to merge

.travis.yml
```diff
@@ -1,3 +1,6 @@
+sudo: required
```
takluyver: Am I right that `sudo: required` is necessary in order to use trusty? If so, a comment to that effect would be good; hopefully that will change at some point in the future, and then we can stop requiring sudo again.

janschulz: done

takluyver: Thanks. Then I think this is fine, but I'm tired, so I'll leave it for others to have a look at in case I'm overlooking something.

PR #359, opened 2016-08-11
correct 'unconfined' metadata conversion to HTML attribute

paulovn:
When converting to HTML a Notebook that contains cells with images, if the cell contains image metadata with the `unconfined` attribute, the resulting HTML is wrong.

Specifically, this:

```
     "metadata": {
      "image/png": {
       "height": 2200,
       "unconfined": true,
       "width": 5500
      }
```
produces this:
```
width=5500
height=2200class="unconfined"
>
</div>
```

Notice the lack of space between the _width_ and the _class_ attributes. This causes the browser to ignore them, resulting in an image squeezed into the `<div>` (instead of the intended behaviour, which for "unconfined" is to provide scrollbars and render the image at full size).

The patch is a simple modification of the `basic` HTML template to avoid suppressing the whitespace between both attributes, therefore creating the correct HTML:

```
width=5500
height=2200 
class="unconfined"
>
</div>
```
(BTW, the `unconfined` attribute is also ignored for SVG images, but that cannot be corrected in the template since SVG elements are processed whole; instead the "unconfined" class attribute should be added to the `<svg>` element, but I don't know how to do that)

### DIFF
--- a/nbconvert/templates/html/basic.tpl
+++ b/nbconvert/templates/html/basic.tpl
@@ -144,7 +144,7 @@ width={{width}}
 {%- if height is not none %}
 height={{height}}
 {%- endif %}
-{%- if output | get_metadata('unconfined', 'image/png') -%}
+{%- if output | get_metadata('unconfined', 'image/png') %}
 class="unconfined"
 {%- endif %}
 >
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks, this looks good to me

michaelpacer:
@takluyver is there a reason this shouldn't be included in the 5.0 milestone and merged? Do you think the SVG image piece needs to be solved first? Is there a way to do using simple templates as opposed to having to parse the entire DOM to find where the appropriate `<svg>` tag is and then insert it there?

takluyver:
No, I think this is fine, I was just leaving it for others to have a look. Or maybe I forgot to hit merge.

This is clearly a bugfix, so I marked it 4.3, but if we don't do a 4.3 release and go straight to 5.0, all issues marked 4.3 can be changed to 5.0.

PR #360, opened 2016-08-11
[WIP] use new KernelClient.execute_interactive

minrk:
in execute preprocessor

Significantly simplifies the preprocessor, relying on upstream implementation.

WIP because it's not released yet.

I hit two missing features that we should either drop here or add there:
1. When we interrupt-on-timeout, we need to flush outputs,
   which suggests that flushing outputs with the hook
   should be a method on its own.
2. ExecutePreprocessor allows a separate timeout for no output,
   not just a total timeout for the whole execution.

cf https://github.com/jupyter/jupyter_client/pull/185

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -7,11 +7,6 @@ and updates outputs"""
 import os
 from textwrap import dedent
 
-try:
-    from queue import Empty  # Py 3
-except ImportError:
-    from Queue import Empty  # Py 2
-
 from traitlets import List, Unicode, Bool, Enum, Any
 
 from nbformat.v4 import output_from_msg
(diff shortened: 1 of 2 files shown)


mpacer:
@minrk is this still active?

minrk:
I'd like to come back to it at some point, but it's far from urgent. Feel free to close if it's in the way.

takluyver:
@minrk I believe the released version of jupyter_client has execute_interactive, if you want to come back to this.

minrk:
I sure do! Trying to finish up some JupyterHub stuff right now, but I hope to find time for this at some point.

akhmerov:
Is this PR still relevant?

SylvainCorlay:
@minrk wow, that is indeed removing a lot a code!

Is this something you would like to rebase and include in the next release?

MSeal:
Closing this issue as the code in question has been entirely rewritten and refactored in 5.5 and I don't believe this is relevant anymore. Reopen with rebase if there's still interest in something I missed :)

PR #335, opened 2016-07-13
Add packages required by IRkernel/repr

flying-sheep:
The [IRkernel](https://github.com/IRkernel/IRkernel)’s display machinery, powered by the [repr](https://github.com/IRkernel/repr) R package, by default requires two LaTeX packages.

Since there is

1. no way to specify required packages during kernel runtime, and
2. templates aren’t kernel specific

this project is the right place for adding LaTeX packages. Eventually it would be nice to have a way to register notebook-wide data that can be used by templates, and therefore allow for kernels and extension packages to influence LaTeX packages, HTML scripts, and the like. this idea is also discussed in IRkernel/IRdisplay#14

PR #367, opened 2016-08-18
Remove minted from Latex base template

takluyver:
It requires a flag to disable a security feature, which we've decided not to use.

Closes gh-350

Alternative to #353 

Ping @amueller @janschulz @flying-sheep

### DIFF
--- a/nbconvert/templates/latex/base.tplx
+++ b/nbconvert/templates/latex/base.tplx
@@ -62,7 +62,6 @@ This template does not define a docclass, the inheriting class must define this.
     \usepackage{longtable} % longtable support required by pandoc >1.10
     \usepackage{booktabs}  % table support for pandoc > 1.12.2
     \usepackage[inline]{enumitem} % IRkernel/repr support (it uses the enumerate* environment)
-    \usepackage{minted}           % highlighting support for IRkernel/repr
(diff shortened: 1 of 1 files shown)


janschulz:
This then "reopens" https://github.com/jupyter/nbconvert/pull/335 

irkernel/repr right now depends on minted: https://github.com/IRkernel/repr/blob/master/R/repr_function.r#L37 (this is not in then PR in https://github.com/IRkernel/repr/pull/77 which only switches the default to use a r function to highlight)

takluyver:
Right, it effectively pushes the issue back to IRkernel.

flying-sheep:
> this is not in then PR in IRkernel/repr#77 which only switches the default to use a r function to highlight

what do you mean? if we merge IRkernel/repr#77, we have a default that works, and once somebody changes the option, they’ll have to go the route `jupyter nbconvert --to=text *.ipynb; pdflatex -shell-escape *.tex`

merging both things would be the pragmatic thing that “makes the ‘download PDF’ button work”

takluyver:
Merging this because the test failure emails are getting old, and this failure might be obscuring other failures. I think we have agreement on this, but if we want to do something else, we can always change it later.

PR #370, opened 2016-08-22
Use traitlets instead of genutils for import_item

juhasch:
Fixes https://github.com/ipython-contrib/jupyter_contrib_nbextensions/issues/723


michaelpacer:
Could you add a test for the case for which you ran into the error in Python 2.7 with ipython.genutils? 

The change looks fine, it's just it seems like this was a bug that wasn't detected in the current testing suite so it's worth fleshing that out.

juhasch:
The tests in traitlets are actually already using unicode, so I believe no new tests are required:
https://github.com/ipython/traitlets/blob/master/traitlets/utils/tests/test_importstring.py

takluyver:
Thanks @juhasch

Issue #366, opened 2016-08-18
rst template does not respect language notebook is written when creating code blocks


michaelpacer:
Based on @takluyver in https://github.com/jupyter/nbconvert/issues/47#issuecomment-113670318 it seems that the rst template is wrong. 

We need to change 
https://github.com/jupyter/nbconvert/blob/master/nbconvert/templates/rst.tpl#L12 to infer the language of the notebook, but should it be inferring the language the notebook is written in on the basis of `metadata.kernelspec.language`, `metadata.language_info.codemirror_mode.name` or `metadata.language_info.name`?

@takluyver or @Carreau can you point to another instance where the jinja template inspects the notebook metadata directly? I'd rather learn jinja bit by bit rather than trying to understand it all at once, so basing this off a known example would be better than trying to figure out how to do this from scratch (especially if there is more than one way to do this).

takluyver:
I think the language names in rst are used to identify pygments lexers, so it should use `pygments_lexer` first and then fall back to `name` if that's not defined. Both `pygments_lexer` and `codemirror_mode` are used to specify overrides if name doesn't identify the correct higlighting for the respective systems.

michaelpacer:
So in the test notebook files, many of them have neither a `pygments_lexer` nor a `codemirror_mode` key in their JSON representation. GitHub seems to be displaying them without difficulty (e.g., [notebook2](https://github.com/jupyter/nbconvert/blob/master/nbconvert/tests/files/notebook2.ipynb) has no mention of `pygments` `codemirror` or even an associated kernel), so I'm guessing these are not required by the notebook format spec, and am not sure where the formatting instructions are coming from. 

Should I be using some other set of notebooks other than the test notebooks to be thinking through this problem? I figured since they were backing our tests that they were the notebooks I should turn to in order to ensure that I'm following all currently supported versions of the notebook formatting spec. 

Also, if you can't point me to an example of where jinja pulls arguments from the notebook's metadata, could you point me to the pieces of code that handle processing the notebook through jinja? I can try to figure out how that works and reverse engineer the changes that need to be made to fix this case?

takluyver:
>  many of them have neither a pygments_lexer nor a codemirror_mode key ...  I'm guessing these are not required by the notebook format spec

[...]

PR #372, opened 2016-08-22
Language specificity in codeblocks in rst template

michaelpacer:
This addresses #366, does not generalise it to markdown. Shouldn't be hard to do but perhaps we want that as a different PR.

### DIFF
--- a/nbconvert/templates/rst.tpl
+++ b/nbconvert/templates/rst.tpl
@@ -12,8 +12,6 @@
 {{".. code:: "-}}
 {%- if nb.metadata.language_info.pygments_lexer -%}
     {{ nb.metadata.language_info.pygments_lexer }}
-{%- elif nb.metadata.language_info.codemirror_mode -%}
-    {{ nb.metadata.language_info.codemirror_mode.name }}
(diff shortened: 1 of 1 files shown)


nbconvert/templates/rst.tpl
```diff
@@ -9,7 +9,14 @@
 
 {% block input %}
 {%- if cell.source.strip() -%}
-.. code:: python
+{{".. code:: "-}}
+{%- if nb.metadata.language_info.pygments_lexer -%}
+    {{ nb.metadata.language_info.pygments_lexer }}
+{%- elif nb.metadata.language_info.codemirror_mode -%}
```
takluyver: I don't think this should try codemirror_mode; we use pygments_mode specifically because docutils, the main renderer of rst, uses pygments for highlighting. It's unlikely that Codemirror will be used, so the codemirror mode shouldn't have priority over the generic language name (also, codemirror mode can be a string or a dictionary, which confuses things more).

michaelpacer:
Addressed the codemirror bit.

michaelpacer:
Anything preventing this merge?

PR #377, opened 2016-08-28
make output_files_dir configurable (fixes #341)

### LINKED ISSUE #341: make 'output_files_dir' configurable
Currently, the 'output_files_dir' resource (which contains the path where figures are saved) is hard-coded in the `NbConvertApp` implementation:

`resources['output_files_dir'] = '%s_files' % notebook_name`

https://github.com/jupyter/nbconvert/blob/cc7a2dd7ccb430c3545b153a5790cf7c289eb4e3/nbconvert/nbconvertapp.py#L332

Such that changing it requires subclassing the `NbConvertApp` or running the whole conversion process manually. It would be useful to make the 'output_files_dir' configurable, so that we can select the directory, where to save the files or disable generating the figure files (by setting `output_files_dir` to `None`). 

For example, I configure the path to export notebooks with figure into Sphinx documentation tree. Currently, I implement it by using the nbconvert API directly  (see this [gist](https://gist.github.com/btel/837b56748698a795a632e081e2e58e23)).

btel:
Add new NbConvertApp configuration option (`output_files_dir`), which allows to set the path in which the exported figures will be saved. Fixes #341.

Created during the EuroSciPy2016 sprint

### DIFF
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -138,6 +138,12 @@ class NbConvertApp(JupyterApp):
             the exporter, and is usually '.nbconvert'."""
     ).tag(config=True)
 
+    output_files_dir = Unicode('{notebook_name}_files',
+         help='''Directory to copy extra files (figures) to.
+               '{notebook_name}' in the string will be converted to notebook
+               basename'''
+    ).tag(config=True)
+
     examples = Unicode(u"""
         The simplest way to use nbconvert is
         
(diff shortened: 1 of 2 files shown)


nbconvert/nbconvertapp.py
```diff
@@ -138,6 +138,11 @@ def _classes_default(self):
             the exporter, and is usually '.nbconvert'."""
     ).tag(config=True)
 
+    output_files_dir = Unicode('%s_files',
```
minrk: Can you use format-strings, so this would be `'{notebook_name}_files'`, rather than `%s`? Below would then become `output_files_dir.format(notebook_name=notebook_name)`

btel:
@minrk Thanks for comments. I updated the PR accordingly.

btel:
Note that the two failing tests are not related to this PR:

```
::TestRSTExporter::test_empty_code_cell <- ../../home/travis/virtualenv/python3.5.0/lib/python3.5/site-packages/nbconvert/exporters/tests/test_rst.py FAILED

::TestRSTExporter::test_export <- ../../home/travis/virtualenv/python3.5.0/lib/python3.5/site-packages/nbconvert/exporters/tests/test_rst.py FAILED
```

minrk:
Yeah, don't worry about those, should be fixed by #378.

PR #378, opened 2016-08-29
handle missing language_info metadata in rst template

minrk:
language_info may not be defined in skeleton notebooks. This should result in no highlighting, not an error.

### DIFF
--- a/nbconvert/templates/rst.tpl
+++ b/nbconvert/templates/rst.tpl
@@ -10,9 +10,9 @@
 {% block input %}
 {%- if cell.source.strip() -%}
 {{".. code:: "-}}
-{%- if nb.metadata.language_info.pygments_lexer -%}
+{%- if 'pygments_lexer' in nb.metadata.get('language_info', {}) -%}
     {{ nb.metadata.language_info.pygments_lexer }}
-{%- elif nb.metadata.language_info.name -%}
+{%- elif 'name' in nb.metadata.get('language_info', {}) -%}
     {{ nb.metadata.language_info.name }}
 {%- endif %}
 


minrk:
Merging so that travis is back to running on master

michaelpacer:
Ah — i'd started working on something like this, but this is much nicer. I'm glad I got to see how you solved this problem after trying my own solution (which involved wrapping it in another if loop first checking for the existence of language_info and then setting them accordingly).

Also I had thought we were supposed to provide a default highlighting something like the previous hardcoded `python` as a fallback; is there a principle/heuristic that I should use regarding when to stick with what had been done before as a default and when to turn it into a new, less opinionated default?

minrk:
That's a choice we can make (I merged a bit quickly so Travis is back to working, but we can refine in new PRs). I have mixed feelings. On the one hand, no metadata resulting in no highlighting makes logical sense to me. On the other, in reality most notebooks that lack this metadata are old IPython notebooks, in which case assuming Python will be correct ~all of the time. Plus, if the notebook is opened and saved one time with a recent version of Jupyter, the issue goes away.

> is there a principle/heuristic that I should use regarding when to stick with what had been done before as a default and when to turn it into a new, less opinionated default?

I would probably go with:

1. Preserve existing behavior unless you have a reason not to (cleaner code doesn't count)
2. Often code is written with assumptions that become untrue over time, in which case we can revisit them and make more appropriate default behavior
3. When in doubt, ask. Someone may recall "we did that because of this condition that doesn't matter anymore" or "we did it because of this obscure case that's still important." Of course, it can always be "I have no idea why we did that," too.

For instance, applying it now: when this template was first written, language_info wasn't part of the notebook format, so we had to make assumptions, and Python was hardcoded in lots of places (non-Python notebooks were not well supported for some time). Now, we have had language_info in the spec for some years, so lacking this information ought to be rare, *especially* for non-Python notebooks, so the question becomes: how should we treat the rare case of missing info? According to the current spec (no language info == no language) or according to the *likely reason* for that missing information (old notebook, probably Python).

Issue #380, opened 2016-08-30
When docs are built sphinx cannot find some attributes associated with exporters


michaelpacer:
This is likely having to do with the recent move to entrypoints.

I'll try to take care of this within the next day or two. 

~/Jupyter/nbconvert/docs/source/api/exporters.rst:68: WARNING: autodoc: failed to import function 'export_custom' from module 'nbconvert.exporters'; the following exception was raised:
Traceback (most recent call last):
  File "~/anaconda3/envs/dev/lib/python3.5/site-packages/sphinx/util/inspect.py", line 109, in safe_getattr
    return getattr(obj, name, *defargs)
AttributeError: module 'nbconvert.exporters' has no attribute 'export_custom'

willingc:
@michaelpacer Seeing a similar error related to autodoc in Jupyter Client docs https://github.com/jupyter/jupyter_client/issues/194

PR #381, opened 2016-08-30
Remove erring function calls

michaelpacer:
addresses #380 

removes no longer existing functions from documentation

### DIFF
--- a/docs/source/api/exporters.rst
+++ b/docs/source/api/exporters.rst
@@ -65,22 +65,6 @@ These functions are essentially convenience functions that
 wrap the functionality of the classes documented in the previous
 section.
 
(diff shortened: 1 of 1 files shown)


takluyver:
To clarify, these are deprecated for 5.0, but still actually exist. I see the import location has moved, I'm just going to open a PR to fix that.

docs/source/api/exporters.rst
```diff
@@ -65,22 +65,6 @@ These functions are essentially convenience functions that
 wrap the functionality of the classes documented in the previous
 section.
```
takluyver: @michaelpacer this description should also be updated along with the change in the things it refers to!

PR #383, opened 2016-08-30
Fix rtd build by use pip install

### LINKED ISSUE #382: Docs won't build on readthedocs because they expect to use easy_install by default
One solution would be to change that upstream, unless there is a reason to use `easy_install` and `bdist_egg`s that I do not know about, that might not be too hard to change.

Instead, could  we change ` setup_py_install: true` to  ` pip_install: true` in the readthedocs.yml?

If we can do that, I think that should avoid the current error caused by using `setup_py_install` currently. 

CC @willingc

michaelpacer:
Fix #382 and should address https://github.com/jupyter/nbconvert/pull/352#issuecomment-243562597

I don't know how to test if this actually works on rtd

### DIFF
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -2,4 +2,4 @@ conda:
     file: docs/environment.yml
 python:
     version: 3.5
-    setup_py_install: true
+    pip_install: true


willingc:
@michaelpacer Thanks. I'll test all in the morning :D

minrk:
one way to test: merge and see what happens!

It is true that pip_install is the right thing to do (always), but particularly important with recent changes in our setup.py.

willingc:
@minrk thanks for the explanation 👍 

@michaelpacer One way to test without merging is to create a personal test RTD account (similar) which builds from your fork. Similar to https://readthedocs.org/projects/test-nbconvert/ which builds from my fork.

takluyver:
This does appear to have got RTD builds working again :+1:

willingc:
I'll be making a few small PRs later today:

- to update the environment.yml so it uses the conda-forge channel since it still uses my personal channel for pandoc which is unneeded since pandoc is now in the forge.

- to update the docs/README.md file to better explain how to build locally

I'll ping you all later for review, additions, and changes when I submit.

PR #385, opened 2016-08-31
Restore deprecated functions for nbconvert.exporters API

takluyver:
Removing these from `__all__` breaks imports from the nbconvert.exporters API, which was their documented location. Let's leave them in place for one import cycle.

I've also updated, fixed and polished the exporter docs.

### DIFF
--- a/docs/source/api/exporters.rst
+++ b/docs/source/api/exporters.rst
@@ -8,6 +8,15 @@ Exporters
    :doc:`/config_options`
      Configurable options for the nbconvert application
 
+.. autofunction:: export
+
+.. autofunction:: get_exporter
+
+.. autofunction:: get_export_names
+
+Exporter base classes
+---------------------
+
 .. autoclass:: Exporter
 
     .. automethod:: __init__
@@ -57,14 +66,3 @@ inherit either directly or indirectly from
 .. autoclass:: PythonExporter
 
 .. autoclass:: RSTExporter
-
-Specialized exporter functions
-------------------------------
-
-These functions are essentially convenience functions that
-wrap the functionality of the classes documented in the previous
-section.
-
-
-
-.. autofunction:: export_by_name
--- a/nbconvert/exporters/__init__.py
+++ b/nbconvert/exporters/__init__.py
@@ -1,4 +1,7 @@
 from .export import *
+# These specific functions are deprecated as of 5.0:
+from .export import (export_custom, export_html, export_slides, export_latex,
+    export_pdf, export_markdown, export_python, export_script, export_rst)
 from .html import HTMLExporter
 from .slides import SlidesExporter
 from .templateexporter import TemplateExporter
--- a/nbconvert/exporters/__init__.py
+++ b/nbconvert/exporters/__init__.py
@@ -1,7 +1,4 @@
 from .export import *
-# These specific functions are deprecated as of 5.0:
-from .export import (export_custom, export_html, export_slides, export_latex,
-    export_pdf, export_markdown, export_python, export_script, export_rst)
 from .html import HTMLExporter
 from .slides import SlidesExporter
 from .templateexporter import TemplateExporter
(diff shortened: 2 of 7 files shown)


michaelpacer:
So I see a slightly different way to do this that avoids the hardcoding solution again (which is what I see as being one of the biggest advantages of using entrypoints as a way of conveying all of the information where it needs to be). 

Is there a way to get the keys from the `setup.py` entrypoints configuration list? because they look to be identical to the keys needed to make the `export_zub`  available as part of the module's global namespace in https://github.com/jupyter/nbconvert/pull/385/commits/02159db8de2d1c5557e54bf8b53a223fbfea0651#diff-592b46ef5ddfed7d0b228485401ea2a7R125. 

if you could (for example) convert 

    [
            'custom=nbconvert.exporters:TemplateExporter',
            'html=nbconvert.exporters:HTMLExporter',
            'slides=nbconvert.exporters:SlidesExporter',
            'latex=nbconvert.exporters:LatexExporter',
            'pdf=nbconvert.exporters:PDFExporter',
            'markdown=nbconvert.exporters:MarkdownExporter',
            'python=nbconvert.exporters:PythonExporter',
            'rst=nbconvert.exporters:RSTExporter',
            'notebook=nbconvert.exporters:NotebookExporter',
            'script=nbconvert.exporters:ScriptExporter'] 

to be the input equivalent to 

     exporter_map = dict(
         custom=TemplateExporter,
         html=HTMLExporter,
         slides=SlidesExporter,
         latex=LatexExporter,
         pdf=PDFExporter,
         markdown=MarkdownExporter,
         python=PythonExporter,
         rst=RSTExporter,
         notebook=NotebookExporter,
         script=ScriptExporter,
     )

which seems like it should be possible, then rather than needing to add the explicit import to the `exporters.py` module it could maintain API compatibility by automatically adding the relevant entries to `__all__` 

by changing

     for name, E in exporter_map.items():
         g['export_%s' % name] = _make_exporter(name, E)
 
 to 

     for name, E in exporter_map.items():
         g['export_%s' % name] = _make_exporter(name, E)
        __all__.append('export_%s' % name)

On the other hand I could see not wanting to modify `__all__` in place, but to have it just manually set. However, this would (I think?) make it possible that any entrypoints declared elsewhere would also be appropriately accessible via the nbconvert.exporters API.

I just don't know enough about entrypoints to be able to implement that in a sane manner.

willingc:
@takluyver This PR looks good. Thanks for all the docstring updates 👍 

@michaelpacer Would you like to go ahead and merge this since it has a bunch of documentation updates? We could flesh out your approach in a separate PR. Thoughts?

takluyver:
I've updated the PR to work as you suggest. Normally I'd rather hardcode 10 names than do this kind of clever dynamic stuff to create them, but since the clever dynamic stuff is already there, we may as well reuse it to add them to `__all__`.

willingc:
Changes look good too. Thanks @takluyver.

michaelpacer:
@willingc apologies on not acting on this sooner — been in a dissertation haze. Happy to merge this now as I would have been before. This is probably a better solution though. 

@takluyver I thought the point of avoiding the hardcoding via entrypoints was to make this kind of functionality available to 3rd party exporters and templates by default. If we had gone with the hardcoding wouldn't that mean that 3rd party exporters wouldn't have been accessible using this approach?

michaelpacer:
Oh…@takluyver I'm fine having merged this, but i now realise what I was asking for was slightly more complicated. I'll open an issue.

willingc:
@michaelpacer Dissertation haze. 🔮crystal ball sees ☁️ haze lifting. Soon it will be done and :sunny:.

michaelpacer:
@willingc Literally submat it in its final form yesterday…now I'm just waiting for the formatting police to finish their inspection (though I will be astounded if they can find anything wrong…given the amount of time I spent on making the [Dissertate](https://github.com/suchow/Dissertate) template to fit the Berkeley guidelines).

Issue #379, opened 2016-08-30
Local documentation build documentation out of date


mpacer:
The second instruction states to build the docs using a `requirements.txt` file that does not currently exist.

Right now there's an error due to the use of a bdist_egg based on however the docs are being built currently, which I have yet to suss out.

willingc:
@michaelpacer The instructions are out of date and written before the use of conda in building on RTD. There is an out of date `requirements.txt` in the `docs` directory.

nbconvert is by far the most complicated build that we have. To build locally, I typically create a conda environment using ``conda env create -f environment.yml`` and then activate it.

Let's work together over the next couple of weeks to refine the instructions for building locally and on RTD and cobble together a short screencast/webcast specific to nbconvert.

michaelpacer:
@willingc  That sounds like a good plan. I'll put it on the 5.1 milestone then.

michaelpacer:
Also, I don't think that there is an out of date `requirements.txt` at least not that I can see…this is partly what evinced the extremity of my initial title, as the first line of the docs readme mentioned a file that was clearly not in the directory like it said it was.

willingc:
You are absolutely correct that the out of date `requirements.txt` is no longer in the repo. Sorry! I had completely forgotten that we removed earlier this year in this PR #222 https://github.com/jupyter/nbconvert/pull/222/files

takluyver:
The bit about `requirements.txt` got fixed - is there anything else to do for this, or can we close it?

willingc:
Don't remember. I'm cool with closing if @mpacer agrees.

MSeal:
Closing ticket as no one has chimed in on other missing out-of-date docs. Please reopen if there's anything still missing.

PR #386, opened 2016-08-31
Update docs readme

michaelpacer:
Pursuant to https://github.com/jupyter/nbconvert/issues/379, this seemed like a good thing to make as a quick fix, with the more substantial restructuring to occur later. 

Notably this requires using conda, which not all users may rely on. I do not know if there is a way to expose the `environment.yml` file to a `pip` command for non-conda users, but it seems like it shouldn't be a long run requirement that you use `conda` in order to build our docs.

cc @willingc

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -9,11 +9,15 @@ is hosted on ReadTheDocs.
 
            $ cd docs
 
-2. Install requirements:
+2. Create conda env (and install relevant dependencies):
 
-           $ pip install -r requirements.txt
+           $ conda env create -f environment.yml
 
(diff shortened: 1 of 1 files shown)


willingc:
@minrk FYI. This Travis build has one of the intermittent kernel dying messages during one of the tests. https://travis-ci.org/jupyter/nbconvert/jobs/156625723

@michaelpacer Thanks for doing this. I agree that it's worthwhile to add instructions on building locally with pip too. I'm going to merge this now, and we can follow up with pip in a future PR. ☀️

PR #387, opened 2016-09-01
Add conda-forge as channel

willingc:
Update `environment.yml` to use condo-forge channel instead of my personal channel which was needed months ago since pandoc was not yet in the forge.


minrk:
>  condo-forge

I see you have the same autocorrect that I do.

willingc:
Darn auto-correct. changelog becomes changeling too. Need to disable at some point.

PR #389, opened 2016-09-01
Add extension for highlighting to correct doc build warnings

willingc:
This corrects the 'IPython lexer' not found warnings when building the docs.


takluyver:
Thanks

PR #398, opened 2016-09-12
Allow specifying templates as paths

### LINKED ISSUE #391: Where to specify custom template file?
Say I've got a template file that I've created on my own, is it possible to pass this to `--template=` without putting it in the templates folder? I haven't been able to find any information online about this. Otherwise, what's the recommended way to use your own template file? I'd like to use one that's located in a different folder from the default templates path.

minrk:
including absolute paths

Detect template-as-path, and ensure directory is on template_path.

Tests added for both relative and absolute paths.

closes #391
closes #345

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -14,7 +14,6 @@ import uuid
 # to move ImportErrors to runtime when the requirement is actually needed
 
 
-# IPython imports
 from traitlets import HasTraits, Unicode, List, Dict, default, observe
 from traitlets.utils.importstring import import_item
 from ipython_genutils import py3compat
@@ -100,8 +99,18 @@ class TemplateExporter(Exporter):
 
     @observe('template_file')
     def _template_file_changed(self, change):
-        if change['new'] == 'default':
+        new = change['new']
+        if new == 'default':
             self.template_file = self.default_template
+            return
+        # check if template_file is a file path
+        # rather than a name already on template_path
+        full_path = os.path.abspath(new)
+        if os.path.isfile(full_path):
+            template_dir, template_file = os.path.split(full_path)
+            if template_dir not in [ os.path.abspath(p) for p in self.template_path ]:
+                self.template_path = [template_dir] + self.template_path
+            self.template_file = template_file
 
     @default('template_file')
     def _template_file_default(self):
@@ -112,7 +121,7 @@ class TemplateExporter(Exporter):
     template_path = List(['.']).tag(config=True, affects_environment=True)
 
     default_template_path = Unicode(
-        os.path.join("..", "templates"), 
+        os.path.join("..", "templates"),
         help="Path where the template files are located."
     ).tag(affects_environment=True)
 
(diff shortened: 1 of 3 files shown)


takluyver:
Looks sensible, but one of the new tests fails:

```
test_relative_template_file
...
>           assert exporter.template.filename == template
E           AssertionError: assert './relative_template.tpl' == '/tmp/tmpbqi1e2v...ive_template.tpl'
E             - ./relative_template.tpl
E             + /tmp/tmpbqi1e2v8/relative_template.tpl
```

michaelpacer:
So I was actually trying to think about how to handle this with entry-points rather than just by passing in paths. I figured that was going to be an easier way for people to distribute templates which was the request in #391. 

I'm not opposed to this, but I'm hesitant to release support for paths without also allowing it to be discovered via entrypoints. 

How I was thinking about it was to make a function that explicitly loaded entry-point declared templates. 

The first thing that needed to be overcome is that we can't directly access the filename/path of any `*.tpl` file in the way that we can with python files with `__file__`. 

The solution I had thought of for that was to actually declare a python file that is in a directory above a `templates` folder that contains the relevant templates below it. Then you can dynamically get a list of the file paths to the templates with something like `glob.glob(__file__+'/templates/*.tpl')`, get their names by splitting and taking off the extension…

That said, that might be overkill for just solving this problem. It really does seem like we should have a better way for people to hook into our template set and add their own.

michaelpacer:
I've been trying to work through this just to figure out which part is actually the error, and I'm having difficulty parsing the intended outcome. 

If i had to guess, the actual test statement `assert exporter.template.filename == template` is not what we want (rather than anything about the core functionality being missing). If we want the exporter to store the relative path directly under filename as a relative path, then it seems to be doing its job. 

template though, based on how it is defined in `template = os.path.join(td, 'relative_template.tpl')` is actually an absolute path because the `td` handle given back by the by the  `testpath.tempdir.TemporaryWorkingDirectory` context seems to be an actual temporary path, and not just a way of specifying the temporary working directory. 

Is that expected behaviour of `testpath.tempdir.TemporaryWorkingDirectory`? If so, is it just that the path shouldn't be `assert exporter.template.filename == template` but `assert exporter.template.filename == './relative_template.tpl'`?

michaelpacer:
I should mention — when I try to test this locally it works without issue. 

This is because, locally, `exporter.template.filename` actually represents a full absolute filepath rather than a relative filepath. 

Now I'm wondering if what's going on is an interaction between travis's container builds and the testpath directory commands…

minrk:
Tests passed locally for me, too. Ensuring we only compare paths with abspath should fix it on Travis and make it more reliable. I think Jinja's just being helpful shortening the name (not sure why it's happening some places and not others), but the location is correct, which is what we care about.

@michaelpacer entrypoints are a great way for people to distribute new collections of templates to other people, that's true. This PR is meant to address the simpler goal of "I have a template, let me use it," which is currently harder than it should be, and what I think #391 is asking for.

takluyver:
I think this looks good, and I think it's ~orthogonal to the question of distributing templates using entry points. So I'm going to merge it.

One possible answer for distributing templates is to recommend that people distribute them along with an exporter that uses each template, and use the entry points for exporters to provide them. We should make sure that's easy and documented, but it's probably no more overhead than writing template-discovery functions to be registered as entry points.

Issue #400, opened 2016-09-12
Correct two warnings when building docs


willingc:
Currently, docs do build, but two warnings are generated when building the api docs. The root cause appears to be documenting ``get_exporter`` in `source/api/exporters.rst`. When commented out, the first warning is gone. Though, the second warning remains. Commenting out `init_single_notebook_resources` in `source/api/nbconvertapp.rst` removes the second warning. Need to investigate why they are warning when not commented out.

**Warnings**
```
reading sources... [100%] usage

/Users/carol/code/jupyter-projects/nbconvert/docs/source/api/exporters.rst:13: WARNING: autodoc: failed to import function 'get_exporter' from module 'nbconvert.exporters'; the following exception was raised:
Traceback (most recent call last):
  File "/Users/carol/anaconda/lib/python3.5/site-packages/Sphinx-1.4.1-py3.5.egg/sphinx/util/inspect.py", line 109, in safe_getattr
    return getattr(obj, name, *defargs)
AttributeError: module 'nbconvert.exporters' has no attribute 'get_exporter'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/carol/anaconda/lib/python3.5/site-packages/Sphinx-1.4.1-py3.5.egg/sphinx/ext/autodoc.py", line 514, in import_object
    obj = self.get_attr(obj, part)
  File "/Users/carol/anaconda/lib/python3.5/site-packages/Sphinx-1.4.1-py3.5.egg/sphinx/ext/autodoc.py", line 410, in get_attr
    return safe_getattr(obj, name, *defargs)
  File "/Users/carol/anaconda/lib/python3.5/site-packages/Sphinx-1.4.1-py3.5.egg/sphinx/util/inspect.py", line 115, in safe_getattr
    raise AttributeError(name)
AttributeError: get_exporter
/Users/carol/anaconda/lib/python3.5/site-packages/nbconvert/nbconvertapp.py:docstring of nbconvert.nbconvertapp.NbConvertApp.init_single_notebook_resources:7: ERROR: Unexpected indentation.
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
```

michaelpacer:
That's really odd given that it doesn't err on `export` or `get_export_names` and if you interactively try to access `nbconvert.exporters.get_exporter` it works just fine.

michaelpacer:
[...]

PR #401, opened 2016-09-12
[WIP] Fix doc build warnings

willingc:
Addresses #400 

@michaelpacer if documenting from the `nbconvert.exporters.export` the warning goes away

### DIFF
--- a/docs/source/api/exporters.rst
+++ b/docs/source/api/exporters.rst
@@ -10,8 +10,6 @@ Exporters
 
 .. autofunction:: export
 
-.. autofunction:: get_exporter
-
 .. autofunction:: get_export_names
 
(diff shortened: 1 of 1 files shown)


michaelpacer:
This is still really odd. https://github.com/jupyter/nbconvert/blob/master/nbconvert/exporters/__init__.py#L1 should make that available at the top level API, since it's specified in `__all__`.

willingc:
I thought `__all__` would have covered it as well. I'll take a look at it more tomorrow on my flight.

Carreau:
That's because of the name conflict `nbconvert.exporter.export` module vs function described in https://github.com/jupyter/nbconvert/issues/415

PR #405, opened 2016-09-14
warn when IPython lexers are unavailable

### LINKED ISSUE #358: The highlight.py filter uses IPython -- setup.py doesn't declare it
While packaging nbconvert for Debian, I was hit by the fact that its setup.py doesn't declare any dep on ipython, even though (1) it does have a dep on it and (2) it needs to be a recent enough ipython.

Indeed, when checking my nbconvert package with my ipython 5.0.0 package available at home, everything went smoothly, but when I sent the result for validation to the Debian build bots, they only found the ancient ipython 2.4.1 package and the tests failed at each place "from IPython.lib.lexers import ..." was used, as there is no module lexers in there.

I suggest setup.py should declare a suitable versioned dep.

minrk:
and fallback on Python, instead of raising.

closes #358

### DIFF
--- a/nbconvert/filters/highlight.py
+++ b/nbconvert/filters/highlight.py
@@ -122,13 +122,25 @@ def _pygments_highlight(source, output_formatter, language='ipython', metadata=N
 
         language = metadata['magics_language']
 
+    lexer = None
     if language == 'ipython2':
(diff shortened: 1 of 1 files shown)


michaelpacer:
LGTM, with @takluyver's +1, I'm merging.

PR #408, opened 2016-09-15
Fix to allow invalidating the cached Jinja environment with recent traitlets

takluyver:
New versions of traitlets now always pass an argument, so it failed without the argument. This was already fixed for the template cache.

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -82,7 +82,7 @@ class TemplateExporter(Exporter):
         return self._template_cached
 
     _environment_cached = None
-    def _invalidate_environment_cache(self):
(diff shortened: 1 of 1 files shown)


minrk:
Weird, I'm not entirely sure how this didn't error as soon as `.observe` was picked up. `observe` shouldn't ever have had the argspec-sniffing that the deprecated `on_trait_change` does.

minrk:
Oh, wait. The `.observe` is only in master, and thus new in 5.0. Nothing to backport for 4.x.

takluyver:
I also think it probably doesn't get used all that often, because the traitlets that trigger it aren't things most people change.

minrk:
Good point.

Issue #413, opened 2016-09-17
get_exporter_names() returns each exporter name twice


michaelpacer:
I'm guessing this has to do with it loading them separately as entrypoints and as manually encoded values.

Issue #361, opened 2016-08-15
Better support metadata overriding script export with externally provided exporter


takluyver:
At IRkernel/IRkernel#394, we're discussing providing a custom script exporter, to be declared in notebook metadata in the `nbconvert_exporter` field. We would provide this as a separate package and register the exporter as an entry point. However, if someone attempts to nbconvert to a script without the exporter package installed, I think it will currently fail.

We could:
- Catch the failure in `ScriptExporter` and fall back to plain script export.
- Add a metadata field to specify a Python package to install for the exporter. However, we'd have to work out how to present this, because doing it the obvious way opens an arbitrary code execution hole from nbconvert.

michaelpacer:
@takluyver Is there a reason for this to be on 4.3 instead of 5.0? I think 5.0 is being released soon and this would seem to be more than a tiny bug fix (i.e., it should involve a few tests regardless of how we decide to implemement it), meaning that it would fit better with the main release. 

As for actually delivering on the feature, is there an example of a custom exporter in the vein of `nbconvert_export_r` per https://github.com/IRkernel/IRkernel/issues/394 to use in the development of this feature? If not, might this even be better as a 5.1 release since developing it and making sure it not only works in that case but for at least two kinds of custom exporter… 

Also, could you say more about the obvious way to do it and why that would create an arbitrary code execution hole? It would seem that the example given in [the "registering custom exporters as entry points" docs](http://nbconvert.readthedocs.io/en/latest/external_exporters.html#registering-a-custom-exporter-as-an-entry-point) would open this hole as well.

takluyver:
Milestones: my assignment wasn't anything definite or final, especially if we're planning nbconvert 5 soonish. I just didn't want it lingering too long.

Examples: at present, [tpllib](https://pypi.python.org/pypi/tpllib/0.0.1) is the only example of a separately packaged exporter I know of, and it doesn't declare an entry point. We should be able to knock together an example easily enough, though.

Security holes: Again, it's a matter of expectations and trust. If you `pip install` an external package for an exporter, we presume you understand the implications of installing a package from the internet.

The 'obvious way' I was referring to would be to add a field to the kernel info metadata, e.g. `"script_exporter_package": "nbconvert_export_r"`. When you did `jupyter nbconvert --to script`, it would automatically `pip install nbconvert_export_r` before trying to load the exporter. This is problematic because nbconvert is not an operation you expect to start installing and executing arbitrary code.

[...]

PR #414, opened 2016-09-17
Custom script exporter[WIP]

michaelpacer:
So this should address #361 at least partially. 

But the code is icky. Try excepts inside other try excepts with if statements embedded…

Also, I don't like the way it's printing three exceptions right now when applied to the included test `ScriptExporterTest.ipynb`, two that we don't want that return a list of exporters(one for `this_doesnt_exist` and one for `do_you_see_me`) and the one that we do want for `do_you_see_me`  (which includes the error message to try to `pip install do_you_see_me`)

It looks like that's happening because there is an error associated with `get_exporter` that we aren't able to suppress. One solution would be to have a flag in `get_exporter()` that can suppress printing an error message. If we were to do that, I'd suggest only turning the flag on for the new custom_exporter_name call to `get_exporter` because there may be cases where there is no `nbconvert_exporter_package` we want the other `get_exporter` to err.  

Also this process revealed #413 

I could use some help in thinking through this more clearly for the sake of having less terrible looking code. But, the fact that those `get_exporter()` calls are both happening suggests to me that, however ugly, the logic behind this is solid.

### DIFF
--- a/nbconvert/exporters/script.py
+++ b/nbconvert/exporters/script.py
@@ -18,7 +18,6 @@ class ScriptExporter(TemplateExporter):
 
     def from_notebook_node(self, nb, resources=None, **kw):
         langinfo = nb.metadata.get('language_info', {})
-        print("I am here") 
         # delegate to custom exporter, if specified
         exporter_name = langinfo.get('nbconvert_exporter')
         if exporter_name and exporter_name != 'script':
@@ -27,10 +26,8 @@ class ScriptExporter(TemplateExporter):
             if exporter_name not in self._exporters:
                 try: 
                     Exporter = get_exporter(exporter_name)
-                    print("i am here")
                     self._exporters[exporter_name] = Exporter(parent=self)
                 except:
-                    print("yes you found me")
                     if langinfo.get('nbconvert_exporter_package',""):
                         external_exporter_name = langinfo.get('nbconvert_exporter_package')
                         try: 
--- a/nbconvert/exporters/script.py
+++ b/nbconvert/exporters/script.py
@@ -7,7 +7,6 @@ from .templateexporter import TemplateExporter
 
 from traitlets import Dict, default
 from traitlets.utils.importstring import import_item
-from .export import ExporterNameError
 
 class ScriptExporter(TemplateExporter):
     
@@ -38,6 +37,7 @@ class ScriptExporter(TemplateExporter):
                             Exporter = get_exporter(external_exporter_name)
                             self._exporters[external_exporter_name] = Exporter(parent=self)
                         except:
+                            from .export import ExporterNameError
                             raise ExporterNameError("The Exporter named {nbconvert_exporter_package} cannot be found, try pip install {nbconvert_exporter_package}.".format(nbconvert_exporter_package=external_exporter_name))
                     else:
                         raise ExporterNameError("The Exporter named {exporter_name} cannot be found.".format(exporter_name=exporter_name)) 
--- a/nbconvert/exporters/script.py
+++ b/nbconvert/exporters/script.py
@@ -27,13 +27,13 @@ class ScriptExporter(TemplateExporter):
                 try: 
                     Exporter = get_exporter(exporter_name)
                     self._exporters[exporter_name] = Exporter(parent=self)
-                except:
+                except ValueError:
                     if langinfo.get('nbconvert_exporter_package',""):
                         external_exporter_name = langinfo.get('nbconvert_exporter_package')
                         try: 
                             Exporter = get_exporter(external_exporter_name)
                             self._exporters[external_exporter_name] = Exporter(parent=self)
-                        except:
+                        except ValueError:
                             raise ExporterNameError("The Exporter named {nbconvert_exporter_package} cannot be found, try pip install {nbconvert_exporter_package}.".format(nbconvert_exporter_package=external_exporter_name))
                     else:
                         raise ExporterNameError("The Exporter named {exporter_name} cannot be found.".format(exporter_name=exporter_name)) 
--- a/nbconvert/exporters/script.py
+++ b/nbconvert/exporters/script.py
@@ -34,7 +34,10 @@ class ScriptExporter(TemplateExporter):
                             Exporter = get_exporter(external_exporter_name)
                             self._exporters[external_exporter_name] = Exporter(parent=self)
                         except ValueError:
-                            raise ExporterNameError("The Exporter named {nbconvert_exporter_package} cannot be found, try pip install {nbconvert_exporter_package}.".format(nbconvert_exporter_package=external_exporter_name))
+                            raise ExporterNameError("The Exporter named "
+                            "{nbconvert_exporter_package} cannot be found"
+                            ", try pip install "
+                            "{nbconvert_exporter_package}.".format(nbconvert_exporter_package=external_exporter_name))
                     else:
                         raise ExporterNameError("The Exporter named {exporter_name} cannot be found.".format(exporter_name=exporter_name)) 
 
--- a/nbconvert/exporters/script.py
+++ b/nbconvert/exporters/script.py
@@ -7,7 +7,7 @@ from .templateexporter import TemplateExporter
 
 from traitlets import Dict, default
 from traitlets.utils.importstring import import_item
-
+from .export import ExporterNameError
 
 class ScriptExporter(TemplateExporter):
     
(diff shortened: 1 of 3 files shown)


nbconvert/exporters/script.py
```diff
@@ -24,10 +24,22 @@ def from_notebook_node(self, nb, resources=None, **kw):
...
             self.log.debug("Loading script exporter: %s", exporter_name)
-            from .export import get_exporter
+            from .export import get_exporter, ExporterNameError
             if exporter_name not in self._exporters:
-                Exporter = get_exporter(exporter_name)
-                self._exporters[exporter_name] = Exporter(parent=self)
+                try: 
+                    Exporter = get_exporter(exporter_name)
+                    self._exporters[exporter_name] = Exporter(parent=self)
+                except:
```
Carreau: try to never leave blank except, they would catch `KeyboardInterrupt`.

nbconvert/exporters/script.py
```diff
@@ -24,10 +24,22 @@ def from_notebook_node(self, nb, resources=None, **kw):
...
+                    if langinfo.get('nbconvert_exporter_package',""):
+                        external_exporter_name = langinfo.get('nbconvert_exporter_package')
+                        try: 
+                            Exporter = get_exporter(external_exporter_name)
+                            self._exporters[external_exporter_name] = Exporter(parent=self)
+                        except:
```
Carreau: Here as well, no blank except if possible.

nbconvert/exporters/script.py
```diff
@@ -24,10 +24,22 @@ def from_notebook_node(self, nb, resources=None, **kw):
...
+                            raise ExporterNameError("The Exporter named {nbconvert_exporter_package} cannot be found, try pip install {nbconvert_exporter_package}.".format(nbconvert_exporter_package=external_exporter_name))
```
Carreau: to avoid long lines you can use multiple single string, they will be automatically join.

```
"This will"
" end up on the same line"
" don't forget the witespace"
```

michaelpacer: should be addressed by 	bf0793c

nbconvert/exporters/script.py
michaelpacer: Fixed by c7ea343

nbconvert/exporters/script.py
michaelpacer: Fixed by c7ea343

nbconvert/exporters/script.py
```diff
@@ -19,15 +18,29 @@ def _template_file_default(self):
...
-                Exporter = get_exporter(exporter_name)
-                self._exporters[exporter_name] = Exporter(parent=self)
+                try: 
+                    Exporter = get_exporter(exporter_name)
+                    self._exporters[exporter_name] = Exporter(parent=self)
+                except ValueError:
+                    if langinfo.get('nbconvert_exporter_package',""):
+                        external_exporter_name = langinfo.get('nbconvert_exporter_package')
+                        try: 
+                            Exporter = get_exporter(external_exporter_name)
```
takluyver: I don't think it should try using the package name to get an exporter. My idea is that there are two separate bits of data:

- `nbconvert_exporter` specifies the exporter name which it tries to look up to convert the notebook to a script
- `nbconvert_exporter_package` is *only* used for the package we suggest the user installs if we can't find `nbconvert_exporter`.

They may have the same value in some cases, but that's OK. That should simplify the code.

Carreau: I don't think we should store tah in each notebook metadata, it feels to me as a recipe for confusion. This seem already that it will be used only for language specific hook, and thus I woudl much rather apply that at the script exporter lever via entry point: Language->Exporter.

First it allows us not to upgrade all notebooks metadata which is always a complicated thing.

Plus having that on a per notebook basis will lead to some notebook being converted in some ways while other won't which will be nightmare if people exchange notebook. 

it seem to me that hook into the script-exporter also make things simpler. In particular if it's in kernelspec **and** you need to have a python-package that match it's a recipe for inconsistency. Plus adding this bunch of metadate make it a de facto standard, while other tools that convert notebook might want to behave differently. 

I'm pretty sure we should even remove our own "nbconvert_exporter" key and make it custom logic in nbconvert.

takluyver: I'm not clear what you're suggesting as the alternative.

One other way we could do it is for `ScriptExporter` to look in a second group of entry points, like `nbconvert.exporters.script`, keyed by the language name already saved in the notebook. That would probably be more like my other suggestion for this issue: there's no suggestion to install a package, but it falls back to `ScriptExporter` if it can't find a more specific exporter.

 We should keep supporting the existing metadata field for some time even if we do deprecate it, because it's explicitly public API.

Carreau: > One other way we could do it is for ScriptExporter to look in a second group of entry points, like nbconvert.exporters.script, keyed by the language name already saved in the notebook. That would probably be more like my other suggestion for this issue: there's no suggestion to install a package, but it falls back to ScriptExporter if it can't find a more specific exporter.

Yes that's why I'm proposing sorry if I missed that.

> We should keep supporting the existing metadata field for some time even if we do deprecate it, because it's explicitly public API.

Of course, and have a deprecation warning if we encounter that.
Falling back to the standard exporter seem fine with me. We can also figure out something at kernel-registration directory level. I just think that embedding that into the notebook themselves is wrong.

takluyver: Embedding data in the notebook like that is definitely ugly, but we've generally lived with it as a compromise so we can view/convert notebooks without the kernel in question installed.

I agree that the entry points approach is neater, and it reduces the security concern over recommending installing a third party package. But let's take a couple of days to mull it over before anyone starts implementing, because I feel like I'm getting tunnel vision on this.

michaelpacer: So, the thought is that `do_you_see_me` is the name of the package that would declare the entrypoint 

```python
setup_args['entry_points']={
  "nbconvert.exporters": [
       'this_doesnt_exist=nbconvert.exporters:ThisDoesntExistExporter'
    ]
}
```

?

takluyver: Right - `nbconvert_exporter` specifies the entry point, and `nbconvert_exporter_package` is the package which includes the entry point. The latter doesn't need to be used programmatically, it's just for showing information to the user.

We think we've come up with a better alternative below, but I think we'd benefit from taking a step back from this for a day or two before we implement either.

michaelpacer: So for this, I don't feel confident in writing the code that succeeds at this without being able to refer to and use an external script exporter. That means including a test for this functionality. 

If we want to go down this route, is there a way to add a something to the entrypoints.nbconvert.exporters.script only for the purpose of testing?

takluyver:
OK, coming back to this. We worked out two possible options:

1. Allow notebook metadata to specify an exporter to be used in place of the basic `script` exporter. This can be installed using the existing `nbconvert.exporters` entry points. We may add a field (as this PR does) to recommend a package to install for that exporter. This is simpler given the infrastructure we already have, but adds more to notebook metadata, and only works for a notebook once it's been opened with a kernel that knows about this mechanism and saved.
2. Use the language name (or some other piece of information, e.g. kernel name) as the key in a new group of entry points, `nbconvert.exporters.script`. This involves less metadata and works without kernels needing to add any information, but it may be less obvious to users that they can install something for a better export.

I think that we were broadly coming round to option 2, but now that we're looking at it again, do we want to reconsider that? Has anyone thought of a fantastic option 3?

michaelpacer:
Talking with @takluyver today we came up with a best (worst?) of both worlds solution

2a) We create a second entrypoint group for scripts exporters that we use on the nbconvert side to find the exporters in question, but then we also give support for kernel authors to include additional metadata in the notebook if they have a preferred exporter such that if no custom exporter is found, an option is suggested.

takluyver:
Thinking more about it, I realised that there are two steps in 2a which can be done separately. I think we should implement the script exporter entry points (option 2), and then separately debate whether to add a metadata field for kernel authors to suggest an exporter package.

This leaves the question of which existing metadata field to use to select an exporter. Language name is the obvious one, and in most cases the most informative, but it's a bit ugly to apply our `ipython2python` transformation to all Python notebooks, even if they may have been made by another Python kernel. We're fibbing a bit by having an extended syntax and still marking the notebooks as being Python. I think we should just live with that inconsistency, though; there should be no harm in running our filter if all the code is valid Python.

Issue #291, opened 2016-04-14
RSTExporter should use ".. math::" instead of ".. raw:: latex"


mgeier:
If LaTeX environments are used in Markdown cells in The Notebook, those are passed to MathJax to be treated as math.
The `RSTExporter`, however, treats it as "raw LaTeX", which means that it disappears if further converted to HTML from there.

A way to implement this can be seen in https://github.com/spatialaudio/nbsphinx/pull/35.

Issue #232, opened 2016-02-05
bad math environment delimiter on conversion to pdf when using equation or align


mkgilbert:
I've noticed that nbconvert doesn't seem to work when there are \begin{equation} or \begin{align} tags in markdown cells. I have created a one-cell notebook with a markdown cell that has the contents:

```
$\begin{equation}
y=x+2
\end{equation}$
```
If I do a download as->pdf, I get the "pdf creation failed" message. The output of the conversion where it fails says:

```
! LaTeX Error: Bad math environment delimiter.

    See the LaTeX manual or LaTeX Companion for explanation.
    Type  H <return>  for immediate help.
     ...

    l.283     \(\begin{equation}
                                 y=x+2 \end{equation}\)
    ?
    ! Emergency stop.
     ...

    l.283     \(\begin{equation}
                                 y=x+2 \end{equation}\)
    !  ==> Fatal error occurred, no output PDF file produced!
    Transcript written on notebook.log.

```
The same thing happens if I try to use align.
Is this a problem with the way I'm using the delimiters?

mgeier:
You should remove the `$` signs and make sure that the LaTeX environment is in a paragraph on it's own.

If it still doesn't work, that's a bug in `nbconvert`. E.g for `rst` output it doesn't work: #291.
HTML and LaTeX output should work, though.

michaelpacer:
@mkgilbert Did removing the `$`s work for you? If so I want to close the issue.

PR #416, opened 2016-09-17
Add @mgeier's math tagging for rst conversion[WIP]

michaelpacer:
Should address #291. 

Some note on review: I just copied that code as it was used in production with nbsphinx. It LGTM, but I've not put it through more extensive testing; I'm assuming that (if it doesn't break our tests) then it does as @mgeier describes in #291 and alludes to in #232.


michaelpacer:
Is there a reason that our pandoc calls inside filters don't include `filter_func`'s by default? If not then, this seems like it should work.

michaelpacer:
Aha…it's because we don't use that at all because our pandoc call is structured slightly differently than [nbsphinx's] (https://github.com/spatialaudio/nbsphinx/blob/master/nbsphinx.py#L723), which has a two step call one to go to json and then a second to convert from json to whichever format is needed. 

This will take a little more work than I thought.

michaelpacer:
Hmmm…one solution would be to perform the intermediate `json` step inside the rst2markdown converter, but that is messy. another option would be to have an alternate execution path for `pandoc` wherein if it has a filter like this then it goes through this intermediate json layer. This way current functionality isn't adjusted. I'm going to investigate that path.

nbconvert/filters/markdown.py
```diff
@@ -102,4 +102,4 @@ def rawlatex2math(text):
         json_data = json.loads(text, object_hook=rawlatex2math_hook)
         return json.dumps(json_data)
 
-    return pandoc(source, 'markdown', 'rst', extra_args=extra_args)
+    return pandoc(source, 'markdown', 'rst', extra_args=extra_args, filter_func=rawlatex2math)
```
michaelpacer: This doesn't have def to refer back to as far as I can tell. no wonder it doesn't work.

nbconvert/filters/markdown.py
```diff
@@ -117,5 +117,7 @@ def rawlatex2math_hook(obj):
     def rawlatex2math(text):
         json_data = json.loads(text, object_hook=rawlatex2math_hook)
         return json.dumps(json_data)
+    
+    temp_text =
```
michaelpacer: This is invalid :(

michaelpacer:
This is going the wrong direction, fix using the new syntax for intermediate actions on the json internals as a pure python process.

michaelpacer:
In particular note: [these results](https://github.com/jupyter/nbconvert/pull/458/files#diff-fa9719444368d3ed717eaea36d19e9a7R67) associated with https://github.com/jupyter/nbconvert/pull/458

nbconvert/filters/markdown.py
```diff
@@ -100,4 +102,21 @@ def markdown2rst(source, extra_args=None):
...
+                ]
+            }]
+        return obj
+
+    def rawlatex2math(text):
+        json_data = json.loads(text, object_hook=rawlatex2math_hook)
+        return json.dumps(json_data)
+    
+
+    return convert_pandoc(source, 'markdown', 'rst', extra_args=extra_args)
```
takluyver: Did you mean to use the new `pandoc_with_filter` function you defined here? At the moment, I don't think it's using the functions functions you define above.

PR #418, opened 2016-09-17
Stop codecov from posting comments.

Carreau:


michaelpacer:
What are the up- and down-sides to this?

Carreau:
upside: CodeCove will complain only if the coverage drop by 10% or more.
downside: CodeCove will complain only if the coverage drop by 10% or more.

michaelpacer:
I'm going to guess you were being humorous with that response. But I still don't understand, so I'll try to rephrase.

What is the upside or the downside to including CodeCove in this way? Why would you want 10% to be the deciding line?

Is it that it is a relative decrease of 10% of what it was before to now, or an absolute decrease for 10% of the codebase (meaning that if we were below 10% it would be impossible for code coverage to ever comment on anything)?

If a 10% drop is intended to be a rare event, what is the advantage of maintaining a state where CodeCove can complain at all? 

If a 10% drop is expected to be a common event, why have a limit on it at all?

Does this mean that it will never alert us to a good change in code coverage meaning we should prioritise certain PRs on the basis of their improvement in code coverage (e.g., testing focused PRs)?

Carreau:
Codecov does comments like [this one](https://github.com/jupyter/nbconvert/pull/417#issuecomment-2478087850) it use to be configurable from their own webpannel but now you have to include a codecov.yml, which we did in the notebook and ipython repository. 

The comment were deemed annoying (especially since they get received by email). The behavior of CodeCov is slightly better than before as it modify its own comments instead of reposting a new one.

I choose the value for uniformity with other repos. 10% is arbitrary but is high enough to likely not trigger false positive on low decrease of coverage because  you removed white lines to technically decreased coverage. And low enough that a thing that would just stop testing a huge chunk of code would be caught. 

While code coverage is nice, because we are a user facing library it is really hard to target 100% coverage so our coverage move up and down you can also use the codecov browser plugin that show you the coverage on the diff as well as total coverage non intrusively.

michaelpacer:
Thank you for the explanation! That is exactly the kind of info that I was unsure about. Consistency across repos seems like a good thing to maintain as well. 

I'll merge.

Carreau:
> I'll merge.

you haven't.

Carreau:
> Thank you for the explanation! That is exactly the kind of info that I was unsure about. Consistency across repos seems like a good thing to maintain as well.

Great ! Sorry to have assumed you kind-of knew all that.

michaelpacer:
Oops, thought I had!

PR #419, opened 2016-09-17
Test on python nightly and allow failures.

Carreau:


michaelpacer:
LGTM. Is there any reason why we shouldn't incorporate this?

Carreau:
See https://github.com/takluyver/flit/pull/90 for some discussions.

michaelpacer:
I'm in favour of this. I'll merge and if there are objections or issues we can revert later.

PR #421, opened 2016-09-19
closes #413

### LINKED ISSUE #413: get_exporter_names() returns each exporter name twice

michaelpacer:
Cherry picked to remove this from other #414 that has more going on.


takluyver:
This looks fine. Hopefully merging it won't cause a conflict with the other PR.

PR #423, opened 2016-09-20
Refactor to avoid cycle

michaelpacer:
So this addresses the same issues as #417 but tries to do so by abstracting the hardcoded, deprecated exporter locating into a separate module from the entrypoint based exporter functionality which is not deprecated and instead is maintained in `base.py`. 

The only thing I don't like is that it introduces a new file that is strictly speaking deprecated because thought it is only acting as a shim for the export module that doesn't conflict with the name-space, all of its contents are deprecated. 

Another solution would be to just leave all the deprecated functionality inside `export.py` since it's deprecated anyway. I just worry that people might still use export and as a result be confused (and the ambiguity will strictly speaking continue to be present either way). I can make that change but it's not what I'd originally proposed, so I'll make the PR originally of what I proposed, and then will change it if that accords with others' intuitions about which is better.


michaelpacer:
Note — the most important change for demonstrating the improvement is in `script.py`; previously if we had moved that import there it would have failed to find `get_exporter` (because of the cyclic dependency).

nbconvert/exporters/exporter_locator.py
```diff
@@ -0,0 +1,86 @@
...
+
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+import warnings
+
+from .base import (export, get_exporter, 
+                   get_export_names , ExporterNameError)
+
+from .exporter import Exporter
```
Carreau: Nitpicking:  Exporter is unused, but I guess that would brea API to remove it ?

Carreau:
LGTM, but need rebase.

nbconvert/exporters/exporter_locator.py
michaelpacer: Exactly. So should we keep it?

Carreau: Sure. Let's merge once you get your rebase right.

Carreau:
BTW mike can you try to install [this](https://chrome.google.com/webstore/detail/codecov-extension/keefkhehidemnokodkdkejapdgfjmijf?hl=en-US)

michaelpacer:
done!

Carreau:
Good. 

> BTW mike can you try to install this

That was to show that with the extensions you can see the files coverage:
screen shot 2016-09-22 at 10 20 13

And per line coverage in the diff:

screen shot 2016-09-22 at 10 20 17

PR #426, opened 2016-09-22
Add warning about unsupported mimetypes.

### LINKED ISSUE #425: don't drop content silently
While debugging a lot of the nbconvert issues I had when writing my book, a common problem was that some content was silently dropped. You need to proof-read quite carefully to see that a particular figure or equation is missing.

Would it be possible to ensure that everything is rendered, or if not, (optionally) a warning is given?
I think what happened was usually that the template / exporter had no way to render a specific mime-type.

michaelpacer:
Closes #425. 

Adds explanation of the expected input structure to documentation of the `DataTypeFilter` `__call__` method.

### DIFF
--- a/nbconvert/filters/datatypefilter.py
+++ b/nbconvert/filters/datatypefilter.py
@@ -18,6 +18,7 @@ NbConvertBase.display_data_priority
 #-----------------------------------------------------------------------------
 
 from ..utils.base import NbConvertBase
+from warnings import warn
 
 __all__ = ['DataTypeFilter']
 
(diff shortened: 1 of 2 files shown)


michaelpacer:
Ok, so part of this ended up requiring a (needed) change to our tests for the datatype filters, which were based on lists, when lists are not actually what is supposed to be passed into the DataTypeFilter function. 

Rather it is supposed to be a dict or – rather – a dict like object (specifically, it's subselection of a NotebookNode which appears to retain its type though it no longer fits the nbformat spec (since it's a subselection), see: http://nbformat.readthedocs.io/en/latest/api.html#nbformat.NotebookNode). So I replaced the list with a dict.

Carreau:
+1, can you rebase/squash a bit ?

michaelpacer:
Rebased and squashed, can I take an approve mark as being ok to merge? Or should I still wait for someone else to merge?

Carreau:
> Rebased and squashed, can I take an approve mark as being ok to merge? Or should I still wait for someone else to merge?

Out of extra-precaution you can wait a bit to merge. Someone might come along. 
When I get blocked, I usually drop a "Self-merge-comment-warning" with a timeline, like "a few hours" "tomorrow", "next time I actually need it" or "next time I see this page", then self-merge if accordingly.

Also I should get used to the "Squash before merging" button of GitHub, and don't ask users to squash.

Carreau:
Also:  Thanks !

PR #428, opened 2016-09-23
Make exporter_map available from nbconvert.exporters.export

takluyver:
The notebook package expects to import it from there, so the rearrangement had broken it.

```
NotebookApp: WARNING: 500 GET /a%40b/nbconvert/html/foo/testnb.ipynb?download=False (127.0.0.1): Could not import nbconvert: cannot import name 'exporter_map'
```


takluyver:
Merging my own PR to fix test failures.

PR #429, opened 2016-09-27
Add command line commands for quickly building the relevant directory/file structures for custom exporters

michaelpacer:
All in title.


Carreau:
Thanks, you could have linked to http://choosealicense.com/

I would also change "You are not required to use this License." to explain that not using a liscence make things by default [not usable](http://choosealicense.com/no-license/).

PR #432, opened 2016-09-28
Add tornado and relative install to doc environment.yml requirements

### LINKED ISSUE #431: Current docs build fails without tornado; does not create an editable install of nbconvert
The current documentation instructions will fail for two reasons:

1. tornado is not included in the environment.yml file and therefore ServePostProcessor cannot be imported resulting in an error. 
2. when you build the environment anew, it will download the stable version of nbconvert, which is likely undesirable if you are building the current docs off of master. 

It is easy enough to solve 1 (see #432).
I am less confident about 2, I'll look into it.

michaelpacer:
Fixes #431 part 1.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -35,6 +35,11 @@ is hosted on ReadTheDocs.
 
    In your browser, go to `http://localhost:8000`.
 
+Note: If you want, you can manually convert this install to an editable install 
+by using ``pip uninstall nbconvert`` and ``pip install -e ..`` while still at the
+top of this ``docs`` directory. This will allow you to rebuild the html on the fly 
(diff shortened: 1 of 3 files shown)


michaelpacer:
f03ed5c fixes #431 part 2.

michaelpacer:
Fixing the last error (unexpected indentation) required a compromise I'm not totally happy with. But for some reason it did not like the 

```
 dict
       text that goes on for more characters
       worth of space than is strictly allowed
           - list item
           - another list item
```

structure, preferring the text that goes on for too many characters to be on a single line. It's not too egregious so I just made the change in this case, but I would like to figure out why it errs on that.

PR #435, opened 2016-09-30
Import audit conclusion

### LINKED ISSUE #424: General import statement audits
Need to go through to check for import statements that are hiding within functions

- [x] nbconvert/nbconvert
- [x] exporters
- [x] filters
- [x] postprocessors (3 ?s found)
- [x] preprocessors (4 ?s found)
- [x] ~~resources (mostly empty)~~
- [x] ~~templates (no `.py` files)~~
- [x] tests (1 ? found)
- [x] utils
- [x] writers

michaelpacer:
Closes #424. 

Pygments, IPython, jupyter-client imports were left alone.

Upgraded required version of tornado to 4.0 (per suggestion by @carreau).

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -17,6 +17,7 @@ import uuid
 from traitlets import HasTraits, Unicode, List, Dict, default, observe
 from traitlets.utils.importstring import import_item
 from ipython_genutils import py3compat
+from jinja2 import TemplateNotFound, Environment, ChoiceLoader, FileSystemLoader
 
 from nbconvert import filters
 from .exporter import Exporter
@@ -178,7 +179,6 @@ class TemplateExporter(Exporter):
         
         This is triggered by various trait changes that would change the template.
         """
-        from jinja2 import TemplateNotFound
 
         if not self.template_file:
             raise ValueError("No template_file specified!")
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -10,10 +10,6 @@ from __future__ import print_function, absolute_import
 import os
 import uuid
 
-# other libs/dependencies are imported at runtime
-# to move ImportErrors to runtime when the requirement is actually needed
-
-
 from traitlets import HasTraits, Unicode, List, Dict, default, observe
 from traitlets.utils.importstring import import_item
 from ipython_genutils import py3compat
@@ -295,7 +291,6 @@ class TemplateExporter(Exporter):
         """
         Create the Jinja templating environment.
         """
-        from jinja2 import Environment, ChoiceLoader, FileSystemLoader
         here = os.path.dirname(os.path.realpath(__file__))
 
         paths = self.template_path + \
--- a/nbconvert/postprocessors/serve.py
+++ b/nbconvert/postprocessors/serve.py
@@ -10,7 +10,6 @@ import webbrowser
 
 from tornado import web, ioloop, httpserver, log
 from tornado.httpclient import AsyncHTTPClient
-
 from traitlets import Bool, Unicode, Int
 
 from .base import PostProcessorBase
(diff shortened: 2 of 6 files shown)


michaelpacer:
@takluyver @minrk  since you had commented on the audit as it was happening — does this look good to you?

nbconvert/exporters/templateexporter.py
```diff
@@ -17,6 +17,7 @@
 from traitlets import HasTraits, Unicode, List, Dict, default, observe
 from traitlets.utils.importstring import import_item
 from ipython_genutils import py3compat
+from jinja2 import TemplateNotFound, Environment, ChoiceLoader, FileSystemLoader
```
takluyver: I think there's another import below that you can remove now that you've added this.

nbconvert/postprocessors/serve.py
```diff
@@ -75,14 +75,10 @@ def postprocess(self, input):
             cdn=self.reveal_cdn,
             client=AsyncHTTPClient(),
         )
-        # hook up tornado logging to our logger
```
takluyver: Let's leave this comment in to clarify slightly what's going on.

willingc:
@michaelpacer I took a quick look at this. I agree with @takluyver's comments. On the suggestion about leaving the logging comment, I would leave it in.

A question for all about style: Do we try to group imports into standard lib, 3rd party, current project?

nbconvert/preprocessors/csshtmlheader.py
```diff
@@ -7,11 +7,19 @@
...
 import hashlib
+import nbconvert.resources
 
 from traitlets import Unicode
 from ipython_genutils.py3compat import str_to_bytes
 from .base import Preprocessor
 
+
+try:
+    from notebook import DEFAULT_STATIC_FILES_PATH
```
minrk: I think this one should stay as a local import.

setup.py
```diff
@@ -198,7 +198,7 @@ def run(self):
     # FIXME: tests still require nose for some utility calls,
     # but we are running with pytest
     'test': ['pytest', 'pytest-cov', 'nose', 'ipykernel', 'testpath'],
-    'serve': ['tornado'],
+    'serve': ['tornado>=4.0,<5.0'],
```
minrk: Probably don't need to pin <5.0, can just mark >=4

takluyver:
> Do we try to group imports into standard lib, 3rd party, current project?

That's often what we do, but not consistently or strictly. Some editors and tools can automatically add to a straightforward alphabetical list of all imports, which is another common pattern.

I also go back and forth on whether to have all the `from` imports grouped after the `import` imports, or to mix them together.

nbconvert/preprocessors/csshtmlheader.py
michaelpacer: Can you explain why?

Carreau: Because otherwise that's make the notebook a hard dependency for all of nbconvert, while as local import it's only a dependency if this preprocessor is actually called.

michaelpacer: Doesn't the try-except fallback specifically make it *not* dependency for all of nbconvert?

michaelpacer:
@willingc should I go through all of the imports for this style check or should that wait for later?

nbconvert/postprocessors/serve.py
michaelpacer: addressed!

nbconvert/exporters/templateexporter.py
michaelpacer: addressed

setup.py
michaelpacer: addressed!

Carreau:
> should I go through all of the imports for this style check or should that wait for later?

I think you can wait. Let's try to get things rolling for now. And groupping import item seem like deciding is the bikeshed should get a glossy or mate painting.

nbconvert/preprocessors/csshtmlheader.py
Carreau: Ok, let's stop the bikeshed, we can move it later as a local import if needed.

Carreau:
Thanks @michaelpacer

willingc:
Thanks @michaelpacer. No need to change. I was just confirming the code style with others earlier :-)

PR #436, opened 2016-09-30
General pandoc filter

amueller:
This introduces a generic pandoc filter as discussed in #355.
I removed ``markdown2latex`` and ``markdown2rst``, though I'm not entirely sure if this is the best way to do it.
How do you guys feel about backward compatibility?


nbconvert/exporters/templateexporter.py
```diff
@@ -25,34 +25,34 @@
...
-        'markdown2html': filters.markdown2html,
-        'ansi2html': filters.ansi2html,
-        'filter_data_type': filters.DataTypeFilter,
-        'get_lines': filters.get_lines,
-        'highlight2html': filters.Highlight2HTML,
-        'highlight2latex': filters.Highlight2Latex,
-        'ipython2python': filters.ipython2python,
-        'posix_path': filters.posix_path,
-        'markdown2latex': filters.markdown2latex,
-        'markdown2rst': filters.markdown2rst,
```
Carreau: Juste as a note `markdown2latex` and `markdown2rst ` are the filters being removed.

Carreau:
> I removed markdown2latex and markdown2rst, though I'm not entirely sure if this is the best way to do it.
> How do you guys feel about backward compatibility?

I'm fine with removing them, I don't think we have many  custom templates. 

Though i'm starting to hand over responsibility to @michaelpacer. 

One possibility would be to alias them with a warning.

michaelpacer:
In general we want to deprecate things for at least one major version release before breaking any APIs. 

Also, it looks like this solution fails to add an extra_args command to recreate the original functionality of  `markdown2htmlpandoc`.

So what I'd do is just shim them and continue to expose them for now. This seems like a good kind of abstraction for the filter layer.

amueller:
@michaelpacer ok sounds good.

You mean the mathjax extra-arg?
Where would that go? The function was never used...

nbconvert/exporters/templateexporter.py
michaelpacer: that's only from the default set of templates.

`markdown2html_pandoc` was also removed elsewhere.

michaelpacer:
So you can just use your `convert_pandoc` in place of the current pandoc command and use the same logic as it had before. I would also add a deprecation warning (look at the `exporter_locator.py` for some examples of how that would look) to the old versions of the function so that we can actually deprecate them for 6.0.

michaelpacer:
When i run the code in this branch locally I'm not getting the errors that we're seeing in the build.

amueller:
Ok so just that I understand, you want to deprecate the ``markdown2latex``, ``markdown2rst`` and ``markdown2html_pandoc`` functions and replace these calls by the ``convert_pandoc`` function?

Hm I'm getting weird unrelated errors locally :-(

amueller:
```
__________________________________________________________ TestNbConvertApp.test_write_figures_to_custom_path ___________________________________________________________

self = <nbconvert.tests.test_nbconvertapp.TestNbConvertApp testMethod=test_write_figures_to_custom_path>

    @dec.onlyif_cmds_exist('pandoc')
    def test_write_figures_to_custom_path(self):
        """
            Check if figure files are copied to configured path.
            """
    
        def fig_exists(path):
            return (len(os.listdir(path)) > 0)
    
        # check absolute path
        with self.create_temp_cwd(['notebook4_jpeg.ipynb',
                                   'containerized_deployments.jpeg']):
            output_dir = tempdir.TemporaryDirectory()
            path = os.path.join(output_dir.name, 'files')
            self.nbconvert(
                '--log-level 0 notebook4_jpeg.ipynb --to rst '
                '--NbConvertApp.output_files_dir={}'
                .format(path))
>           assert fig_exists(path)

tests/test_nbconvertapp.py:435: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

path = '/tmp/tmprkto8i37/files'

    def fig_exists(path):
>       return (len(os.listdir(path)) > 0)
E       FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmprkto8i37/files'
```
Have you seen this before?

michaelpacer:
I've seen something like that before and it has to do with the util that makes temporary files, which I think comes from `from testpath import tempdir`. Do you have an up-to-date version of `testpath`? My guess is that you do…given that it hasn't been updated since February, but it can't hurt to check.

michaelpacer:
Also, the travis build looks to be something related to how you're handling the `extra-args` right now (or at least that's the best that I can tell).

takluyver:
Are you testing on Python 3 or 2? On Python 3, that `TemporaryDirectory` is actually implemented in the stdlib; on Python 2 it's a backport. Either way, I can't see how it would cause a test failure like that.

amueller:
yeah I need to fix the extra-args. I'm testing on Python 3.

michaelpacer:
@Carreau @takluyver @minrk Could one or more of you review this as well? It looks like its good to go from my perspective, even though the tests that rely on tempdir aren't working correctly on @amueller's system.

nbconvert/filters/markdown.py
```diff
@@ -56,20 +59,33 @@ def markdown2latex(source, markup='markdown', extra_args=None):
     out : string
       Output as returned by pandoc.
     """
-    return pandoc(source, markup, 'latex', extra_args=extra_args)
+    warnings.warn("""`markdown2latex` is deprecated in favor of 
+                  `nbconvert.filters.pandoc.generic_pandoc` with 
+                  appropriate arguments since nbconvert 5.0""")
```
Carreau: Nitpicking: `with appropriate arguments` you can give the appropriate new syntax. Also triple quoted string keep indentation:

```
In [4]: warnings.warn("""`markdown2latex` is deprecated in favor of
   ...:                   `nbconvert.filters.pandoc.generic_pandoc` with
   ...:                   appropriate arguments since nbconvert 5.0""")
/Users/bussonniermatthias/anaconda/bin/ipython:3: UserWarning: `markdown2latex` is deprecated in favor of
                  `nbconvert.filters.pandoc.generic_pandoc` with
                  appropriate arguments since nbconvert 5.0
```

You might want to use textwrap.dedent or auto joining of simple strings.

nbconvert/filters/markdown.py
```diff
@@ -56,20 +59,33 @@ def markdown2latex(source, markup='markdown', extra_args=None):
...
+    """
+    Deprecated since version 5.0.
+    
+    Convert a markdown string to HTML via pandoc.
+    """
     extra_args = extra_args or ['--mathjax']
-    return pandoc(source, 'markdown', 'html', extra_args=extra_args)
+    warnings.warn("""`markdown2html_pandoc` is deprecated in favor of 
+                  `nbconvert.filters.pandoc.generic_pandoc` with
+                  appropriate arguments since nbconvert 5.0""")
```
Carreau: same.

nbconvert/filters/markdown.py
```diff
@@ -84,4 +100,7 @@ def markdown2rst(source, extra_args=None):
     out : string
       Output as returned by pandoc.
     """
-    return pandoc(source, 'markdown', 'rst', extra_args=extra_args)
+    warnings.warn("""`markdown2rst` is deprecated in favor of 
+                  `nbconvert.filters.pandoc.generic_pandoc` with
+                  appropriate arguments since nbconvert 5.0""")
```
Carreau: same.

nbconvert/filters/tests/test_markdown.py
```diff
@@ -71,21 +75,25 @@ def test_markdown2latex_markup(self):
...
-            "#latex {{long_line | md2l('markdown', ['--no-wrap'])}}",
+            "#latex {{long_line | md2l(['--no-wrap'])}}",
             "#rst {{long_line | md2r(['--columns', '5'])}}",
         ])
         long_line = ' '.join(['long'] * 30)
         env = Environment()
         env.filters.update({
-            'md2l': markdown2latex,
-            'md2r': markdown2rst,
+            'md2l': lambda code, extra_args: convert_pandoc(
```
Carreau: `md2l` and `md2r` are unused. Are they necessary ?

amueller:
sorry for not following up on this @michaelpacer :-/

michaelpacer:
Its ok! I just wanted to see it move forward so we can also get your ascii_doc converter incorporated. I'd wanted to release 5.0 really soon, likely before the end of the week and this was close enough that I thought I'd just move forward on it. 

Thank _you_ for all the work you've put into it! It just required a few tweaks cause it was already well done.

nbconvert/filters/tests/test_markdown.py
```diff
@@ -175,8 +183,8 @@ def test_markdown2rst(self):
         tokens[1] = r'\*\*test'
 
         for index, test in enumerate(self.tests):
-            self._try_markdown(markdown2rst, test, tokens[index])
-
+            self._try_markdown(partial(convert_pandoc, from_format='markdown',
```
michaelpacer: @Carreau  This is what I mean, in this case it's using named arguments, but in (see other comment)…

nbconvert/templates/latex/document_contents.tplx
```diff
@@ -30,12 +30,12 @@
 
 % Display latex
 ((* block data_latex -*))
-    ((( output.data['text/latex'] | citation2latex | strip_files_prefix | markdown2latex )))
+    ((( output.data['text/latex'] | citation2latex | strip_files_prefix | convert_pandoc('markdown', 'latex'))))
```
michaelpacer: @Carreau (see earlier comment) in this case it's being invoked without a source argument at all, and no named arguments for the format

nbconvert/filters/tests/test_markdown.py
```diff
@@ -45,20 +48,21 @@ class TestMarkdown(TestsBase):
...
+                               self.tokens[index])
 
     @dec.onlyif_cmds_exist('pandoc')
     def test_markdown2latex_markup(self):
         """markdown2latex with markup kwarg test"""
         # This string should be passed through unaltered with pandoc's
         # markdown_strict reader
         s = '1) arabic number with parenthesis'
-        self.assertEqual(markdown2latex(s, markup='markdown_strict'), s)
+        self.assertEqual(convert_pandoc(s, 'markdown_strict', 'latex'), s)
```
michaelpacer: @Carreau  (part 3 of 3) In this case `convert_pandoc` is being invoked with a source argument, but with no named arguments.

This is why I wasn't sure what to include in the warning string.

nbconvert/filters/tests/test_markdown.py
michaelpacer: They are used. see lines 85 & 86 of the same file.

nbconvert/filters/tests/test_markdown.py
Carreau: I'm still confused by what you are asking.

michaelpacer: compare this to https://github.com/jupyter/nbconvert/pull/436/files#diff-fa9719444368d3ed717eaea36d19e9a7R33 and https://github.com/jupyter/nbconvert/pull/436/files#diff-72a549c7ebe4e2a80383f8351ef80d4fR65

In this case from_format is being treated as keyword argument, but in the other cases, it's being implicitly passed as a positional argument (but in one case in the first position and in the other case the second position).

     partial(convert_pandoc, from_format='markdown',to_format='rst')

vs. 

     convert_pandoc('markdown', 'latex')

vs. 

    convert_pandoc(s, 'markdown_strict', 'latex')

Carreau: The second case is only when you are in a template, and you pipe into the filter. 
If you want to be more explicit write the second a case as `... | convert_pandoc('markdown', 'latex') | ... ` than one can see how to use it from within jinja.

michaelpacer: Ok, tried to change it to take this into account…do the new warning strings look ok?

michaelpacer:
I think this is good to merge, and once this is merged, we can move forward on #355.

nbconvert/filters/markdown.py
```diff
@@ -56,20 +54,37 @@ def markdown2latex(source, markup='markdown', extra_args=None):
     out : string
       Output as returned by pandoc.
     """
-    return pandoc(source, markup, 'latex', extra_args=extra_args)
+    warnings.warn("`markdown2latex` is deprecated in favor of "
```
minrk: I would prefer to *not* deprecate these names, as they are as clear as and more concise than the pandoc call with arguments. I'd just leave them as convenient wrappers.

minrk:
Sorry for being slow to come in. I don't see a benefit of removing/deprecating the existing filter names. Leaving them as shorthand for the most common conversions seems like the most useful and convenient way to go.

Carreau:
I'm fine leaving them if min is for leaving them.

michaelpacer:
Ok, I'm still going to use the generic_pandoc call in the actual filters so that people get examples of how to do that, but we can still keep the filter names for those who wish to have a more concise representation.

Carreau:
> Ok, I'm still going to use the generic_pandoc call in the actual filters so that people get examples of how to do that, but we can still keep the filter names for those who wish to have a more concise representation.

Seem like a good compromise.

Carreau:
Thanks All !

minrk:
Great, thanks!

Issue #439, opened 2016-10-03
Troubleshoot build error on RTD


willingc:
RTD isn't happy with the latest `environment.yml` changes and nbsphinx or `../.` (https://github.com/jupyter/nbconvert/pull/432/files).

cc @michaelpacer P.S. If you create an account on RTD, I will add you as an admin/owner of the RTD instance. Just ping me once it is created 🌻

michaelpacer:
@willingc I have one! :) as @mdpacer (an old github handle)

willingc:
@michaelpacer You now have magic powers!

## A 20 second intro to RTD administration (ignore if you already know this)

### Advanced Settings Page

As currently set. The conda `environment.yml` file *should* override these but YMMV.

In fact, YMMV quite a bit with RTD builds. I don't assume that a failed build means that my configuration is incorrect as sometimes RTD can be a bit stubborn about clearing out old cruft. As such, I set up a `test-project` for prototyping builds (i.e. `willingc/test-nbconvert`).

<img width="501" alt="screen shot 2016-10-03 at 10 51 09" src="https://cloud.githubusercontent.com/assets/2680980/19047739/ec16232a-8957-11e6-9e36-991ae9cfa913.png">

### Builds Page

Click on a particular build to find out detail about how it failed.

You may also select a branch from the pull down and start a build.

<img width="483" alt="screen shot 2016-10-03 at 10 52 05" src="https://cloud.githubusercontent.com/assets/2680980/19047869/9322a2ec-8958-11e6-8272-55002b3d1bb6.png">

### Versions Page

Be nice to this page. It's your friend.

On this page, you can click "edit" to enable or disable building of a branch.

Clicking the "wipe" button **should** clear out the cruft of old builds and what not. In practice, that does not always happen especially if RTD pushed changes recently.

My approach is typically the following if a build has failed but I think my code is fine:

- Wipe branch (on versions page)
- Rebuild branch
- Wipe branch, latest, and stable
- Edit branch and disable and submit. Then reenable and submit.
- Take a break and try again in a couple of hours. :sunny:

**Versions Main Menu**

<img width="491" alt="screen shot 2016-10-03 at 11 09 43" src="https://cloud.githubusercontent.com/assets/2680980/19048152/f5e9ada2-8959-11e6-898f-fdcad74bf88e.png">


**Versions tab from Admin Menu** - does differ from Versions Main Menu

[...]

PR #440, opened 2016-10-03
Doc improver

michaelpacer:
This builds for my repo on rtd. https://readthedocs.org/projects/nbconvert-mp/builds/4469568/

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -7,38 +7,40 @@ is hosted on ReadTheDocs.
 
 1. Change directory to documentation root:
 
-           $ cd docs
+        $ cd docs
 
 2. Create conda env (and install relevant dependencies):
 
-           $ conda env create -f environment.yml
+        $ conda env create -f environment.yml
 
 3. Activate the newly built conda environment `nbconvert_docs`
 
-           $ source activate nbconvert_docs
+        $ source activate nbconvert_docs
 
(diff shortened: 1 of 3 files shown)


michaelpacer:
As pointed out by @Carreau, we can't actually release this solution because it will kill the ability to have documentation with branches.

michaelpacer:
with the previous caveat of being unreleasable, this seems to address #439

michaelpacer:
Yay! So on my rtd branch, this most recent version works and it won't be strictly tied to any particular release, but will build it relative to the current version of the file.

I'm guessing that a _lot_ of our issues in the past with rtd have been secretly linked to the fact that we were not building the correct version of nbconvert with which to run the docs by using pip.

Is this the solution used in other places? to include a pip call directly in RTD? 

Also, since I used the `subprocess.run()` command, we need to ensure that the python in the env is at least python 3.5 (which it seems rtd was already doing for me).

willingc:
@michaelpacer I'm going to merge this as the code looks good to me and it's building on your test site. We do use subprocess.run() in the ipywidgets docs build. 

My test on RTD is most grumpy. Probably will just delete it:
screen shot 2016-10-03 at 17 01 17

Overall, let's see how it goes after merging :tada: 🏄

PR #443, opened 2016-10-10
Add more informative error message if --to pdf fails

### LINKED ISSUE #411: Add TeX requirements to install docs
The current docs don't mention anything about installing TeX or having TeX installed but it is required to convert to PDF, etc. I understand that installing TeX differs depending on the user's OS (e.g. mactex for OS X), so I'm creating an issue vs. creating a PR because I'm not sure what the suggested TeX packages are for other OSs. It seems to me that this should be included right alongside the pandoc install instructions.

michaelpacer:
Should fix the last bit of #411 with TeX not giving an informative error when it cannot find the specified TeX interpreter (`xelatex`).

closes #411.


Carreau:
Thanks !

PR #454, opened 2016-11-01
Add pipeline overview to architecture documentation, fix README [WIP]

michaelpacer:
I'm planning on looking at the writeup of the class structure more closely as I tried to extract much of the higher order description into the opening preamble.

I also want to add a few images to illustrate the structure.

### DIFF
--- a/docs/source/architecture.rst
+++ b/docs/source/architecture.rst
@@ -4,14 +4,62 @@
 Architecture of nbconvert
 =========================
 
-This is a high-level outline of the structure and objects in nbconvert,
-and how they are used in the pipeline of converting a notebook to any given format.
+This is a high-level outline of the basic workflow, structures and objects in nbconvert.
+Specifically, this exposition has a two-fold goal: 
+    
+    #. to alert you to the affordances available for customisation or direct contributions 
+    #. to provide a map of where and when different events occur, which should aid in tracking down bugs.
+    
 
+A detailed pipeline exploration
+===============================
+
+Nbconvert takes in a notebook, which is a JSON object, and operates on that object. 
+
+This can include operations that take a notebook and return a notebook.
+For example, that operation could be to execute the notebook as though it were a continuous script; if it were executed ``--in-place`` then it would overwite the current notebook.
+Or it could be that we wish to systematically alter the notebook, for example by clearing all output cells.
+Format agnostic operations on cell content that do not violate the nbformat spec can be interpreted as a notebook to notebook conversion step; such operations can be performed as part of the preprocessing step.
+
+But often we want to have the notebook's structured content in a different format.
+Importantly, in many cases the structure of the notebook should be reflected in the structure of the output, adapted to the output's format.
+For that purpose, the original JSON structure of the document is crucial scaffolding needed to support this kind of structured output.
+In order to maintain structured, it can be useful to apply our conversion programmatically on the structure itself.
+To do so, when converting to formats other than the notebook, we use the `jinja`_ templating engine.
+
+The basic unit of structure in a notebook is the cell.
+Accordingly, since our templating engine is capable of expressing structure, the basic unit in our templates will often be specified at the cell level.
+Each cell has a certain type; the three most important cell types for our purposes are code, markdown, and raw NbConvert.
+Code cells can be split further into their input and their output.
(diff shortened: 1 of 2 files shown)


docs/source/architecture.rst
```diff
@@ -4,14 +4,62 @@
...
+
+A detailed pipeline exploration
+===============================
+
+Nbconvert takes in a notebook, which is a JSON object, and operates on that object. 
+
+This can include operations that take a notebook and return a notebook.
+For example, that operation could be to execute the notebook as though it were a continuous script; if it were executed ``--in-place`` then it would overwite the current notebook.
+Or it could be that we wish to systematically alter the notebook, for example by clearing all output cells.
+Format agnostic operations on cell content that do not violate the nbformat spec can be interpreted as a notebook to notebook conversion step; such operations can be performed as part of the preprocessing step.
```
Carreau: I would emphase preprocessing as is / will be the term that refer to notebook->notebook conversions.

docs/source/architecture.rst
```diff
@@ -4,14 +4,62 @@
...
+
+But often we want to have the notebook's structured content in a different format.
+Importantly, in many cases the structure of the notebook should be reflected in the structure of the output, adapted to the output's format.
+For that purpose, the original JSON structure of the document is crucial scaffolding needed to support this kind of structured output.
+In order to maintain structured, it can be useful to apply our conversion programmatically on the structure itself.
+To do so, when converting to formats other than the notebook, we use the `jinja`_ templating engine.
```
Carreau: Jinja is not mandatory, one could have an exporter that does not use Jinja , we might want to slight rephrase that maybe.

docs/source/architecture.rst
```diff
@@ -4,14 +4,62 @@
...
+Accordingly, since our templating engine is capable of expressing structure, the basic unit in our templates will often be specified at the cell level.
+Each cell has a certain type; the three most important cell types for our purposes are code, markdown, and raw NbConvert.
+Code cells can be split further into their input and their output.
+Operations can also occur separately on input and output and their respective subcomponents.
+Markdown cells and raw NbConvert cells do not have analogous substructure.
+
+The template's structure then can be seen as a mechanism for selecting content on which to operate.
+Because the template operates on individual cells, this has some upsides and drawbacks.
+One upside is that this allows the template to have access to the individual cell's metadata, which enables intelligently transforming the appropriate content. 
+The transformations occur as a series of replacement rules and filters.
```
Carreau: we can literally link filter using intesphinx to jinja docs.

docs/source/architecture.rst
```diff
@@ -4,14 +4,62 @@
...
+One downside is that this makes operations that require global coördination (e.g., cross referencing across cells) somewhat challenging to implement as filters inside templates.
+
+Note that all that we've described is happening in memory. 
+This is crucial in order to ensure that this functionality is available when writing files is more challenging.
+Nonetheless, the reason for using nbconvert almost always involves producing some kind of output file.
+We take the in-memory object and write a file appropriate for the output type.
+
+The entirety of heretofore described process can be described as part of an ``Exporter``. 
+``Exporter``s often involves ``Preprocessor``s, ``filters``, ``templates`` and ``Writer``s. 
+These classes and functions are described in greater detail below.
```
Carreau: One more note is that exporter are/should be stateless and can be reused.

In a further refinement we can speak about state in the form of the resource dictionary, that one can use.

Carreau:
Thanks.

docs/source/architecture.rst
michaelpacer: I'd thought through that and had decided against it. The reason I hadn't emphasised it in that way was that there is also a "notebook" exporter, so technically that is how nbconvert converts notebooks to notebooks (even if that's primarily a wrapper for preprocessors). And putting in a qualifier made it sound confusing.

Can you think of a way to say that without saying that preprocessing is the term that covers all notions of notebook to notebook conversion?

docs/source/architecture.rst
michaelpacer: What do you mean by linking filters?

docs/source/architecture.rst
michaelpacer: I'm still not 100% sure about way that the resource dictionary interacts with everything else. We should chat about that more.

PR #458, opened 2016-11-05
Link resolution

### LINKED ISSUE #11: Anchor links and conversion to LaTeX
Converting markdown like `[Header](#Header)` to LaTeX just strips out the anchor link instead of inserting `\ref{header}` (even though `\label{header}` is actually inserted next to the header)

michaelpacer:
Close #11 by passing an intermediate JSON representation through a filter that directly manipulates the underlying pandoc AST as part of a 2 step conversion process.

It also creates a general utility for walking through this representation in other cases based on the  https://github.com/jgm/pandocfilters library with a new utility (`applyJSONFilters`) that we are mirroring until/if  https://github.com/jgm/pandocfilters/pull/49 is merged.

@takluyver

### DIFF
--- a/nbconvert/exporters/latex.py
+++ b/nbconvert/exporters/latex.py
@@ -45,7 +45,8 @@ class LatexExporter(TemplateExporter):
     output_mimetype = 'text/latex'
 
     def default_filters(self):
-        yield from super(LatexExporter, self).default_filters()
+        for x in super(LatexExporter, self).default_filters():
+            yield x 
         yield ('resolve_references', resolve_references)
 
     @property
--- a/nbconvert/exporters/latex.py
+++ b/nbconvert/exporters/latex.py
@@ -45,7 +45,7 @@ class LatexExporter(TemplateExporter):
     output_mimetype = 'text/latex'
 
     def default_filters(self):
-        yield from super(LatexExporter,self).default_filters()
+        yield from super(LatexExporter, self).default_filters()
         yield ('resolve_references', resolve_references)
 
     @property
--- a/nbconvert/exporters/latex.py
+++ b/nbconvert/exporters/latex.py
@@ -10,7 +10,7 @@ from traitlets.config import Config
 
 from nbconvert.filters.highlight import Highlight2Latex
 from .templateexporter import TemplateExporter
-
+from ..filters.filter_links import resolve_references
 
 class LatexExporter(TemplateExporter):
     """
@@ -43,6 +43,10 @@ class LatexExporter(TemplateExporter):
     template_extension = Unicode(".tplx").tag(config=True)
 
     output_mimetype = 'text/latex'
+    
+    def default_filters(self):
+        yield from super(TemplateExporter,self).default_filters()
+        yield ('resolve_references', resolve_references)
 
     @property
     def default_config(self):
--- a/nbconvert/exporters/latex.py
+++ b/nbconvert/exporters/latex.py
@@ -9,8 +9,8 @@ from traitlets import Unicode, default
 from traitlets.config import Config
 
 from nbconvert.filters.highlight import Highlight2Latex
+from nbconvert.filters.filter_links import resolve_references
 from .templateexporter import TemplateExporter
-from ..filters.filter_links import resolve_references
 
 class LatexExporter(TemplateExporter):
     """
@@ -43,9 +43,9 @@ class LatexExporter(TemplateExporter):
     template_extension = Unicode(".tplx").tag(config=True)
 
     output_mimetype = 'text/latex'
-    
+
     def default_filters(self):
-        yield from super(TemplateExporter,self).default_filters()
+        yield from super(LatexExporter,self).default_filters()
         yield ('resolve_references', resolve_references)
 
     @property
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -50,7 +50,6 @@ default_filters = {
         'prevent_list_blocks': filters.prevent_list_blocks,
         'get_metadata': filters.get_metadata,
         'convert_pandoc': filters.convert_pandoc,
-        'wrapped_convert_link': wrapped_convert_link,
 }
 
 
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -17,7 +17,7 @@ from jinja2 import TemplateNotFound, Environment, ChoiceLoader, FileSystemLoader
 
 from nbconvert import filters
 from .exporter import Exporter
-
+from ..filters.filter_links import wrapped_convert_link
 #Jinja2 extensions to load.
 JINJA_EXTENSIONS = ['jinja2.ext.loopcontrols']
 
@@ -50,6 +50,7 @@ default_filters = {
         'prevent_list_blocks': filters.prevent_list_blocks,
         'get_metadata': filters.get_metadata,
         'convert_pandoc': filters.convert_pandoc,
+        'wrapped_convert_link': wrapped_convert_link,
 }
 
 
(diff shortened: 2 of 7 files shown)


nbconvert/filters/filter_links.py
```diff
@@ -0,0 +1,33 @@
...
+    return applyJSONFilters([convert_link], source)
+
+def convert_link(key, val, fmt, meta):
+    """
+    """
+    
+    if key == 'Link':
+        target = val[2][0]
+        # Links to other notebooks
+        m = re.match(r'(\d+\-.+)\.ipynb$', target)
```
takluyver: The cross-notebook linking only works as part of bookbook, so we should probably take it out of here to avoid confusion.

nbconvert/filters/filter_links.py
```diff
@@ -0,0 +1,33 @@
...
+        if m:
+            return RawInline('tex', 'Section \\ref{sec:%s}' % m.group(1))
+            
+        # Links to sections of this or other notebooks
+        m = re.match(r'(\d+\-.+\.ipynb)?#(.+)$', target)
```
takluyver: Ditto here, the first part of this only makes sense for bookbook joining multiple notebooks together.

nbconvert/filters/pandoc.py
```diff
@@ -23,4 +24,8 @@ def convert_pandoc(source, from_format, to_format, extra_args=None):
     out : string
       Output as returned by pandoc.
     """
+#    mydir = os.path.dirname(os.path.abspath(__file__))
+#    filter_links = os.path.join(mydir, 'filter_links.py')
+#    if extra_args is not None:
+#        extra_args.extend(['--filter',filter_links])
```
takluyver: You probably didn't mean to commit this?

nbconvert/templates/latex/document_contents.tplx
```diff
@@ -64,7 +64,7 @@
 
 % Render markdown
 ((* block markdowncell scoped *))
-    ((( cell.source | citation2latex | strip_files_prefix | convert_pandoc('markdown', 'latex') )))
+    ((( cell.source | citation2latex | strip_files_prefix | convert_pandoc('markdown', 'json',extra_args=[]) | wrapped_convert_link | convert_pandoc('json','latex'))))
```
takluyver: Open API question - do we want to have a function which encapsulates these last three steps into one:

- pandoc to json
- apply Python filters
- pandoc from json

nbconvert/filters/pandoc.py
michaelpacer: Yes meant to delete that is left over from the efficiency prototyping

michaelpacer: Yes meant to delete that is left over from the efficiency prototyping

nbconvert/utils/pandoc.py
```diff
@@ -109,6 +109,28 @@ def check_pandoc_version():
                        RuntimeWarning, stacklevel=2)
     return ok
 
+try: 
+    from pandocfilters import applyJSONFilters
+except ImportError:
```
takluyver: John did a new release of pandocfilters for us, so we can rely on `>=1.4.1` and drop the falllback here.

michaelpacer: Sounds good.

nbconvert/templates/latex/document_contents.tplx
michaelpacer: I was thinking about doing that, but I figured that should be a different PR that demonstrates the generality of this, whereas this is more of a bugfix.

nbconvert/exporters/templateexporter.py
```diff
@@ -50,6 +50,7 @@
         'prevent_list_blocks': filters.prevent_list_blocks,
         'get_metadata': filters.get_metadata,
         'convert_pandoc': filters.convert_pandoc,
+        'wrapped_convert_link': wrapped_convert_link,
```
takluyver: I'd like the name here to be a bit more descriptive: `wrapped` is an implementation detail, and it's not clear what it means (wrapped in what?). A descriptive name might be like `pandoc_internal_links_to_latex_refs`. That is a bit wordy, so better suggestions welcome.

I also wonder if it should only be registered for the latex exporter - then the name could be a bit less precise. Adding a filter for an exporter works like this (example from bookbook):

```python
def default_filters(self):
    yield from super(MyHTMLExporter, self).default_filters()
    yield ('markdown2html', markdown2html_custom)
```

nbconvert/utils/pandoc.py
```diff
@@ -109,6 +109,22 @@ def check_pandoc_version():
                        RuntimeWarning, stacklevel=2)
     return ok
 
+    def applyJSONFilters(actions, source, format=""):
```
takluyver: We shouldn't need this function now that it's in pandoc.

nbconvert/exporters/templateexporter.py
michaelpacer: Should be addressed

nbconvert/filters/filter_links.py
```diff
@@ -0,0 +1,26 @@
...
+import re
+
+from pandocfilters import RawInline, applyJSONFilters
+
+def resolve_references(source):
+    return applyJSONFilters([resolve_one_reference], source)
+
+def resolve_one_reference(key, val, fmt, meta):
+    """
+    """
```
takluyver: Let's either put a docstring in or remove the empty docstring (I'm happy for this to have no docstring).

nbconvert/exporters/templateexporter.py
michaelpacer: hmmm, I don't think that this idiom works exactly as is (also I think this is not quite how bookbook does it) I'll work on it.

nbconvert/exporters/latex.py
```diff
@@ -43,6 +43,10 @@ def _template_skeleton_path_default(self):
     template_extension = Unicode(".tplx").tag(config=True)
 
     output_mimetype = 'text/latex'
+    
+    def default_filters(self):
+        yield from super(TemplateExporter,self).default_filters()
```
takluyver: We should be passing in the *current* class here, i.e. `LatexExporter`, not `TemplateExporter`.

Python 3 improved on this, so you can just call `super().default_filters()` as we do in bookbook. But nbconvert still has to support Python 2 for now.

michaelpacer: For some reason even when running in my Python 3 environment, that doesn't work. with the current code. 

However, the Python 2 compatible way of doing of it does work. Any idea why?

michaelpacer:
I do not like that this causes a ridiculous slowdown in the tests. I imagine this will negatively impact anyone who is converting a large number of notebooks. 

I'm ok merging this but we should make it a priority then figuring out if we can do some kind of conditional filtering (something like adding a piece of metadata to a cell if it has a link in it (or even a specifically internal link in it) as part of a postprocessing step when saving the notebook. I'm not sure but my impression is that this may be appropriate in the https://github.com/jupyter/nbformat repo? Or if that does no analysis of the content of the notebook file, something as part of the saving mechanism inside the notebook itself.

nbconvert/exporters/latex.py
takluyver: Not sure off the top of my head. If you're interested in working it out, we can try to debug on the Hacktrain tomorrow.

nbconvert/exporters/latex.py
```diff
@@ -44,6 +44,10 @@ def _template_skeleton_path_default(self):
 
     output_mimetype = 'text/latex'
 
+    def default_filters(self):
+        yield from super(LatexExporter, self).default_filters()
```
takluyver: Sorry, I forgot when I gave the example that `yield from` is Python 3 only. To make this work on Python 2, we'll need to expand it:

```python
for f in super(LatexExporter, self).default_filters():
    yield f
```

takluyver:
Well, we knew it was going to be slower, right? We can look at ways to mitigate that, like maybe offering an option to skip the link processing entirely.

I don't like the idea of doing stuff on save to support this:

- It's extra complexity and extra slowness on every save for something which is only relevant for converting notebooks to latex. Most notebooks will never be converted to latex.
- It requires a change in every tool writing notebooks (e.g. nteract, Pycharm), and they would have to inspect Markdown, which is currently unnecessary.

PR #467, opened 2016-11-14
Create functional epub exporter

michaelpacer:
This epub exporter has a few problems:

 - [ ] cross links break (problem with pandoc's normalising of internal references when converting to epub  without normalising linked headers)
 - [ ] images alt text is printed directly
 - [ ] images that are relatively linked are not included (this problem is shared with the pdf exporter, which also works inside a temporary directory)

I think the first one should probably be fixed upstream, the second should be dealt with via a filter to just remove alt text for images, and the last should become its own issue for pdf and epub export.


michaelpacer:
Removing the alt text from the images is sufficient for us not *adding* anything to the links that we create, but it won't prevent any alt text for images defined elsewhere.

takluyver:
Are you hoping to put this in 5.0, or a later 5.x release?

michaelpacer:
Well…if we're going to release a second beta (which I think we should if we're going to be changing the id mechanisms), this might as well go in with that.

At least, that's my inclination.

michaelpacer:
@Carreau mentioned that this solution to the image alt text was not going to work since it creates invalid image tags with no alt text.

michaelpacer:
Ok, more precisely, [here's a resource](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img) that states more directly what the image alt text means for the image tag as a whole. 

The image tag _can_ have no alt text, and it _can_ have an empty string, however those mean different things. 

An image tag with no alt text is stating that there is no equivalent to the image and that the image is a key part of the content. 

An image tag with alt text states that the image is not a key part of the content and can be skipped over by non-visual browsers. 

Either of these may produce accessibility difficulties. At the same time, the epub rendering where it always places the alt text immediately after the image seems to be the wrong approach. There must be another way.

michaelpacer:
It looks like there's no way to encode alt text free images in markdown. If you just include nothing, then it's an empty string. and so pandoc will accept it.

Regardless, an empty string and no value for alt seem like good options. Now we need to decide on which.

Also, this default might make for an interesting potential feature based on a better version of what the markdown to epub pandoc converter was doing; making the default typographic meaning of an alt text in published form (at least PDF, html, & epub).

mpacer:
Talking with @takluyver, we should move this into its repo & project, and use an external exporter at first, so that progress can be made and bugs can be reported.

PR #469, opened 2016-11-15
refactor add_anchor to isolate reuable id generation

michaelpacer:
This isn't quite ready, as it breaks. But I wanted to put it up here to get comments on whether more should be put in here. E.g., should we do anything else that would make our mechanism more similar to the pandoc [autolinking mechanism](http://pandoc.org/MANUAL.html#header-identifiers)?

Not sure, but right now it's erring because the wrong things are being passed around.

### DIFF
--- a/nbconvert/filters/strings.py
+++ b/nbconvert/filters/strings.py
@@ -77,12 +77,14 @@ def html2text(element):
     return text
 
 
-def convert_header_id(header_contents):
-    """Convert header contents to valid id value. Takes html as input, returns string.
+def _convert_header_id(header_contents):
+    """Convert header contents to valid id value. Takes string as input, returns string.
+    
+    Note: this may be subject to change in the case of changes to how we wish to generate ids.
 
     For use on markdown headings.
     """
-    return html2text(header_contents).replace(' ', '-')
+    return header_contents.replace(' ', '-')
 
 def add_anchor(html):
     """Add an id and an anchor-link to an html header
@@ -94,7 +96,7 @@ def add_anchor(html):
     except Exception:
         # failed to parse, just return it unmodified
         return html
-    link = convert_header_id(h)
+    link = _convert_header_id(html2text(h))
     h.set('id', link)
     a = ElementTree.Element("a", {"class" : "anchor-link", "href" : "#" + link})
     a.text = u'¶'
--- a/nbconvert/filters/strings.py
+++ b/nbconvert/filters/strings.py
@@ -77,17 +77,12 @@ def html2text(element):
     return text
 
 
-def convert_header_id(html):
+def convert_header_id(header_contents):
     """Convert header contents to valid id value. Takes html as input, returns string.
 
     For use on markdown headings.
     """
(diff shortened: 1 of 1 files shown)


michaelpacer:
Ok, no longer broken. Just needed to not treat it as a potentially raw html element. 

So as it stands it expects to receive a string that has already processed and extracted the contents, and it passes a string back this should make it easier for other methods to interact with it.

takluyver:
Hmm, I hadn't realised quite how trivial this transformation is. I think we're unlikely to change it, because changing it would break internal links in existing notebooks. Maybe we should just do the same `s.replace(' ', '-')` in the couple of places we need it.

While we're thinking about it, it would be good to check that it's working with various punctuation characters as well.

michaelpacer:
I would strongly prefer to have it isolated as its own function.

Having a single function that is invoked everywhere will make it easier to
propagate any changes we might want to make as well as making it explicitly
invoked so people don't need to hunt it down if they are looking for this
feature.

Also as you've already pointed out we've not really tested it against other
kinds of punctuation.

But punctuation coverage can be a separate PR and the set of needed to be
covered cases should be listed in an issue (which I'm happy to help figure
out).

It doesn't introduce any new public API, so that leaves few reasons to not
include this.

I'd like to get this merged ASAP so I can start using it in its current
form elsewhere in a way that is future proof from further changes (such as
better punctuation handling).

michaelpacer:
Also…we should see if we can support something like [pandoc's solution](http://pandoc.org/MANUAL.html#extension-auto_identifiers) to the problem where multiple headers have the same text:

> The exception is when several headers have the same text; in this case, the first will get an identifier as described above; the second will get the same identifier with -1 appended; the third with -2; and so on.

I don't think this can be done at a filter level though, since it would need to operate on the whole document to be able to keep track. But again…once this PR is closed, I'll make a new issue detailing this and linking to the new function. 

Additionally, right now we are creating invalid identifiers meaning that some of this functionality will need to be fixed (e.g., first character must be a letter, and we're not stripping punctuation or numbers), meaning that the transformation will have to be beefed up. 

But this already is an improvement without even including more in the transformation. In fact, I'd suggest that we attempt to make a coördinated change in the notebook (isolating this functionality) so that we can more easily check that https://github.com/jupyter/notebook/blob/bdee765dd32775e85c8460997f95ecea7d910dec/notebook/static/notebook/js/textcell.js#L393 and `convert_header_id` are in sync and will work together. 

To save future searching, here's a link to the commit in the pandoc repo that this was originally introduced with: https://github.com/jgm/pandoc/commit/c435e9cda7eb73cf4d8c446bb829079dddfaedbc?diff=unified#diff-2db348243eee8710717037a8cd14c4d3R1482

michaelpacer:
Here's where the transformation code I think exists in pandoc: https://github.com/jgm/pandoc/blob/eced02d70e0a4e8d7dfa4c373a6c4d8a4cc08407/src/Text/Pandoc/Readers/Markdown.hs#L636

PR #470, opened 2016-11-15
do not suppress kernel errors

minrk:
This can be important info when things go wrong

alternately:

- only allow kernel output on debug
- enable explicit suppression on `--quiet` etc.


takluyver:
Seems reasonable. We can always add more complexity (e.g. making it optional) later.

PR #472, opened 2016-11-18
Change ipython_genutils.tempdir calls to testpath.tempdir and tempfile

michaelpacer:
Elsewhere we speak of removing dependencies on ipython_genutils. I ran a grep and found the cases where I know we were using them needlessly, specifically related to creating temporary directories and replaced them with `testpath.tempdir` if it was a `TemporaryWorkingDirectory` and the standard library `tempfile` if it was a `TemporaryDirectory` since `testpath.tempdir` was just importing that itself anyway.

### DIFF
--- a/nbconvert/exporters/pdf.py
+++ b/nbconvert/exporters/pdf.py
@@ -9,7 +9,7 @@ import sys
 
 from ipython_genutils.py3compat import which, cast_bytes_py2
 from traitlets import Integer, List, Bool, Instance, Unicode
-from ipython_genutils.tempdir import TemporaryWorkingDirectory
+from testpath.tempdir import TemporaryWorkingDirectory
 from .latex import LatexExporter
 
 class LatexFailed(IOError):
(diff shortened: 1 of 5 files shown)


michaelpacer:
oops. `tempfile` only works in this way on python3.2 and above. 

Can't wait for the python3 filled future.

Fortunately, `testpath.tempdir` should have our backs. At least this `testpath.tempdir` only version runs locally on python2 with py.test.

takluyver:
:+1:

`testpath.tempdir` comes from a copy of `IPython.utils.tempdir` before the big split, so it should be a drop-in replacement.

michaelpacer:
Ok, is this ready to be merged? There wasn't anything more I was going to add to this PR.

minrk:
Only one thing missing: moving testpath to a standard requirement in setup.py. It's only a tests requirement at the moment.

michaelpacer:
Ok. That should be covered now.

minrk:
Thanks!

PR #479, opened 2016-12-01
Removed utf8 packages

ManifoldFR:
They became redundant with the switch to XeLaTeX for compiling. I also switched `amsmath` with `mathtools` which automatically loads the former with some fixes.

### DIFF
--- a/nbconvert/templates/latex/base.tplx
+++ b/nbconvert/templates/latex/base.tplx
@@ -13,8 +13,14 @@ This template does not define a docclass, the inheriting class must define this.
     ((* block docclass *))((* endblock docclass *))
     
     ((* block packages *))
-    % Nicer default font (+ math font) than Computer Modern for most use cases
-    \usepackage{mathpazo}
+    \usepackage{polyglossia}
+	\usepackage{fontspec}
+    \usepackage{mathtools} % Equations
+    \usepackage{amssymb} % Equations
+	\usepackage{unicode-math}
+	\setmainfont{XITS}
+	\setmathfont{Latin Modern Math}
+	\setmathfont[range=\mathbb]{xits-math.otf}
 
     % Basic figure setup, for now with no caption control since it's done
     % automatically by Pandoc (which extracts ![](path) syntax from Markdown).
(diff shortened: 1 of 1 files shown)


michaelpacer:
@takluyver Do we have any tests that check for coverage across a decent range of utf-8 chars that are found in the default fonts that we're already including in the travis build? I can go through each test individually, but if you know one that is specifically testing utf-8 compatibility and therefore watching for this kind of change, that'd be great. 

I'd prefer for there to be a test rather than needing to directly verify that removing any particular package will not break existing successful exports.

takluyver:
We try to test with some non-ascii characters to check that it doesn't blow up the first time an ø appears, but I don't think we have anything that really checks a wide range of characters.

michaelpacer:
A really wide range of characters isn't supported anyway right now. I was just thinking something from Latin Extended A since I'm pretty sure those are included in the fonts that we're using (e.g., the long letter s: ſ; I didn't actually check that we are covering that with the current fonts, but I can if needed). 

But perhaps what we should do is add characters to that test that won't be output (because we in many cases don't have the right fonts installed but on which we can be sure anything that can't handle utf-8 will choke).

takluyver:
It looks like the file `Skip Exceptions.ipynb` has some non-ascii characters in it, so we can make tests with that.

nbconvert/templates/latex/base.tplx
```diff
@@ -13,14 +13,14 @@ This template does not define a docclass, the inheriting class must define this.
...
-	\usepackage{polyglossia}
-	\usepackage{fontspec}
+    \usepackage{polyglossia}
+    \usepackage{fontspec}
     \usepackage{mathtools} % Equations
     \usepackage{amssymb} % Equations
-	\usepackage{unicode-math}
-	\setmainfont{XITS}
-	\setmathfont{Latin Modern Math}
-	\setmathfont[range=\mathbb]{xits-math.otf}
```
ManifoldFR: Somehow, the 4-spaces got converted into double-tabs, hence an error when travis checked.

Issue #481, opened 2016-12-05
Custom mime type


SylvainCorlay:
When converting to html Would it be possible to convert datatypes in the form of `application/****+json`
to `script` tags if the same mime type containing the corresponding json?

For example, I guess

```json
   "outputs": [
    {
     "data": {
      "application/vnd.jupyter.widget-view+json": {
       "model_id": "11a0e82dc6cf47b3aa0cf241705fd85b"
      }
     },
     "metadata": {},
     "output_type": "display_data"
    }
```

would become

```html
<div class="output">
<div class="output_area">
    <div class="prompt"></div>
    <script type="application/vnd.jupyter.widget-view+json">
    {
       "model_id": "11a0e82dc6cf47b3aa0cf241705fd85b"
    }
    </script>
</div>
</div>
```

SylvainCorlay:
So I guess

```json
   "outputs": [
    {
     "data": {
      "application/vnd.jupyter.widget-view+json": {
       "model_id": "11a0e82dc6cf47b3aa0cf241705fd85b"
      }
     },
     "metadata": {},
     "output_type": "display_data"
    }
```

would become

```html
<div class="output">
<div class="output_area">
    <div class="prompt"></div>
    <script type="application/vnd.jupyter.widget-view+json">
    {
       "model_id": "11a0e82dc6cf47b3aa0cf241705fd85b"
    }
    </script>
</div>
</div>
```

SylvainCorlay:
^ ping @michaelpacer

michaelpacer:
My intuition is that we would need to support that in the [display priority skeleton template](https://github.com/jupyter/nbconvert/blob/560a1fc/nbconvert/templates/skeleton/display_priority.tpl) and (possibly) the [filter_data_type filter](https://github.com/jupyter/nbconvert/blob/1c8a514901557335cf24b70fb427d461d32e6243/nbconvert/filters/datatypefilter.py) since that seems to be how we're sorting through the mimetypes in output right now.

The current approach is based on a bunch of manually encoded mimetypes which I believe is due to the static nature of the current Jinja2 based template architecture. My guess is that to support this we would need to be able to programmatically generate templates in memory (which is something that @carreau and @takluyver and I have talked about a bit). Doing that would allow a much more than just custom mimetypes (e.g., better foreign language support in PDF and LaTeX since we could inject the appropriate commands into the preamble). But it probably be one of the biggest overhauls to the current architecture that we've handled in nbconvert in a while. 

[...]

PR #482, opened 2016-12-06
New mimetypes

michaelpacer:
Responding to #481 for widget related mimetypes. 

Does not add them to the rest of the templates. 

Markdown needs particular consideration as it is the basis of the epub exporter in #467 (not our html exporter). Incorporating widgets for epubs may require splitting the markdown exporter utility from the markdown-to-epub exporter functionality (possibly only requiring a slightly different template format).


nbconvert/templates/html/basic.tpl
```diff
@@ -209,3 +209,31 @@ var element = $('#{{ div_id }}');
 </script>
 </div>
 {%- endblock -%}
+
+{%- block data_widget_state scoped %}
+{% set div_id = uuid4() %}
+{% set content = output.data['application/vnd.jupyter.widget-state+json'] %}
+<div id="{{ div_id }}"></div>
+<div class="output_subarea output_widget_state {{extra_class}}">
+<script type="tapplication/vnd.jupyter.widget-state+json">
```
SylvainCorlay: typo `tapplication -> application`

nbconvert/templates/html/basic.tpl
```diff
@@ -209,3 +209,31 @@ var element = $('#{{ div_id }}');
...
+</script>
+</div>
+{%- endblock data_widget_state -%}
+
+{%- block data_widget_view scoped %}
+{% set div_id = uuid4() %}
+{% set content = output.data['application/vnd.jupyter.widget-view+json'] %}
+<div id="{{ div_id }}"></div>
+<div class="output_subarea output_widget_view {{extra_class}}">
+<script type="tapplication/vnd.jupyter.widget-view+json">
```
SylvainCorlay: same (`tapplication`)

nbconvert/templates/html/basic.tpl
```diff
@@ -209,3 +209,31 @@ var element = $('#{{ div_id }}');
...
+{% set div_id = uuid4() %}
+{% set content = output.data['application/vnd.jupyter.widget-state+json'] %}
+<div id="{{ div_id }}"></div>
+<div class="output_subarea output_widget_state {{extra_class}}">
+<script type="tapplication/vnd.jupyter.widget-state+json">
+var element = $('#{{ div_id }}');
+{
+"model_id": "{{content.model_id}}"
+}
+</script>
```
SylvainCorlay: The content of `application/vnd.jupyter.widget-state+json` is actually not `{ model_id : {{ content.model_id }}}` but has a complet inner structure.

Why not simply dumping the entire content there?

michaelpacer: Oh. I thought you wanted that `model_id` specifically. Well, thank you for making the problem more specific which turned out to be a great way to get me to think about how to solve this problem.

SylvainCorlay: I have an example of notebook with these new mime types here: https://gist.github.com/SylvainCorlay/d6d8cd44e032523d07e102f5dc0bb1df

A bunch of output areas have content of mime type
```
"application/vnd.jupyter.widget-view+json": {
       "model_id": "b59af7d27c594924957e1aa26257751f"
      }
```

SylvainCorlay: Side-note: the json schema for `widget-view` allows for two additional integer fields besides the `model_id`, namely `version_major` and `version_minor` but are not used at the moment, not is it actually added in the current implementation.

SylvainCorlay:
@michaelpacer thanks a lot for tackling this.

michaelpacer:
Happy to. 

Open questions: 

Do you want contents to be wrapped in another set of curly braces or do you want it to be dumped with nothing wrapping it? 

Should I be actually setting the value to some variable so that you can do something with it? 
    If yes, is that a variable with a particular type or (more likely, since it's a naming problem) a name that should be stable enough so that you know how to access it (i.e., API decisions)?

Would it be better to just put that other js in another `<script>` tag?

michaelpacer:
Also, this is starting to seem like a potential security hole. @carreau can you help develop some test cases that try to exploit this to do terrible things if they were to be exposed on a computer as is?

SylvainCorlay:
> Also, this is starting to seem like a potential security hole. @carreau can you help develop some test cases that try to exploit this to do terrible things if they were to be exposed on a computer as is?

I don't think it can be a security issue since the type is not `javascript` and is not executable code...

Up to the front-end to embed the required javascript to render this content, or not...

michaelpacer:
Ok. That is really helpful. It makes me think that you just want me to put the content there bare and that might be easier than trying to access it from a variable. You'll just take the content of the script tag, because it's not really acting as a script tag in the sense of an executable script, but more as a data structure that other executable code can directly access.

SylvainCorlay:
> Also, this is starting to seem like a potential security hole. @carreau can you help develop some test cases that try to exploit this to do terrible things if they were to be exposed on a computer as is?

I guess that your approach for the `view` reading the model_id prevents any form of injection. (so this was fine).

The `application/vnd.*+json` mime type is not executed by the browser. Up to the UI to add additional script tags to perform the rendering.

SylvainCorlay:
> Do you want contents to be wrapped in another set of curly braces or do you want it to be dumped with nothing wrapping it?

So what is needed is to convert

```json
{
    "application/vnd.jupyter.widget-state+json" : FOOBAR
}
```

into

```html
<script, type="application/vnd.jupyter.widget-state+json">
FOOBAR
</script>
```

SylvainCorlay:
I don't understand your last commit. Is it some nbconvert specific thing?

nbconvert/templates/html/basic.tpl
michaelpacer: in response to the bit regarding the notebook metadata, thank you for the link. I'm going to try to apply the current version of nbconvert to that notebook, but am running into issues that have to do with my own system.

About the second, is there anything you would like to have done with the contents of either `version_major` or `version_minor`or is the current "dumping it all in between two curly braces" approach going to cover that as well? My intuition is that it does cover that, but I could be wrong.

michaelpacer:
Its message is a slight misunderstanding of what you wanted, but now it's being fixed.

michaelpacer:
{
    "application/vnd.jupyter.widget-state+json" : FOOBAR
}

is not the same as 

{
    "application/vnd.jupyter.widget-state+json" : {
           FOOBAR
     }
}

But the current approach would return your second, and not your first because I think it gets the contents as itself being a dict (rather than a string with this particular set of values), which I do not believe will include the brackets, so I need to add them manually for it to return what you want in the former case.

michaelpacer:
Just tried running nbconvert on that file and it doesn't recognise it as a valid notebook file? Does it violate the stable release version of nbformat?

SylvainCorlay:
> So what is needed is to convert
>
> {
>    "application/vnd.jupyter.widget-state+json" : FOOBAR
> }
> into
> 
> <script, type="application/vnd.jupyter.widget-state+json">
> FOOBAR
> </script>

So I stand by this. FOOBAR happens to be an object.

SylvainCorlay:
> Just tried running nbconvert on that file and it doesn't recognise it as a valid notebook file? Does it violate the stable release version of nbformat?

We require nbformat 4.2.

michaelpacer:
The issue is, if I don't know that data type and it is interpreted directly, I can't add formatting to it to make it valid… for example, I'm pretty sure a string would not be given the appropriate `"…"` if it were included either. It's not getting it as a json object per se but interpreting it using python's dictionary structure because all Jinja code should be interpreted as python. Anything inside `{{}}` is python code that is executed and returned.

michaelpacer:
What nbconvert is doing shouldn't be thought of as string processing but rather data interpretation and structure regeneration. 

I could attempt something with string processing, but that would be inconsistent with how we handle other mimetypes.

michaelpacer:
sorry for a stupid question; but what's the fastest way of checking nbformat's version number without listing everything using `pip list`? 

`jupyter-nbformat --version` doesn't work and there was pushback when I suggested creating it.

michaelpacer:
[NbConvertApp] Converting notebook nbconvert/tests/files/Widget_List.ipynb to html
[NbConvertApp] ERROR | Notebook JSON is invalid: {'model_id': 'b59af7d27c594924957e1aa26257751f'} is not valid under any of the given schemas

:/ I think i might need a newer version of `ipywidgets`.

SylvainCorlay:
> [NbConvertApp] Converting notebook nbconvert/tests/files/Widget_List.ipynb to html
[NbConvertApp] ERROR | Notebook JSON is invalid: {'model_id': 'b59af7d27c594924957e1aa26257751f'} is not valid under any of the given schemas
> 
> :/ I think i might need a newer version of ipywidgets

Yes, ipywidgets master requires nbformat 4.2 that was just released.

michaelpacer:
Any idea why it overwrite my editable install? I opened an issue: https://github.com/ipython/ipywidgets/issues/951 but I figured I'd ask.

michaelpacer:
```
[NbConvertApp] ERROR | Notebook JSON is invalid: {'model_id': 'b59af7d27c594924957e1aa26257751f'} is not valid under any of the given schemas

Failed validating 'oneOf' in display_data['properties']['data']['additionalProperties']:

On instance['cells'][8]['outputs'][0]['data']['application/vnd.jupyter.widget-view+json']:
{'model_id': 'b59af7d27c594924957e1aa26257751f'}
```

Can't figure out what script to tell it to highlight for that, but that's the error I'm getting with an up to date nbformat and ipywidgets.

SylvainCorlay:
> Any idea why it overwrite my editable install? I opened an issue: ipython/ipywidgets#951 but I figured I'd ask.

That is before 4.2 is more recent that 4.2dev. Even if you pulled from upstream in your editable install after the release, you need to re-run `pip install -e` on your editable install for pip to see that the currently installed version is more recent than 4.2

michaelpacer:
I did that (ran `pip install -e .` twice for nbformat after rebasing on upstream/master), and it's still overwriting it :/

SylvainCorlay:
Then you should see `4.3.0dev`...

michaelpacer:
Hrm. Foiled by my own inconsistent system for my failure to keep a consistent system for upstream vs. origin. Sorry about that.

SylvainCorlay:
> Hrm. Foiled by my own inconsistent system for my failure to keep a consistent system for upstream vs. origin. Sorry about that.

yeah, happens sometimes...

michaelpacer:
Weird: the error i'm getting strange:

```
[NbConvertApp] Converting notebook nbconvert/tests/files/Widget_List.ipynb to html
/Users/cocosci/Jupyter/nbconvert/nbconvert/filters/datatypefilter.py:41: UserWarning: Your element with mimetype(s) dict_keys(['application/vnd.jupyter.widget-view+json']) is not able to be represented.
    mimetypes=output.keys())
```

It looks like the warning is printing some of it's own signature…which must be wrong. But I also am am a little confused the dict_keys() object being there.

SylvainCorlay:
What I did in the context of nbsphinx is: `https://github.com/spatialaudio/nbsphinx/pull/75/files`

SylvainCorlay:
(wondering in which timezone you are...)

michaelpacer:
PST trouble sleeping and wanted to address this.

michaelpacer:
Although now that you've drawn my attention to it, I just yawned.

michaelpacer:
in terms of `nbsphinx`, I'm not sure if it's the use of macros but it looks like there are a number of clever things to be drawn from there. But that might be bigger than this. 

Regardless, I think I might be able to do something like that, but I get the feeling that nbconvert may end up looking a little clunkier.

I'll look at this with a fresh mind tomorrow; sorry to disappoint for continued progress tonight

michaelpacer:
Actually, the nbsphinx thing was easy enough to mirror so long as I wasn't passing down the data in exactly the same way.

SylvainCorlay:
I think that this looks good. I will discuss this with @minrk after our current meeting is finished.

michaelpacer:
@SylvainCorlay Do you want me to further mirror the nbsphinx solutions that you're describing in https://github.com/spatialaudio/nbsphinx/pull/76? 

Also, it wouldn't be hard to add this to the rst exporter using the same mechanisms in the nbsphinx exporter that way if someone was using nbconvert instead of nbsphinx as part of an rst workflow, they would not get divergent behaviour.

SylvainCorlay:
> @SylvainCorlay Do you want me to further mirror the nbsphinx solutions that you're describing in spatialaudio/nbsphinx#76?
> 
> Also, it wouldn't be hard to add this to the rst exporter using the same mechanisms in the nbsphinx exporter that way if someone was using nbconvert instead of nbsphinx as part of an rst workflow, they would not get divergent behaviour.

I am not sure that we should add widgets to the default rst exporter if it is not meant to be used in sphinx specifically where we have such additional directives.

IMO The place where it is the most interesting here is the html exporter for the nbviewer usecase, which would be a huge improvement over the current situation.

michaelpacer:
Ok sounds good.

SylvainCorlay:
@michaelpacer we got widgets to render on RTD:

http://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html#Tabs

The last missing piece in your PR is the widget manager state which is (optionally) saved in notebook metadata.

We used the `footer` section to add the relevant html tag.

https://github.com/spatialaudio/nbsphinx/blob/master/nbsphinx.py#L245

nbconvert/templates/html/basic.tpl
```diff
@@ -209,3 +209,27 @@ var element = $('#{{ div_id }}');
...
+
+{%- block data_widget_view scoped %}
+{% set div_id = uuid4() %}
+{% set datatype = output.data|filter_data_type %} 
+<div id="{{ div_id }}"></div>
+<div class="output_subarea output_widget_view {{extra_class}}">
+<script type="text/javascript">
+var element = $('#{{ div_id }}');
+</script>
+<script type="{{datatype}}">{{output.data[datatype]}}</script>
```
SylvainCorlay: I think that this need to be json-serialized, see the change in https://github.com/spatialaudio/nbsphinx/pull/76

SylvainCorlay:
Can't wait for this to work. This is a major milestone for widgets and the oldest open issue.

michaelpacer:
So I can check that I understand, we want `application/vnd.jupyter.widget-state+json` to be both included from outputs and grabbed (separately) from the notebook metadata?

Is there a reason you put it in a footer rather than a header?

nbconvert/templates/html/basic.tpl
michaelpacer: Ok, that makes a lot of sense. That was what I was worried about when discussing its being interpreted not as a true json object, but rather a json object as interpreted by python, which wouldn't necessarily be included in the same way.

michaelpacer:
If we do want both of them to be present…and they conflict do we care which one takes priority?

michaelpacer:
hmm… this continues to be broken even after I fix the call naming scheme

the error i'm getting is odd, and I'm not entirely sure how to figure out what's going wrong because the object is buried so deep in the call stack that directly accessing it is nigh impossible as far as I can tell

but here's the traceback until we get to the part where it's nbconvert: 

```
  File "~/Jupyter/nbconvert/nbconvert/exporters/../templates/html/basic.tpl", line 236, in block "data_widget_view"
    {{ output.data[datatype] | json_dumps }}
  File "~/anaconda3/envs/dev/lib/python3.5/json/__init__.py", line 230, in dumps
    return _default_encoder.encode(obj)
  File "~/anaconda3/envs/dev/lib/python3.5/json/encoder.py", line 198, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "~/anaconda3/envs/dev/lib/python3.5/json/encoder.py", line 256, in iterencode
    return _iterencode(o, 0)
  File "~/anaconda3/envs/dev/lib/python3.5/json/encoder.py", line 179, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: Undefined is not JSON serializable
```

michaelpacer:
When I run this locally, I'm not seeing interactive widgets, I have fetch and merged the latest version of ipywidgets so I'm not sure why.

Also, is there an easy way to make it so that the json that is spat out by the metadata call isn't just one line with no line breaks? I.e., is there a pretty print version of it? I know that in many cases its worth minifying it. However, when debugging, this makes it difficult to figure out what's going on when Chrome cuts off the string in the web dev view with an ellipsis (`…`). Currently this line is causing vim to halt for 5 secs before being able to render it.

michaelpacer:
ok, so just adding json_dumps(indent="") works enough to add new lines, but chrome is still cutting off the string early in the dev view, but now I can view it without difficulty in vim.

But could you check if it's working for you? Everything seems like it should be fixed now.

SylvainCorlay:
> So I can check that I understand, we want application/vnd.jupyter.widget-state+json to be both included from outputs and grabbed (separately) from the notebook metadata?
> 
> Is there a reason you put it in a footer rather than a header?

For now, this mime type will only be met in metadata, but we discussed the possibility for widgets to put the widget state in some output (and also discussed the possibility to add global notebook-level output to the notebook format).

There is no particular reason to put this in the footer other than the rest of the page may show before this is loaded which makes the user experience faster.

michaelpacer:
> Nbviewer needs to add a script tag for the widget embedder for this to render.

I wasn't using nbviewer, I was just working with the html export and a local server. Should I just embed that script tag in the html template as well?

Where do you define that script tag for the widget embedder?

SylvainCorlay:
> Where do you define that script tag for the widget embedder?

`https://unpkg.com/jupyter-js-widgets@^2.0.16/dist/embed.js`

michaelpacer:
Where is that being included in the template by nbsphinx? a search isn't coming up with anything?

SylvainCorlay:
> Where is that being included in the template by nbsphinx? a search isn't coming up with anything?

It is not. We will add it in nbsphinx when we have a stable release of `ipywidgets 6.0`. The jupyter-js-widgets `2.0.x` npm packages are tagged with `next` so that they don't come up by default.

The only place where it is currently included is in the `jupyter_sphinx.embed_widgets` extension which is another means of embedding widgets in the documentation...

michaelpacer:
> It is not. 

Ok, I feel more sane now :). 

> We will add it in nbsphinx when we have a stable release of `ipywidgets 6.0`. The jupyter-js-widgets `2.0.x` npm packages are tagged with `next` so that they don't come up by default.

That makes sense, but then I don't understand how rtd is rendering live widgets. Also I don't know much about npm packaging, so I'm not sure what `next` means beyond your description or what the equivalent of `--pre` for pip would be if I wanted to get this working locally. 

> The only place where it is currently included is in the `jupyter_sphinx.embed_widgets` extension which is another means of embedding widgets in the documentation...

Is that what you're currently using? I saw that you were including that in your requirements.txt but I wasn't sure what jupyter-sphinx beyond the discussion in the ipywidgets docs as an alternative solution to embedding (but that seemed at cross purposes with what we were trying to accomplish here).

So, if it doesn't matter that I can't see these locally, it isn't yet stable enough to release, does that mean that (for now) this PR covers all that we need?

SylvainCorlay:
> That makes sense, but then I don't understand how rtd is rendering live widgets. Also I don't know much about npm packaging, so I'm not sure what next means beyond your description or what the equivalent of --pre for pip would be if I wanted to get this working locally.

For the RTD of ipywidgets, I am using both `jupyter_sphinx` and `nbsphinx`. One of them at least must include the tag...

npm tags are an hugly hack to come around an issue with pre-releases in npm. For example typescript "2.0.0" was in fact 2.1. All the 2.0.x where tagged "next".

> So, if it doesn't matter that I can't see these locally, & it isn't yet stable enough to release meaning it shouldn't be included in the base template, does that mean that (for now) this PR covers all that we need?

I think that this covers what we need for now. I don't know if the javascript tag should be in nbviewer or the template.

SylvainCorlay:
@michaelpacer @takluyver I tested this locally and it works beautifully. 

🎉 ✨ 👍 💯 😄 🐴 🐰

SylvainCorlay:
@takluyver 

merge

minrk:
Sorry I didn't reply earlier. I see that you pushed out 4.3 with just this change. Great! But I also see that 5.0 has had zero issues open for some time, and has put out a beta two months ago. @michaelpacer do you want to cut 5.0 now? Or is there something else you'd like to finish first?

michaelpacer:
The header id thing has been a project since the dev meeting, as it is a
breaking change and 5.0 is a major version.

My hope was to get a second beta out there of 5.0, since this it is more
breaking of a change than XeLaTeX.

But, you're right; it might make sense to just release 5.0, as is, and
start a beta release for 6.0 when're the header ids are fixed. Even if that
means we release 6.0b less than a month after 5.0, with few minor versions.
Having a long beta to test something like this is a good idea since it'll
give more chances for people to test there notebooks.

@carreau we should look into a bot that people can enable for their repos
that will test whether people's notebooks will break with an upcoming
release and create an issue describing the breakage. I know it's
non-trivial from an analytics point of view and the bot point of view, but
it also seems like a nice instance of a more generally useful pattern when
changing things like a template, schema or format. Proactive warnings &
compatibility tests any time you can statically analyze a document in this
way.

E.g., anyone with a header like:

    # header content {fancier name}

We can know ahead of time is going to experience breakage in their cross
references to that header in some output formats with 6.0b.

Carreau:
yeah, I [have created an organisation](https://github.com/jupyterdownstream) where I'm planning to enable travis cron weekly and a repository where we allow anyone to add a test to see if we break a downstream project. It on my todo list.

minrk:
Since the headers are a new thing, I think they should go in after we release 5.0, which has been awaiting release for a while. Do you want to do a release candidate with the current state, or make 5.0 final? Most big changes should spend a while in master, so they have a chance to undergo some revisions. Merging a new change like that would delay 5.0 quite a bit, and I think it's overdue as it is.

michaelpacer:
Ok that sounds reasonable… I can release later today if that's what people
would prefer; I'm not sure if that's too soon after 4.3 is all.

takluyver:
I don't think there's any particular problem with doing one release soon after another. Though I'm a bit surprised we did a 4.3 when 5.0 is already in beta... *checks commit history* Aha, @SylvainCorlay, have you been applying the hotel whiteboard theory again to get features you want in rapid releasses? ;-)

In general, we don't do long betas - if things look OK after a week or two, we make a final release. It's a bit tough making releases, because there's always something else you want to get in. But if we make releases fairly frequently, then your favourite change missing one release is not so distressing.

SylvainCorlay:
> Aha, @SylvainCorlay, have you been applying the hotel whiteboard theory again to get features you want in rapid releases? ;-)

i have no recollection of the event in question.

michaelpacer:
Hrm… I'll have to learn what this hotel whiteboard theory is at the next in-person dev meeting :smile:

Madhu94:
@mpacer @SylvainCorlay Is this meant to work for any custom mime-type or just the widget views mime-type?

PR #487, opened 2016-12-13
Fixing the widget calls, conditionally embedding the script tag

michaelpacer:
This makes widgets actually work in nbconvert, this closes https://github.com/jupyter/nbviewer/pull/650 for now. And finishes up closing #481. 

Also, made it so that it wouldn't err if a notebook is passed in that does not have the widgets metadata.

This shouldn't break anything and should be a high priority merge to fix a potential break in #482.

### DIFF
--- a/nbconvert/templates/html/basic.tpl
+++ b/nbconvert/templates/html/basic.tpl
@@ -242,7 +242,7 @@ var element = $('#{{ div_id }}');
 
 {%- block footer %}
 {% set mimetype = 'application/vnd.jupyter.widget-state+json'%} 
-{% if mimetype in nb.metadata.widgets %}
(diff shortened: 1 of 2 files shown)


michaelpacer:
@SylvainCorlay This should solve the nbviewer problem (if nbviewer were to work off of nbconvert's master… which I don't think it does. 

Btw, regarding releases, we're in a weird spot. There are breaking changes that I'm working on that I wanted to include in the 5.0b. But I'm starting to think that we should just release 5.0 and I'll try to aim at a really quick turn before getting to 6.0.

SylvainCorlay:
@michaelpacer thanks! And thank you for the release!

Issue #484, opened 2016-12-09
Docs haven't built for a month


michaelpacer:
Somehow, the docs are first installing master, and then they are overwriting that with stable:

```
Installing collected packages: nbconvert
  Found existing installation: nbconvert 5.0.0.dev0
    Uninstalling nbconvert-5.0.0.dev0:
      Successfully uninstalled nbconvert-5.0.0.dev0
  Running setup.py develop for nbconvert
Successfully installed nbconvert-4.2.0
```

Not sure what changed.

https://readthedocs.org/projects/nbconvert/builds/4618074/

michaelpacer:
Sphinx did a beta release on nov 6, which was when the failure started. 

I don't think it's related, but it might be.

michaelpacer:
@willingc Do you know of anything that's been happening in this space that might change how our installation commands are read?

willingc:
@michaelpacer Hmm... I'll be up your way next week. I'll take a look this weekend and we can follow up next week. I didn't see anything in my quick scan about why that might be the case.

I've gone ahead and made the stable release the default in RTD and opened a PR to add the stable badge to the README #485.

michaelpacer:
Is it possible to similarly include a badge for the latest branch as well? It would be useful to know that any PRs are not breaking RTD in terms of the current version of master. 

I feel like we had a conversation about actually testing for RTD breakage as part of travis-CI and it was proving difficult, so a badge to even get a rapid visual check of this would probably be useful. 

Also, I just noticed that strangely it was getting nbconvert from conda not from pypi… maybe something changed in terms of how rtd handles conda envs?

michaelpacer:
Command that breaks is 

```bash
python /home/docs/checkouts/readthedocs.org/user_builds/nbconvert/conda/latest/bin/sphinx-build -T -E -b readthedocs -d _build/doctrees-readthedocs -D language=en . _build/html
```

Which isn't an install command…but a build command. Why is it installing something on a build at all?

michaelpacer:
Aha…

From the sphinx docs for the command `sphinx-build` 

> -E
> Don’t use a saved environment (the structure caching all cross-references), but rebuild it completely. The default is to only read and parse source files that are new or have changed since the last run.

I wonder if that's the cause…

willingc:
> Also, I just noticed that strangely it was getting nbconvert from conda not from pypi… maybe something changed in terms of how rtd handles conda envs?

[...]

PR #488, opened 2016-12-15
Fix doc build in conf.py

michaelpacer:


michaelpacer:
@Carreau This branch now builds docs correctly, see here: https://readthedocs.org/projects/nbconvert/builds/4792479/

basically there were a couple of holdover bugs that (before releasing) I should have made a pr for from https://github.com/michaelpacer/nbconvert/tree/unbreak_docs (but after squashing, cause those were a lot of attempts to get things to work on rtd that mostly ended up not mattering). 

Upside: because of fighting with rtd all of yesterday, I knew exactly which two things needed to change to get the docs to build on master & to close #484 (at least provisionally).

docs/environment.yml
```diff
@@ -7,7 +7,7 @@ dependencies:
 - nbformat
 - jupyter_client
 - ipython
-- sphinx>=1.3.6
+- sphinx>=1.3.6,!=1.5.0,!=1.5.1
```
takluyver: Are we confident that 1.5.2 will work?

willingc: @takluyver I would say hopeful of 1.5.2 but not 100% confidence. @michaelpacer You could make this 1.3.6 <= sphinx < 1.5.0 if you wish. Or leave as is. I don't have a strong preference either way.

willingc:
Closes #484

docs/environment.yml
takluyver: OK. Let's remember to keep an eye on it when Sphinx 1.5.2 comes out to see if it breaks again.

I've just spotted that RTD has email notifications for failures. I'll sign myself up for that,

michaelpacer: https://github.com/spatialaudio/nbsphinx/pull/85 should also take care of this, it works on both 1.5.0 and 1.5.1 (so long as we have docutils < 1.12 which we will inherit from nbsphinx (we could make this an explicit dependency if we want to)).

But also, the change is already merged on stable for sphinx, so I expect that it will be in 1.5.2. The sphinx maintainers just forgot to port the PR from master to stable before releasing. That's why I was confident that it would be fixed, because it already was fixed, just not merged in the right place at the time of release.

Issue #490, opened 2016-12-16
`extra_loaders` broken in new release?


matthew-brett:
Hi,

As of today, my project tests started to fail because my code for using an in-memory jinja template for nbconvert is no longer working.  Here's a test snippet that worked yesterday, but not today (with version 5.0):

```
from jinja2 import DictLoader

try:
    from nbformat import v4
    import nbconvert
except ImportError:
    from IPython.nbformat import v4
    from IPython import nbconvert


my_loader = DictLoader({'my_template': "{%- extends 'rst.tpl' -%}"})


class MyExporter(nbconvert.RSTExporter):
    template_file = 'my_template'


exporter = MyExporter(extra_loaders=[my_loader])

nb = v4.new_notebook()

out, resources = exporter.from_notebook_node(nb)
```

In nbconvert 5.0, this gives `TemplateNotFound: my_template.tpl`.   I believe you now assume that the template is a filename, and do not allow specifying the template via jinja2 loaders.   This is inconvenient for me, because I want to build my template in memory.

takluyver:
It looks like we assume that it should have an extension, but I think we still respect extra_loaders. Does it work if you use `my_template.tpl` as the key in the dict? Alternatively, I think you could set `template_extension = ''`.

matthew-brett:
Ah yes, you're right, I'm sorry for my lazy diagnostics.   Maybe it would be worth relaxing the requirement for a `tpl` extension again, on the basis that the template need not be a file?

takluyver:
Have a look at #491 :-)

michaelpacer:
If #491 works for your case @matthew-brett, then I'll merge it as is.

matthew-brett:
Too ungrateful to suggest adding a test, perhaps from my original report?

michaelpacer:
Not hard to do if all you care is that it doesn't _break_ to use an memory template (thank you for giving such a nice MWE): e696578 takes care of that (fails on master, not on #491 for the right reasons). 

But it doesn't actually check that the content is appropriately converted. It happens to in this case, but it seems like it'd be out of scope for this test (and there should be a separate discussion about which content should be tested).

willingc:
@michaelpacer Check out the recent doc builds on nbconvert and notebook master. Ideas?

michaelpacer:
this push should fix it https://github.com/spatialaudio/nbsphinx/pull/94

michaelpacer:
[...]

PR #491, opened 2016-12-16
Better way to add optional extension to specified template file

takluyver:
Using a custom Jinja loader which wraps the FileSystemLoader. This means we only add the file extension for the default loader, and assume anyone who is providing extra loaders knows what they're doing.

Closes gh-490


michaelpacer:
Will this solution break on files ending with `.tplx`?

takluyver:
I don't think so. Nothing I did was specific to any particular ending.

nbconvert/exporters/templateexporter.py
```diff
@@ -54,6 +56,29 @@
...
+    """A template loader which optionally adds a given extension when searching.
+
+    Constructor takes two arguments: *loader* is another Jinja loader instance
+    to wrap. *extension* is the extension, which will be added to the template
+    name if finding the template without it fails. This should include the dot,
+    e.g. '.tpl'.
+    """
+    def __init__(self, loader, extension):
+        self.loader = loader
+        self.extension = extension
```
michaelpacer: I think if you change this to 

     def __init__(self, loader, extension_list):
          self.loader = loader
          self.extension_list = extension_list 

then you could accommodate (below) something that is similar to:

    if any([template.endswith(ext) for ext in self.extension_list]):
        raise
    return self.loader.get_source(environment, template+self.extension_list[0])

which would default to returning `.tpl`

and then later on you'd need to do something like:

    template_extension_list = List(['.tpl','tplx']).tag(config=True, affects_environment=True)

michaelpacer:
under template_extension it expects `.tpl` as the extension rather than as either `.tpl` or `.tplx`. 

My guess is the reason why this works with only one string is that we just overwrite the Unicode traitlet for LaTeX & pdf export to be `.tplx`. 

Right now we have no instance where we might use some templates with either extension, but is that necessarily the case?

michaelpacer:
Or alternatively, could we change it so that nbconvert didn't expect templates to have any particular extension? I'm trying to figure out where the extension is actually being _used_ to see if this is possible, and is it just in the code where we're finding and loading the templates that come with nbconvert by default?

It seems there should be an elegant way of handling in memory and on disk templates that doesn't require pretending like in memory templates were passed in with a file extension.

michaelpacer:
Realized we should have someone else look over the code than @takluyver and I. 

@Carreau @minrk should this solution for in memory templates be ok?

takluyver:
The file extension for Latex templates is different because we use a different template syntax there, to avoid clashing with Latex syntax. So those template files are effectively a different format, and you can't mix and match the two kinds. So I don't think it's necessary for a single loader to support multiple extensions.

If I was writing this code from scratch, I wouldn't bother automatically adding extensions; I think it's clearer what's going on if users have to supply the extension themselves. But since it already works without users providing the extension, I want to keep it working.

> It seems there should be an elegant way of handling in memory and on disk templates that doesn't require pretending like in memory templates were passed in with a file extension.

That's what this does. We only wrap the `FileSystemLoader` in this machinery to add extensions, so if you supply an in-memory template as in #490, the code does not deal with extensions at all.

michaelpacer:
@takluyver Ok, that makes sense. I misunderstood how this was working. 

You said it ignores the extension, but it looks like it doesn't ignore the extension, but rather it adds if it's not present in https://github.com/jupyter/nbconvert/pull/491/files#diff-43c2e238630461c0a2c422a0b09fae55R77. But I think that's only going to happen if you use the FileLoader. So this also has a feature improvement (since people can specify templates without their extensions now), right?

The reason why none of that machinery is extension sensitive if you use another loader is because those loaders are just prepended onto the list and thus are operating independently of this change, right? 

Do we have separate machinery for interpreting `.tplx` templates then? It seems to be mostly the same but the only difference is that the LatexExporter is specifying a different template extension in https://github.com/jupyter/nbconvert/blob/master/nbconvert/exporters/latex.py#L43. Is that why it counts as a different format?

Since `.tplx` are effectively a different format, I added a test specifically for `.tplx` templates.

nbconvert/exporters/templateexporter.py
michaelpacer: Ok, I see why now that this is unnecessary. It's handled via the LaTeXExporter, which means that this mechanism will adapt to whichever value to which we set that traitlet.

michaelpacer:
I said I was going to be able to send the patch release tomorrow when this gets in and it only needs a few more eyes. 

I just broke everything with the last push, and I will be confident that once I can get this expected failure to actually fail, I will understand this well enough to merge it. That may take more than tomorrow as I thought I'd get further than I did tonight. Still made progress though, but I'm going to edit the title to include [WIP].

michaelpacer:
Ok @takluyver so the place where I'm stuck is getting to directly pass the loader that is to be used so that it errs in the correct place. 

As you can see in the most recent version of test_result.txt the failure is coming back as expected, but from the wrong place.

takluyver:
> So this also has a feature improvement (since people can specify templates without their extensions now), right?

It was already possible to specify a template file without the extension, because of [this code](https://github.com/jupyter/nbconvert/pull/491/files#diff-43c2e238630461c0a2c422a0b09fae55L194) adding the extension. But we were doing that regardless of where the template was coming from. This pushes the extension handling down a level so it only happens when the template is being loaded from the filesystem.

> The reason why none of that machinery is extension sensitive if you use another loader is because those loaders are just prepended onto the list and thus are operating independently of this change, right?

Right. In normal usage, we set up just a single `FileSystemLoader` to get templates from files. That loader is what's being wrapped in the new `ExtensionTolerantLoader`. In Matthew's case, he is adding a `DictLoader` to the start of the list. That will be asked for the template before our own loader, and because it's not wrapped in `ExtensionTolerantLoader`, it will only get the name as it was specified, with no additions.

(If someone wanted the extension-adding behaviour for their own loader, they could also wrap it in `ExtensionTolerantLoader`. But I'm not planning to document or support it as a public API.)

> Do we have separate machinery for interpreting .tplx templates then?

It uses the same machinery, but we tell Jinja2 to look for different markers [using this code](https://github.com/jupyter/nbconvert/blob/5759515c480ae339c31573739f10a356ab4ce652/nbconvert/exporters/latex.py#L87-L93). Specifying different syntax like this is a documented feature of Jinja2, but I think it's not very widely used.

test_result.txt
```diff
@@ -0,0 +1,65 @@
+============================= test session starts ==============================
```
takluyver: Was there a particular reason you committed the test results? This isn't something we normally do.

nbconvert/exporters/tests/test_templateexporter.py
```diff
@@ -114,6 +117,43 @@ def test_relative_template_file(self):
...
+        exporter = MyExporter(extra_loaders=[my_loader])
+        nb = v4.new_notebook()
+        out, resources = exporter.from_notebook_node(nb)
+
+
+    @pytest.mark.xfail(strict=True, raises=TemplateNotFound)
+    def test_in_memory_template_failure_to_find(self):
+        # Loads in an in memory template using jinja2.DictLoader
+        # creates a class that uses this template with the template_file argument
+        # converts an empty notebook using this mechanism
```
takluyver: The test here doesn't match up to its description; I'm not sure which you intend it to be ;-)

test_result.txt
michaelpacer: Yes I know we normally don't do this, but I was trying to allow you to see what I was seeing as I was stuck. The hope was that in seeing the error wherever it was I ended up getting stuck you could suggest a remedy. The idea is to put the test in here for now while writing the code and remove it later.

michaelpacer: The goal was to regain one of the advantages of pair coding that is lost when we only push successful changes on Github.

takluyver: Aha, gotcha. Good idea, but I'd suggest doing that by putting the log in a gist and linking to it. Anything that's committed to the repository ends up in everyone's copy of the git history for ever, unless we fiddle around with rebasing and squashing commits, so we tend not to commit things that we intend to delete again.

nbconvert/exporters/tests/test_templateexporter.py
takluyver: Going from this commit message:

> Now it's happening in the Choice Loader and not the ExtensionTolerantLoader

I think I see the confusion. `ChoiceLoader` is what wraps the series of loaders to try, so that the user can provide `extra_loaders` with higher priority than our default loader. ChoiceLoader will try loaders in sequence. If one of those loaders throws `TemplateNotFound`, ChoiceLoader [catches that](https://github.com/pallets/jinja/blob/2.8/jinja2/loaders.py#L397) and moves on to try the next loader. If all loaders fail like that, ChoiceLoader finally raises TNF itself.

So the TNF exception you get from using our `TemplateExporter` class will always come from ChoiceLoader, because it catches the exceptions from individual loaders.

michaelpacer: Also how would you specify the extra_loaders value if you weren't including it in the definition of the Exporter itself? It looked like it didn't like directly setting it because of it's identity as a traitlet List.

michaelpacer:
Ok I'm at an impasse at figuring out how to switch out ChoiceLoader, since it's fairly deep in the environment stack.

michaelpacer:
squashed away the tests, sad to change history, but in this case I think it makes sense.

nbconvert/exporters/templateexporter.py
```diff
@@ -55,6 +57,32 @@
...
+        try:
+            return self.loader.get_source(environment, template)
+        except TemplateNotFound:
+            print(template,str(template.endswith(self.extension)))
+            if template.endswith(self.extension):
+                raise TemplateNotFound(template)
+            return self.loader.get_source(environment, template+self.extension)
+
+    def list_templates(self):
+        return self.loader.list_templates()
```
takluyver: Nothing in our code currently uses it, but it's part of the API for template loaders in Jinja, and since it's easy to implement, I implemented it.

nbconvert/exporters/tests/test_templateexporter.py
takluyver: I think it should be sufficient to test it from this level, though I think we can simplify the test a bit.

I'm not sure exactly what you're asking about setting extra_loaders. It should work with either of these patterns:

```python
# Either:
exporter = TemplateExporter(extra_loaders=[foo])

# Or:
exporter = TemplateExporter()
exporter.extra_loaders = [foo]
```

nbconvert/exporters/tests/test_templateexporter.py
```diff
@@ -114,6 +117,46 @@ def test_relative_template_file(self):
...
+        class MyExporter(TemplateExporter):
+            self.template_file = 'my_template.tpl'
+        
+        with tempdir.TemporaryDirectory() as td:
+            template = os.path.join(td, 'my_template.tpl')
+            test_output = 'absolute!'
+            with open(template, 'w') as f:
+                f.write(test_output)
+            exporter = MyExporter()
+            exporter.template_file = template + exporter.template_extension
```
takluyver: I think we can simplify this a bit. All we need to test it, I think, is a template name that we can be reasonably sure won't exist, like `does_not_exist.tpl`. Then we don't need to subclass TemplateExporter, or create a temporary directory, or create a template that does exist. We can just instantiate `TemplateExporter(template_file='does_not_exist.tpl')` and check that trying to use it raises TemplateNotFound.

nbconvert/exporters/tests/test_templateexporter.py
michaelpacer: The second `exporter.extra_loaders = [foo]` was giving me an error, I can't access it anymore (because squash) and it's not giving me an error here, so I must have been doing something differently and incorrectly.

michaelpacer:
Took a little bit of figuring out how the load_templates() functions work in Jinja2, since there are actually 2 different kinds of them but you'd never know it by looking at their API docs ( `Environment.load_templates(extensions=None, filter_func=None)`&  *`Loader.load_templates()`).

nbconvert/exporters/tests/test_templateexporter.py
```diff
@@ -114,6 +117,35 @@ def test_relative_template_file(self):
...
+        class MyExporter(TemplateExporter):
+            template_file = 'my_template'
+        
+        exporter = MyExporter(extra_loaders=[my_loader])
+        nb = v4.new_notebook()
+        out, resources = exporter.from_notebook_node(nb)
+
+
+    @pytest.mark.xfail(strict=True, raises=TemplateNotFound)
+    def test_in_memory_template_failure_to_find(self):
```
takluyver: This test name is misleading, because it's not doing anything with in-memory templates.

Once that's fixed, I think this is good.

michaelpacer:
Sorry for the delay on this, it should all be good now.

michaelpacer:
@takluyver ping, I'm thinking that if this and a couple of the other PRs get in it'll be good for a 5.1 release (which then will also let me shift back rtd to stable).

nbconvert/exporters/tests/test_templateexporter.py
```diff
@@ -114,6 +117,35 @@ def test_relative_template_file(self):
...
+        
+        class MyExporter(TemplateExporter):
+            template_file = 'my_template'
+        
+        exporter = MyExporter(extra_loaders=[my_loader])
+        nb = v4.new_notebook()
+        out, resources = exporter.from_notebook_node(nb)
+
+
+    @pytest.mark.xfail(strict=True, raises=TemplateNotFound)
```
takluyver: As a stylistic point, I think it's nicer to use `pytest.raises()`, making it specific to the statement we expect to raise an error, rather than marking the whole test as an expected failure. More info in docs here:

http://doc.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

I'm not going to hold up merging for this, though.

Issue #492, opened 2016-12-16
Check about nbsphinx and whether it works on sphinx 1.5.2 


michaelpacer:
As of #484 nbsphinx had been incompatible with sphinx 1.5.0 and 1.5.1, with a fix expected to be included in 1.5.2 as described here: https://github.com/sphinx-doc/sphinx/pull/3224. 
However, https://github.com/spatialaudio/nbsphinx/pull/85 may fix the incompatibility with 1.5.1 so it may be enough to get around this problem to rely on a more recent version of nbsphinx.

mgeier:
I've just released `nbsphinx` 0.2.12: https://pypi.python.org/pypi/nbsphinx/0.2.12

I hope it works now with Sphinx 1.5.1.
It should also work with Sphinx 1.5.0, but only when forcing `docutils<0.13.1`.

michaelpacer:
Ok #493 and #494 should address this.

PR #494, opened 2016-12-19
bring up to date with nbsphinx's compatibility with sphinx 1.5.1

michaelpacer:
Same as #493, but for 5.x not master. 

https://github.com/jupyter/nbconvert/issues/492#issuecomment-268060877 points out that we can update to the latest sphinx so long as we update to the latest nbsphinx.

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -7,9 +7,9 @@ dependencies:
 - nbformat
 - jupyter_client
 - ipython
-- sphinx>=1.3.6,!=1.5.0,!=1.5.1
+- sphinx>=1.5.1
 - sphinx_rtd_theme
 - tornado
 - entrypoints
 - pip:
-    - nbsphinx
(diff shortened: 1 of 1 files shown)


takluyver:
5.x doesn't currently have any commits that aren't on master. Our usual practice is to only make maintenance branches when we merge the first thing that won't be in the maintenance releases (i.e. for this branch, when we merge the first 6.0 only change). So I'm going to delete the 5.x branch for now to reduce the maintenance overhead.

michaelpacer:
This also was the branch to which our readthedocs pointed. And Matthias pointed out that it is bad to have it pointing to latest anyway since then it is liable to break. That was why I hadn't switched back to latest. This can be solved with a patch release that I'll release tomorrow.

Basically I want to get in the fix for in-memory representations of templates (#491), but that needs another pair of eyes, unless you think the tests look good as is.

Issue #495, opened 2016-12-20
ImportError: No module named jupyter_client.manager


stefanw:
In a fresh environment:

```
$ pip install nbconvert
$ python -c "import nbconvert"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "lib/python2.7/site-packages/nbconvert/__init__.py", line 6, in <module>
    from . import preprocessors
  File "lib/python2.7/site-packages/nbconvert/preprocessors/__init__.py", line 10, in <module>
    from .execute import ExecutePreprocessor
  File "lib/python2.7/site-packages/nbconvert/preprocessors/execute.py", line 21, in <module>
    from jupyter_client.manager import KernelManager
ImportError: No module named jupyter_client.manager
```

So without the `extras_require` named `execute` `nbconvert` crashes on import time.

takluyver:
Thanks, I've opened #499 to address this.

PR #499, opened 2016-12-21
Don't import jupyter_client unless we're executing the notebook

takluyver:
Closes gh-495

Also added `jupyter_client` as an explicit dependency for the tests, since it's imported in test code.

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -18,7 +18,6 @@ from nbformat.v4 import output_from_msg
 from .base import Preprocessor
 from ..utils.exceptions import ConversionException
 from traitlets import Integer
-from jupyter_client.manager import KernelManager
 
 
 class CellExecutionError(ConversionException):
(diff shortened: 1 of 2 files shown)


michaelpacer:
Why does setting the default value as a string work when previously it was being passed a class object?

Happy to merge, just wanted to know.

takluyver:
The `Type` and `Instance` traitlets have some special functionality that you can supply classes as a string containing the full importable name of the class. They import it and get the actual class object when the object they're attached to is instantiated. Here's the code in `Type`:

https://github.com/ipython/traitlets/blob/4.3.1/traitlets/traitlets.py#L1601

michaelpacer:
So, based on https://github.com/conda-forge/nbconvert-feedstock/pull/7/commits/1fce020e01e10968d16ffcae54cb39d8348dcca7 it suggests we should not have `jupyter_client` as *any* kind of dependency (strictly for tests or optionally)… I'm not sure how to fix that though while maintaining all current functionality.

minrk:
I believe everything is correct now. It's a dependency for a feature, which is expressed via the setuptools `extras_require` field.

PR #500, opened 2016-12-21
Add block to allow customizing html output prompt

eddiejessup:
For generating HTML reports I usually want to completely hide cells' output, apart from a select few types such as HTML and images. I can almost accomplish this with a custom template inheriting from `full.tpl`, except that I can't hide the output prompt without hiding the cell output itself. I can if I add this block.

### DIFF
--- a/nbconvert/templates/html/basic.tpl
+++ b/nbconvert/templates/html/basic.tpl
@@ -53,6 +53,7 @@ In&nbsp;[&nbsp;]:
 
 {% block output %}
 <div class="output_area">
+{% block output_area_prompt %}
 {%- if output.output_type == 'execute_result' -%}
     <div class="prompt output_prompt">
 {%- if cell.execution_count is defined -%}
@@ -64,6 +65,7 @@ In&nbsp;[&nbsp;]:
     <div class="prompt">
 {%- endif -%}
     </div>
+{% endblock output_area_prompt %}
 {{ super() }}
 </div>
 {% endblock output %}


takluyver:
Thanks, I think this is a reasonable addition. 

Can you add it to the [template structure doc](https://github.com/jupyter/nbconvert/blob/master/docs/source/template_structure.html), which is rendered at:
http://nbconvert.readthedocs.io/en/stable/customizing.html#Template-structure

michaelpacer:
Should be good to go.

docs/source/template_structure.html
```diff
@@ -61,28 +61,38 @@
 
 <div class="template_block">outputs
     <div class="template_block">output
-        <div class="template_block">execute_result</div>
+        <div class="template_block">output_area_prompt
+            <div class="template_block">execute_result</div>
+        </div>
```
takluyver: I don't think this is quite right:

1. The new block goes before the output content blocks (like `execute_result`), not wrapping them.
2. The new block is only in the HTML template; there's a separate section below for extra blocks in the HTML template.

eddiejessup:
Hey @michaelpacer, based on @takluyver's comments I dropped your commit, but thanks for helping out, I definitely wouldn't have noticed I needed to update the notebook otherwise!

@takluyver, I think the HTML structure is now correct. I wasn't sure how best to state the block inheritance, so I just went for an explicit `{% super %}`, hope this doesn't make the document show too much implementation guts. If you can think of a nicer way to express this, I'd be happy to change it.

takluyver:
Thanks! I think `{% super %}` is too much implementation guts. Can we say something like 'output type as above'? There's already an 'as above' in the Latex section below. Better wording welcome.

eddiejessup:
@takluyver thanks for the feedback! I think all done.

takluyver:
Thanks :-)

michaelpacer:
This may have broken rtd.

eddiejessup:
@michaelpracer hey, sorry what's rtd mean?

eddiejessup:
@michaelpacer

michaelpacer:
Readthedocs I.e., http://nbconvert.readthedocs.io

I got an email warning of its failure this morning and this is the pr that
merged around then.

eddiejessup:
Thanks for catching this. I'm not familiar with how readthedocs works, where can I see its build details? I can see that it looks like the ipython notebook with the template structure is shown there, but the PR changes aren't reflected, even in its 'latest' version.

takluyver:
I think Mike has got it working again, but for the record, you can see RTD build logs here: https://readthedocs.org/projects/nbconvert/builds/

michaelpacer:
I think it ended up being independent of this PR per se (it was just what triggered the most recent failure), so no worries.

PR #502, opened 2016-12-24
decode a string containing unicode chars to unicode

srinivasreddy:


nbconvert/exporters/pdf.py
```diff
@@ -163,9 +163,9 @@ def from_notebook_node(self, nb, resources=None, **kw):
         )
         # set texinputs directory, so that local files will be found
         if resources and resources.get('metadata', {}).get('path'):
-            self.texinputs = resources['metadata']['path']
+            self.texinputs = str_to_unicode(resources['metadata']['path'])
```
takluyver: The general rule with unicode is that we should convert this as soon as possible, i.e. where it's passed in to nbconvert, rather than here. Have a look in `Exporter.from_filename`.

nbconvert/exporters/pdf.py
```diff
@@ -163,9 +163,9 @@ def from_notebook_node(self, nb, resources=None, **kw):
...
         else:
-            self.texinputs = os.getcwd()
+            self.texinputs = str_to_unicode(os.getcwd())
```
takluyver: The py3compat module should have a function `getcwd` which points to `os.getcwdu()` on Python 2, which is neater than converting.

michaelpacer:
Looks like `str_to_unicode` needs to be used slightly differently given that this errs on python 2.7 with an issue regarding ascii encoding. If you look in https://github.com/ipython/ipython_genutils/blob/master/ipython_genutils/py3compat.py you'll see it's mapped to `decode` on python2.7.

```python
def decode(s, encoding=None):
    encoding = encoding or DEFAULT_ENCODING
    return s.decode(encoding, "replace")
```

Try running `py.test --pyargs nbconvert` in a python 2 environment (or even faster would be to find the specific tests that are failing and drill into the directory structure to test only them, though then you wouldn't use `--pyargs`. Those approaches should help when debugging this.

srinivasreddy:
@michaelpacer  What do you think?

michaelpacer:
LGTM, and it's passing the tests. Merging.

PR #505, opened 2017-01-04
Add links to tests by having them in the example notebook

michaelpacer:
I noticed none of the link code included in https://github.com/jupyter/nbconvert/pull/458 was being tested. 

Figured I'd preëmptively include links for both pandoc compatible and incompatible automatic header-id generation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,7 @@ addons:
 before_install:
     - git clone --quiet --depth 1 https://github.com/minrk/travis-wheels travis-wheels
 install:
-    # - wget https://hackage.haskell.org/package/pandoc-1.19.1/pandoc-1.19.1.tar.gz && tar -xzvf pandoc-1.19.1.tar.gz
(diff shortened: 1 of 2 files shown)


michaelpacer:
Ok I think this is breaking because travis is using the older version of pandoc as installed https://github.com/jupyter/nbconvert/blob/master/.travis.yml#L30, we should update it since it's on version 1.15.1 and the most recent pandoc version is 1.19.1.

michaelpacer:
Right now I'm running into issues that I cannot reproduce locally. They seem to be related to finding pandoc after it is unzipped, since the pandoc command is not found. My guess is that this has to do with the way that we were unpacking it before (using what I think is a rackspace cdn), which isn't the standard installation procedure (at least in terms of the link given in the pandoc docs). 

I'm going to figure out how to build travis locally and then will take a crack at debugging it more directly.

michaelpacer:
@Carreau | @minrk | @takluyver Can I get an external review? I figure the only thing that could be in disagreement about this is shifting to pandoc's newest version, with the API against which we've been writing code. 

Note: our current code fails when using the old API (since travis is using pandoc v1.15, just including the links in the example notebook causes failures).

PR #506, opened 2017-01-04
Pin python version to 3.5 for RTD so that docs will build

michaelpacer:
I'm not entirely sure why, but when readthedocs attempts to use python 3.6 to build our docs, it fails. 

I think this may relate to dependencies that do not expect to be interpreted via python 3.6. See the error log at https://readthedocs.org/projects/nbconvert/builds/4854852/.

Or the issue I filed with conda at https://github.com/conda/conda/issues/4168

What really confuses me though is that this included some dependencies that list `3.6*` as a potential version number… The only reason I thought to notice this is that I saw `py36_` in the installation file names. 

Solution: pinning the version at 3.5 until we figure out how to better solve this problem. (Note, the rtd build succeeds with this change, see https://readthedocs.org/projects/nbconvert-mp/builds/4855978/)

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -2,7 +2,7 @@ name: nbconvert_docs
 channels:
   - conda-forge
 dependencies:
-- python>=3.5
+- python==3.5
 - pandoc
 - nbformat
 - jupyter_client


michaelpacer:
Hmm this fixed this for my personal repo but not the jupyter one.

michaelpacer:
Oh actually it did work! It just was the last to be merged so the other two merges still failed & sent emails saying so.

Issue #452, opened 2016-10-23
Conversion back from executable script to ipynb notebook


meta-coder:
Is there a way to convert an executable (Python) script to notebook? Assume a suitable format for embedding cell meta-data in executable scripts.
Something like conf file formats may be:
```
# [code]
...
...
# [markdown]
...
...
```

takluyver:
We don't have a tool to do that, but it shouldn't be too hard to write a script to do it if you want to. You can use the `nbformat` library to construct a notebook.

clouds56:
But it's hard to convert the python script file generated by nbconvert back to notebook.

e.g. consider the flowing code, which is an valid output from nbconvert

    # 1. this is a markdown comment
    
    # In[1]:
    
    some_python_code()
    
    # 2. this is a comment in code cell
    
    # In[2]:
    
    you_could_not_the_different(markdown_comment, comment_in_code)

could we have some special format for markdown cell? something like

    # Markdown:
    # this is a markdown cell

    # this is comment in code

michaelpacer:
@clouds56 Including additional text would break expectations about what the exporter is doing. However, we could programmatically use triple quoted string literals to indicate Markdown, which are less commonly used (except for docstrings).

It'd be easy enough to detect docstrings (as they must follow function definitions), and we could expect that people hold to the convention that any in-line string literals that not assigned, and therefore are acting as comments (which is technically frowned upon anyway) should be indented to align with the indentation level of the lines immediately preceding it. 

This wouldn't be perfect, but I think it would hit the 80/20 solution. Does that sound like a good approach @takluyver @Carreau @minrk?

michaelpacer:
Note also, that if #507 is merged as is (to solve #503), then this will currently lose that information. 

What we could do instead is add a line comment immediately after the magic command rewriting that would allow us to recover the original command.

michaelpacer:
Ok now #507 is more compatible with reconstructing the notebook, it will just have to know the convention that in the case of matplotlib magic commands (in scripts `get_ipython.magic(matplotlib …)`), if there is a comment immediately following that begins with `# nbconvert removed: …` it will need to replace the content that follows back into the original command. 

A useful note for implementation: currently, markdown blocks can be distinguished from the lack of a `#` with a new line between blocks that are prefixed by `#`.

I don't know the full variety of cases where nbconvert might want to deliberately change the content of code to be more amenable to processing as a script, but there's probably a more general form for this kind of convention that could work well. 

[...]

PR #507, opened 2017-01-05
Check for and remove gui based mpl backends

### LINKED ISSUE #503: `%matplotlib inline` fails when converted to script
This might be known, but didn't see a bug for it. Consider a notebook `test.pynb` containing the magic `%matplotlib inline`and converted to a script `test.py`:

    jupyter nbconvert --to script test.ipynb

`test.py` now looks like

    get_ipython().magic('matplotlib inline')

However, `get_ipython()` is undefined and the script will be unusable. nbconvert version 5.0.0

mpacer:
This should close #503. 

I tried to make the regex flexible going forward, such that if matplotlib ever has more gui incompatible backends (or if someone wants to write a template that excludes more back ends), we/they can just plug them in the filter (where they flow down to the the regex generating function) rather than needing to hard code them.

Also, updated example notebook to a more standard kernel name, more recent nbformat version, &c. and included examples of different matplotlib backend calls.

### DIFF
--- a/nbconvert/filters/strings.py
+++ b/nbconvert/filters/strings.py
@@ -188,7 +188,39 @@ def get_lines(text, start=None,end=None):
     # Return the right lines.
     return "\n".join(lines[start:end]) #re-join
 
-def ipython2python(code):
+def _mpl_magic_regex_generator(forbidden_backends=None):
+    """Create regex for detecting incompatible backends in mpl magics.
+    
+    Parameters
+    ----------
+
+    forbidden_backens: List
+        List of strings, each of which is a forbidden matplotlib backend. 
+    """
+    if forbidden_backends is None:
+        forbidden_backends = ["notebook","inline"]
+    no_mpl_be_str = "|".join(forbidden_backends)
+    mpl_magic_backend_extractor = re.compile(r'(^%matplotlib \w*)({})'.format(no_mpl_be_str))
+    return mpl_magic_backend_extractor
+
+mpl_magic_regex = _mpl_magic_regex_generator()
+
+def _remove_gui_mpl_backends(code):
+    """Remove incompatible backends from mpl magics.
+    
+    Parameters
+    ----------
+
+    code : str
+        IPython code with matplotlib magic
+    """
+    lines = code.split('\n',maxsplit=1)
+    if mpl_magic_regex.match(lines[0]):
+        lines[0] = mpl_magic_regex.match(lines[0]).group(1)
+        code = "\n".join(lines)
+    return code
+
+def ipython2python(code, no_gui=False):
     """Transform IPython syntax to pure Python syntax
 
     Parameters
@@ -206,6 +238,8 @@ def ipython2python(code):
         )
         return code
     else:
+        if no_gui and "%matplotlib" in code:
+            code = _remove_gui_mpl_backends(code)
         isp = IPythonInputSplitter(line_input_checker=False)
         return isp.transform_cell(code)
 
--- a/nbconvert/filters/strings.py
+++ b/nbconvert/filters/strings.py
@@ -221,11 +221,12 @@ def _remove_browser_mpl_backends(code, browser_backends=None):
     else: 
         mpl_magic_regex_local = mpl_magic_regex
 
-    lines = code.split('\n',1)
-    notebook_magic = mpl_magic_regex_local.match(lines[0])
-    if notebook_magic: 
-        lines[0] = notebook_magic.group(1)
-        lines[0] += "\n# nbconvert removed: " + notebook_magic.group(2)
+    lines = code.split('\n')
+    magic_line = [i for i,x in enumerate(lines) if mpl_magic_regex_local.match(lines[x])]
+    if len(magic_line)!=0: 
+        # only change first backend declaration (since the rest shouldn't work anyway)
+        lines[magic_line[0]] = mpl_magic_regex_local.match(lines[magic_line[0]]).group(1)
+        lines[magic_line[0]] += "\n# nbconvert removed: " + mpl_magic_regex_local.match(lines[magic_line[0]]).group(2)
         code = "\n".join(lines)
     return code
 
--- a/nbconvert/filters/strings.py
+++ b/nbconvert/filters/strings.py
@@ -222,8 +222,10 @@ def _remove_gui_mpl_backends(code, bad_backends=None):
         mpl_magic_regex_local = mpl_magic_regex
 
     lines = code.split('\n',1)
-    if mpl_magic_regex.match(lines[0]) and bad_backends is None:
-        lines[0] = mpl_magic_regex_local.match(lines[0]).group(1)
+    notebook_magic = mpl_magic_regex_local.match(lines[0])
+    if notebook_magic: 
+        lines[0] = notebook_magic.group(1)
+        lines[0] += "\n# nbconvert removed: " + notebook_magic.group(2)
         code = "\n".join(lines)
     return code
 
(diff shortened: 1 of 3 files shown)


michaelpacer:
Seems like a good idea to indicate what was originally, in case someone ever wanted to reconstruct the original notebook from the exported format, based on #452.

michaelpacer:
@takluyver @Carreau @minrk Forgot to ping y'all on this one. 

Is there a reason not to do this? I was a little worried about losing the information before, but this seems pretty clean of a change now that it's still recoverable. I'm up for changing how we represent that information, but the core idea seems to be solid.

nbconvert/filters/strings.py
```diff
@@ -188,14 +188,59 @@ def get_lines(text, start=None,end=None):
...
+    """Create regex for detecting incompatible backends in mpl magics.
+    
+    Parameters
+    ----------
+
+    forbidden_backens: List
+        List of strings, each of which is a forbidden matplotlib backend. 
+    """
+    if forbidden_backends is None:
+        forbidden_backends = ["notebook", "inline"]
```
takluyver: The code elsewhere is talking about GUI backends, but I would use that term to refer to e.g. the Qt4 and GTK backends, which work with GUI toolkits.

takluyver:
I'm not entirely keen that this is embedding more knowledge of Python, and of IPython specifically, into nbconvert. That's generally the opposite of the way we're trying to move.

michaelpacer:
We have a python specific converter already, and right now functions that are useful in the notebook create broken exported scripts. That seems suboptimal. If we were to get rid of all IPython/python specific exporting machinery (including the exporter we ship & the ipython2python filter), I'd agree with you, but I'm pretty sure that's not actually desired.

nbconvert/filters/strings.py
michaelpacer: Do you have a suggested name for these backends?

nbconvert/filters/strings.py
```diff
@@ -188,14 +188,59 @@ def get_lines(text, start=None,end=None):
...
+    bad_backends: list of str
+        List of strings, each of which is a forbidden matplotlib backend.
+    """
+    if bad_backends is not None:
+        mpl_magic_regex_local = _mpl_magic_regex_generator(bad_backends)
+    else: 
+        mpl_magic_regex_local = mpl_magic_regex
+
+    lines = code.split('\n',1)
+    notebook_magic = mpl_magic_regex_local.match(lines[0])
```
jpeacock29: Since magic can occur anywhere in a cell this would miss cases like,

```python
pass
%matplotlib inline
```
We could check every line, but then we have to grapple with multiline strings.

nbconvert/filters/strings.py
```diff
@@ -188,14 +188,59 @@ def get_lines(text, start=None,end=None):
...
     return "\n".join(lines[start:end]) #re-join
 
-def ipython2python(code):
+def _mpl_magic_regex_generator(forbidden_backends=None):
+    """Create regex for detecting incompatible backends in mpl magics.
+    
+    Parameters
+    ----------
+
+    forbidden_backens: List
```
jpeacock29: Typo: "backens" -> "backends". Also, is there a reason the list of backends is called `forbidden_backends` here and `bad_backends` elsewhere?

minrk:
I'm concerned about adding special-casing of matplotlib to nbconvert. In general, we are trying to *remove* IPython-specific logic from Jupyter projects, and this adds a whole new class of IPython-specific feature: conditional code modification. Previously, we've only done syntax-translation, where the code doesn't change at all, we just turn IPython syntax into the Python syntax it corresponds to. And this was added before Jupyter existed, and probably wouldn't be accepted now, as it is too kernel-specific.

If more IPython-specific logic is desirable, I think a better place for IPython-specific manipulation of code is a separate, IPython-provided Exporter, which can be accessed by the nice new entrypoints mechanism.

Carreau:
I tend to agree with @takluyver and @minrk. We should not try to go through the slippery slope of automatically fixing things, or we should find a better way to do it. 

We have the same problem with `ipynb.fs` which can't import IPython-specific syntax. 
As magics are (mostly) syntactic sugar, we could _try_ to actually parse the final (i)python file.
And either:
  - fail – not valid Python
  - actually convert it to real python syntax.

The second case is harder as we would likely have to refactor the magics to have the ability to get rid of the `get_ipython()` when possible.

Though, it is true that `matplotlib inline` is a common case, but is it common enough in notebooks that get exported.

We might hit a case of practicality beats purity here.

Another ugly-ish solution would be to inject a `get_ipython()` which return a no-op getattr that just raise a warning.

takluyver:
Once matplotlib 2 comes out (which will be *any day now*, as it has been for a year or so ;-), it should no longer be necessary to invoke `%matplotlib inline`, because it will pick a suitable default backend when you load matplotlib in the kernel. Of course, lots of people will still do it because they're used to it, but it lowers the priority a little.

Carreau:
> Once matplotlib 2 comes out (which will be any day now, as it has been for a year or so ;-), it should no longer be necessary to invoke %matplotlib inline, because it will pick a suitable default backend when you load matplotlib in the kernel. Of course, lots of people will still do it because they're used to it, but it lowers the priority a little.

True, I forgot that !

In a couple of versions of to should we `warn('Hey ! You're using IPython 7.3 with matplotlib 2.5, matplotlib inline is not necessary anymore !')`

michaelpacer:
All of this seems sensible but doesn't change the fact that for now I think
this change should be included to solve this common case. I think fixing
this is no more “ugly” than having the `ipython2python` filter in the first
place.

I like the idea of migrating the python exporter to the ipython install,
but I think that it's not a bad idea. However I think that would require
solving the other pr (https://github.com/jupyter/nbconvert/pull/414) which
involves calling up language specific exporters. It stalled because there
were a couple of suggested approaches, and I wasn't sure what to move
forward on.

I thought that %matplotlib inline was actually required to be the first
line of a cell in which it appeared. It seems I upgraded a best practice to
a protocol in my mind. That should be fixed later today along with the
white space escaping.

JamiesHQ:
@mpacer : what's the next step on this? We're running into this issue in the Data Carpentry Reproducible Science Workshop. The notebook we're trying to convert to .py has *%matplotlib inline* and is failing. If the fix is not currently ready to go, is there a stopgap?

jpeacock29:
I know the current patch would incorrectly change magic found in a multiline string. I'm working on a fix for that by modifying the transform [escaped_commands()](https://github.com/ipython/ipython/blob/master/IPython/core/inputtransformer.py#L259) used in [IPythonInputSplitter](https://github.com/ipython/ipython/blob/master/IPython/core/inputsplitter.py#L559). There are also concerns of adding IPython specific code to Jupyter. I'd contend this is a case of practicality beats purity, since people are clearly encountering this issue.

choldgraf:
FWIW, I'm running on MPL 2.0 master and it did not automagically pick the inline backend in the notebook:

choldgraf:
Actually I should show a pick w/ the backend:

image

It **is** calling an inline backend, and if I explicitly have the cell return `fig` then it does get plotted inline:

image

So let me know if this is:

1. Desired behavior
1. A MPL bug and then I'll open something over there
1. A jupyter bug and then I'll keep chatting here :)

mpacer:
@choldgraf You should be bringing this up in the notebook repo rather than here. This PR is about what to do in the case that %matplotlib inline (or the like) is used in a notebook and we want to export it to a script. 

Just as a heads up, the anti-special casing `%matplotlib inline` camp is gaining ground. I'm still a little split on the issue, but @fperez, @carreau, @takluyver & @minrk are in overall agreement in being against including this in our default python exporter. 

To clarify, my thought is that we already have the `ipython2python` code and this is just a fairly common use case for an ipython magic that fails even if someone uses `ipython` to run their converted script. If we were going to eliminate this special casing code from our codebase, I'd be more inclined to agree with the "purity over practicality" argument, but without that change it seems like this is not that much less pure with a huge boost in practicality.

Carreau:

PR #509, opened 2017-01-08
Use pytest.raises not pytest.mark.xfail

michaelpacer:
Addresses comment in #491 by @takluyver.


takluyver:
Thanks, that's exactly what I was thinking of. :-)

PR #511, opened 2017-01-10
Allow kernel authors to suggest a package for script export

takluyver:
Part of gh-361

Now that we use entry points to find exporters, kernel authors could provide a custom script exporter, and specify it in language_info metadata. But currently, if that package is not installed, nbconvert will fail with an unhelpful error message. This is a problem especially for kernels which are not distributed as Python packages (e.g. R, Julia)

With this PR, if the specified exporter cannot be found, nbconvert will display a warning and fall back on plain script export. It also looks for a new field in language_info, called 'nbconvert_exporter_dist', to provide a suggestion of a package to install for better script export. If we do go for this, it will need to be documented in the messaging docs.

### DIFF
--- a/nbconvert/exporters/base.py
+++ b/nbconvert/exporters/base.py
@@ -26,6 +26,7 @@ __all__ = [
     'get_exporter',
     'get_export_names',
     'ExporterNameError',
+    'ExporterNotFound',
 ]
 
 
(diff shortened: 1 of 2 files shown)


michaelpacer:
Is there a reason to open a new PR rather than building on https://github.com/jupyter/nbconvert/pull/414?

takluyver:
Oh, sorry, I just forgot that existed.

PR #513, opened 2017-01-14
Add option to configure the IOPub timeout

jhamrick:
We're still having trouble with getting IOPub timeouts in some of the nbgrader tests, so it would be nice to be able to configure this option to be higher if necessary.

cc @lgpage


lgpage:
Just to add this is mainly an issue for tests on windows CI servers.

michaelpacer:
LGTM, merging.

Issue #504, opened 2017-01-03
issues with latex <


minrk:
reported at https://github.com/jupyter/nbviewer/issues/647 by @hzzyyy, with handy notebook for reproducing the bugs:

https://nbviewer.jupyter.org/github/hzzyyy/try/blob/master/latex%20parenthesis%20bug.ipynb

michaelpacer:
So it looks like we need to treat `<` and possibly `>` as a special case, using a character entities in html (`&lt;` and `&gt;`).

michaelpacer:
Also, for some reason nbconvert is still using 4.3.0 instead of 5.0.0? Is there a reason?

michaelpacer:
This appears to be a well known issue with MathJax, where having no spaces around `<` and `>` is [explicitly discouraged in the MathJax docs](http://docs.mathjax.org/en/latest/tex.html#tex-and-latex-in-html-documents).

So to support this is possible, but then we're explicitly flouting that recommendation and implicitly encouraging people to use nonstandard conventions for writing math in MathJax (as opposed to pure LaTeX).

michaelpacer:
Also, technically for pandoc output it should not accept some of these markdown cells as successfully specified math delimiters:

from [pandoc math docs](http://pandoc.org/MANUAL.html#math)

> **The opening $ must have a non-space character immediately to its right, while the closing $ must have a non-space character immediately to its left,** and must not be followed immediately by a digit.

zasdfgbnm:
https://github.com/jupyter/nbconvert/pull/514

PR #514, opened 2017-01-18
Fix bug when there is "<" in the math formula

zasdfgbnm:
The easiest way to trigger the bug is to add a markdown cell and put $$<k'>$$.
When the notebook is converted to html, the "<" is left unescaped,
which makes browsers parse this as a html tag

### DIFF
--- a/nbconvert/filters/markdown_mistune.py
+++ b/nbconvert/filters/markdown_mistune.py
@@ -8,6 +8,7 @@ Used from markdown.py
 from __future__ import print_function
 
 import re
+import html
 
 import mistune
 
@@ -104,20 +105,19 @@ class IPythonRenderer(mistune.Renderer):
         html = super(IPythonRenderer, self).header(text, level, raw=raw)
         return add_anchor(html)
 
-    def escape_lt(self,text):
-        return text.replace('<','&lt;')
+    def escape_html(self,text):
+        return html.escape(text,quote=False)
 
-    # Pass math through unaltered - mathjax does the rendering in the browser
     def block_math(self, text):
-        return '$$%s$$' % self.escape_lt(text)
+        return '$$%s$$' % self.escape_html(text)
 
     def latex_environment(self, name, text):
-        name = self.escape_lt(name)
-        text = self.escape_lt(text)
+        name = self.escape_html(name)
+        text = self.escape_html(text)
         return r'\begin{%s}%s\end{%s}' % (name, text, name)
 
     def inline_math(self, text):
-        return '$%s$' % self.escape_lt(text)
+        return '$%s$' % self.escape_html(text)
 
 def markdown2html_mistune(source):
     """Convert a markdown string to HTML using mistune"""
--- a/nbconvert/filters/markdown_mistune.py
+++ b/nbconvert/filters/markdown_mistune.py
@@ -104,15 +104,20 @@ class IPythonRenderer(mistune.Renderer):
         html = super(IPythonRenderer, self).header(text, level, raw=raw)
         return add_anchor(html)
 
+    def escape_lt(self,text):
+        return text.replace('<','&lt;')
+
     # Pass math through unaltered - mathjax does the rendering in the browser
     def block_math(self, text):
-        return '$$%s$$' % text
+        return '$$%s$$' % self.escape_lt(text)
 
     def latex_environment(self, name, text):
+        name = self.escape_lt(name)
+        text = self.escape_lt(text)
         return r'\begin{%s}%s\end{%s}' % (name, text, name)
 
     def inline_math(self, text):
-        return '$%s$' % text
+        return '$%s$' % self.escape_lt(text)
 
 def markdown2html_mistune(source):
     """Convert a markdown string to HTML using mistune"""
--- a/nbconvert/filters/markdown_mistune.py
+++ b/nbconvert/filters/markdown_mistune.py
@@ -8,7 +8,7 @@ Used from markdown.py
 from __future__ import print_function
 
 import re
-import html
+import cgi
 
 import mistune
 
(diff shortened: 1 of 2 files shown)


zasdfgbnm:
https://github.com/jupyter/nbconvert/issues/504

nbconvert/filters/markdown_mistune.py
```diff
@@ -104,15 +104,20 @@ def header(self, text, level, raw=None):
         html = super(IPythonRenderer, self).header(text, level, raw=raw)
         return add_anchor(html)
 
+    def escape_lt(self,text):
+        return text.replace('<','&lt;')
```
takluyver: Shouldn't we also be translating `>`? There's a function [html.escape()](https://docs.python.org/3/library/html.html#html.escape) which we might want to use.

takluyver:
Could you add a test for this in `nbconvert.filters.tests.test_markdown`? Thanks!

nbconvert/filters/markdown_mistune.py
zasdfgbnm: I didn't find any bug triggered by ">". But it is a good idea to escape all these special characters.

zasdfgbnm:
I will add the test

nbconvert/filters/markdown_mistune.py
```diff
@@ -104,15 +104,20 @@ def header(self, text, level, raw=None):
...
+
     # Pass math through unaltered - mathjax does the rendering in the browser
```
michaelpacer: This will no longer be true, we're deciding to follow a different standard than MathJax in this case, as MathJax does not always treat "<" and ">" as characters to be escaped (which leads to the "bugs" that this PR is trying to address).

zasdfgbnm:
It should work now.

takluyver:
Thanks. This looks OK to me now, but I'll let @michaelpacer have another look over it.

michaelpacer:
I just realised that I had had those comments in a sort of "staged" sense for the past day or so…

zasdfgbnm:
@michaelpacer You are actually making good suggestions. What I was doing on the test was just trying not to make too much changes on existing code. If @takluyver thinks it is better to restructure several tests in a simple PR, I have no problem doing this job.

michaelpacer:
@zasdfgbnm That instinct makes sense and in general I'm guessing is a good practice. I just think in this case it may merit a bigger change to the tests. 

That said, some of my comments can be addressed without changing the structure of the tests just by changing the set of cases in `test_markdown2html_math_escape`.

nbconvert/filters/tests/test_markdown.py
```diff
@@ -118,8 +118,8 @@ def test_markdown2html_heading_anchors(self):
         ]:
             self._try_markdown(markdown2html, md, tokens)
 
-    def test_markdown2html_math(self):
-        # Mathematical expressions should be passed through unaltered
+    def test_markdown2html_math_noescape(self):
+        # Mathematical expressions not containing <, >, & should be passed through unaltered
```
takluyver: Unifying them is probably ideal, but I'd merge a PR if that was the only thing outstanding.

nbconvert/filters/tests/test_markdown.py
```diff
@@ -131,7 +131,26 @@ def test_markdown2html_math(self):
                 ]
         for case in cases:
             self.assertIn(case, markdown2html(case))
-    
+
+    def test_markdown2html_math_escape(self):
+        # all the "<", ">", "&" must be escaped correctly
+        cases = [ "$a<b&b<lt$",
+                  "$a<b&lt;b>a;a-b<0$",
+                  "$<k'>$"]
```
takluyver: +1 to adding a few more test cases.

nbconvert/filters/tests/test_markdown.py
```diff
@@ -171,7 +190,8 @@ def test_markdown2html_math_paragraph(self):
         ]
 
         for case in cases:
-            self.assertIn(case, markdown2html(case))
+            s = markdown2html(case)
+            self.assertIn(case,self._unescape(s))
```
takluyver: It's a bit ugly, but it was a bit of an ugly test before - it checks that the Markdown rendering doesn't change these samples. So it only works with samples that don't use any Markdown syntax (other than the math syntax we add). I wouldn't particularly ask @zasdfgbnm to fix this. Obviously we welcome improvements to the test, but that's probably best as a later PR.

michaelpacer: Yes that was the original test before. Currently this transformation does not leave these samples unchanged, which is why the unescaping needs to happen. I get that to make the test pass as it is it's easier to not have to concern yourself with the escaped bits.

Also, if the idea of this test is that the math processing leaves markdown unaffected, we should probably include some `<`, `>` and `&` in the markdown section and not automatically unescape them (because the math processing is leaking out into the markdown processing). 

Also, this test should include a 

```latex

\begin{}
… 
\end{}

```

(i.e., including the new lines around the declaration) since, not being wrapped in any number of `$`, there is an absence of the easiest cues to delimiting math vs. markdown.

zasdfgbnm:
@takluyver @michaelpacer Sorry for the long wait. I was busy the past days. I merged the nonescape cases and escape cases for math formula. I also added test case for `$$...$$` and table as @michaelpacer  says. I didn't do anything to the `_unescape`, because I have no idea on exactly what to do to improve it.

zasdfgbnm:
Actually I don't think `_unescape` is something ugly only to make test passes as @michaelpacer  said. On python 3 there is a `unescape` on official library that does the same thing, but it's not on python 2. That's all the reason I write my own `_unescape`

takluyver:
Thanks @zasdfgbnm - I'm happy enough to live with the test being a bit ugly for now.

Issue #517, opened 2017-01-22
Make extension tolerant loader more tolerant of non-extended templates


michaelpacer:
Regression introduced with #491 to deal with extension free templates. 

Should be fixable by adding try excepts on secondary template; solution should allow either specifying an extension free template or passing it and treat the two as equivalent (this was behaviour before 5.1.1).

takluyver:
I'm having trouble understanding what the issue is (both here and on #518, which I assume is about this). Some concrete examples would help!

takluyver:
Here's the example from nbsphinx that is going wrong:

```python
loader = jinja2.DictLoader({'nbsphinx-rst.tpl': RST_TEMPLATE})
        super(Exporter, self).__init__(
            template_file='nbsphinx-rst', extra_loaders=[loader],
            ...
```

I.e. defining a template in a custom loader with an extension, then specifying it without the extension.

1. Do we want to do extension munging for custom loaders? I wrote in #491 "assume anyone who is providing extra loaders knows what they're doing." (by which I meant, they will specify the extension exactly as it is). There's a case to be made for simply fixing it in nbsphinx. If we do want to muck around with file extensions from custom loaders, we can either wrap all loaders in `ExtensionTolerantLoader`, or get rid of ETL entirely and go back to doing it in TemplateExporter.
2. Is there any reason to allow the opposite (template declared without extension, then specified with)? #518 adds this, but I haven't seen any examples of people doing that, I don't believe it worked before, and it's extra complexity.

takluyver:
I've just made a PR fixing it in nbsphinx (in a slightly different way to Mike's existing PR). If @mgeier is happy to merge that and do a new release soon, I think I'm inclined not to change anything in nbconvert. I'd like to try to keep the ambiguity around template names as little as possible, by minimising the number of cases in which we try with a variant of the name specified.

mgeier:
I just merged @takluyver's PR. I can make a release shortly, say, tomorrow if nothing else comes up.

takluyver:
Thanks :-)

michaelpacer:
To me handling, this kind of extension munging makes for a more pleasant user experience for at least some users. 

Yes, more advanced users are the only ones who are using this today, and they _could_ figure out what's going wrong, but should they __have__ to? If we can handle it (& we can), why not ease friction wherever we can?

[...]

PR #518, opened 2017-01-22
add test to avoid nbsphinx issue with template extensions

michaelpacer:
Fixes issue #517 but does so in a way I don't like, I'm going to simplify the logic next.

This does mean you can't have both a template without and with a file extension and expect straightforward importing behaviour (but on disk templates should have extensions anyway).

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -209,7 +209,6 @@ class TemplateExporter(Exporter):
         """
         self.log.debug("Attempting to load template %s", template_file)
         self.log.debug("    template_path: %s", os.pathsep.join(self.template_path))
-        pass
 
     def _load_template(self):
         """Load the Jinja template object from the template file
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -204,6 +204,12 @@ class TemplateExporter(Exporter):
         self.observe(self._invalidate_template_cache,
                      list(self.traits(affects_template=True)))
 
+    def _log_template_loading(self,template_file):
+        """ abstract away some of the loggging for finding templates
+        """
+        self.log.debug("Attempting to load template %s", template_file)
+        self.log.debug("    template_path: %s", os.pathsep.join(self.template_path))
+        pass
 
     def _load_template(self):
         """Load the Jinja template object from the template file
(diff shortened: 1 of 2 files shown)


willingc:
@michaelpacer This looks good to me. Ran fine locally for me too.

nbconvert/exporters/templateexporter.py
```diff
@@ -203,6 +204,12 @@ def __init__(self, config=None, **kw):
         self.observe(self._invalidate_template_cache,
                      list(self.traits(affects_template=True)))
 
+    def _log_template_loading(self,template_file):
+        """ abstract away some of the loggging for finding templates
+        """
+        self.log.debug("Attempting to load template %s", template_file)
+        self.log.debug("    template_path: %s", os.pathsep.join(self.template_path))
+        pass
```
Carreau: extra `pass`

Carreau:
+1

nbconvert/exporters/templateexporter.py
```diff
@@ -217,9 +223,26 @@ def _load_template(self):
...
+                    return self.environment.get_template(template_file)
+                except TemplateNotFound:
+                    raise TemplateNotFound(template_file)
+            else:
+                try:
+                    template_file = self.template_file + self.template_extension
+                    self._log_template_loading(template_file)
+                    return self.environment.get_template(template_file)
+                except TemplateNotFound:
+                    raise TemplateNotFound(template_file)
```
takluyver: Having a bunch of checks and fallbacks here is precisely what I was trying to avoid by making the ExtensionTolerantLoader, so I'd rather not put them all back in here.

willingc: @takluyver It wasn't totally clear to me what extension tolerant meant. I would recommend adding a section to the class docstring to clarify that these checks are not needed when using ExtensionTolerantLoader. Perhaps a quick rationale sentence as the 2nd paragraph in the docstring before constructor paragraph.

michaelpacer: One of the primary use cases that this broke was nbsphinx. That's fixed now, but it's not obvious to me what is different between 

```python
from jinja2 import DictLoader
my_loader = DictLoader({'my_template': "{%- extends 'rst.tpl' -%}"})
    template_file = 'my_template'
exporter = MyExporter(extra_loaders=[my_loader])
```
and 
```python
loader = jinja2.DictLoader({'nbsphinx-rst': RST_TEMPLATE})
      super(Exporter, self).__init__(
          template_file='nbsphinx-rst', extra_loaders=[loader],          
          config= traitlets.config.Config(
                 {'HighlightMagicsPreprocessor': {'enabled': True}}),		                 
          filters={	
                 'markdown2rst': markdown2rst,
                 'get_empty_lines': _get_empty_lines,
                 'extract_toctree': _extract_toctree,
                 'get_output_type': _get_output_type,
                 'json_dumps': json.dumps,
             })		             
 		 
```

And extension tolerant loader seems to not be activated in that case; my first attempt at this was to add additional logic to ExtensionTolerantLoader to look for the case where the extension was present and could instead be removed, but despite the logic being correct (as evidence by logging) trying to loading from there didn't work and that failure seemed to be related to the `_load_template` function.

michaelpacer: I suppose the lack of clarity of what I attempted is in part due to my trying to clean up my commit history and overwriting the earlier stuff I did adding additional logic to ExtensionTolerantLoader. I'll try do that less going forward.

takluyver: I don't think there's any significant difference between those, which don't use an extension either to declare or select the template. I think that both would be broken in 5.0, and fixed by #491.

The case that broke in nbsphinx was with `nbsphinx-rst.tpl` as the name where the template was defined.

takluyver:
I don't think we should change this functionality - I've explained why in #517 . I'm happy to clarify the docstring of ExtensionTolerantLoader, though.

nbconvert/exporters/templateexporter.py
takluyver: I rarely go digging into the commit history of a PR to work out what the author was thinking. There are just too many PRs for that kind of detailed inspection ;-). IMHO, the important thing is something I know you're good at: writing. When you open an issue or PR, write a clear description of the problem at hand. If a PR relates to an issue, link it ("fixes issue #517" instead of "fixes the issue"!).

michaelpacer: The second custom loader was broken with #491, the former was not. That's the problem this PR is fixing.

takluyver: Yep, the version that was broken by #491 was like this:

```python
loader = jinja2.DictLoader({'nbsphinx-rst.tpl': RST_TEMPLATE})
        super(Exporter, self).__init__(
            template_file='nbsphinx-rst', extra_loaders=[loader],
            ...
```

Here, the template is defined with the `.tpl` extension, and selected without it. That's different from the first example, where it is both defined and called without the extension.

michaelpacer:
Arguing over this is not a productive use of our time when the original source of the problem is fixed. 

In #517 it was said that there aren't too many people that are using the style of a template declaration that used in nbconvert but doesn't after the release of 5.1.1. Hopefully that is correct, because if they did use it, 5.1.1 will break their code.

PR #519, opened 2017-01-23
use pytest.mark.parameterize not yield tests from nose

michaelpacer:
Got tired of the deprecation warnings, and after a little bit of looking into it, it wasn't too hard to fix. 

In fact the `@pytest.mark.parametrize` decorator may be useful in other contexts (now that I know how it works). 

But this is all I'm going to do in this pr.


willingc:
Yay test enhancements! Ran fine locally and is passing here. Thanks @michaelpacer :sunny:

PR #520, opened 2017-01-27
Configurable filter

### LINKED ISSUE #412: In exported Jupyter Notebooks, how do I disable the ¶?
Following up from my [SO question](http://stackoverflow.com/questions/39518136/in-exported-jupyter-notebooks-how-do-i-disable-the?noredirect=1#comment66369347_39518136), is there a way to not put pilcrows on headers?

mpacer:
This closes #412 (replacing #508), and introduces a mechanism for allowing other configurable properties to be added to existing filters without breaking API compatibility by using callable functions.

It introduces a deprecation warning for non-html export in an environment that does not contain [`mistune`](https://github.com/lepture/mistune) as it is a required dependency for installing nbconvert at least as of 4.0 (i.e., before the move to the jupyter GitHub organisation). 

It also introduces a deprecation warning for html export that does not use this mechanism (including `markdown2html_pandoc`) because we cannot guarantee that other mechanisms will be able to respect the configurable parameters that people may include, resulting in future confusion. We have already provided for those wanting this functionality in custom templates through the use of `convert_pandoc(source, "markdown", "html", extra_args=extra_args)`.

Thanks to @Carreau for helping me navigate the many hurdles that needed to be jumped in order to make this functionality possible.

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -54,7 +54,43 @@ default_filters = {
         'get_metadata': filters.get_metadata,
         'convert_pandoc': filters.convert_pandoc,
         'json_dumps': json.dumps,
+        'multiplier' : MyDummyFilterMultiplier,
 }
+
+class Adder:
+
+
+    def __init__(self, adding):
+        self.adding = adding
+
+    def __call__(self, other):
+        return other + self.adding
+
+plusone = Adder(1)
+
+plusone(6) # 7
+
+
+class ConfigurableFilter(Configurable):
+
+    def __init__(self, config=None, parents=None):
+        super().__init__(****)
+
+    def __call__(self, *args):
+        return doyoustuff(*args,)
+
+
+class MyDummyFilterMultiplier(ConfigurableFilter):
+
+    count = Integer(1).tag(config=True)
+
+    def __call__(self, thing:str):
+        return thing*self.count
+
+
+
+
+
 
 
 class ExtensionTolerantLoader(BaseLoader):
@@ -332,6 +368,10 @@ class TemplateExporter(Exporter):
 
         # Add default filters to the Jinja2 environment
         for key, value in self.default_filters():
+            if issubclass(value, ConfigurableFilter):
+                value = value(parent=self) # this is calling __init__
+
+            value('heyho') # this is caling __call__
             self._register_filter(environment, key, value)
 
         # Load user filters.  Overwrite existing filters if need be.
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -22,9 +22,9 @@ from nbconvert import filters
 from .exporter import Exporter
 
 try:
-    from nbconvert.filters.markdown_mistune import Markdown2Html_Mistune as md2html
+    from nbconvert.filters.markdown_mistune import Markdown2Html_Mistune as _md2html
 except ImportError:
-    md2html = filters.markdown2html
+    _md2html = filters.markdown2html
 
 
 #Jinja2 extensions to load.
@@ -32,7 +32,7 @@ JINJA_EXTENSIONS = ['jinja2.ext.loopcontrols']
 
 default_filters = {
         'indent': filters.indent,
-        'markdown2html': md2html,
+        'markdown2html': _md2html,
         'markdown2asciidoc': filters.markdown2asciidoc,
         'ansi2html': filters.ansi2html,
         'filter_data_type': filters.DataTypeFilter,
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -21,12 +21,18 @@ from jinja2 import (
 from nbconvert import filters
 from .exporter import Exporter
 
+try:
+    from nbconvert.filters.markdown_mistune import Markdown2Html_Mistune as md2html
+except ImportError:
+    md2html = filters.markdown2html
+
+
 #Jinja2 extensions to load.
 JINJA_EXTENSIONS = ['jinja2.ext.loopcontrols']
 
 default_filters = {
         'indent': filters.indent,
-        'markdown2html': filters.markdown2html,
+        'markdown2html': md2html,
         'markdown2asciidoc': filters.markdown2asciidoc,
         'ansi2html': filters.ansi2html,
         'filter_data_type': filters.DataTypeFilter,
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -54,43 +54,7 @@ default_filters = {
         'get_metadata': filters.get_metadata,
         'convert_pandoc': filters.convert_pandoc,
         'json_dumps': json.dumps,
-        'multiplier' : MyDummyFilterMultiplier,
 }
-
-class Adder:
-
-
-    def __init__(self, adding):
-        self.adding = adding
-
-    def __call__(self, other):
-        return other + self.adding
-
-plusone = Adder(1)
-
-plusone(6) # 7
-
-
-class ConfigurableFilter(Configurable):
-
-    def __init__(self, config=None, parents=None):
-        super().__init__(****)
-
-    def __call__(self, *args):
-        return doyoustuff(*args,)
-
-
-class MyDummyFilterMultiplier(ConfigurableFilter):
-
-    count = Integer(1).tag(config=True)
-
-    def __call__(self, thing:str):
-        return thing*self.count
-
-
-
-
-
 
 
 class ExtensionTolerantLoader(BaseLoader):
@@ -368,10 +332,6 @@ class TemplateExporter(Exporter):
 
         # Add default filters to the Jinja2 environment
         for key, value in self.default_filters():
-            if issubclass(value, ConfigurableFilter):
-                value = value(parent=self) # this is calling __init__
-
-            value('heyho') # this is caling __call__
             self._register_filter(environment, key, value)
 
         # Load user filters.  Overwrite existing filters if need be.
--- a/nbconvert/filters/markdown.py
+++ b/nbconvert/filters/markdown.py
@@ -8,6 +8,7 @@ markdown within Jinja templates.
 
 from __future__ import print_function
 import re
+import warnings
 
 try:
     from .markdown_mistune import markdown2html_mistune
@@ -16,7 +17,6 @@ except ImportError as e:
     # store in variable for Python 3
     _mistune_import_error = e
 
-    import warnings
     warnings.warn(
             "Conversion to other formats without mistune has been deprecated as of nbconvert 5.2."
             "It has been required for html export since at least nbconvert 4.0.",
@@ -69,7 +69,14 @@ def markdown2html_pandoc(source, extra_args=None):
     """
     Convert a markdown string to HTML via pandoc.
     """
-
+    
+    warnings.warn(
+            "markdown2html_pandoc has been deprecated as of nbconvert 5.2."
+            "Using markdown2html_pandoc will not respect some configuration parameters."
+            "If you wish to continue using pandoc for html conversion, please use"
+            "convert_pandoc(source, 'markdown', 'html', extra_args=extra_args) directly."
+            FutureWarning)
+   
     extra_args = extra_args or ['--mathjax']
     return convert_pandoc(source, 'markdown', 'html', extra_args=extra_args)
 
--- a/nbconvert/filters/markdown.py
+++ b/nbconvert/filters/markdown.py
@@ -11,14 +11,22 @@ import re
 
 try:
     from .markdown_mistune import markdown2html_mistune
+    markdown2html = markdown2html_mistune()
 except ImportError as e:
     # store in variable for Python 3
     _mistune_import_error = e
 
+    import warnings
+    warnings.warn(
+            "Conversion to other formats without mistune has been deprecated as of nbconvert 5.2."
+            "It has been required for html export since at least nbconvert 4.0.",
+            FutureWarning)
     def markdown2html_mistune(source):
         """mistune is unavailable, raise ImportError"""
         raise ImportError("markdown2html requires mistune: %s"
                           % _mistune_import_error)
+
+    markdown2html = markdown2html_mistune
 
 from .pandoc import convert_pandoc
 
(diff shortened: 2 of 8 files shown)


nbconvert/exporters/templateexporter.py
```diff
@@ -21,12 +21,18 @@
 from nbconvert import filters
 from .exporter import Exporter
 
+try:
+    from nbconvert.filters.markdown_mistune import Markdown2Html_Mistune as md2html
+except ImportError:
+    md2html = filters.markdown2html
```
Carreau: make it private, as it's neither a class nor a function.

nbconvert/filters/markdown_mistune.py
```diff
@@ -114,6 +123,15 @@ def latex_environment(self, name, text):
...
+        super(Markdown2Html_Mistune, self).__init__(**kw)
+
+    def __call__(self, source):
+        """Convert a markdown string to HTML using mistune"""
+        renderer = IPythonRenderer(escape=False)
+        # when called appropriately by register filter, it will inherit from template's config
+        renderer.local_config = self.config
+        return MarkdownWithMath(renderer=renderer).render(source)
+
+markdown2html_mistune = Markdown2Html_Mistune
```
Carreau: A) That's API breaking-ish. as you go from `function` to `class`, so likely construct it here, (potentially wrap in a function and deprecate.

nbconvert/filters/markdown.py
```diff
@@ -8,18 +8,26 @@
 
 from __future__ import print_function
 import re
+import warnings
 
 try:
     from .markdown_mistune import markdown2html_mistune
+    markdown2html = markdown2html_mistune()
```
Carreau: B) ... and then no need to instantiate here anymore.

Carreau:
(sorry, I reviewed 2 days ago, but it was not submitted)

nbconvert/filters/markdown_mistune.py
takluyver: It would definitely be an API break as it stands; calling `markdown2html_mistune` no longer does the conversion.

takluyver:
@michaelpacer I've made #522 to illustrate what I meant. Just as I finished that, I saw that it's quite similar to #508 - sorry for not looking at that earlier. I think #508 was ultimately a better starting point for this work than this current PR, so I regret that I didn't see it before you did this.

I imagine my reviewing behaviour is frustrating, so allow me to explain a bit how it works. When I have time to look at issues and PRs, I go through the new ones since my last check. A significant fraction don't need a response from me at all, so they're easily cleared. In other cases I can see what to do next, so I take a minute or two to respond to each as I come to it. The remainder need proper thought and attention, and these I mark as unread to come back to after I've cleared the simple issues. Of course, if I've got other things to do, that means it's the trickier ones that pile up. Unfortunately, your pull requests almost always go into the third category, because I want to take the time to look carefully at them, and you're unlikely to get stuck on some trivial technical issue.

Carreau:
One of the reason we went with trying a Configurable Filter is that the logic in nbconvert already accept it, but we have close to no example as to how to make one.

One of the advantage of configurable filters (over templates), and having example thereof is that it allows pure-python modification only of the filter without touching surrounding Exporters.

nbconvert/filters/strings.py
```diff
@@ -86,27 +88,34 @@ def _convert_header_id(header_contents):
...
+        h.set('id', link)
+        a = ElementTree.Element("a", {"class" : "anchor-link", "href" : "#" + link})
+        a.text = self.anchor_link_text
+        h.append(a)
+        # Known issue of Python3.x, ElementTree.tostring() returns a byte string
+        # instead of a text string.  See issue http://bugs.python.org/issue10942
+        # Workaround is to make sure the bytes are casted to a string.
+        return py3compat.decode(ElementTree.tostring(h), 'utf-8')
+
+add_anchor = AddAnchor
```
michaelpacer: are you imagining that this also would have a deprecation message?

```python
def add_anchor():
    warnings.warn("deprecation message")
    return AddAnchor()
```

and also, since elsewhere we create the object with configuration parameters, would we want to pass those in as optional arguments? 

```python
def add_anchor(local_config=""):
    warnings.warn("deprecation message")
    return AddAnchor(local_config=local_config)
```

In which case do we really want to say that this is deprecated since we could use it instead of calling AddAnchor directly in the IpythonRenderer call?

nbconvert/filters/markdown_mistune.py
michaelpacer: Could you say more about how you would wrap it in a function and deprecate it?

do you just mean

```python
def markdown2html_mistune():
    warnings.warn("deprecation message")
    return Markdown2Html_Mistune()
```
?

willingc:
> Unfortunately, your pull requests almost always go into the third category, because I want to take the time to look carefully at them, and you're unlikely to get stuck on some trivial technical issue.

This comment from @takluyver illustrates well the complexity and details that I face with nbconvert. It's not an easy thing to develop or review. We are incredibly fortunate to have @michaelpacer who is incredibly detail oriented working on the many ways that text can be displayed and manipulated.

What might be helpful for reviewers, or at least for me, would be to summarize the key goals (real-world use problem solved) and key changes in bullet format. The additional detailed prose is valuable too; yet, it helps to have short bullets when I am context shifting from another project back to here. For example,

### Key goals (real world problems solved)

- Enable others to extend an existing nbconvert **filter**
    - add new configurable filter properties
    - retain compatability with current API
- Why?
    - to provide end users a way to customize information that will be passed to an exporter
    - to customize a filter's output so that it can be used by multiple exporters (html, epub, etc)

### Key changes in PR

-
-
- 

I'll leave the key changes for you folks to identify. Also if there are one or two generic functional tests that (independent of design approach) test the desired functionality that would help as well. For example, `test_custom_filter_with_builtin_exporters_output` would test an input with all of the stock exporters. 

One last note to perhaps simplify would be to break out any deprecations that are independent of design approach to their own PRs.

michaelpacer:
## Key goals (real world problems solved)

- Make anchor text postceding the header content a configurable value using a Unicode traitlet (currently it is always ¶)  
    - This allows having custom text (e.g., ★ by using `c.AddAnchor.anchor_link_text = "★"` in your `jupyter_nbconvert_config.py`)
    - This allows having no anchor text (set `c.AddAnchor.anchor_link_text = ""`), closing #412
- Make directly configurable filters possible
    - This allows 3rd party imports of the filters to have access to these configuration options as well.
    - This requires changing the structure of a filter to be a callable class rather than a function.
    - This also requires changing the structure of calls to the filters that occur outside of jinja templates to be inside of classes that can be passed configuration parameters when called.  
- Provide concrete example of how to create a configurable filter class to which we can point people.
    - This case was especially complicated because the filter is not used directly, but rather is used as part of another filter. 
- Not breaking existing API.
- Deprecate old API which will not be able to have access to this functionality.
- Add new filter configuration options to the documentation. 


# Key changes in PR:

- `AddAnchor` is added, a callable class that duplicates the previous behaviour of add_anchor. 
    - Inherits from `LoggingConfigurable` to be able to receive configuration parameters. 
- `Markdown2Html_Mistune` is added, a callable class that applies configuration options and then duplicates previous functionality of `markdown2html_mistune`
    - Inherits from `LoggingConfigurable` to be able to receive configuration parameters.
- Change the filter declaration to be a class rather than a function so that Exporters will pass their configuration arguments.
- The code to generate configuration documentation now includes filters.


@willingc is that along the lines of what you're talking about?


## Re: deprecations

I can remove the deprecation warnings, but the one on importing without mistune is closely tied to the changes made in this PR, since it's a fallback that a proper install should never hit. 

The other isn't as crucial to deprecate (html conversion via pandoc) so I would be more comfortable removing it. I was doing this in this PR because that function is not used anywhere in our code, is more or less a shim around the `convert_pandoc` function, and would not be able to access this configuration option (since it does not use our add_anchor code for html conversion). It is more directed at 3rd parties who are using the `markdown2html_pandoc` function to make it more explicit that they are using a "nonstandard" html conversion mechanism that does not support all the features nbconvert claims to support.

willingc:
### Key changes in PR:

- Adds **AddAnchor**, a callable class
    - Duplicates the previous behaviour of `add_anchor`.
    - Inherits from **LoggingConfigurable** to be able to receive configuration parameters.
- Adds **Markdown2Html_Mistune**, a callable class
    - Applies configuration options
    - Duplicates previous functionality of `markdown2html_mistune`
    - Inherits from `LoggingConfigurable` to be able to receive configuration parameters.
- Changes the filter declaration to be a class rather than a function
    -  Enables *Exporters* to pass their configuration arguments *to ???**.
- Adds code to generate configuration documentation to include filters.

---

#### Key goals and problem statement

- Enable others to extend an existing nbconvert **filter**
    - add new configurable filter properties
    - retain compatibility with current API
- Why?
    - provide end users a way to customize information that will be passed to an exporter
    - customize a filter's output so that it can be used by multiple exporters (html, epub, etc)

#### Design approach

- Expand **anchor text link** values *from* current ¶ *to* no text ("") or any other single character value such as "★"
    - Make anchor text postceding the header content a configurable value using a Unicode traitlet (currently it is always ¶)
    - Allows having custom text (e.g., ★ by using c.AddAnchor.anchor_link_text = "★" in your jupyter_nbconvert_config.py)
    -  Allows having no anchor text (set c.AddAnchor.anchor_link_text = ""), closing #412
- Make directly configurable filters possible
    - This allows 3rd party imports of the filters to have access to these configuration options as well.
- Provide concrete example of how to create a configurable filter class to which we can point people.
    - This case was especially complicated because the filter is not used directly, but rather is used as part of another filter.

#### Major design and code structure impacts of this PR

- Requires changing the structure of a filter to be a callable class rather than a function.
- Requires changing the structure of calls to the filters that occur outside of jinja templates to be inside of classes that can be passed configuration parameters when called.

#### Implementation constraints

- Not breaking existing API.
- Deprecate old API which will not be able to have access to this functionality.
- Add new filter configuration options to the documentation.

#### Deprecations

As referenced [above](https://github.com/jupyter/nbconvert/pull/520#issuecomment-276787316) 

---

@michaelpacer Yes, you are on the right path. Remember that we, other team members reviewing your code, are not as close to the implementation details as you are. One thing that makes a PR effective is its ability to guide the reviewer through the change.

I like to set the context of the PR by succinct bullets on what code has changed (high level). I then follow that with background on:
- real world problem and what is driving the change
- what design approach the PR takes
- major changes in design approach that the PR introduces
- any constraints on implementation
- deprecations
- questions for others (if it is a WIP PR)

I've taken a swipe at reorganizing your content a bit and making the statements more imperative. [Link](https://blogs.atlassian.com/2016/07/written-unwritten-guide-pull-requests/) to an article about PR effectiveness and helping your reviewer be successful. Writing pull requests is hard. Helping your reviewer (who almost always will not know the code specifics as well as you).

Carreau:
Let me try to get my view on #522 (Thomas') vs this one. 

In pure implementation Thomas' Pr is simpler. It add the configuration option on the `Exporter`. The `Exporter` have many `filters`, and the `Exporter` pass down its new configuration option to the required `filter`. 

Thomas PR make things relatively simple on our side, but made a couple of other more complicated for a potential user who wants to extend nbconvert. If you need to extend/configure ina different way you need to provide your own `Exporter` and your own `filter`. Which may be fine, but can be a tiny bit more complicated. I'm unsure how many people would use it though. 

Mike PR is more complicated, in particular as we jump through many hoops not to break backward compatibility. Meaning that a `filter` that was a function should now behave _sometime_ as a class, sometime as a function.

There is (IMHO) a better separation of concern in the approach Mike it taking. If you want to configure a filter, you just need to write a new `filter` and make it configurable. Regardless of what exporter uses it. In particular it now uses a piece of code that was there and most likely not used. The fact that filter can be either `String|function|class|Configurables`  and that it should "just work". 

Though as I guided Mike through most of the PR so I'm definitively biased. I think that Mike (and my) approach make more sens in the long term. 

Still being devil advocate, I don't know how many people are using these advance features of nbconvert and how useful they will be.

takluyver:
> Thomas PR... If you need to extend/configure ina different way you need to provide your own Exporter and your own filter

It remains possible to specify custom filters without custom exporters, using the `TemplateExporter.filters` config option.

My rationale for doing it the way I did basically amounts to: I don't much like our config system, and I'd rather not spread it into every bit of the code. I find code using standard Python mechanisms like parameters and attributes easier to understand and less buggy than code involving traitlets.

willingc:
I spent about an hour this morning looking at both PRs:

- #522 is less complexity from a code standpoint (which I think we all agree on)
- #508 does provide good encapsulation by keeping filter changes within filters. While I like #522 for its simplicity, it limits the ability to change the anchor text link to understand where/how to configure it.

Since I'm having a difficult time choosing one approach over the other, I'm asking myself:
- **how important is it to close #412?** (For example, would we be doing this if this issue was not open? If not, perhaps neither PR is needed at this time.)

takluyver:
It's possibly not all that important, I think it's the only request we've had for that, and we've been putting the pilcrow by headers for a long time. The significance is maybe that markdown conversion is something with a fair bit of flexibility but which can't be controlled by templates, the primary way that we tell people to customise nbconvert behaviour.

takluyver:
A (much) more ambitious proposal would be to parse markdown and then pass at least block-level elements into the template, so you could customise how headers looked by overriding a header block in the template, rather than re-parsing the HTML produced from Markdown.

nbconvert/filters/markdown_mistune.py
```diff
@@ -102,7 +107,11 @@ def block_code(self, code, lang):
 
     def header(self, text, level, raw=None):
         html = super(IPythonRenderer, self).header(text, level, raw=raw)
-        return add_anchor(html)
+        if hasattr(self, 'local_config'):
```
takluyver: Using `hasattr` where the code already knows what the object is is a code smell, in my opinion. In this case, I'd instantiate `IPythonRenderer` with a config object that can be passed down. If we want it to be optional, the default can be an empty Config instance.

nbconvert/filters/markdown.py
```diff
@@ -8,18 +8,24 @@
...
 try:
     from .markdown_mistune import markdown2html_mistune
 except ImportError as e:
     # store in variable for Python 3
     _mistune_import_error = e
 
+    warnings.warn(
+            "Conversion to other formats without mistune has been deprecated as of nbconvert 5.2."
+            "It has been required for html export since at least nbconvert 4.0.",
+            FutureWarning)
```
takluyver: Is it important to get rid of this possibility? As far as I'm aware, we don't need mistune for any other export formats. I don't think it's hurting us to leave this try/except here.

PR #521, opened 2017-01-27
Traceback in cell execution error

mupdt982:
We're executing Notebooks during CI to make sure the content is always up-to-date.

However, if the execution fails, we get the following error message fro nbconvert:

```
[NbConvertApp] Converting notebook /u/matej/Test.ipynb to notebook
[NbConvertApp] Executing notebook with kernel: python2
[NbConvertApp] ERROR | Error while converting '/u/matej/Test.ipynb'
...UNINFORMATIVE INTERNAL STACK TRACE...
File "/u/matej/git/open-source/nbconvert/nbconvert/preprocessors/execute.py", line 113, in preprocess_cell
raise CellExecutionError(msg)
CellExecutionError
[1] 110973 exit 1 jupyter nbconvert --to notebook --stdout --execute ~/Test.ipynb
```

The message only says `CellExecutionError`, which makes it hard to pinpoint the cell and code where things went wrong.

This patch adds a traceback to the error message. It now looks like this:

```
[NbConvertApp] Converting notebook /u/matej/Test.ipynb to notebook
[NbConvertApp] Executing notebook with kernel: python2
[NbConvertApp] ERROR | Error while converting '/u/matej/Test.ipynb'
...UNINFORMATIVE INTERNAL STACK TRACE...
File "/u/matej/git/open-source/nbconvert/nbconvert/preprocessors/execute.py", line 113, in preprocess_cell
raise CellExecutionError(msg)
CellExecutionError: An error occurred while executing the following cell:
------------------
import dakjdnsak
------------------
ImportError: No module named dakjdnsak
[1] 111347 exit 1 jupyter nbconvert --to notebook --stdout --execute ~/Test.ipynb
```

### DIFF
--- a/nbconvert/utils/exceptions.py
+++ b/nbconvert/utils/exceptions.py
@@ -13,6 +13,4 @@
 
 class ConversionException(Exception):
     """An exception raised by the conversion process."""
-    
-    def __init__(self, message):
-        super(ConversionException, self).__init__(message)
+    pass
--- a/nbconvert/utils/exceptions.py
+++ b/nbconvert/utils/exceptions.py
@@ -13,5 +13,6 @@
 
 class ConversionException(Exception):
     """An exception raised by the conversion process."""
-
-    pass
+    
+    def __init__(self, message):
+        super(ConversionException, self).__init__(message)
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -28,6 +28,7 @@ class CellExecutionError(ConversionException):
     failures gracefully.
     """
     def __init__(self, traceback):
+        super(CellExecutionError, self).__init__(traceback)
         self.traceback = traceback
 
     def __str__(self):
--- a/nbconvert/preprocessors/tests/test_execute.py
+++ b/nbconvert/preprocessors/tests/test_execute.py
@@ -1,3 +1,5 @@
+# coding=utf-8
+
 """
 Module with tests for the execute preprocessor.
 """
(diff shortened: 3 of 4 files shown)


michaelpacer:
This looks like useful functionality and a straightforward change. 

I'd prefer to see a test to make sure that we never break this functionality once it's included.

I could imagine testing against the content of a traceback might be an issue… but it seems like that would be the appropriate approach for testing this feature.

takluyver:
I think it should be sufficient to test that the error string is shown. I.e. have a cell:

```python
raise Exception("Can you see me?")
```

Then run that in nbconvert and assert that the exception shows "Can you see me?".

nbconvert/utils/exceptions.py
```diff
@@ -13,5 +13,6 @@
 
 class ConversionException(Exception):
     """An exception raised by the conversion process."""
-
-    pass
\ No newline at end of file
+    
+    def __init__(self, message):
+        super(ConversionException, self).__init__(message)
```
Carreau: Is that necessary ?  Wouldn't message be passed to `Exception` if there is no init ?

mupdt982: Oh, yes, that's right. Will remove.

takluyver:
@mupdt982 it looks like we're waiting on a couple of changes here (test, remove unnecessary `__init__`). Just wanted to check you haven't forgotten about it. :-)

mupdt982:
@takluyver: I've got a fix now, will push soon.

mupdt982:
Please take another look.

- removed the explicit super constructor call,
- added tests (I reused existing notebooks for the tests; I also added a test that verifies utf-8 is handled correctly).

takluyver:
Thanks, that looks good to me.

PR #531, opened 2017-02-14
Add nbconvert.exporters.script entrypoints

takluyver:
As discussed in gh-414, these are looked up by the language name in the notebook metadata. They override the default plain code export when you use `--to script`.

Closes gh-414

### DIFF
--- a/nbconvert/exporters/script.py
+++ b/nbconvert/exporters/script.py
@@ -3,20 +3,37 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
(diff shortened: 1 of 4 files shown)


Carreau:
+1

which milestone ?

michaelpacer:
I really like this. Merging now. 

I'll put it on 5.2? This isn't a bugfix, and it doesn't break backwards compatibility, so it seems appropriate for the next point release.

Carreau:
BTW, if 5.1 is out you can "Edit milestone" set the date and "close" it, so nothing else can be added.

PR #542, opened 2017-03-05
Make the slide exporter a little more configurable

yuvipanda:
- Allow configuring which theme to use for reveal.js
- Allow configuring where to load jQuery / require_js / font-awesome from (same reason as #315) 
- Move default for loading font-awesome to cdnjs to match default for other URLs


michaelpacer:
This lgtm.

PR #547, opened 2017-03-10
Pass --reveal-prefix option to ServePostProcessor

gbrener:
ServePostProcessor has an attribute called reveal_prefix that does not get set when users provide the --reveal-prefix option over the command-line. This is easy to observe, for example, if you execute `jupyter nbconvert --reveal-prefix` with a directory named something other than "reveal.js", or put the reveal.js directory outside of the current one. This PR fixes the issue by passing the --reveal-prefix value along to the ServePostProcessor instance if the option was provided.

### DIFF
--- a/nbconvert/preprocessors/tests/test_clearoutput.py
+++ b/nbconvert/preprocessors/tests/test_clearoutput.py
@@ -12,6 +12,13 @@ from ..clearoutput import ClearOutputPreprocessor
 class TestClearOutput(PreprocessorTestsBase):
     """Contains test functions for clearoutput.py"""
 
+    def build_notebook(self):
+        notebook = super(TestClearOutput, self).build_notebook()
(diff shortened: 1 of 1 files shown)


michaelpacer:
This currently is failing our svg test in the `sanitizehtml` preprocessor which is already happening on jenkins (@takluyver)

@yuvipanda — any thoughts?

mpacer:
i think this needs to be rebased to get the correct pip in travis.

mpacer:
It looks like your rebase wasn't quite correct — did you rebase on the upstream master?

takluyver:
For the record @mpacer , I think that closing and reopening a PR will trigger Travis to rebuild it, testing the merge with the latest master, so it shouldn't be necessary to ask for a rebase for something like this. I might be wrong about that, though.

Issue #557, opened 2017-03-26
Slides broken in master, defaults values are empty


damianavila:
Trying to use slides from master gives me a blank page.
This is cause by empty values for the recently added config options at https://github.com/jupyter/nbconvert/pull/542

```
$ jupyter nbconvert lecture_1.ipynb --to slides --post serve
[NbConvertApp] Converting notebook lecture_1.ipynb to slides
Python 3.6.0 |Continuum Analytics, Inc.| (default, Dec 23 2016, 12:22:00)
Type "copyright", "credits" or "license" for more information.

IPython 5.3.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: resources
Out[1]:
ResourcesDict(None,
              {'config_dir': '/home/damian/.jupyter',
               'metadata': ResourcesDict(None,
                             {'modified_date': 'March 25, 2017',
                              'name': 'lecture_1',
                              'path': ''}),
               'output_extension': '.slides.html',
               'output_files_dir': 'lecture_1_files',
               'reveal': {'font_awesome_url': '',
                'jquery_url': '',
                'require_js_url': '',
                'theme': '',
                'url_prefix': 'reveal.js'},
               'unique_key': 'lecture_1'})
```

It seems putting the `default` value inside the traitlets object is not working... and looking into the traitlets docs, I do not see an option to pass a key:value thing where key is `default`.

I push the following PR: https://github.com/jupyter/nbconvert/pull/558, which is adding the default value a the first parameter to pass and everything seems to be OK now.

@mpacer @takluyver @yuvipanda

PR #558, opened 2017-03-26
Passing default values correctly, otherwise they are all empty.

damianavila:
Fixes https://github.com/jupyter/nbconvert/issues/557


mpacer:
Based on this https://github.com/ipython/traitlets/blob/master/traitlets/traitlets.py#L489 I think it might have needed to be `default_value` rather than `default` but this should work too.

mpacer:
Merging

damianavila:
> Based on this https://github.com/ipython/traitlets/blob/master/traitlets/traitlets.py#L489 I think it might have needed to be default_value rather than default but this should work too.

Nice to know.

> Merging

Thank you!

PR #562, opened 2017-03-30
No ascii only

mpacer:
Now that we use X∃LaTeX this shouldn't be an issue.


takluyver:
Deprecation cycle just in case something outside the project is using it? It may well be that nothing is, but I think deprecating things before removing them is a good habit to get into.

Also, the tests are failing; I don't know if it's related.

Carreau:

PR #563, opened 2017-04-03
support display_id in execute preprocessor

minrk:
Added in protocol v5.1, supported in notebook 5.0.

display_id allows updating existing outputs, so keep track of mapping of display_id to outputs.

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -11,7 +11,7 @@ try:
 except ImportError:
     from Queue import Empty  # Py 2
 
-from traitlets import List, Unicode, Bool, Enum, Any, Type, default
+from traitlets import List, Unicode, Bool, Enum, Any, Type, Dict, default
 
 from nbformat.v4 import output_from_msg
 from .base import Preprocessor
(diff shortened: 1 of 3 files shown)


takluyver:
You've got a merge conflict on this - looks like they were old commits but you just got round to making a PR?

nbconvert/preprocessors/execute.py
```diff
@@ -154,6 +153,8 @@ class ExecutePreprocessor(Preprocessor):
         config=True,
         help='The kernel manager class to use.'
     )
+    
+    _display_id_map = {}
```
takluyver: Can we have a comment here describing the structure of the data stored in this?

minrk: Good call. Added.

takluyver:
Travis has decided to run now, and there's a failure on Python 2, when it tries to start a Python 3 kernel for the new notebook. Have we done something to get round that for the other notebooks?

minrk:
> Have we done something to get round that for the other notebooks?

I looked around and only saw that the test notebooks lack kernel metadata, so launch implicitly with the default kernel. Seems like we should have a better mechanism for that, but I followed the rest in this PR.

takluyver:
This looks good to me, but I'll give @mpacer a chance to look at it as well.

mpacer:
I had looked at this after the initial commits but didn't say anything because I didn't know the context of this. I don't see anything wrong at first glance, but what is this accomplishing in the larger picture (which will help me think about the code in greater depth).

takluyver:
Recent versions of IPython and notebook added a mechanism for updating (replacing, really) an output that has already been displayed. When you display an output, you can give it an ID (or ask IPython to assign it a random ID and return that to you). That ID is sent in the metadata of the output message. If another output message comes later with the same display ID in its metadata, the frontend will replace the previous output with that ID, rather than adding a new one. This can be used to make a progress bar, for instance. (I've deliberately handwaved the details a bit because I don't remember them, but hopefully they're documented in the [message spec](http://jupyter-client.readthedocs.io/en/latest/messaging.html).)

This PR adds support for this when nbconvert is executing a notebook.

mpacer:
kk. LGTM.

PR #566, opened 2017-04-06
Zip postprocessor

mpacer:
Create a zip postprocessor.

I feel like I did something slightly shady with how I managed the loops for the files, but I think it should work as is.

Question: by default should it clean up the files that would have otherwise been created? It feels like it kind of should…

### DIFF
--- a/nbconvert/postprocessors/pack.py
+++ b/nbconvert/postprocessors/pack.py
@@ -16,7 +16,7 @@ class ZipPostProcessor(PostProcessorBase):
         zip_filename = base_name + '.zip'
         zippath = os.path.join(dirname,zip_filename)
         with ZipFile(zippath, mode='w', compression=ZIP_DEFLATED) as zipf:
-            zipf.write(input, arcname = base_name)
+            zipf.write(input, arcname = filename)
             for output_path, _, output_filenames in os.walk(os.path.join(dirname,output_files_dir)):
                 for output_filename in output_filenames: 
                     zipf.write(output_path, arcname = os.path.join(output_files_dir,output_filename))
--- a/nbconvert/postprocessors/pack.py
+++ b/nbconvert/postprocessors/pack.py
@@ -1,0 +1,24 @@
+from __future__ import print_function
+
+from .base import PostProcessorBase
+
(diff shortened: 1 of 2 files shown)


mpacer:
Starting to think that we should have a pack writer that works over byte streams in addition to the pack post-processor, that way we can use it directly and configurably from the notebook server rather than the current `require_zip` function. Also in terms of other packing formats to support I'm thinking just tar for now (maybe tar.gz too?) since that seems to cover the most common formats.

minrk:
> by default should it clean up the files that would have otherwise been created? 

Yes, I think so.

> Also in terms of other packing formats to support I'm thinking just tar for now (maybe tar.gz too?)

I would only add tar.gz (not uncompressed .tar) for now.

nbconvert/postprocessors/pack.py
```diff
@@ -0,0 +1,60 @@
...
+    # need to actually write this and look at the api
+    # def compression_type = DottedStringImport 
+
+    def postprocess(self, input):
+        dirname, filename = os.path.split(input)
+        base_name = os.path.splitext(filename)[0]
+        #  print(self.config.output_files_dir)
+        #  output_files_dir = (self.config.NbConvertApp.output_files_dir).format(notebook_name=base_name)
+        output_files_dir = '{notebook_name}_files'.format(notebook_name=base_name)
+        tar_filename = base_name + '.tar'
```
minrk: When one sees `tar`, one generally means `tar.gz`, so I would assume gzip-compression in this one (`mode='w:gz'`) below. Alternately, you could have a `TarPostProcessor.compression` config trait with choices: `gz`, `bz2`, looking something like:

```python
tar_filename = base_name + '.tar.' + self.compression
...
with TarFile(tarpath, mode='w:%s' % self.compression) as tarf:
    ...
```

mpacer: I was specifically trying to use tar in the browser as that was seeming to play nicer with javascript encodings (still not sure why) for downloading objects.

mpacer:
in light of potentially deprecating postprocessors, i'm going to close this.

PR #567, opened 2017-04-10
get mathjax from cdnjs

minrk:
cdn.mathjax.org is [shutting down](https://www.mathjax.org/cdn-shutting-down/)

Should probably get a backport release. Not sure how far back.


takluyver:
Looks good to me. I'd suggest we just aim to roll up 5.2 fairly soon - I don't think a 5.1.1 is going to get installed any quicker than 5.2 would.

pkra:
Update to 2.7.1, please?

damianavila:
> Update to 2.7.1, please?

PR in progress here: https://github.com/jupyter/nbconvert/pull/584

PR #569, opened 2017-04-17
Remove output metadata in ClearOutputPreprocessor.

tillahoffmann:
This PR removes metadata associated with outputs in the `ClearOutputPreprocessor`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ before_install:
     - git clone --quiet --depth 1 https://github.com/minrk/travis-wheels travis-wheels
 install:
     - wget https://github.com/jgm/pandoc/releases/download/1.19.1/pandoc-1.19.1-1-amd64.deb && sudo dpkg -i pandoc-1.19.1-1-amd64.deb
+    - pip install --upgrade setuptools pip
     - pip install -f travis-wheels/wheelhouse . codecov coverage
     - pip install nbconvert[execute,serve,test]
     - python -m ipykernel.kernelspec --user
--- a/nbconvert/preprocessors/clearoutput.py
+++ b/nbconvert/preprocessors/clearoutput.py
@@ -20,5 +20,5 @@ class ClearOutputPreprocessor(Preprocessor):
             # Remove metadata associated with output
             if 'metadata' in cell:
                 for field in ['collapsed', 'scrolled', 'ExecuteTime']:
-                    del cell.metadata[field]
+                    cell.metadata.pop(field, None)
         return cell, resources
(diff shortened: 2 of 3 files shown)


takluyver:
This seems reasonable, though I hope this doesn't become an ever-expanding list.

minrk:
What sets ExecuteTime?

tillahoffmann:
ExecuteTime is set by the [ExecuteTime](http://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions/execute_time/readme.html) extension. I wasn't sure about which approach to take because updating the preprocessor for each extension seems odd but ExecuteTime seems like a reasonably popular extension.

mpacer:
I'm going to suggest that we make this a configurable traitlet rather than a static list that we are providing, and then individual extensions can determine whether they want to exclude their own metadata from outputs by default or not.

@tillahoffmann if you look at https://github.com/tillahoffmann/nbconvert/blob/c73eec79199584c0663189c28a7c15139f7ec515/nbconvert/preprocessors/extractoutput.py#L38 you'll see an example of this as used by the extract output preprocessor to describe the set of outputs that are to actually be extracted.

Then I'd remove the ExecuteTime entry here and make a PR on `jupyter_contrib_nbextensions` (using an approach similar to that described in the ExecuteTime docs https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/master/src/jupyter_contrib_nbextensions/nbextensions/execute_time/readme.md#options).

mpacer:
You'll need to update the test, also it'd be good to test the traitlet functionality directly (so, for example you would want to have a custom value in a test notebook and configure the traitlet instance in the test to watch for and exclude that value).

tillahoffmann:
Regarding the traitlet test: The `ExecutePrepreprocessor` seems to configure the traitlet by just setting an attribute on the instance (https://github.com/jupyter/nbconvert/blob/master/nbconvert/preprocessors/tests/test_execute.py#L75). Is that what you had in mind or is there a better way to configure the traitlet?

takluyver:
Yep, setting the attribute is the standard way to change traitlet values.

tillahoffmann:
Ah, seems that the tests are not failing because of `pip` but IPython dropped support for 2.7 in version 6: https://travis-ci.org/jupyter/nbconvert/jobs/223900749#L748

takluyver:
Yup, I just merged another pull request into master to use a newer version of pip on Travis, which installs the correct version of IPython.

takluyver:
This looks OK to me, I'll give others a chance to have another look over it too.

mpacer:
This looks good to me now.

PR #571, opened 2017-04-20
Upgrade pip on Travis CI

takluyver:
To avoid it trying to install an incompatible version of IPython on Python 2.


takluyver:
Tests passing; merging to fix tests on master.

PR #572, opened 2017-04-20
Force ipython5 for python2.

tillahoffmann:
Tests fail for python2 because [ipython6 no longer supports python2](http://blog.jupyter.org/2017/04/19/release-of-ipython-6-0/).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,10 +74,10 @@ class FetchCSS(Command):
     user_options = []
     def initialize_options(self):
         pass
-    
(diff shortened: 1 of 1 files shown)


tillahoffmann:
Fixed by pip upgrade.

takluyver:
Thanks :-). I was trying to sneak my PR in before you updated your other one, so that you'd benefit from it automatically, but I wasn't quite quick enough.

PR #573, opened 2017-04-21
Add convenience function for executing notebooks

takluyver:
Using the `ExecutePreprocessor` API is slightly awkward if you want to execute a notebook from a Python script. This provides a more convenient `executenb` function to do it.

Prompted by takluyver/nbparameterise#4

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -334,3 +334,25 @@ class ExecutePreprocessor(Preprocessor):
                 outs.append(out)
 
         return outs
+
+
+def executenb(nb, cwd=None, **kwargs):
+    """Execute a notebook's code, updating outputs within the notebook object.
+    
+    This is a convenient wrapper around ExecutePreprocessor. It returns the
+    modified notebook object.
+    
+    Parameters
+    ----------
+    nb : NotebookNode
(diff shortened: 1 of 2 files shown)


mpacer:
LGTM

mpacer:
Hmm actually do you have any thoughts on how the API would expose the ability to pass configuration values (e.g., other preprocessors that would be run before running nbexecute? I feel like that could also be useful from the perspective of nbparameterize.

mpacer:
If you don't want to make that part of this PR, no worries, and I'll just merge as is.

takluyver:
> how the API would expose the ability to pass configuration values (e.g., other preprocessors that would be run before running nbexecute)?

At least for the case of other preprocessors running before execution, I would suggest running them before calling the function. There's no need to design a new API for running a series of operations in order, because code already defines a series of operations to run in order. ;-)

```python
# Complex
do_y(..., pre_function=do_x)

# Simple
do_x(...)
do_y(...)
```

We may want to provide similar convenience functions for other preprocessors, so they're easy to call from Python code. But I'd take that on a case-by-case basis.

I notice that this branch has a merge conflict, so I'll resolve those.

mpacer:
Fair enough and fair enough. LGTM, merging.

PR #577, opened 2017-04-28
Use resolve_references filter in Latex report template

takluyver:
@ketch alerted me in takluyver/bookbook#5 that we're overriding the markdowncell block in `report.tplx`, but haven't updated it to use the new resolve_references filter. David, do you have time to try this?

CC @mpacer


ketch:
That works correctly for me.  Thanks!

mpacer:
LGTM

PR #583, opened 2017-05-08
Add startup_timeout configurable to execute preprocessor

parente:
The execute preprocessor has a hardcoded 60 second timeout for a kernel to report that it is ready. In a busy environment or underpowered machine, a longer time is sometimes needed. This PR makes it configurable by the user but keeps the default of 60s.

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -212,13 +212,13 @@ class ExecutePreprocessor(Preprocessor):
 
         # from jupyter_client.manager import start_new_kernel
 
-        def start_new_kernel(kernel_name='python', **kwargs):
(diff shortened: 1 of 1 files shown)


nbconvert/preprocessors/execute.py
```diff
@@ -202,14 +212,13 @@ def preprocess(self, nb, resources):
 
         # from jupyter_client.manager import start_new_kernel
 
-        def start_new_kernel(startup_timeout=60, kernel_name='python',
```
takluyver: I think the intention of this function is that at some point it can be replaced with the `jupyter_client` import above. So let's leave this as is, and pass in `startup_timeout` where it's called a few lines below.

parente: Changed as requested.

mpacer:
lgtm merging.

PR #586, opened 2017-05-13
add language cell-magic highlighting to more exports

mpacer:
LaTeX exporter had the ability to inherit cell magics, but nothing was actually being done to enable that in the template because it wasn't passing the cell metadata. 

@Carreau

### DIFF
--- a/nbconvert/templates/asciidoc.tpl
+++ b/nbconvert/templates/asciidoc.tpl
@@ -11,11 +11,11 @@
 {%- endif -%}
 [source
 {%- if 'magics_language' in cell.metadata  -%}
-    {{, cell.metadata.magics_language}}
+   , {{ cell.metadata.magics_language}}
 {%- elif 'pygments_lexer' in nb.metadata.get('language_info', {}) -%}
-    {{, nb.metadata.language_info.pygments_lexer }}
+   , {{ nb.metadata.language_info.pygments_lexer }}
 {%- elif 'name' in nb.metadata.get('language_info', {}) -%}
-    {{, nb.metadata.language_info.name }}
+   , {{ nb.metadata.language_info.name }}
 {%- endif -%}]
 ----
 {{ cell.source}}
--- a/nbconvert/templates/asciidoc.tpl
+++ b/nbconvert/templates/asciidoc.tpl
@@ -9,7 +9,14 @@
     +*In[]:*+
 {%- endif -%}
 {%- endif -%}
-[source{% if nb.metadata.language_info %}, {{ nb.metadata.language_info.name }}{% endif %}]
+[source
+{%- if 'magics_language' in cell.metadata  -%}
+    {{, cell.metadata.magics_language}}
+{%- elif 'pygments_lexer' in nb.metadata.get('language_info', {}) -%}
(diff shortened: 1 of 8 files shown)


nbconvert/templates/asciidoc.tpl
```diff
@@ -9,7 +9,14 @@
     +*In[]:*+
 {%- endif -%}
 {%- endif -%}
-[source{% if nb.metadata.language_info %}, {{ nb.metadata.language_info.name }}{% endif %}]
+[source
+{%- if 'magics_language' in cell.metadata  -%}
+   , {{ cell.metadata.magics_language}}
+{%- elif 'pygments_lexer' in nb.metadata.get('language_info', {}) -%}
+   , {{ nb.metadata.language_info.pygments_lexer }}
```
takluyver: Is asciidoc usually rendered by something using pygments for highlighting? We do this for rst because we know that the main processor of rst, docutils, uses pygments. But if it's likely to be processed by another tool, we shouldn't give special status to the pygments highlighter name.

nbconvert/templates/markdown.tpl
```diff
@@ -8,7 +8,14 @@
 {%- endblock output_prompt %}
 
 {% block input %}
-```{% if nb.metadata.language_info %}{{ nb.metadata.language_info.name }}{% endif %}
+```
+{%- if 'magics_language' in cell.metadata  -%}
+    {{ cell.metadata.magics_language}}
+{%- elif 'pygments_lexer' in nb.metadata.get('language_info', {}) -%}
+    {{ nb.metadata.language_info.pygments_lexer }}
```
takluyver: I don't think markdown is usually rendered by something using pygments for highlighting, so we shouldn't be giving priority to the pygments lexer name here.

takluyver:
Other than those comments, :+1:

mpacer:
Asciidoc seems to be based on pygments as well: http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#source-code

As far as markdown goes, what was previously there was what we use for the main language identity inside the notebook.

My only thought is that pandoc is expecting to see not [pygments](https://bitbucket.org/birkenfeld/pygments-main) but [skylighting](https://github.com/jgm/skylighting) language names, and that github is using [Linguist](https://github.com/github/linguist) and I don't know how compatible those three are. I could imagine markdown being transformed into any of the three… my gut is to go with pygments because we are a python based tool and let people who have issues come to us and we make changes as needed.

mpacer:
Created [this stackexchange question](http://stackoverflow.com/questions/43985848/what-are-the-differences-between-skylighting-pygments-and-linguist) in order to see if there's any community knowledge that we can tap into.

nbconvert/templates/asciidoc.tpl
mpacer: Also, n.b., until this PR we _weren't_ doing this for rst.

mpacer: Yes, asciidoc is usually rendered using pygments for highlighting.

takluyver:
OK, this looks good to me, then.

PR #588, opened 2017-05-16
Add width tags for jpeg and png when present

mscuthbert:
The display(Image(..., retina=True)) command creates metadata about the image explaining the size that it should be displayed at in the notebook.  This information, if present, can be carried into the notebook and converted to rst for use in documenters such as Sphinx.  If not present, retina images appear in .rst (and from there to other formats such as HTML) as twice their intended size.

### DIFF
--- a/nbconvert/templates/rst.tpl
+++ b/nbconvert/templates/rst.tpl
@@ -50,14 +50,14 @@
 
 {% block data_png %}
 .. image:: {{ output.metadata.filenames['image/png'] | urlencode }}
-{%- if output.metadata['image/png'].get('width', '') %}
+{%- if 'image/png' in output.metadata and output.metadata['image/png'].get('width', '') %}
    :width: {{ output.metadata['image/png'].width}}px
 {% endif -%}
 {% endblock data_png %}
 
 {% block data_jpg %}
 .. image:: {{ output.metadata.filenames['image/jpeg'] | urlencode }}
-{%- if output.metadata['image/jpeg'].get('width', '') %}
+{%- if 'image/jpeg' in output.metadata and output.metadata['image/jpeg'].get('width', '') %}
    :width: {{ output.metadata['image/jpeg'].width}}px
 {% endif -%}
 {% endblock data_jpg %}
--- a/nbconvert/templates/rst.tpl
+++ b/nbconvert/templates/rst.tpl
@@ -50,16 +50,26 @@
 
 {% block data_png %}
 .. image:: {{ output.metadata.filenames['image/png'] | urlencode }}
-{%- if 'image/png' in output.metadata and output.metadata['image/png'].get('width', '') %}
-   :width: {{ output.metadata['image/png'].width}}px
-{% endif -%}
+{%- set width=output | get_metadata('width', 'image/png') -%}
+{%- if width is not none %}
+   :width: {{ width }}px
+{%- endif %}
+{%- set height=output | get_metadata('height', 'image/png') -%}
+{%- if height is not none %}
+   :height: {{ height }}px
+{%- endif %}
 {% endblock data_png %}
 
 {% block data_jpg %}
 .. image:: {{ output.metadata.filenames['image/jpeg'] | urlencode }}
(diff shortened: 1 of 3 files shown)


takluyver:
Thanks, this looks good to me. I assume you've tested this fix and it works?

mscuthbert:
Yes, tested with both PNG and JPEG w/ and w/o retina=True flag.

takluyver:
Great :-). I'd like @mpacer to have a quick look over this, but I'm happy for it to go in.

mpacer:
Is `width` special or would `height` do as well? If so why probably it'd help to check and set that as well. 

It looks like we have a `get_metadata` filter that addresses this problem, as you can see in the  [html basic template](https://github.com/jupyter/nbconvert/blob/master/nbconvert/templates/html/basic.tpl#L142).  I think since we have a supported API that we use elsewhere for this purpose and which is tested directly, using that will be more foolproof going forward.

I'd slightly prefer it if we could test that this is adhered to so that this doesn't happen in the future. 
Could you modify the current rst exporter tests to include this check? Here's an [html exporter tests](https://github.com/jupyter/nbconvert/blob/master/nbconvert/exporters/tests/test_html.py#L63) that is basically checking for the same thing. 

An aside: there's no way to update the image metadata in `pngmetadata.ipynb` directly, but it's already there as the file to use in your test.

nbconvert/templates/rst.tpl
```diff
@@ -50,10 +50,16 @@
 
 {% block data_png %}
 .. image:: {{ output.metadata.filenames['image/png'] | urlencode }}
+{%- if 'image/png' in output.metadata and output.metadata['image/png'].get('width', '') %}
```
mpacer: It's cleaner to use the [`get_metadata` filter](https://github.com/jupyter/nbconvert/blob/master/nbconvert/filters/metadata.py#L3). You can find an example of its use for an analogous purpose in the [html template](https://github.com/jupyter/nbconvert/blob/master/nbconvert/templates/html/basic.tpl#L149).

nbconvert/templates/rst.tpl
```diff
@@ -50,10 +50,16 @@
...
+   :width: {{ output.metadata['image/png'].width}}px
+{% endif -%}
```
mpacer: You should also add height information if it is present.

nbconvert/templates/rst.tpl
```diff
@@ -50,10 +50,16 @@
...
 {% endblock data_png %}
 
 {% block data_jpg %}
 .. image:: {{ output.metadata.filenames['image/jpeg'] | urlencode }}
+{%- if 'image/jpeg' in output.metadata and output.metadata['image/jpeg'].get('width', '') %}
```
mpacer: Same as above, use `get_metadata` and add height.

mscuthbert:
Thanks for "teaching me to fish" -- I've made these changes and added a test equivalent to the metadata test in test_html.  I now feel like I understand the rst remplate system better so may be able to make some further improvements in the future.  THANKS! @mpacer

mpacer:
Happy to help — thank _you_ for contributing!

It's good as it stands, but if you want I'm confident that by tweaking the test to be closer to testing the behaviour you actually care about it can be even better. Happy to help with that too.

But that's your call — if you want to call it "done" it's ready to be merged.

mpacer:
@mscuthbert Did you want to keep working on this and learn a bit more or do you want me to merge it as is and make the changes myself?

mpacer:
Also, I'm ok with you leaving the eclipse stuff there for now, but you should know that you could have just added them to [.git/info/exclude](https://help.github.com/articles/ignoring-files/#explicit-repository-excludes).

mscuthbert:
Hi @mpacer -- I'll try to get the changes in there soon.  Thanks for the poke!

mpacer:
@mscuthbert, is it ok if you make a separate PR to improve your test? I'm planning on releasing nbconvert today and I want to include the work you've done already. Does that seem reasonable?

mpacer:
I'm just going to go ahead with that — it isn't too much overhead to make a new PR and then this feature will be available for everyone

PR #593, opened 2017-05-24
Better releases

mpacer:
In the process of making a release, thought I'd add some better documentation (rather than just following a hodgepodge of notebook and ipython release docs, and also wanted to add explicit checks for the version number to avoid nonvalid version numbers that have caused problems in the past.

@Carreau I figure you'll care the most of everyone about this.

Just want to get this settled and then I'll make the next release. I have these thoughts every time I want to make a release and this time I just wanted to deal with some of the loose ends that are always left dangling before making the actual release.

Also, technically we weren't using a pep440 compliant version number before I don't know why it wasn't causing a problem but hey, no harm in doing so just briefly before releasing.


docs/source/development_release.rst
```diff
@@ -0,0 +1,80 @@
...
+Making an ``nbconvert`` release
+===============================
+
+This document guides a contributor through creating a release of ``nbconvert``.
+
+
+Assign all merged PRs to milestones
+-----------------------------------
+
+Go to GitHub and assign all PRs that have been merged to milestones.
```
takluyver: We can provide a handy link here to PRs merged with no milestone:
https://github.com/jupyter/nbconvert/pulls?utf8=%E2%9C%93&q=is%3Amerged%20is%3Apr%20no%3Amilestone%20

docs/source/development_release.rst
```diff
@@ -0,0 +1,80 @@
...
+
+This would ask you for confirmation before removing all untracked files. 
+
+Make sure the ``dist/`` folder is clean and avoid stale builds from
+previous attempts.
+
+Create the release
+------------------
+
+#.  Update the changelog to account for all the PRs assigned to this milestone.
```
takluyver: Should this explicitly mention where the changelog file is?

docs/source/development_release.rst
```diff
@@ -0,0 +1,80 @@
...
+-----------------------
+
+#.  Push directly on master, including --tags separately
+    
+    .. code:: bash
+
+        git push upstream 
+        git push upstream --tags
+
+#.  Push tages on master forgetting to push ``--tags`` too.
```
takluyver: This looks like a confused copy of the point immediately above?

docs/source/development_release.rst
```diff
@@ -0,0 +1,80 @@
+.. _nbconvert_release:
+
+Making an ``nbconvert`` release
```
takluyver: Have you checked a built copy of this? I can't remember if backticks in headings work.

nbconvert/_version_tools.py
```diff
@@ -0,0 +1,83 @@
+# import pep440 # cannot be imported as described below
+import re
```
takluyver: This file looks like it's copied from somewhere - can we have a comment at the top clearly saying where, including what version (or commit, if necessary)? We may also need to include a copyright notice and license text.

setup.py
```diff
@@ -197,13 +197,20 @@ def run(self):
...
+extra_requirements = {
     # FIXME: tests still require nose for some utility calls,
     # but we are running with pytest
-    'test': ['pytest', 'pytest-cov', 'nose', 'ipykernel', 'jupyter_client'],
+    'test': ['pytest', 'pytest-cov', 'nose', 'ipykernel', 'jupyter_client>=4.2'],
     'serve': ['tornado>=4.0'],
     'execute': ['jupyter_client>=4.2'],
 }
+s = [] 
+[s.extend(values) for values in extra_requirements.values()]
```
takluyver: I don't like using list comprehensions for side effects. I'd either do this with a regular for loop:

```python
for reqs in extra_requirements.values():
    s.extend(reqs)
```

Or get clever with `sum`:

```python
s = sum(extra_requirements.values(), [])
```

setup.py
```diff
@@ -197,13 +197,20 @@ def run(self):
...
+extra_requirements['all'] = list(set(s))
+
+extras_require = setuptools_args['extras_require'] = {
+    **extra_requirements
```
takluyver: This syntax is new in Python 3.5. If we need a copy of the dictionary, I think `extra_requirements.copy()` should work. I'm not sure that we do need a copy, though - I think you can just set them to the same object.

docs/source/development_release.rst
mpacer: Yes.

docs/source/development_release.rst
mpacer: yes i think I was in the middle of splitting one thing in the original notebook, but then realised that I could just have the raw code as two lines.

nbconvert/_version_tools.py
mpacer: It's @Carreau's pep440 to which you've committed, we can add a copyright notice if you like. Basically I'd just use it, but for the fact that it can't be built with setuptools.

setup.py
mpacer: I did not know that sum could be applied to lists. I'm intrigued because that could allow: `extra_requirements['all'] = sum(extra_requirements.values(), [])`. Which is perhaps a bit too clever but totally does the trick.

mpacer:
@takluyver do you or @Carreau have any idea about why this makes entrypoints unavailable?

nbconvert/_version_tools.py
takluyver: Yeah, put an explicit Github link and a commit ID, for good measure.

I don't personally mind about the copyright notice & license text, and @Carreau may well not mind either. In general, though, if we're including code from another project, we should include those things - it's a typical condition of the licenses which allow you to redistribute that code.

setup.py
takluyver: Yep, I think that would be OK.

I believe that sum can accept anything which you can combine with the `+` operator. If you're not using it for numbers, though, you need to give the second argument for the starting value.

takluyver:
I don't know what was going on with entrypoints, but it seems to be working now.

setup.py
```diff
@@ -197,13 +197,15 @@ def run(self):
...
+extra_requirements = {
     # FIXME: tests still require nose for some utility calls,
     # but we are running with pytest
-    'test': ['pytest', 'pytest-cov', 'nose', 'ipykernel', 'jupyter_client'],
+    'test': ['pytest', 'pytest-cov', 'nose', 'ipykernel', 'jupyter_client>=4.2'],
     'serve': ['tornado>=4.0'],
     'execute': ['jupyter_client>=4.2'],
 }
+extra_requirements['all'] = sum(extra_requirements.values(), [])
+extras_require = setuptools_args['extras_require'] = extra_requirements
```
takluyver: I don't think we need to have this called both `extra_requirements` and `extras_require`. There's nothing in the code after this using it by either name; its only in `setuptools_args` that it's used.

mpacer: Taken care of.

nbconvert/_version_tools.py
Carreau: You may also just copy past the `is_cannonical` from [pep 440 itself](https://www.python.org/dev/peps/pep-0440/#appendix-b-parsing-version-strings-with-regular-expressions) that was added recently to make this kind of things easier. 

I'll make my pep440 easier to vendor.

mpacer: Got it from pep440 itself Appendix B.

PR #595, opened 2017-05-26
"…to spite the face"

### LINKED ISSUE #592: Actually remove nose dependency
This will take a little while, but I want to actually lose our dependency on nose since we use py.test for almost everything.

mpacer:
closes #592. 

Finally, without any `nose`, scent is less prominent.


takluyver:
LGTM

Issue #78, opened 2015-07-31
Review the scroll functionality in the slideshow


damianavila:
@bollwyvl has raised a lot of concern about this and I agree with most of them, I think we should revisit our previous decision to have scrolling slides or make something like a `long slide` for people interested in that option.

Check @bollwyvl 's points below:

I think the phrase I used before was that they "break the medium"... slides "unintentionally" bigger (i.e. not infinite like prezi, etc.) than the screen...
- are not supported on any other slide platform
- magnify the "i'm going to read you this slide" presentation style... most people can read well north of 300 words per minute in their mother tongue, and showing someone a dense slide, they'll likely tune you out until they've read it
- break URLs: no way to link to a particular scrolled location. not that the reveal ones are great to begin with
- are not liked by reveal, hence all our troubles
  - will likely never printable or exportable, one of the use cases i value a lot but haven't been able to handle succinctly... even without scrolling slides... but at least it is theoretically doable!
- which require a scroll wheel/bar during a talk is distracting to the speaker and the listeners, and content will be in a half-revealed state, which again breaks focus for both parties
  - if the content really is big, like a big diagram or visualization, then temporary full screen/scroll/zoom on that content is a better fit... with text and margins so big, it's hard to even get 80 character code on a screen
- break touch gestures (unless you've fixed that!)

perhaps an explicit longslide slidetype that, if found, would enable the scrolling, while slide would conform to the reveal model would be less surprising.

akhmerov:
For a reference, this is previous discussion of the same topic: ipython/ipython/issues/3589

Personally I completely agree with all @bollwyvl's arguments, and I found the scrolls in presentations to be a huge pain which required a lot of manual tweaking.

yadongli:
I run into the scrollbar problem when migrating my old IPython notebook slides to the new jupyter version, and suddenly my formulas on the slides are surrounded by scrollbars in both x and y directions, I've tried to put the following text into my custom.css, and regenerate the slides.html using nbconvert, but the scroll bars still show up. Any suggestions on how can I remove the scrollbars from the html/reveal slides generated from nbconvert?

.reveal {
  overflow-y: hidden;
}

akhmerov:
@yadongli this issue is about different scrollbars. Your problem is rather related to #38, but different. Do you mind opening a new issue with an example notebook that reproduces the problem? (as far as I can tell, this issue wasn't reported yet)

yadongli:
[...]

PR #600, opened 2017-05-31
Update reveal, fix scrolling and add transition option

damianavila:
* Better per-slide scrolling (defaulting to False but with options to enable it with 
`--SlidesExporter.reveal_scroll=True`)
* Upgrade to latest reveal.js (3.5.0)
* Add option to pass the transition we want.

### DIFF
--- a/nbconvert/exporters/slides.py
+++ b/nbconvert/exporters/slides.py
@@ -114,7 +114,7 @@ class SlidesExporter(HTMLExporter):
 
     reveal_scroll = Bool(False,
         help="""
-        Whether to enable/disable scrolling per slide basis
+        If True, enable scrolling within each slide
         """
     ).tag(config=True)
 
--- a/nbconvert/exporters/slides.py
+++ b/nbconvert/exporters/slides.py
@@ -103,12 +103,12 @@ class SlidesExporter(HTMLExporter):
         """
     ).tag(config=True)
 
-    reveal_transition = Unicode('linear',
+    reveal_transition = Unicode('slide',
         help="""
         Name of the reveal.js transition to use.
 
-        The list of themes that ship by default with reveal.js are:
-        default, cube, page, concave, zoom, none.
+        The list of transitions that ships by default with reveal.js are:
+        none, fade, slide, convex, concave and zoom.
         """
     ).tag(config=True)
 
--- a/nbconvert/exporters/slides.py
+++ b/nbconvert/exporters/slides.py
@@ -6,7 +6,7 @@
 from copy import deepcopy
 from warnings import warn
 
-from traitlets import Unicode, default
+from traitlets import Bool, Unicode, default
 
 from .html import HTMLExporter
 
@@ -103,6 +103,21 @@ class SlidesExporter(HTMLExporter):
         """
     ).tag(config=True)
 
+    reveal_transition = Unicode('linear',
+        help="""
+        Name of the reveal.js transition to use.
+
+        The list of themes that ship by default with reveal.js are:
+        default, cube, page, concave, zoom, none.
+        """
+    ).tag(config=True)
+
+    reveal_scroll = Bool(False,
+        help="""
+        Whether to enable/disable scrolling per slide basis
+        """
+    ).tag(config=True)
+
     require_js_url = Unicode(
         "https://cdnjs.cloudflare.com/ajax/libs/require.js/2.1.10/require.min.js",
         help="""
@@ -146,6 +161,8 @@ class SlidesExporter(HTMLExporter):
             resources['reveal'] = {}
         resources['reveal']['url_prefix'] = self.reveal_url_prefix
         resources['reveal']['theme'] = self.reveal_theme
+        resources['reveal']['transition'] = self.reveal_transition
+        resources['reveal']['scroll'] = self.reveal_scroll
         resources['reveal']['require_js_url'] = self.require_js_url
         resources['reveal']['jquery_url'] = self.jquery_url
         resources['reveal']['font_awesome_url'] = self.font_awesome_url
--- a/nbconvert/templates/html/slides_reveal.tpl
+++ b/nbconvert/templates/html/slides_reveal.tpl
@@ -85,7 +85,6 @@ if( window.location.search.match( /print-pdf/gi ) ) {
 /* Overrides of notebook CSS for static HTML export */
 .reveal {
   font-size: 160%;
-  overflow-y: scroll;
 }
 .reveal pre {
   width: inherit;
@@ -190,8 +189,7 @@ require(
             progress: true,
             history: true,
 
-            theme: Reveal.getQueryHash().theme, // available themes are in /css/theme
-            transition: Reveal.getQueryHash().transition || 'linear', // default/cube/page/concave/zoom/linear/none
+            transition: "{{resources.reveal.transition}}",
 
             // Optional libraries used to extend on reveal.js
             dependencies: [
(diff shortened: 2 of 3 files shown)


nbconvert/exporters/slides.py
```diff
@@ -103,6 +103,21 @@ def _reveal_url_prefix_default(self):
         """
     ).tag(config=True)
 
+    reveal_transition = Unicode('linear',
+        help="""
+        Name of the reveal.js transition to use.
+
+        The list of themes that ship by default with reveal.js are:
+        default, cube, page, concave, zoom, none.
```
takluyver: Should this say 'list of *transitions*'? And shouldn't `linear` be in the list, as it's set as the default?

nbconvert/exporters/slides.py
```diff
@@ -103,6 +103,21 @@ def _reveal_url_prefix_default(self):
...
+        """
+    ).tag(config=True)
+
+    reveal_scroll = Bool(False,
+        help="""
+        Whether to enable/disable scrolling per slide basis
```
takluyver: Probably clearest to say something like:

> If True, enable scrolling within each slide

If this is disabled, does scrolling move between slides? Or just do nothing?

nbconvert/templates/html/slides_reveal.tpl
```diff
@@ -211,13 +209,29 @@ require(
...
+        function setScrollingSlide(scroll) {
+
+            if (scroll === "True") {
+              var h = $('.reveal').height() * 0.95;
+              var hpx = "" + h + "px";
+              $('section.present').find('section')
+                .filter(function() {
+                  return $(this).height() > h;
+                })
+                .css('height', hpx)
```
takluyver: Does it work to use a height like `95%`, or even [CSS calc](https://developer.mozilla.org/en-US/docs/Web/CSS/calc) (`calc(100% - 20px)`), rather than doing calculations in JS?

nbconvert/templates/html/slides_reveal.tpl
```diff
@@ -211,13 +209,29 @@ require(
...
+                .css('overflow-y', 'scroll')
+                .css('margin-top', '20px');
+            }
+        }
+
+        Reveal.addEventListener( 'slidechanged', function( event ) {
+          // check and set the scrolling slide every time the slide change
+          var scroll = "{{resources.reveal.scroll}}";
```
takluyver: Making the boolean into a string is a bit ugly. We could either:

- Have the condition in the template language, `{% if resources.reveal.scroll %}`
- Convert it to a JS boolean `{{ resources.reveal.scroll | json_dumps }}`

nbconvert/exporters/slides.py
damianavila: Yep... copy pasted stuff and forgot to fix it before pushing.

nbconvert/exporters/slides.py
damianavila: >If True, enable scrolling within each slide

I agree.

>If this is disabled, does scrolling move between slides? Or just do nothing?

Just do nothing... scrolling to move between slides can be configured with other reveal.js parameter (`mousewheel` IIRC), but we don't surface that in our template.

nbconvert/templates/html/slides_reveal.tpl
damianavila: I did not try to use CSS calc... will take a look in your suggestions.

nbconvert/templates/html/slides_reveal.tpl
damianavila: > bit ugly

Very ugly... Need to be fixed... I like the `if` approach... I will try that.

nbconvert/exporters/slides.py
damianavila: Fixed and changed the default value to slide (new name for linear transition).

nbconvert/exporters/slides.py
damianavila: Fixed.

nbconvert/templates/html/slides_reveal.tpl
damianavila: > Does it work to use a height like 95%

Nop, it cut the stuff at the very top.

> even CSS calc

Neither... in both cases I get the cut off

slide2

It seems I need to specify the the absolute height in pixel to make it work.

nbconvert/templates/html/slides_reveal.tpl
damianavila: I change my mind and I went with the json_dumps approach (and simplify the structure a little bit).

damianavila:
Push some other minor changes to have a nicer scrollbar and a shared color between several pieces such as controls, progressbar and scrollbar.

damianavila:
Btw, I cancelled the first job because I push to origin by mistake (instead of remote branch).

nbconvert/templates/html/slides_reveal.tpl
takluyver: Gah, CSS. It's probably doing the relative height relative to the wrong parent.

One more thing to try: `calc(100vh - 20px)`. vh units are 1% of the height of the viewport, so 100vh should be the full height available in the window.

nbconvert/templates/html/slides_reveal.tpl
```diff
@@ -149,6 +183,19 @@ a.anchor-link {
 .rendered_html p {
   text-align: inherit;
 }
+::-webkit-scrollbar
+{
+  width: 6px;
+  height: 6px;
+}
```
takluyver: I didn't know CSS can override the scrollbar. Are you sure you want to do that? It seems like the kind of thing that people might get angry about.

mpacer: I don't have strong intuitions around this, what would the downsides be?

nbconvert/templates/html/slides_reveal.tpl
damianavila: Great, `vh` tested and worked. Pushed a commit fixing it. Thanks for letting learn about this CSS calc stuff!

nbconvert/templates/html/slides_reveal.tpl
damianavila: It is overriding stuff on the slideshow... and this scroll is there specifically to scroll the slide.
Btw, this gives me the opportunity to make the scroll closer to other elements... otherwise you get the "system" view which it is not fitting at all the the slide feeling.

nbconvert/templates/html/slides_reveal.tpl
mpacer: This looks along the lines of the kinds of things to get around `'auto'` heights in the case of animations. There should be a way to work natively in pixels rather than treating it as a string manipulation problem though. I forget what that is off the top of my head though.

damianavila: I just pushed a commit using @takluyver suggested approach which get rid of the awkward string manipulation.

damianavila:
What I don't understand is why we need to do this kind of custom scrolling in the first place. What is it about our stuff is breaking the standard scroll?

People using nbconverted slides wanted from the very beginning to have scrolling slides. And we provide that even when reveal.js itself discouraged any use of scrolling inside the slides. As a consequence, in repetitive cases, the scrolling stuff broke. In this PR I am trying to make the scrolling stuff just a configurable thing and if the user really wanted, they can have it.

You have here more stuff on the scrolling discussion if you are interested: https://github.com/jupyter/nbconvert/issues/78

mpacer:
Ok I'll merge.

damianavila:
Thanks @mpacer and @takluyver for you helpful review.

PR #602, opened 2017-06-05
Patch better rst test

mscuthbert:
Tests for actual presence of correctly formatted output rather than
just some metadata. As shown so patiently by @mpacer in #601.

### DIFF
--- a/nbconvert/exporters/tests/test_rst.py
+++ b/nbconvert/exporters/tests/test_rst.py
@@ -4,6 +4,7 @@
 # Distributed under the terms of the Modified BSD License.
 
 import io
+import re
 
 import nbformat
 from nbformat import v4
@@ -61,3 +62,14 @@ class TestRSTExporter(ExportersTestsBase):
         (output, resources) = RSTExporter().from_filename(
             self._get_notebook(nb_name="pngmetadata.ipynb"))
         assert len(output) > 0
+        check_for_png = re.compile(
+            r'.. image:: .*?\n\s+(:.*?: \d+px)\n\s+(:.*?: \d+px)\n\s*\n', 
+            re.MULTILINE)
+        result = check_for_png.search(output)
+        assert result is not None
+        attr_string = result.group(1)
+        assert ':width:' in attr_string
+        assert 'px' in attr_string
+        attr_string = result.group(2)
+        assert ':height:' in attr_string
+        assert 'px' in attr_string
--- a/nbconvert/exporters/tests/test_rst.py
+++ b/nbconvert/exporters/tests/test_rst.py
@@ -63,8 +63,8 @@ class TestRSTExporter(ExportersTestsBase):
             self._get_notebook(nb_name="pngmetadata.ipynb"))
         assert len(output) > 0
         check_for_png = re.compile(
-            r'.. image::.*?\n\s+(?s:(.*?))\n\s*\n', 
-            re.MULTILINE)
+            r'.. image::.*?\n\s+(.*?)\n\s*\n', 
+            re.DOTALL)
         result = check_for_png.search(output)
         assert result is not None
         attr_string = result.group(1)
--- a/nbconvert/exporters/tests/test_rst.py
+++ b/nbconvert/exporters/tests/test_rst.py
@@ -63,13 +63,11 @@ class TestRSTExporter(ExportersTestsBase):
             self._get_notebook(nb_name="pngmetadata.ipynb"))
         assert len(output) > 0
         check_for_png = re.compile(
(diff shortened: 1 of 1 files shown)


mpacer:
Hey, this is looking great! This should definitely work for this particular case, but in the vein of more robust testing, it seems that where what we had before was expecting too little in the way of specifics, this is too much in the way of specifics. I.e., there are things we might add to the rst image metadata someday and if it's possible we should write the test so that it fails only if what we're adding actually is invalid and breaks the behaviour of setting the image metadata. 

Some thoughts with this : 

1. since you are explicitly using and checking the content of result, you probably don't need to explicitly test that it's not None (it's not an expensive operation but I thought I might as well point that out)
2. won't this fail if it has a non-numerical metadata value before the `:width:` tag? 
3. is it only valid to specify the width and height in pixels?
4. Won't this fail if you specify the height first and the width second?
    * one thought would be to grab everything with the pattern `:.*?: \w+?\n` using a non-capturing grouping operation `(?:)` to apply a `*` to the whole bunch followed an arbitrary amount of spaces and another newline (so something like `((?::.*: \w+?\n)*?)\s*\n`) that way all of the metadata objects that you might want are captured in a single group.
5. You probably don't need to assign the value of `result.group(x)` to `attr_string`, you can just check directly `assert ":width:" in result.group(1)` (I know I did… but you don't *have* to)
    * alternatively, it may be somewhat clearer if the two groups did have different meaning to assign them specifically and assert that the elements are present (e.g., `width_string = result.group(1)` `height_string = result.group(2)`) but in this case they don't have different syntax/semantics so it may not be clearer to do that

NB: Since we're using this as an exploration of the codebase and testing converted documents, I'm being trying to give a bit more care than you would want to give to the average test. If you want me to ask for less, I'm happy to oblige.

mscuthbert:
Ah, okay -- no problem with patching it.  In music21 we tend to do a fail-fast philosophy where if more metadata classes or something are added later we want the test to fail to make sure that this is reflected in the tests.  I also tend to do very general tests first, like checking that output exists (since checking for None is basically a zero-time operation) so that I know immediately how bad the failure is (was there no output whatsoever or was it in an incorrect format).

For PNG testing, I think that px is going to be the only valid format, but that's easily removed.

mpacer:
And that's probably a good point about the cheapness of checking for None & the use of general tests to know how bad a test is. 

I think it's fair to have width be required to be a pixel value for now. Technically [rst](http://docutils.sourceforge.net/docs/ref/rst/directives.html#image) allows a percentage as the :width: value, where that implies it's the percentage of the line width inside of which the image is being displayed. But nothing's using that right now including the IPython display mechanism, so I think it's fine not to code for that now.

mscuthbert:
Great so I think it should be good to go if you think so.

mpacer:
I've been trying to figure out why that error is occurring for py<=3.5 by looking at the source, docs, and issues and I'm a little stumped. I'll dive into it in a bit.

mscuthbert:
closing and reopening to rerun the tests and see if it was a problem with this patch or something on the nbconvert master at the time.

mpacer:
Ok, there's one part of the regex that I don't follow:

1. What is `.*?` accomplishing? 

And one part that I can identify as causing the error: you're actually using a python 3.6 only feature

>(?imsx-imsx:...)
> (Zero or more letters from the set 'i', 'm', 's', 'x', optionally followed by '-' followed by one or more letters from the same set.) The letters set or removes the corresponding flags: re.I (ignore case), re.M (multi-line), re.S (dot matches all), and re.X (verbose), for the part of the expression. (The flags are described in Module Contents.)
>
> New in version 3.6.

Which explains why it only works in 3.6 :P

Unfortunately, it means that we can't use this feature as nbconvert will need to be compatible with py2 for now.

mscuthbert:
Ah!  I didn't see that -- I'll get a patch in that doesn't use that fancy Py3.6 feature soon as I can.

mpacer:
An aside: I didn't know that that _was_ a feature, so thank you for giving me one more reason to love being on the latest python.

mpacer:
Sounds good!

mscuthbert:
That and having the Regexp flags available as an IntEnum -- this will be great when 3.6 can be assumed a few years from now.

mscuthbert:
python3.4 test failure because out of sockets -- try second time...

mscuthbert:
The new Regex seems to work on all versions -- the first test had a py3.4 failure which had to do with being out of sockets -- this seems to be a test system failure; it did not occur in rst conversion, and a second run of the tests worked on all systems.

mpacer:
w00t, then I'll merge it! Thank you!

PR #604, opened 2017-06-12
Fixed AttributeError when getting slide_type

Happyholic1203:
To reproduce this AttributeError:
1. Run jupyterlab: `jupyter lab`
2. Create an ipynb
3. Edit a cell and choose **Cell Tools** from the leftmost panel
4. Set the cell to **slide** type

Now, the metadata looks like this:
```json
{
  "collapsed": true,
  "slideshow": "slide"
}
```

Now, it's obvious that the original code would raise an `AttributeError`.

I got this all the time when using jupyterlab.
Maybe this should be fixed in jupyterlab, but I think adding a safe guard can't be harmful.
After all, you never know maybe one day it can become `"slideshow": None`

### DIFF
--- a/nbconvert/exporters/slides.py
+++ b/nbconvert/exporters/slides.py
@@ -16,7 +16,12 @@ def prepare(nb):
 
     for cell in nb.cells:
         # Make sure every cell has a slide_type
-        cell.metadata.slide_type = cell.metadata.get('slideshow', {}).get('slide_type', '-')
+        try:
+            slide_type = cell.metadata.get('slideshow', {}).get(
+                'slide_type', '-')
+        except AttributeError:
+            slide_type = '-'
+        cell.metadata.slide_type = slide_type
 
     # Find the first visible cell
     for index, cell in enumerate(nb.cells):


takluyver:
Ping @damianavila for slideshow metadata coordination :-)

damianavila:
@Happyholic1203 @takluyver I think this should be fixed in JupyterLab itself... @Happyholic1203 did you open an issue about this in that repo?

mpacer:
So I'm hesitant to remove this Error. Or rather we may want to instead of "fixing" this, raise an error that explains that the users' notebook's metadata seems to be incorrectly specified. Since that's really the problem, not that the attribute is missing.

damianavila:
> raise an error that explains that the users' notebook's metadata seems to be incorrectly specified

Sounds like a good idea.

SylvainCorlay:
@Happyholic1203 I would be happy to include this fix (and the suggestions), but the things have moved around quite a bit since you opened your PR - and there are conflicts. Would you be ok with opening an updated PR to master?

MSeal:
Closing as #1329 has been updated for 6.0 alpha

PR #609, opened 2017-06-23
Improve, normalise mathjax handling, yapf the file

### LINKED ISSUE #544: Same bug as notebook/#759 regarding correctly delimiting mathjax
This needs a solution similar to https://github.com/jupyter/notebook/pull/2220.

mpacer:
Closes #544.


mpacer:
Does anyone have any issues with this PR? Or can I merge?

takluyver:
For the future, I'd suggest keeping functional changes separate from code formatting, because it's harder to review changes to the code if they're mixed with changes to the style. As this is already here, though, go for it.

(We've also mostly discouraged large formatting changes, because the changes make things like git blame less useful. But as you're doing most of the work on nbconvert at the moment, I think you get to decide things like that)

bollwyvl:
@mpacer will take a look at this!

mpacer:
This is not likely to be problematic. It just aims to match the notebook in terms of its functionality; but always happy to have more review.

PR #611, opened 2017-06-23
Transforms all mistune mathjax logic to be inline

mpacer:
This is substantially less complicated in terms of the underlying logic doing the parsing. 

That said, I'm not sure we're testing all of the ways that this could cause things to break. But, if this did cause breakage, my guess is it would teach us what tests we _should_ have.

Also, it avoids the inelegance of the previous simultaneous inline-block-math and block-block-math approach.

Worth mentioning, if you look at the output text and had two block-block-math blocks back to back
e.g.,

```

$$a + b$$

\\[a +b\\]

```

Way mistune does preprocessing would eat the whitespace between them and transform it into

```
"<p>$$a+b$$$$a+b$$<\p>"
```

which would still be parsed correctly from mathjax's perspective, but it's more than a little weird if you look at the raw text directly.


mpacer:
Note while this deals with markdown & mathjax, it doesn't solve the probem for LaTeX, that should probably be a separate PR.

mpacer:
NB: #617 deals with LaTeX output.

mpacer:
@bollwyvl this is the one that i actually meant but the other one too!

PR #616, opened 2017-06-28
Add remove regex default to TemplateExporter, don't remove cells with outputs but no source

### LINKED ISSUE #615: RemoveRegexPreprocessor removes cells with no source, even if they have output
If you have a cell that has no input but it does have output content, it is removed upon export.

My intuition about this feature was that this was supposed to remove empty cells not cells whose input is empty.

I was going to propose making this a part of TemplateExporter, but I think we first should decide whether we want the behaviour that currently exists or something more nuanced.

I recognise, it's not a great practice to end up with a notebook in this state. But if anyone were to do it, I don't think nbconvert by default should be ripping out their outputs.

Really the context in which I could imagine someone doing this would be as a hacky way to hide input, which should be implemented separately. But my intuition would be that the effectively desired behaviour of the hacky practice would be to apply the "hide input" to these empty cells w/ output (rather than removing them entirely). So implementing that version would require solving the problem of hiding input on a per-cell basis (which will likely be closely tied to tag-based filtering).

mpacer:
Closes #615. 

This enables RegexRemovePreprocessor in the TemplateExporter. 

It also changes its behaviour to not remove cells that _do_ have output even if they match the regular expression (which, by default, is the empty string).

@tillahoffmann is this consistent with your original intended use of this preprocessor (keeping in mind that tag-based input filtering should also be possible soon)?

### DIFF
--- a/nbconvert/preprocessors/regexremove.py
+++ b/nbconvert/preprocessors/regexremove.py
@@ -55,11 +55,8 @@ class RegexRemovePreprocessor(Preprocessor):
         pattern = re.compile('|'.join('(?:%s)' % pattern
                              for pattern in self.patterns))
 
-        # check that output is not present if it is a code cell
-        code_w_output = cell.cell_type == 'code' and cell.get('outputs',[]) != []
-
-        # Filter out cells that meet the pattern and have no output
-        return code_w_output or not pattern.match(cell.source)
+        # Filter out cells that meet the pattern and have no outputs
+        return cell.get('outputs') or not pattern.match(cell.source)
 
     def preprocess(self, nb, resources):
         """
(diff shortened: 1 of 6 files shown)


nbconvert/preprocessors/regexremove.py
```diff
@@ -39,19 +40,36 @@ class RegexRemovePreprocessor(Preprocessor):
 
     patterns = List(Unicode, default_value=[r'\Z']).tag(config=True)
 
-    def preprocess(self, nb, resources):
+    def check_conditions(self, cell):
         """
-        Preprocessing to apply to each notebook. See base.py for details.
+        Checks that a cell matches the pattern and that (if a code cell)
```
tillahoffmann: The comment here doesn't quite match the code below: The function checks that the cell does _not_ match the pattern _or_ is a code cell with output.

As an aside, are there any other cell types that can have output? We might just be able to check `not pattern.match(cell.source) or cell.get('outputs')`. Then `cell.get('outputs')` will evaluate to false if the key doesn't exist or the list is empty.

mpacer: No other cell types have outputs, so we should be in the clear.

PR #617, opened 2017-06-29
use pandoc's tex_math_double_backslash extension

mpacer:
This addresses issues related to how we process `\\(\\)` and `\\[\\]` for LaTeX and PDF output.

I think this closes an outsized number of issues given the number of lines that needed to be changed, but I'd need to track down those individually later.


minrk:
Sounds good!

PR #618, opened 2017-06-29
add traitlet to ServePostProcessor that configures the browser used

### LINKED ISSUE #614: Choose webbroswer
There is no way to choose `webbroswer` in the config files for `nbconvert`'s serve post processor. It should be selectable, just like `jupyter notebook` (see https://github.com/jupyter/notebook/blob/master/notebook/notebookapp.py). This would also help mitigate the jupyter/notebook#2438 issue for `nbconvert`, as well. This would be set in https://github.com/jupyter/nbconvert/blob/master/nbconvert/postprocessors/serve.py.

In `jupyter_nbconvert_config.py` there should be a setting such as `c.ServePostProcessor.browser`.

mpacer:
This closes #614. 

I took some of the design of how the notebook implemented this and tried to abstract away only those parts that we needed. 

In particular, I didn't run the `browser.open` command using `threading.thread`, since multi-threading didn't seem necessary, but that might be wrong. Could use others' insights.

### DIFF
--- a/nbconvert/postprocessors/serve.py
+++ b/nbconvert/postprocessors/serve.py
@@ -47,17 +47,21 @@ class ServePostProcessor(PostProcessorBase):
     ).tag(config=True)
 
     browser = Unicode(u'', 
-                      help="""Specify what command to use awhen opening a browser.
-                      If not specified, the default browser will be determined 
-                      by the `webbrowser` standard library module, which allows 
-                      setting of the BROWSER environment variable to override it.
+                      help="""Specify what browser should be used to open slides. See
+                      https://docs.python.org/3/library/webbrowser.html#webbrowser.register
+                      to see how keys are mapped to browser executables. If 
+                      not specified, the default browser will be determined 
+                      by the `webbrowser` 
+                      standard library module, which allows setting of the BROWSER 
+                      environment variable to override it.
                       """).tag(config=True)
 
     reveal_cdn = Unicode("https://cdnjs.cloudflare.com/ajax/libs/reveal.js/3.5.0",
-        help="""URL for reveal.js CDN."""
-    ).tag(config=True)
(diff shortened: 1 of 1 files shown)


takluyver:
I think that wrapping it in a thread is needed in some situations, because `browser.open()` might try to wait for the process it launches to exit before returning. Looking at the code of the webbrowser module, I think this would affect you if you set the BROWSER env variable and didn't already have the browser running when Python tried to launch it.

minrk:
I believe it's only on Linux where `open` can block, and there only when the browser application is not already running, as Thomas mentioned.

takluyver:
I think it can happen on any system if you have the BROWSER environment variable set; that's registered as a `GenericBrowser` regardless of your platform. It's a corner case, but I suspect enough people will run into it to be worth considering - especially because the failure mode will be confusing: the browser will start and the server won't be running.

nbconvert/postprocessors/serve.py
```diff
@@ -84,7 +97,14 @@ def postprocess(self, input):
         print("Serving your slides at %s" % url)
         print("Use Control-C to stop this server")
         if self.open_in_browser:
-            webbrowser.open(url, new=2)
+            try: 
+                browser = webbrowser.get(self.browser or None)
+                threading.Thread(target=(browser.open(url, new=2))).start()
```
takluyver: This won't actually do it in a thread - you're calling `browser.open()` in the main thread, and passing its return value as the `target=` parameter.

You can either use a lambda, like we [do in notebook](https://github.com/jupyter/notebook/blob/5.0.0/notebook/notebookapp.py#L1398), or pass args and kwargs separately, like this:

```python
Thread(target=browser.open, args=(url,), kwargs={'new': 2})
```

Both ways are a bit ugly, unfortunately.

PR #619, opened 2017-07-01
Add clear-output as a convenience flag per @rjleveque's in-person suggestion

mpacer:
Talking with @fperez and @rjleveque today pointed out that nbconvert should be able to apply the ClearOutputPreprocessor easily on a file. However doing so is unnecessarily verbose, e.g.:

```
jupyter nbconvert --ClearOutputPreprocessor.enabled=True --inplace example.ipynb
```

This creates a flag that makes this much more concise.
```
jupyter nbconvert --clear-output example.ipynb
```


mpacer:
Does anyone have an objection to this?

takluyver:
Looks good to me. Don't forget to mention it in the release notes :-)

PR #631, opened 2017-07-25
Normalise <ipython-input-...> in traceback in test

takluyver:
The hashes have changed in IPython master, which is causing a test failure on Jenkins.


mpacer:
could you rebase on master? this seems to conflict

takluyver:
Thanks for taking care of the conflict - I thought I pulled from master before making the branch, but maybe I forgot.

Issue #621, opened 2017-07-07
error when converting to pdf/reST 


Constatza:
When trying to download the file  as .pdf or .rst the following error is raised:

```
500 : Internal Server Error
The error was:
nbconvert failed: 'charmap' codec can't decode byte 0x9a in position 127: 
character maps to <undefined>
```
I am running Windows and have folders with Greek Alphabet, could this cause the problem?

mpacer:
That is likely part of the problem. If you try to convert on the command line does it work?

Constatza:
Yeah, tried once the same issue.

takluyver:
If you're not already, we recommend using python 3 if you have non ASCII
file names, especially on Windows.

Constatza:
That won't work either, I use Anaconda with python 3.6.

Constatza:
Initially it raised an error because I hadn't installed pandoc, but since I installed it it raises the same error.

mpacer:
Does it work on *any* of the exporters? E.g., html, markdown, asciidoc, notebook, python…

or is it broken on every one of them?

mpacer:
Basically, if it works on at least one of them, we can say confidently that it's an issue with the particular exporters, and if it works on none of them, it's likely to be an issue with nbconvertapp & how it reads file names.

Constatza:
It works for every exporter except for the .pdf and .rst

Constatza:
The home directory of jupyter is : C\users\Kωνσταντίνος . Should I try to change the user's name? I wouldn't like to try that, unless you guys are more or less sure that the folder's name causes the issue.

Constatza:
Apparently, it seems the error is caused by the notebook's content. I just tried to convert a notebook with no greek markdown text as title and it worked fine. Hence the greek titles are the problem.
So, do you any suggestion to overcome this?

takluyver:
Can you show the whole of the traceback you get when you do it on the command line with the greek text? That should show us where in the code the error comes from.

Constatza:
```
C:\Users\Κωνσταντίνος\Google Drive\PythonNotebooks>jupyter nbconvert --to pdf olives.ipynb
[NbConvertApp] Converting notebook olives.ipynb to pdf

Traceback (most recent call last):
  File "C:\ProgramData\Anaconda3\Scripts\jupyter-nbconvert-script.py", line 11, in <module>
[...]

PR #638, opened 2017-08-03
Fix getting pandoc version number

takluyver:
Allowing for non-ascii home directory on Windows. We only need to decode the first line, not the whole output.

Closes gh-621

### DIFF
--- a/nbconvert/utils/pandoc.py
+++ b/nbconvert/utils/pandoc.py
@@ -75,11 +75,10 @@ def get_pandoc_version():
         if not which('pandoc'):
             raise PandocMissing()
 
-        out = subprocess.check_output(['pandoc', '-v'],
-                                      universal_newlines=True)
+        out = subprocess.check_output(['pandoc', '-v'])
         out_lines = out.splitlines()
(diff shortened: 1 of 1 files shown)


mpacer:
Why did you get rid of `universal_newlines=True`? Just curious.

mpacer:
Is the issue that we just need to get it to be in the defaultlocale so that we can use that information to implicitly decode the string to `ascii`?

takluyver:
`universal_newlines` in subprocess actually means 'decode the output from this process using the locale default encoding'. It's presumably like that for historical reasons, because it makes absolutely no sense now.

The issue in #621 was that pandoc produces output that's not in the locale encoding. It may be using UTF-8 in all cases, but I don't think that really matters. The version number which we're trying to get is probably always going to be made up of ascii characters.

So the change here tells subprocess not to decode the output, so we get bytes back from it. Then we take the first line, and decode that ourselves to get the version number. The [bytes.splitlines()](https://docs.python.org/3/library/stdtypes.html#bytes.splitlines) method already copes with the different newline conventions, so we don't need subprocess to handle that.

mpacer:
Got it!

PR #639, opened 2017-08-05
Specify default template files with extension

takluyver:
This should make issues like ipython/ipython#10731 less likely. At present, if you have a file in the working directory called `python` or `rst` or `article` (and so on) with no extension, nbconvert will try to use that as the template. Honestly, I'm surprised this hasn't come up more.

With this change, it will only look for the default template files with their extension. So you can still override it with a file called `python.tpl`, for instance. But that's less likely to be there accidentally than `python`.

### DIFF
--- a/nbconvert/exporters/html.py
+++ b/nbconvert/exporters/html.py
@@ -36,7 +36,7 @@ class HTMLExporter(TemplateExporter):
 
     @default('template_file')
     def _template_file_default(self):
-        return 'full'
+        return 'full.tpl'
     
     output_mimetype = 'text/html'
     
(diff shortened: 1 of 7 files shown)


mpacer:
Merging, but should we go through the trouble to also check that a .tpl is actually given in all cases (perhaps excepting CLI's)?

takluyver:
IIRC we made the extension required at one point and broke some people's code. We could go through a proper deprecation cycle to make it required, but I don't think it's worth it. Our config system also makes it difficult to treat command line options differently from options in config files or third party Python code.

PR #640, opened 2017-08-07
Tag remove preprocessor

mpacer:
This PR implements tag based element filtering using a Preprocessor to remove entire cells, entire output areas, and individual outputs.

Cell removal is enabled by adding a tag to `tags` in the cell's `metadata` and then adding the tag to `TagRemovePreprocessor.remove_cell_tags`.

Output area removal is enabled by adding a tag to `tags` in the cell's `metadata` and then adding the tag to `TagRemovePreprocessor.remove_all_outputs_tags`.

Individual output removal is enabled by adding a tag to a `tags` field created in an output's `metadata`, and then adding the tag to `TagRemovePreprocessor.remove_single_output_tags`.

I have an idea talked through with @Carreau on how to additionally filter inputs (which would create invalid notebook objects), but that can't be done with a preprocessor (since preprocessors can only produce valid notebooks). Therefore, I'll leave that to another PR.

### DIFF
--- a/nbconvert/preprocessors/tagremove.py
+++ b/nbconvert/preprocessors/tagremove.py
@@ -38,7 +38,7 @@ class TagRemovePreprocessor(ClearOutputPreprocessor):
         """
 
         # Return true if any of the tags in the cell are removable.
-        return not any([tag in cell.get('tags', [])
+        return not any([tag in cell.get('metadata', {}).get('tags', [])
                         for tag in self.remove_cell_tags])
 
     def preprocess(self, nb, resources):
@@ -63,15 +63,16 @@ class TagRemovePreprocessor(ClearOutputPreprocessor):
         Apply a transformation on each cell. See base.py for details.
         """
 
-        if any([tag in cell.get('tags', [])
-                for tag in self.remove_all_output_tags]):
+        if (any([tag in cell.get('metadata', {}).get('tags',[])
+                for tag in self.remove_all_output_tags])
+            and cell.cell_type == 'code'):
             cell.outputs = []
             cell.execution_count = None
             # Remove metadata associated with output
             if 'metadata' in cell:
                 for field in self.remove_metadata_fields:
                     cell.metadata.pop(field, None)
-        if cell.outputs:
+        if cell.get('outputs', {}):
             cell.outputs = [self.preprocess_output(output,
                                                    resources,
                                                    cell_index,
@@ -84,7 +85,7 @@ class TagRemovePreprocessor(ClearOutputPreprocessor):
                             ]
         return cell, resources
 
-    def check_output_conditions(self, output, resources, 
+    def check_output_conditions(self, output, resources,
                                 cell_index, output_index):
         """
         Checks that an output has a tag that indicates removal.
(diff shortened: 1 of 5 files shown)


nbconvert/preprocessors/tagremove.py
```diff
@@ -0,0 +1,101 @@
...
+    remove_all_output_tags: removes entire output areas on cells
+                            tagged with these values
+    remove_single_output_tags: removes individual output objects on
+                               outputs tagged with these values
+
+    """
+
+    remove_cell_tags = List(Unicode, default_value=[]).tag(config=True)
+    remove_all_outputs_tags = List(Unicode, default_value=[]).tag(config=True)
+    remove_single_output_tags = List(Unicode, default_value=[]).tag(config=True)
```
takluyver: Since the order doesn't matter here, I think you could make them sets instead of lists. That means we can use set operations to check if any tag matches:

```python
self.remove_cell_tags.intersection(cell.get('metadata', {}).get('tags', []))
```

nbconvert/preprocessors/tagremove.py
```diff
@@ -0,0 +1,101 @@
...
+        if (any([tag in cell.get('metadata', {}).get('tags', [])
+                 for tag in self.remove_all_outputs_tags])
+            and cell.cell_type == 'code'):
+            cell.outputs = []
+            cell.execution_count = None
+            # Remove metadata associated with output
+            if 'metadata' in cell:
+                for field in self.remove_metadata_fields:
+                    cell.metadata.pop(field, None)
+        if cell.get('outputs', {}):
```
takluyver: It doesn't really matter to the code, but for the sake of clarity, I'd make the fallback value a list rather than a dict. Or maybe even `None`.

nbconvert/preprocessors/tagremove.py
```diff
@@ -0,0 +1,101 @@
...
+
+        Returns: Boolean.
+        True means output should *not* be removed.
+        """
+        return not any([tag in output.get('metadata', {}).get('tags', [])
+                        for tag in self.remove_single_output_tags])
+
+    def preprocess_output(self, output, resources,
+                          cell_index, output_index):
+        return output, resources
```
takluyver: Are you planning some future extension here, or can this method be discarded?

takluyver:
This basically looks good, I've just noted a few places where I think the code can be clearer/simpler. :-)

nbconvert/preprocessors/tagremove.py
mpacer: Done.

nbconvert/preprocessors/tagremove.py
mpacer: Done

nbconvert/preprocessors/tagremove.py
mpacer: I was partially trying to give people a handle on preprocessing individual outputs if they wanted to subclass (I thought it might be helpful for things like tag based preprocessing… but I'm realising now that something like that is more broad ranging and should probably be put into a different PR).

mpacer:
Should we be considering a major release soon? I want to add a new class to handle the input version of this, so maybe we should have it target 6.0?

takluyver:
Historically with IPython we did feature releases like '6.0' and bugfix rollups like '6.1', but since the big split we've tended more towards semver style versioning, so we increment the major version to indicate breaking changes. To some extent this is up to the people working on each individual project, though.

I'm not the biggest fan of semver - I think it ignores the complexity of real development, where almost every change can [break something](https://xkcd.com/1172/) - but I know that when we're aiming for a new .0 release, it tends to become a Big Deal, and everyone wants their thing backported to the previous series. So I'm inclined to leave 6.0 until we actually want to break something.

mpacer:
K @takluyver would you say this is good to go?

mpacer:
@Carreau @minrk I'll merge if either of you look it over and give a 👍.

PR #650, opened 2017-08-18
normalize base64 output data in tests

minrk:
checks for base64 the lazy way, by trying to decode it first.

avoids spurious failure when equivalent base64 data appears changed due to changes in presence of newlines, etc.

cc @takluyver this is the only failure on jack-of-none: https://travis-ci.org/jupyter/test-grid/jobs/266020642

### DIFF
--- a/nbconvert/preprocessors/tests/test_execute.py
+++ b/nbconvert/preprocessors/tests/test_execute.py
@@ -7,6 +7,7 @@ Module with tests for the execute preprocessor.
 # Copyright (c) IPython Development Team.
 # Distributed under the terms of the Modified BSD License.
 
+from binascii import b2a_base64, a2b_base64
 import copy
 import glob
 import io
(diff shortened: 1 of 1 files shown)


mpacer:
looks like there's something wrong with python-nightly and gcc, but that the python2 error actually has something to do with the way the base64 was encoded… Seemingly it lost some padding `=`s at its end?

minrk:
Should be fixed, now.

mpacer:
Sounds good. I still am confused as to why both `base64` and `binascii` exist as separate modules but this fix looks good.

minrk:
From the docs: there are three APIs for base64 encoding in two modules.

- binascii implements efficient, low-level b2a, a2b encodings, and is meant to be primarily used via other modules
- base64 provides 'high level' API wrappers around it
- base64.encode/decodebytes is the 'legacy' API in base64, which encodes base64 as multi-line text
- base64.b64encode/decode is the 'modern' API (new in Python 2.4), and probably what we should be using in general

PR #654, opened 2017-08-26
Added license file.

tdalseide:


mpacer:
If we're going to merge this we should probably get rid of the `COPYING.md` so we don't need to keep them in sync.

tdalseide:
There, that should make it a bit more apparent what happened.

Carreau:
This should also have updated the `Manifest.in` file. See #665.

Issue #23, opened 2015-05-17
Cann't download css files


dongweiming:
Install notebook but display this error:

```python
    Complete output from command /Users/dongweiming/.virtualenvs/jupyter_notebook/bin/python -c "import setuptools, tokenize; __file__='/Users/dongweiming/.virtualenvs/jupyter_notebook/src/nbconvert/setup.py'; exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" develop --no-deps:
    running develop
    running css
    Failed to download css from https://cdn.jupyter.org/notebook/4.0.0-dev/style/style.min.css: [Errno socket error] [Errno 1] _ssl.c:504: error:14077438:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert internal error
    error: None
    
    ----------------------------------------
    Command "/Users/dongweiming/.virtualenvs/jupyter_notebook/bin/python -c "import setuptools, tokenize; __file__='/Users/dongweiming/.virtualenvs/jupyter_notebook/src/nbconvert/setup.py'; exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" develop --no-deps" failed with error code 1 in /Users/dongweiming/.virtualenvs/jupyter_notebook/src/nbconvert
```

i have to use `http://` instead of `https://`

takluyver:
This looks like a duplicate of #21, though it's interesting that it works with http. Let's continue the discussion on that.

Issue #16, opened 2015-05-04
Warn if JS has been stripped


Carreau:
It would be nice to find a way when you remove Javascript from display_priority to have a prefiter and a template to ass a warning that JS has been disbled. 

Assigning to me, but I'll probably not have a look in the couple few weeks.

captainsafia:
If you don't have time for this @Carreau, I can take a look at it.

Might be a good way to get familiar with the `nbconvert` codebase.

Carreau:
Sure, I'm flying to France in 3h so you are safe for a few days.

Envoyé de mon iPhone

MSeal:
Don't believe this is still relevant. Reopen if that's incorrect.

PR #15, opened 2015-05-03
Fix propagation of fragments across multiple slides/subslides.

damianavila:
This fixes #9 

Essentially, we avoid labelling "-" cells with a previous data-fragment-index if we find a new slide/subslide. The fragments (and the group of fragments and "-" cells) only have sense inside a slide/subslide... it is an "intraslide" property and the data-fragment-index information has to be bounded to the slide/subslide containing the fragment/s (or fragments / "-" sequence).

I think the fix should be back-ported to the next 3.2 because right now you have unwanted fragments in your slides and that's very annoying.

Thanks!

PR #662, opened 2017-09-01
add feature to improve docs by having links to prs

mpacer:
Per @Carreau's suggestion, I've added the ability to autolink to prs/issues by including a custom role.

I do not believe that this will work in notebooks, because there is no way to specify the custom role in markdown. But this still seems like an improvement for our changelogs (at the least).


mpacer:
The only change from the ipython way of doing this is to not include the "PR" prefix because when multiple PRs are on the same lines this looks really bad & I don't want to try to batch the processing.

takluyver:
We [made a package](https://pypi.python.org/pypi/sphinxcontrib_github_alt) of our GH Sphinx extension so that we could use it across different repos without duplicating it. Is the thing about the 'PR' prefix important enough to make it worth duplicating the code again

mpacer:
didn't know about that… couldn't that be included with ghpro rather than as a separate package? They seem conceptually similar…

mpacer:
Part of the reason I ask if we could move it over there is that that would also give us the opportunity to make this at least mildly more configurable and modular, so that it wouldn't be an issue to remove the PR prefix (which is inconsistent with how GitHub formats these links today and looks pretty strange when more than one PR are referenced on a single line).

takluyver:
ghpro is a command line tool for doing operations on a Github repo. sphinxcontrib-github-alt is a Sphinx extension for referring to Github issues and PRs from Sphinx docs. It doesn't make any sense to me to join those into one package.

It's already under the Jupyter org (https://github.com/jupyter/sphinxcontrib_github_alt), so if we need some extra customisation, there's nothing stopping us doing that. I'm not convinced that the prefix thing is worth spending time on,  but I'm happy to look over a pull request if you think it's worth doing.

Carreau:
> It doesn't make any sense to me to join those into one package.

I agree with thomas on this one, the conceptual use case for GhPro and sphinxcontrib_github_alt is quite different. The only relevant part is that the magic `:gh:` role may need to contact GitHub to know if a number is an issue or a PR.

  I would be happy to have the `PR/issue` prefix configurable. Second option, is as you said to make batch processsing, make `:gh:` take multiple parameters and render only the first with a prefix. For example `` :gh:`4,8,15,16,23,42` `` would render as `PR #4, #8,  #15, #16, #23 and #42` it should not be too hard.

takluyver:
I've taken out the duplicate copy of the Sphinx github extension, and set it to use the packaged one instead. Any objections to merging this? We can make improvements to the extension as well, but I think having issues & PRs linked is already a nice step.

Issue #663, opened 2017-09-02
With nbconvert 5.3 you cannot use the ExecutePreprocessor and filter outputs at the same time


mpacer:
`TagRemovePreprocessor` currently handles removing cells, inputs, all outputs and individual outputs. It runs before the `ExecutePreprocessor` runs. 

As a result, once `ExecutePreprocessor` runs, it recreates the outputs that should have been filtered out by `TagRemovePreprocessor`. 

The simplest solution is to rerun `TagRemovePreprocessor` but that seems wasteful. 

The other solution would be to create a subclass of `TagRemovePreprocessor` called something like `OutputTagRemovePreprocessor` that inherits the traitlet values, but overwrites the `preprocess_cell` method to only filter outputs (adhering to both `remove_all_outputs_tags` and `remove_single_output_tags`). Then invoke this new preprocessor after all other preprocessors have run. This would make sense to appear in the base `Exporter` class.

mpacer:
@pxhanus Are you interested in tackling this one?

jzf2101:
@mpacer can we add this one to hacktoberfest?

pxhanus:
@mpacer Yes, I've got this! I'll let you know if I have any questions.

mpacer:
@pxhanus, why don't you start by just making a PR that readds the `TagRemovePreprocessor`?

jzf2101:
@finwickipedia

dhanus:
@jzf2101 @fenwickipedia I am still working on this. Thanks for checking!

pxhanus:
@jzf2101 @fenwickipedia I am still working on this. Thanks for checking!

fcollonval:
There is a temporary fix proposed in https://github.com/jupyter/nbconvert/issues/1300#issuecomment-656477456

PR #666, opened 2017-09-05
Include LICENSE in manifest.in

Carreau:
@mpacer,  you are the release manager of nbconvert , do you want to make a patch release with just this ? We can have further discussion about other issues with manifest.in.


mpacer:
I can do the patch release with this and the bad fix to #663.

mpacer:
eh i won't include the fix to #663 in this just for the sake of ease

Issue #667, opened 2017-09-05
Manifest.in and misc files not included


Carreau:
The [jupyter-nbconvert](https://github.com/jupyter/nbconvert/blob/d884adaa350d3dbcb4e35a82afbec223e00280c1/scripts/jupyter-nbconvert) file seem to not be included neither in sdist nor in wheel (to check), it is autogenerated by setuptools ?

More generally, I think we should check our repositories with [check-manifest](https://pypi.python.org/pypi/check-manifest) to avoid this kind of issues and decide what should (and should not) be in sdist.

Current output:

```bash
$ check-manifest
lists of files in version control and sdist do not match!
missing from VCS:
  nbconvert/resources/style.min.css
missing from sdist:
  .mailmap
  MANIFEST.in
  codecov.yml
  nbconvert/templates/skeleton/Makefile
  nbconvert/tests/README.md
  readthedocs.yml
  scripts
  scripts/jupyter-nbconvert
suggested MANIFEST.in rules:
  include *.in
  include *.yml
  recursive-include nbconvert *.md
  recursive-include nbconvert Makefile
```

mpacer:
should the Manifest itself be included? Should the scripts not be included or are they built automatically because of some kind of setuptools magic?

mpacer:
@takluyver do you have any thoughts on this? Should I be adding anything to Manifest.in other than the LICENSE file added in #666?

takluyver:
setuptools will generate the script if you install with setuptools. Pip uses setuptools, and that's the recommended way to install, but we include the `jupyter-nbconvert` script for installation not using setuptools. On several projects, though, we're now moving to require setuptools and explicitly disable egg building (which seems to cause most of the unpleasantness with setuptools). We should either do that (and thus get rid of the script), or include the script in setuptools.

I've never included MANIFEST.in itself - I thought it did that automatically, but I haven't checked.

Generally I'd say that the sdist should contain everything you might need to install or develop the package - but not generated files. My idea of an sdist is similar to a git tag, but more stable.

mpacer:
Ok… so for the 5.3.1 patch release I should add the scripts & scripts/jupyter-nbconvert but then I should open an issue about deprecating its support?

mpacer:
Marking this as closed in order to keep the discussion about any particular files (e.g., `Makefile`, `tests/README`, or the `MANIFEST.in` itself) in their own issues.

PR #670, opened 2017-09-06
Explicitetly exclude or include all files in Manifest.

Carreau:
And test for it.
Finish up #667

I used the following criteria:
> If Internet were to disapear which file would still be relevant?

So I excluded the one that configure services we use.


mpacer:
Looks good to me. Merging.

Issue #249, opened 2016-02-19
how to change pdf title


randy3k:
Besides changing the name of the ipynb file, is there any way to change the title of the saved pdf?

I know that I could first convert the file to .tex and compile the latex file by myself. But this is not what I want. I am looking for an automatic solution.

Carreau:
There is no (simple) automatic solution. 
You need a custom conversion pipeline. 

The simple way is to use a preprocessor. append it to the list of preprocessor add change `resource['metadata']['name']` to the title you like.

michaelpacer:
Now there is a way for you to do this: use the --output flag 

For example:

```bash
jupyter nbconvert mynotebookname.ipynb --to pdf --output yourpdftitle.pdf
```

liushapku:
@randy3k, when you mentioned pdf title, is it the title in the first page of the pdf, not the filename of the pdf as mentioned by @mpacer? Did you get any way of doing this?

randy3k:
@liushapku yes, it is what I meant. Not sure why @mpacer closed it. I had to convert it to `.tex` and fixed it myself.

mpacer:
sorry! Misunderstood!

liushapku:
@randy3k I have the same need. Currently, I use the API and call `exporter.from_notebook_node(nb, resources={'metadata': {'name': title}})` to achieve this. An alternative way is to call `preprocessor.preprocess` and pass the resource as the second argument. Hope this feature can be added to the cmdline tool.

mpacer:
#672 will make it easier if you set the title metadata in the notebook itself (meaning that `exporter.from_notebook_node(nb, resources={})` would work if you had already set the notebook metadata.title field.

agarnitin86:
Hi,
I am new to this. Can someone explain in a bit detail, how to convert ipynb to .tex to pdf file with custom title. I an using jupyter for creating ipynb files.

JacquotLeHaricot:
@agarnitin86 I know my answer comes late, but for the others that would have had the same question :
- File > Export Notebook As... > Latex
- you will get a zip file ; extract it somewhere
- open the .tex file inside : you will see something like 
`% Document parameters
    \title{TITLE_YOU_WANT_TO_CHANGE}`
- edit TITLE_YOU_WANT_TO_CHANGE to the title of your choice

whitstd:
I used the following commands in a Makefile (Ubuntu):

```
notebook:
[...]

PR #672, opened 2017-09-08
if nb.metadata.title is set, default to that for notebook

mpacer:
There is a title field in the notebook metadata as part of the nbformat v4 spec. We should use this feature if it is available.

This should solve #249.


mpacer:
@Carreau can I get a review?

takluyver:
Looks fine to me.

PR #674, opened 2017-09-13
Simple API for in-memory templates

### LINKED ISSUE #673: Offer api for more convenient in memory template definition &/or docs
After working through this process with @ian-r-rose today (& @choldgraf this summer), I've come to the conclusion that we should have a more convenient API for specifying in-memory templates. 

In particular, I think the ideal API would involve the following steps: 
1. Defining a template string
2. passing the template string to a `TemplateExporter` subclass.

And that's it! 

That fully specifies the template the person wants and where they want it. So why should we be asking for more? Anything more complicated is offloading the burden of implementation on users. That excludes many of the people Jupyter is best poised to reach out to. (Aside: many thanks to @willingc & @yuvipanda for getting me to think along these lines!)

But before we can make that happen… we need to figure out where we stand.

# Current situation: complicated, intricate, arbitrary steps

Right, now it takes multiple steps (at least 7 as laid out below), a good deal of understanding (knowledge of multiple traitlets, the traitlet types, an understanding that Jinja template loaders exist & how they work, &c.) and a number of arbitrary decisions about features in order to get an in-memory template working correctly in a custom exporter.

The full process will probably be worth documenting explicitly (I started this effort below). I'm not making a PR yet because I want more of an idea as to where this is going to go.

But in looking at this preliminary documentation, it becomes clear *really* quickly that this is an unnecessarily complicated set of choices for the simplest possible version of this task. 

What's great though is I think we can reduce this complexity substantially. I'll give it a try tomorrow.

## How to create and assign an in-memory template (as of 5.3.1)

1. Create in-memory template text e.g., 
    ```python
    my_template_text = "{%- extends 'rst.tpl' -%}"
    ```

2. Decide on string for template name, a pseudo file-path (possibly with or without an extension)
    ```python
    my_template_name_no_ext = "my_template_name"
    my_template_name =  my_template_name_no_ext + ".tpl"
    ```

3. wrap template name and text in dictionary, the key is the name to the text
    ```python
    my_template_dict  = {my_template_name: my_template_text}
    ```

4. Explicitly load DictLoader from jinja2 
    ```python
    from jinja2 import DictLoader
    ```
    **NB**: Knowing to import this requires knowing:
    1. nbconvert uses jinja2 for its templates
    2. jinja2 uses template loaders to add templates to its environment
    3. nbconvert explicitly requires you to create the jinja2 environment by expressing its loaders.
    4. DictLoader exists.
    5. DictLoader is what you probably want for this case.
        1. DictLoader will create a pseudo filename in the environment so that it based on the keys in the dict passed.

    **NB**: Writing a template – especially if you're just taking existing templates and modifying them without taking the time to fully understanding them – does not require you to know any of those things. Given how complicated nbconvert is (even for _me_), I feel like the less we ask people to understand in order to use it, the more happy & capable everyone involved will feel and be.

5. create a jinja loader instance that can load the template via `jinja2.DictLoader`
```python
my_template_loader = DictLoader(my_template_dict)
```

6. Assign template name with or without extension to `TemplateExporter.template_file`
    This can be achieved in multiple ways.
    1. Assign as a class attribute directly ([inspiration](https://github.com/jupyter/nbconvert/issues/490#issue-196123843))
        ```python
            template_file = my_template_name
            # or template_file = my_template_name_no_ext 
        ```
        I will note that I'm not going to include the option of `my_template_name_no_ext` anymore in these examples… but we should probably drive that home if it does come down to docs.
    2. as part of the __init__ (always part of MyExporter, doesn't need to be passed in) [inspiration](https://github.com/spatialaudio/nbsphinx/blob/f8543afc78c5c6f4aeb24db23f64fe88f981c406/nbsphinx.py#L514-L546)
        ```python
               def __init__(self, *args, **kwargs):
                   super(MyExporter, self).__init__(template_file=my_template_name)
        ```
    3. as part of the constructor call inside of a script (i.e., without explicitly tying the exporter to your template file, it just happens to get it on this particular instance)
        ```python
        #Either, directly
        MyExporter(template_file = my_template_name)
        # or as a config option
        MyExporter(config = {
               "TemplateExporter":{
                     "template_file": my_template_name}
               }
        })
        ```

7. Add my_template_loader to the `List` `TemplateExporter.extra_loaders`
    This can be achieved in multiple ways (similar to 6 above…)
    1. Assign as a class attribute directly 
        ```python
            extra_loaders = [my_template_loader]
        ```
    2. as part of the `__init__` 
        ```python
               def __init__(self, *args, **kwargs):
                   super(Exporter, self).__init__(extra_loaders=[my_template_loader])
        ```
    3. as part of the constructor call inside of a script
        ```python
        #Either, directly
        MyExporter(extra_loaders = [my_template_loader])
        # or as a config option
        MyExporter(config = {
               "TemplateExporter":{
                     "template_loader": [my_template_loader]}
               }
        })
        ```

cc @takluyver @Carreau

takluyver:
Closes #673.

The new API is a method `exporter.use_string_template(s)`, which you pass the string template into.

Example to use it with a subclass:

```python
import sys
from nbconvert.exporters import HTMLExporter

POPPING_TEMPLATE = """
{%- extends 'full.tpl' -%}

{%- block markdowncell -%}
<div style="background-color: hsl(85, 86%, 76%);">
{{ super() }}
</div>
{%- endblock markdowncell -%}
"""

class PopExporter(HTMLExporter):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.use_string_template(POPPING_TEMPLATE)

if __name__ == '__main__':
    out, res = PopExporter().from_filename(sys.argv[1])
    with open('pops.html', 'w') as f:
        f.write(out)
```

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -16,8 +16,8 @@ from traitlets.config import Config
 from traitlets.utils.importstring import import_item
 from ipython_genutils import py3compat
 from jinja2 import (
-    TemplateNotFound, Environment, ChoiceLoader, FileSystemLoader, BaseLoader
-)
+    TemplateNotFound, Environment, ChoiceLoader, FileSystemLoader, BaseLoader,
+    DictLoader)
 
 from nbconvert import filters
 from .exporter import Exporter
@@ -244,6 +244,15 @@ class TemplateExporter(Exporter):
         self.observe(self._invalidate_template_cache,
                      list(self.traits(affects_template=True)))
 
+        self._str_template_container = {}
+
+    def use_string_template(self, tpl):
+        """Use a template from a string instead of loading from a file.
+
+        Pass the template string into this method.
+        """
+        self._str_template_container['<memory>'] = tpl
+        self.template_file = '<memory>'
 
     def _load_template(self):
         """Load the Jinja template object from the template file
@@ -374,6 +383,7 @@ class TemplateExporter(Exporter):
              os.path.join(here, self.template_skeleton_path)]
 
         loaders = self.extra_loaders + [
+            DictLoader(self._str_template_container),
             ExtensionTolerantLoader(FileSystemLoader(paths), self.template_extension)
         ]
         environment = Environment(
--- a/nbconvert/exporters/tests/test_templateexporter.py
+++ b/nbconvert/exporters/tests/test_templateexporter.py
@@ -133,6 +133,19 @@ class TestExporter(ExportersTestsBase):
         nb = v4.new_notebook()
         out, resources = exporter.from_notebook_node(nb)
 
+    def test_in_memory_template_simple(self):
+        # New simpler API for providing an in-memory template
+        tpl = ("{%- extends 'markdown.tpl' -%}\n"
+               "{% block markdowncell scoped %}\n"
+               "In-memory template\n"
+               "{% endblock markdowncell %}\n")
+        exporter = TemplateExporter()
+        exporter.use_string_template(tpl)
+
+        nb = v4.new_notebook()
+        nb.cells = [v4.new_markdown_cell("boo")]
+        out, resources = exporter.from_notebook_node(nb)
+        assert "In-memory template" in out
 
     def test_fail_to_find_template_file(self):
         # Create exporter with invalid template file, check that it doesn't


mpacer:
I had a slightly different API in mind, and I'm going to try to write it from scratch but if I find myself using enough patterns similar to yours I'll just rebase it off of this branch.

PR #676, opened 2017-09-15
Set BIBINPUTS and BSTINPUTS environment variables when making PDF

takluyver:
@sobester had some difficulties converting a notebook to PDF using bibtex references. Converting to tex and then running latex manually worked. My guess is that bibtex couldn't find the .bib file because nbconvert doesn't copy it to the temporary directory from which it runs latex.

While looking through this code today, I noticed that we use an environment variable `TEXINPUTS` to tell Latex to look for related files in the directory where the original notebook is. A bit of searching reveals that bibtex uses two similar environment variables, `BIBINPUTS` (for bibliography files) and `BSTINPUTS` (for style files). If my guess is correct, adding the notebook directory to these should resolve the problem with using bibtex.

### DIFF
--- a/nbconvert/exporters/pdf.py
+++ b/nbconvert/exporters/pdf.py
@@ -27,6 +27,18 @@ class LatexFailed(IOError):
         u = self.__unicode__()
         return cast_bytes_py2(u)
 
+def prepend_to_env_search_path(varname, value, envdict):
+    """Add value to the environment variable varname in envdict
+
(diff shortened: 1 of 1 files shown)


takluyver:
I couldn't find any good docs about these environment variables, but they're described in various posts, e.g. [this SE answer](https://tex.stackexchange.com/a/120473).

mpacer:
This looks fine, but I'd still like a way to directly inform both LaTeXExporter and PDFExporter about where the bib file is directly (i.e., without using environment variables as a signalling mechanism).

PR #684, opened 2017-10-02
Per cell exception

### LINKED ISSUE #680: Allow errors on a cell-by-cell basis
In some tutorial notebooks we have in [astropy/astropy-tutorials](https://github.com/astropy/astropy-tutorials), we intentionally raise an exception to demonstrate a feature or concept to the user. But we also want to be able to test the notebooks by running them during the nbconvert phase (using nbsphinx). Right now, we can allow errors on a per-notebook basis using `--allow-errors`, but it would be useful if we could specify (via the cell-level metadata) that certain cells are allowed to fail. 

Is this a feature that you would support in nbconvert?

cc @eteq

(spatialaudio/nbsphinx#125 and astropy/astropy-tutorials#150 are related)

mpacer:
Closes #680.

### DIFF
--- a/nbconvert/preprocessors/tests/test_execute.py
+++ b/nbconvert/preprocessors/tests/test_execute.py
@@ -213,7 +213,8 @@ class TestExecute(PreprocessorTestsBase):
 
     def test_raises_exception_cell_tag(self):
         """
-        Check that conversion continues if ``cell_tags`` is False.
+        Check that conversion continues if ``raises-exception`` is present in
+        the cell in which an exception is raised.
         """
         current_dir = os.path.dirname(__file__)
         filename = os.path.join(current_dir, 'files',
--- a/nbconvert/preprocessors/tests/test_execute.py
+++ b/nbconvert/preprocessors/tests/test_execute.py
@@ -36,7 +36,7 @@ def _normalize_base64(b64_text):
         return b64encode(b64decode(b64_text.encode('ascii'))).decode('ascii')
     except (ValueError, TypeError):
         return b64_text
-    
+
 class TestExecute(PreprocessorTestsBase):
     """Contains test functions for execute.py"""
     maxDiff = None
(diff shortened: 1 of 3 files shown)


nbconvert/preprocessors/tests/test_execute.py
```diff
@@ -195,19 +197,36 @@ def timeout_func(source):
...
-            assert u"# üñîçø∂é".encode('utf8', 'replace') in str(exc.value)
+            self.assertIsInstance(str(exc.value), str)
+            if sys.version_info >= (3, 0):
+                assert u"# üñîçø∂é" in str(exc.value)
+            else:
+                assert u"# üñîçø∂é".encode('utf8', 'replace') in str(exc.value)
+
+    def test_raises_exception_cell_tag(self):
+        """
+        Check that conversion continues if ``cell_tags`` is False.
```
takluyver: This docstring doesn't seem quite right.

mpacer: agreed.

nbconvert/preprocessors/tests/test_execute.py
```diff
@@ -195,19 +197,37 @@ def timeout_func(source):
...
+            self.assertIsInstance(str(exc.value), str)
+            if sys.version_info >= (3, 0):
+                assert u"# üñîçø∂é" in str(exc.value)
+            else:
+                assert u"# üñîçø∂é".encode('utf8', 'replace') in str(exc.value)
+
+    def test_raises_exception_cell_tag(self):
+        """
+        Check that conversion continues if ``raises-exception`` is present in
+        the cell in which an exception is raised.
```
takluyver: This one is actually testing `force_raise_errors`, though.

Can we split this into two tests: one with this name and docstring, and another one with this code (`test_force_raise_errors`)? ;-)

mpacer: Technically the test of it on its own implicitly occurs earlier in the global "convert everything!"  test: https://github.com/jupyter/nbconvert/blob/0a262a3a00a1680c0b30c95b0b9837ed4f8ff063/nbconvert/preprocessors/tests/test_execute.py#L122-L137

But you're right that this should explicitly test `force_raise_errors` instead

takluyver: Aha,  great. In that case, if you just want to fix the test name and docstring,  I think this is ready to go.

PR #685, opened 2017-10-03
Img filename metadata

### LINKED ISSUE #671: Feature request: custom image file names
By default nbconvert names generated image files using a fixed prefix and the cell number. While this is simple and nicely general it doesn't work well in scenarios where you want to re-use the generated images as part of a different workflow. As an example: I am working on a LaTeX document that includes figures generated from a jupyter notebook. No problem as long as the notebook doesn't change, but as soon as I insert cells at the beginning of the notebook all image file names change and have to be adapted manually in the LaTeX file.

I've had a long look through the documentation but I haven't been able to find a way to change that behaviour.

My suggestion: optionally let image file names be generated either from a heading preceding the cell containing the image or from a meta tag in the image cell itself. I've changed my local copy of nbconvert to use a tag "name" in place of the cell number if present (2 lines of code), but a proper implementation should probably make that optional and easily configurable.

mpacer:
closes #671 

this should overwrite files if they are already present.

It also will add an extension to the file if it is not already present.

I'm open to changing either of those behaviours.

### DIFF
--- a/nbconvert/preprocessors/extractoutput.py
+++ b/nbconvert/preprocessors/extractoutput.py
@@ -88,6 +88,8 @@ class ExtractOutputPreprocessor(Preprocessor):
                         ext = '.' + mime_type.rsplit('/')[-1]
                     if out.metadata.get('filename', ''):
                         filename = out.metadata['filename']
+                        if not filename.endswith(ext):
+                            filename+=ext
                     else:
                         filename = self.output_filename_template.format(
                                     unique_key=unique_key,
--- a/nbconvert/preprocessors/extractoutput.py
+++ b/nbconvert/preprocessors/extractoutput.py
@@ -108,9 +108,13 @@ class ExtractOutputPreprocessor(Preprocessor):
 
                     if filename in resources['outputs']:
                         raise ValueError(
-                            "Your filename: {} appears more than once. "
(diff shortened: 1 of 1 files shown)


takluyver:
This looks fine to me. I might even go further and throw an error if two outputs in one notebook have the same filename.

mpacer:
How does that check feel? It looked like `resources.ouptuts` was acting as the global store of filenames, so this seemed like the right approach.

nbconvert/preprocessors/extractoutput.py
```diff
@@ -102,6 +106,13 @@ def preprocess_cell(self, cell, resources, cell_index):
                     out.metadata.setdefault('filenames', {})
                     out.metadata['filenames'][mime_type] = filename
 
+                    if filename in resources['outputs']:
+                        raise ValueError(
+                            "Your filename: {} appears more than once. "
```
takluyver: I think this message could be a bit clearer - imagine a user who has got a notebook from someone else and is trying to convert it. Even if someone wrote it themselves, they might accidentally duplicate a filename by copying a cell.

Let's mention that this is a filename associated with an *output*, which nbconvert is now saving to a separate file, and that the name has come from output metadata.

takluyver:
Love the detailed error message! :-)

PR #692, opened 2017-10-18
Log execution of notebooks.

tillahoffmann:
I sometimes want to see the internals of my notebooks when executing them as a batch job. 

This PR adds a `log_execution` traitlet to the `ExecutePreprocessor` which defaults to `False`. If `log_execution=True`, the input and output of the cells are logged when `nbconvert` is invoked with `--execute`. 

A few questions:
* I have only covered a few `output_type`s for logging. Did I miss any?
* The printed output looks a bit funny because the logger prepends `[NbConvertApp]` to each statement. That could be fixed manually by indenting by 14 characters, but it seems there must be a better way.

# Example

screen shot 2017-10-18 at 16 33 31

```
$ jupyter nbconvert --execute --ExecutePreprocessor.log_execution=True --allow-errors Untitled.ipynb
[NbConvertApp] Converting notebook Untitled.ipynb to html
[NbConvertApp] Executing notebook with kernel: python3
[NbConvertApp] In [1]: print("Hello world.")
        1
[NbConvertApp] Hello world.

[NbConvertApp] Out[1]: 1
[NbConvertApp] In [2]: from matplotlib import pyplot as plt
        %matplotlib inline
        plt.plot((0, 1), (0, 1))
[NbConvertApp] Out[2]: [<matplotlib.lines.Line2D at 0x10dbe4828>]
[NbConvertApp] <matplotlib.figure.Figure at 0x10c311860>
[NbConvertApp] In [3]: assert False
[NbConvertApp] ---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-3-6e6df518a476> in <module>()
----> 1 assert False

AssertionError:
[NbConvertApp] Writing 283927 bytes to Untitled.html
```

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -4,7 +4,7 @@ and updates outputs"""
 # Copyright (c) IPython Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-from textwrap import dedent
+from textwrap import dedent, indent
 
 try:
     from queue import Empty  # Py 3
@@ -194,6 +194,15 @@ class ExecutePreprocessor(Preprocessor):
             )
     ).tag(config=True)
 
+    log_execution = Bool(False,
+        help=dedent(
+            """
+            If `False` (default), the output of executed cells is discarded.
+            If `True`, the output of executed cells is logged.
+            """
+        )
+    ).tag(config=True)
+
     kernel_manager_class = Type(
         config=True,
         help='The kernel manager class to use.'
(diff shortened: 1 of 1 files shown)


takluyver:
I wouldn't normally add a new config option to control what gets logged; instead I would log it at a lower level, e.g. `log.debug()` and when you want to see it, use the existing config options to make that visible.

For nicer presentation, it might be better to build a separate tool based on the [execute_interactive](https://github.com/jupyter/jupyter_client/blob/8f7a86571323430a666d9ecf5522bc2b0bcf667c/jupyter_client/blocking/client.py#L213) method Min added to jupyter_client; this will `print()` output, and it should be easy to `print()` input manually.

tillahoffmann:
@takluyver, just to clarify: Are you suggesting keeping the `log_execution` flag and using `execute_interactive` with `silent=not log_execution`, dropping `log_execution` and substituting `self.log.debug` for self.log.info` in this PR, or a third option I couldn't parse?

takluyver:
I'm roughly suggesting two options:

1. If it's for debugging,  and it's not terribly important that the output looks nice, drop `log_execution` and use `self.log.debug()` to record what's happening. (You can also get more fancy with numeric levels for logging - debug is 10, info 20, and so on - you can use numbers in between to get more control).
2. If you want to show what's happening in the terminal as a primary purpose of running the notebook,  write a separate script, not part of nbconvert at all, which uses `execute_interactive()` to show the output.

tillahoffmann:
It would be great to be able to combine the interactive execution or execution with display of the inner workings of the notebook with the convenience and flexibility of `nbconvert`.

Here's a bit more information on my use case: I process a number of datasets using the same notebook using a `Makefile` that looks something like the following.

```Makefile
DATASETS = dataset_a dataset_b ...
TARGETS = $(DATASETS:=.html)

all : $(TARGETS)

$(TARGETS) : %.html : %
    DATASET=$* jupyter nbconvert --execute --allow-errors --out $@
```

This gives me 
* The awesomeness of Jupyter (thank you for all the hard work!) at development time.
* Reproducible processing of all datasets at runtime.
* Automatic reports by exporting to html.

We use a similar setup to train our machine learning models.

The only thing I'm missing is some introspection to see what's going on while the build is executing. Hence, the desire to integrate the functionality into `nbconvert`. But I do understand that my particular use case may be too specific to warrant inclusion in the more general-purpose tool `nbconvert. 

@takluyver, would be great to hear your thoughts. And if you think the above is overkill, I'll change this PR to use `self.log.debug` without the flag.

mpacer:
@tillahoffman if you were to create a custom exporter for the sole purpose of providing a configurable logger preprocessor based on the execute preprocessor, declaring an entry point, it should be accessible from nbconvert. 

And actually… You may not even need the whole entry point apparatus… If it turns out that you actually _do_, we should make it easier to enable custom preprocessors.

pvieito:
Any update on this? I am interested in this feature and could open a new Pull Request adding support for logging the cell outputs in debug mode. It would simply add the following code before [the `force_raise_errors` phase](https://github.com/jupyter/nbconvert/blob/6dd08eb1590c0f6339f21f2ad2332f4cb4ede36d/nbconvert/preprocessors/execute.py#L381):

```python
for out in outputs:
    if out['output_type'] == 'stream':
        self.log.debug(out['text'])
    elif out['output_type'] == 'display_data':
        self.log.debug(out['data']['text/plain'])
    elif out['output_type'] == 'execute_result':
        self.log.debug(
            'Out[%d]: %s', out['execution_count'], out['data']['text/plain']
        )
```

MSeal:
I could look at including a modified version of this that we use in papermill wrapping nbconvert: https://github.com/nteract/papermill/blob/master/papermill/preprocess.py#L84-L95. Any concerns with using the app conf settings in this PR to enable logging like we do in the link @mpacer ?

PR #694, opened 2017-10-22
added shebang for python

### LINKED ISSUE #690: Use shebang when converting to python script?
Using these commands:
`jupyter nbconvert --to python [YOUR_NOTEBOOK].ipynb`
or even
`jupyter nbconvert --to script [YOUR_NOTEBOOK].ipynb`,
I would propose, that every converted python scritpt from Ipython/Jupyter
should contain the python shebang `#!/usr/bin/env python` as their first line.

geniusupgrader:
fixes #690.
The shebang should start on the first line,
and that is why there is no newline between the template and the shebang

### DIFF
--- a/nbconvert/exporters/tests/test_python.py
+++ b/nbconvert/exporters/tests/test_python.py
@@ -21,4 +21,4 @@ class TestPythonExporter(ExportersTestsBase):
         """Can a PythonExporter export something?"""
         (output, resources) = self.exporter_class().from_filename(self._get_notebook())
         self.assertIn("coding: utf-8", output)
-        self.assertIn("#!/usr/bin/env ipython", output)
+        self.assertIn("#!/usr/bin/env python", output)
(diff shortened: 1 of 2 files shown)


takluyver:
If you put dashes in the template syntax, I think it should suppress the newlines so the template can be written the obvious way.

```
{%- block header -%}
```

geniusupgrader:
ok, now with dashes

mpacer:
LGTM, could you add a test to watch for this when converting to python?

mpacer:
Also, it shouldn't be `python` but `ipython` since we rely on ipython as part of the python exporter to handle a number of ipython magics

mpacer:
If you've allowed edits from maintainers I'm going to push these changes and will merge this. If not… i'll make a PR against yours and will need you to merge it first and then I can merge it here.

takluyver:
Hmm, I'd actually say that the python shebang was OK. Yes, if you use magics you will get a Python script that can only be run inside IPython, but not all notebooks will use magics, and we don't really encourage using IPython to run scripts when it's not necessary.

I don't feel strongly about this, though.

mpacer:
Maybe we could detect if a magic is used? It seems like a lot of overhead just to determine whether it should be an `ipython` or `python`.

takluyver:
I'm inclined to just let it be `python` - even if that won't necessarily work all the time, I think it represents the intent of converting to a `.py` script. Even run in IPython, you can't guarantee that code from a notebook will work - it might depend on displaying rich output, for instance.

PR #696, opened 2017-10-23
Enable ANSI underline and inverse

mgeier:
This is a companion PR to https://github.com/jupyter/notebook/pull/2967.

### DIFF
--- a/nbconvert/filters/tests/test_ansi.py
+++ b/nbconvert/filters/tests/test_ansi.py
@@ -41,6 +41,9 @@ class TestAnsi(TestsBase):
             'hel\x1b[0;32mlo': 'hel<span class="ansi-green-fg">lo</span>',
             'hellø': 'hellø',
             '\x1b[1mhello\x1b[33mworld\x1b[0m': '<span class="ansi-bold">hello</span><span class="ansi-yellow-intense-fg ansi-bold">world</span>',
+            'he\x1b[4mll\x1b[24mo': 'he<span class="ansi-underline">ll</span>o',
+            '\x1b[35mhe\x1b[7mll\x1b[27mo': '<span class="ansi-magenta-fg">he</span><span class="ansi-default-inverse-fg ansi-magenta-bg">ll</span><span class="ansi-magenta-fg">o</span>',
+            '\x1b[44mhe\x1b[7mll\x1b[27mo': '<span class="ansi-blue-bg">he</span><span class="ansi-blue-fg ansi-default-inverse-bg">ll</span><span class="ansi-blue-bg">o</span>',
         }
 
         for inval, outval in correct_outputs.items():
@@ -61,6 +64,7 @@ class TestAnsi(TestsBase):
             'hello\x1b[01;34mthere': r'hello\textcolor{ansi-blue-intense}{\textbf{there}}',
             'hello\x1b[001;34mthere': r'hello\textcolor{ansi-blue-intense}{\textbf{there}}',
             '\x1b[1mhello\x1b[33mworld\x1b[0m': r'\textbf{hello}\textcolor{ansi-yellow-intense}{\textbf{world}}',
+            'he\x1b[4mll\x1b[24mo': 'he\\underline{ll}o',
         }
 
         for inval, outval in correct_outputs.items():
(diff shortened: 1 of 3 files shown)


mpacer:
You're going to have to give a lot more context than just saying it is a companion. 

What is the purpose of this particular change? Please explain for those who will need to debug this in the future.

Note: you made an API breaking change, these are private methods but still… we try to avoid breaking APIs. Why does it need to be a positional argument?

mgeier:
@mpacer Thanks for the feedback!

> You're going to have to give a lot more context than just saying it is a companion.

I wanted to wait for feedback on the `notebook` issue and if and when that gets accepted, this PR is just about implementing the same thing for `nbconvert`.

> What is the purpose of this particular change? Please explain for those who will need to debug this in the future.

Do you mean here in the Github comments or somewhere in the code docstring/comments?

The purpose is in the title of this PR.

The `notebook` already supports ANSI "underline", and it somewhat supports "inverse", but instead of inverting foreground and background, a dotted line is drawn around the text. In https://github.com/jupyter/notebook/pull/2967 I'm trying to actually invert the colors instead.

> You removed some documentation, is that documentation wrong?

Yes, since #259.

> If it is, is it that the functionality should change or the documentation should change?

The functionality *did* change in #259 and the documentation should follow suit at some point.

> I am not a fan of the multiple lines of documentation that describe different pieces of functionality. Please break them up into individual lines for the individual code definitions.

I don't understand what you mean by that.

It would probably be best to get rid of most of the documentation, because it will not be as accurate as the code anyway, and it is very likely to fall out of sync (which it already did).

What about removing the description about the actual codes from the docstring and instead adding a few comments in the code where it is not obvious what certain codes do?

mgeier:
I forgot to address this:

> Note: you made an API breaking change, these are private methods but still… we try to avoid breaking APIs. Why does it need to be a positional argument?

It doesn't need to be a positional argument. I can make it a keyword argument if that's what you want, but it won't really help. The problem is that I'm calling the converter functions with one more argument in `_ansi2anything()`. If somebody uses `_ansi2anything()` with their own converter function, the keyword argument won't help.

But as you say, those are private functions. And I made them private intentionally. They are only there as an implementation detail for internal code re-use and they are not meant to be used externally.

In the future, we might even add more ANSI formatting, which will require further "breaking" changes.

mgeier:
https://github.com/jupyter/notebook/pull/2967 was merged, I've updated this PR accordingly: 006153c18eb561bbc40c7b46317d2f3c97bd630f

I don't know when/how the new CSS classes from the `notebook` will be available to `nbconvert`. If you need me to add temporary CSS definitions, please tell me where I should add them.

mgeier:
@mpacer Ping?

takluyver:
I think this looks OK, and I'm not worried about changing the API of these private functions.

Can you add a few tests for the new functionality in `nbconvert.filters.tests.test_ansi`?

mgeier:
@takluyver 

I've shortened a docstring (a6e4ba27e033921baff046b32a29c1ef7b365f16) as mentioned in https://github.com/jupyter/nbconvert/pull/696#issuecomment-339574322.

And I've added some test cases in e443e85ad6c66ed2af4f204af767e656a6f3e7b6.

What about the new CSS classes?

mgeier:
I've changed the last commit to avoid a problem with Python 2.x and `r'\u'` (197dceea7c6caa3cdbbe8e6d06a3c5a48ed96b19) and now the tests on Travis-CI pass.

takluyver:
Grant is getting ready to release notebook 5.3 in the next couple of days, which will include your PR https://github.com/jupyter/notebook/pull/2967 . Once that's done, you can update the version of the notebook it gets the CSS from in `setup.py`, as part of this PR.

takluyver:
Thanks!

@mpacer do you want to have another look at this before we press the green button?

mgeier:
Ping?

mgeier:
I've rebased this PR onto `master` to avoid a conflict where I updated the Jupyter CSS to 5.3.0 while in the meantime it was updated to 5.4.0 in #748.

nbconvert/filters/tests/test_ansi.py
```diff
@@ -61,6 +64,7 @@ def test_ansi2latex(self):
             'hello\x1b[01;34mthere': r'hello\textcolor{ansi-blue-intense}{\textbf{there}}',
             'hello\x1b[001;34mthere': r'hello\textcolor{ansi-blue-intense}{\textbf{there}}',
             '\x1b[1mhello\x1b[33mworld\x1b[0m': r'\textbf{hello}\textcolor{ansi-yellow-intense}{\textbf{world}}',
+            'he\x1b[4mll\x1b[24mo': 'he\\underline{ll}o',
```
takluyver: Is it convenient to test the conversion of inverted mode to latex as well?

mgeier: It is a bit unwieldy, but why not: dab959638199ce562534efcdd9a9d7eb9dfc31bd

takluyver:
Thanks!

PR #701, opened 2017-11-06
add log silencing command line flag

mpacer:
Was using nbconvert as part of a demo and it was useful to have a concise way of specifying that there should be no log output.

Errors should still raise, but I figure that other people have wanted this even though I didn't see any issues asking for it.


mpacer:
nvm log-level is already attached, that's easier and better.

PR #715, opened 2017-12-04
Add Markdown block lexer/parser for LaTeX blocks

### LINKED ISSUE #404: Broken latex equations on convert
Some latex code works fine in Jupyter notebook, but is not compiled correctly on export. e.g.

    For the bispectrum contributions, we have
    $$
    \langle \phi(l_1)\phi(l_2)\omega(l_3)\rangle = (2\pi)^2 \delta (l_1+l_2+l_3) b_{l_1,l_2,l_3}^{\phi\phi\omega}
    $$
    where
    $$
     b_{l_1,l_2,l_3}^{\phi\phi\omega} = -8\, l_1\times l_2 \,l_1\cdot l_2 \int_0^{\chi_*} d\chi \frac{W(\chi,\chi_*)^2}{\chi^2} \int_0^{\chi}d\chi' \frac{W(\chi',\chi)W(\chi',\chi_*)}{{\chi'}^2} \left[
     P_\psi\left(\frac{l_1}{\chi},z(\chi)\right) P_\Psi\left(\frac{l_2}{\chi'},z(\chi')\right)
    - (l_1\leftrightarrow l_2)
    \right]
    $$
    Now defining
    $$
    M_*(l,l') \equiv l^4\int_0^{\chi_*}d\chi \frac{W(\chi,\chi_*)^2}{\chi^2} P_\Psi\left(\frac{l}{\chi}, z(\chi)\right) C_{l'}^\kappa(\chi,\chi_*)
    $$
    we have
    $$
    b_{l_1l_2l_3}^{\kappa\kappa\omega} = -\sin 2\phi_{21}\left[ M_*(l_1,l_2)-M_*(l_2,l_1)\right]
    $$

In the notebook it looks like this:
image

After HTML export it looks like this:

image

danilobellini:
Fix #404

### DIFF
--- a/nbconvert/filters/markdown_mistune.py
+++ b/nbconvert/filters/markdown_mistune.py
@@ -21,12 +21,43 @@ from pygments.util import ClassNotFound
 from nbconvert.filters.strings import add_anchor
 
 
+class MathBlockGrammar(mistune.BlockGrammar):
+    block_math = re.compile(r"^\$\$(.*?)\$\$|^\\\\\[(.*?)\\\\\]", re.DOTALL)
+    latex_environment = re.compile(r"^\\begin\{([a-z]*\*?)\}(.*?)\\end\{\1\}",
+                                   re.DOTALL)
+
+
 class MathInlineGrammar(mistune.InlineGrammar):
     inline_math = re.compile(r"^\$(.+?)\$|^\\\\\((.+?)\\\\\)", re.DOTALL)
     block_math = re.compile(r"^\$\$(.*?)\$\$|^\\\\\[(.*?)\\\\\]", re.DOTALL)
     latex_environment = re.compile(r"^\\begin\{([a-z]*\*?)\}(.*?)\\end\{\1\}",
                                    re.DOTALL)
     text = re.compile(r'^[\s\S]+?(?=[\\<!\[_*`~$]|https?://| {2,}\n|$)')
+
+
+class MathBlockLexer(mistune.BlockLexer):
+    default_rules = (['block_math', 'latex_environment']
+                     + mistune.BlockLexer.default_rules)
+
+    def __init__(self, rules=None, **kwargs):
+        if rules is None:
+            rules = MathBlockGrammar()
+        super(MathBlockLexer, self).__init__(rules, **kwargs)
+
+    def parse_block_math(self, m):
+        """Add token for a $$math$$ block"""
+        self.tokens.append({
+            'type': 'block_math',
+            'text': m.group(1) or m.group(2)
+        })
+
+    def parse_latex_environment(self, m):
+        """Add token for a \begin{.} ...LaTeX stuff... \end{.} block"""
+        self.tokens.append({
+            'type': 'latex_environment',
+            'name': m.group(1),
+            'text': m.group(2)
+        })
 
 
 class MathInlineLexer(mistune.InlineLexer):
@@ -53,7 +84,16 @@ class MarkdownWithMath(mistune.Markdown):
     def __init__(self, renderer, **kwargs):
         if 'inline' not in kwargs:
             kwargs['inline'] = MathInlineLexer
+        if 'block' not in kwargs:
+            kwargs['block'] = MathBlockLexer
         super(MarkdownWithMath, self).__init__(renderer, **kwargs)
+
+    def output_block_math(self):
+        return self.renderer.block_math(self.token["text"])
+
+    def output_latex_environment(self):
+        return self.renderer.latex_environment(self.token["name"],
+                                               self.token["text"])
 
 
 class IPythonRenderer(mistune.Renderer):
--- a/nbconvert/filters/tests/test_markdown.py
+++ b/nbconvert/filters/tests/test_markdown.py
@@ -137,6 +137,18 @@ class TestMarkdown(TestsBase):
             "$$a<b&b<lt$$",
             "$$a<b&lt;b>a;a-b<0$$",
             "$$<k'>$$",
+            """$$x
+=
+2$$""",
+            r"""$$
+ b_{l_1,l_2,l_3}^{\phi\phi\omega} = -8\, l_1\times l_2 \,l_1\cdot l_2 \int_0^{\chi_*} d\chi \frac{W(\chi,\chi_*)^2}{\chi^2} \int_0^{\chi}d\chi' \frac{W(\chi',\chi)W(\chi',\chi_*)}{{\chi'}^2} \left[
+ P_\psi\left(\frac{l_1}{\chi},z(\chi)\right) P_\Psi\left(\frac{l_2}{\chi'},z(\chi')\right)
+- (l_1\leftrightarrow l_2)
+\right]
+$$""",
+            r"""\begin{equation*}
+x = 2 *55* 7
+\end{equation*}""",
             """$
 \\begin{tabular}{ l c r }
   1 & 2 & 3 \\


takluyver:
There was a Markdown block lexer before, and it was removed in #611. Let's make sure we figure out properly what's needed, so we don't just bounce back and forth between adding and removing it.

danilobellini:
There's no way to remove the block lexer without breaking the whole process. As I described in the commit message, the inline lexer doesn't even get a string with the whole equation as a single input, the block math regex won't match.

It's strange to fix a bug opened in 2016 that seems to have emerged recently (only in the release 5.3.0, as it seems).

Nevertheless, it's something I felt recently, I found it quite annoying when this notebook was no longer shown correctly in nbviewer (like several other notebooks I wrote): http://nbviewer.jupyter.org/github/danilobellini/scientific-literature/blob/master/1936-Fisher/Fisher_1936.ipynb

mpacer:
You said in your commit message that the block parser was splitting the equations. That couldn't be the case as there was no block parser previously. The inline lexer might have an inappropriate regex, but we should fix the _regex_ rather than reintroduce the BlockParser. 

Do you have another reason than "it's impossible" to use the BlockParser rather than an inline parser that allows for new lines in its content? That is what we do in the `latex_environment` regex, and it works without issue (otherwise the current tests would fail). It sounds like what needs to be changed is the interpretation of math block inline lexer to match something closer to the `latex_environment` variable.

mpacer:
Ok. So I have a game plan and it similar to but slightly different than what you currently have here. I have built up an additional explanation of the problem at the bottom of this comment.

My proposal is as follows: 
1. We use the `MathBlockLexer` to preëmpt other block rules from splitting up our math blocks. 
2. The processing of the `MathBlockLexer` exclusively passes the text contents of the blocks directly to the `MathInlineLexer` (without processing it). 
    - This means the `MathBlockLexer` has different regular expressions than the MathInlineLexer.
    - This means that `parse_block_math` and `parse_latex_environment` should only have `type` and `text` keys.
    - This means that `output_block_math` and `output_latex_environment` need to call `self.inline(self.token["text"])` not `self.renderer`.
    - Because the signature of the processing is simpler, it may be better to implement this as a single `multiline_math` type object that operates over a sequence of alterations that take the union of the  
3. The `MathInlineLexer` handles parsing the semantics of the expression.
    - This means that we have only one location that is doing all of our parsing.

Purpose:
The original aim of #611 was to implement a simpler logic for parsing math. In particular, it created a single code path for creating mathematical expressions. The approach outlined above solves the problem @danilobellini regarding block parsing while maintaining the uniqueness of the parsing path. This proposal additionally enables distinguishing the roles of the `MathBlockLexer` and the `MathInlineLexer` as they have completely different aims.

<details>
<summary>Explanation of the problem: from the top down(click to view)</summary>

`mistune` first parses the file into a nested hierarchy of blocks, where the parsing begins at the top level and works its way down. Only after blocks are parsed and a fairly low level is reached (e.g., a paragraph or a list item) is the inline lexer called directly.

<details>
<summary>Code for applying block-level parsing </summary>
That is what is (more or less) laid out here:

https://github.com/lepture/mistune/blob/92b7f32664bad1a4b3740ee81eda47e5246e780f/mistune.py#L1021-L1023
```python
    def output(self, text, rules=None):
        self.tokens = self.block(text, rules)
        self.tokens.reverse()
```
</details>

Inline lexers are only activated if the block level processing passes some of the contents of the block to the `self.inline()` method (which uses the `InlineLexer` as a callable class). 

<details>
<summary>Examples of where <code>self.inline()</code> is used in the <code>mistune.Markdown</code> processor</summary>
Inline parsing is only directly called inside some of the block processing methods: 
- [`Markdown().tok_text`][tok_text] 
- [`Markdown().output_heading`][output_heading]
- [`Markdown().output_paragraph`][output_paragraph]
- [`Markdown().output_open_html`][output_open_html] 
- [`Markdown().output_table`][output_table]. 
</details>

Crucially (for this bug), lists are identified based on the presence of a fairly complicated set of regexes after which the content is split into a sequence of list items (see [`BlockLexer().parse_list_block`][parse_list_block] to see the lexer logic,
 and [`Markdown().output_list`][output_list] to see where it is invoked). 

The upshot of this is we can have content that _would_ be correctly parsed if it were directly passed into the InlineLexer that still is not parsed correctly. Specifically, it is not parsed correctly because the inline content can be split apart by the block rules inherited from `mistune.BlockLexer`. 

I believe that this is what @danilobellini meant by "the lexer doesn't even see the multiple lines".
</details>

[parse_list_block]: https://github.com/lepture/mistune/blob/92b7f32664bad1a4b3740ee81eda47e5246e780f/mistune.py#L271-L319

[output_list]:
https://github.com/lepture/mistune/blob/92b7f32664bad1a4b3740ee81eda47e5246e780f/mistune.py#L1097-L1102

[output_heading]:
https://github.com/lepture/mistune/blob/92b7f32664bad1a4b3740ee81eda47e5246e780f/mistune.py#L1053-L1058

[tok_text]: https://github.com/lepture/mistune/blob/92b7f32664bad1a4b3740ee81eda47e5246e780f/mistune.py#L1041-L1045

[output_table]: 
https://github.com/lepture/mistune/blob/92b7f32664bad1a4b3740ee81eda47e5246e780f/mistune.py#L1065-L1089

[output_paragraph]: https://github.com/lepture/mistune/blob/92b7f32664bad1a4b3740ee81eda47e5246e780f/mistune.py#L1143-L1144

[output_open_html]: https://github.com/lepture/mistune/blob/92b7f32664bad1a4b3740ee81eda47e5246e780f/mistune.py#L1134-L1141

nbconvert/filters/markdown_mistune.py
```diff
@@ -21,6 +21,12 @@
 from nbconvert.filters.strings import add_anchor
 
 
+class MathBlockGrammar(mistune.BlockGrammar):
+    block_math = re.compile(r"^\$\$(.*?)\$\$|^\\\\\[(.*?)\\\\\]", re.DOTALL)
```
mpacer: As described in my comment, this should be including the delimiters in its content to be passed through to the inline parser.

mpacer: It also would make sense to give this rule a new name, since this conflicts with the current meaning of `block_math` in the `InlineLexer`. One possibility would be `multiline_math`.

nbconvert/filters/markdown_mistune.py
```diff
@@ -21,6 +21,12 @@
...
+    latex_environment = re.compile(r"^\\begin\{([a-z]*\*?)\}(.*?)\\end\{\1\}",
```
mpacer: This should not exist as a separate regular expression. 
It should be included in the previous one(suggested to be called `multiline_math`) as an alternation.

mpacer: Additionally, once it is combined with the previous one, it should include its delimiters only grab a single group.

nbconvert/filters/markdown_mistune.py
```diff
@@ -29,6 +35,31 @@ class MathInlineGrammar(mistune.InlineGrammar):
     text = re.compile(r'^[\s\S]+?(?=[\\<!\[_*`~$]|https?://| {2,}\n|$)')
 
 
+class MathBlockLexer(mistune.BlockLexer):
+    default_rules = (['block_math', 'latex_environment']
```
mpacer: reduce to one new rule.

nbconvert/filters/markdown_mistune.py
```diff
@@ -29,6 +35,31 @@ class MathInlineGrammar(mistune.InlineGrammar):
...
+        super(MathBlockLexer, self).__init__(rules, **kwargs)
+
+    def parse_block_math(self, m):
+        """Add token for a $$math$$ block"""
+        self.tokens.append({
+            'type': 'block_math',
+            'text': m.group(1) or m.group(2)
+        })
+
+    def parse_latex_environment(self, m):
```
mpacer: This method can be dropped entirely once they are combined.

nbconvert/filters/markdown_mistune.py
```diff
@@ -29,6 +35,31 @@ class MathInlineGrammar(mistune.InlineGrammar):
...
+class MathBlockLexer(mistune.BlockLexer):
+    default_rules = (['block_math', 'latex_environment']
+                     + mistune.BlockLexer.default_rules)
+
+    def __init__(self, rules=None, **kwargs):
+        if rules is None:
+            rules = MathBlockGrammar()
+        super(MathBlockLexer, self).__init__(rules, **kwargs)
+
+    def parse_block_math(self, m):
```
mpacer: Change this to have the correct name for what you decide to modify the main regex to. In the case of `multiline_math` it would be `parse_multiline_math`.

nbconvert/filters/markdown_mistune.py
```diff
@@ -29,6 +35,31 @@ class MathInlineGrammar(mistune.InlineGrammar):
...
+        """Add token for a $$math$$ block"""
```
mpacer: Docstring needs to change to be more generic and explain its role as a pass-through.

nbconvert/filters/markdown_mistune.py
```diff
@@ -29,6 +35,31 @@ class MathInlineGrammar(mistune.InlineGrammar):
     text = re.compile(r'^[\s\S]+?(?=[\\<!\[_*`~$]|https?://| {2,}\n|$)')
 
 
+class MathBlockLexer(mistune.BlockLexer):
```
mpacer: This should have a comment describing the role of the `MathBlockLexer` as a means of preëmpting mistune's block grammar rules. It should also describe its general role as a pass-through, delegating the parsing to the `MathBlockLexer`.

nbconvert/filters/markdown_mistune.py
```diff
@@ -29,6 +35,31 @@ class MathInlineGrammar(mistune.InlineGrammar):
...
+
+    def __init__(self, rules=None, **kwargs):
+        if rules is None:
+            rules = MathBlockGrammar()
+        super(MathBlockLexer, self).__init__(rules, **kwargs)
+
+    def parse_block_math(self, m):
+        """Add token for a $$math$$ block"""
+        self.tokens.append({
+            'type': 'block_math',
```
mpacer: type name will need to change to match whatever the rule is called. for example
```python
    'type': 'multiline_math'
```

nbconvert/filters/markdown_mistune.py
```diff
@@ -29,6 +35,31 @@ class MathInlineGrammar(mistune.InlineGrammar):
...
+            'text': m.group(1) or m.group(2)
```
mpacer: This will need to include the possibility of the latex_environment style multiline_math object and so will need to be:
```python
    'text': m.group(1) or m.group(2) or m.group(3)
```

nbconvert/filters/markdown_mistune.py
```diff
@@ -53,8 +84,17 @@ class MarkdownWithMath(mistune.Markdown):
...
         if 'inline' not in kwargs:
             kwargs['inline'] = MathInlineLexer
+        if 'block' not in kwargs:
+            kwargs['block'] = MathBlockLexer
         super(MarkdownWithMath, self).__init__(renderer, **kwargs)
 
+    def output_block_math(self):
+        return self.renderer.block_math(self.token["text"])
+
+    def output_latex_environment(self):
```
mpacer: Can be dropped with new approach.

nbconvert/filters/markdown_mistune.py
```diff
@@ -53,8 +84,17 @@ class MarkdownWithMath(mistune.Markdown):
     def __init__(self, renderer, **kwargs):
         if 'inline' not in kwargs:
             kwargs['inline'] = MathInlineLexer
+        if 'block' not in kwargs:
+            kwargs['block'] = MathBlockLexer
         super(MarkdownWithMath, self).__init__(renderer, **kwargs)
 
+    def output_block_math(self):
```
mpacer: This should have its name changed to map the new name (e.g., `output_multiline_math`).

nbconvert/filters/markdown_mistune.py
```diff
@@ -53,8 +84,17 @@ class MarkdownWithMath(mistune.Markdown):
...
+        return self.renderer.block_math(self.token["text"])
```
mpacer: This should be `return self.inline(self.token["text"])` which will internally handle delegation to the different renderers.

nbconvert/filters/tests/test_markdown.py
```diff
@@ -137,6 +137,18 @@ def test_markdown2html_math(self):
             "$$a<b&b<lt$$",
             "$$a<b&lt;b>a;a-b<0$$",
             "$$<k'>$$",
+            """$$x
+=
+2$$""",
+            r"""$$
```
mpacer: Does this example need to be so complicated? Please don't use an example that requires a raw literal string to be legible.

nbconvert/filters/tests/test_markdown.py
```diff
@@ -137,6 +137,18 @@ def test_markdown2html_math(self):
...
+ b_{l_1,l_2,l_3}^{\phi\phi\omega} = -8\, l_1\times l_2 \,l_1\cdot l_2 \int_0^{\chi_*} d\chi \frac{W(\chi,\chi_*)^2}{\chi^2} \int_0^{\chi}d\chi' \frac{W(\chi',\chi)W(\chi',\chi_*)}{{\chi'}^2} \left[
+ P_\psi\left(\frac{l_1}{\chi},z(\chi)\right) P_\Psi\left(\frac{l_2}{\chi'},z(\chi')\right)
+- (l_1\leftrightarrow l_2)
+\right]
+$$""",
+            r"""\begin{equation*}
```
mpacer: What does this add over first example?

https://github.com/jupyter/nbconvert/blob/3e203ce4c5989192908720e646f3c41cd2dea038/nbconvert/filters/tests/test_markdown.py#L124-L128

mpacer: It doesn't need to be a raw string so please use a standard literal string and escape anything that needs to be escaped.

nbconvert/filters/tests/test_markdown.py
mpacer: What is this example adding over the previous examples?

mpacer:
@danilobellini if you would prefer to not make these changes, let me know and I'm happy to make them for you.

danilobellini:
That's getting way too bureaucratic. The goal in #611 is the root of all evil here, please get rid from it as much as you can.

There's a bug, I want it fixed, else the nbviewer remains unuseful for me. The GitHub parsing for notebooks isn't perfect when it comes to LaTeX (`\lambda`, `\require`, `\beta` and `\pmb` don't get shown correctly) but it at least doesn't break completely when I put a LaTeX block that has an unpretentious line starting with `-` or `=`, nor when an environment has the `*` symbol like in `\begin{matrix*}` (the 3 tests I've included in this PR that breaks nbconvert). I'm here to help, not to complain, that's why I wrote a fix, I'm not asking for help.

I wrote a humble fix, a single small commit, no style-changing lines, no deletions, just some insertions. I'm okay if you tell any sound reason to update it, but I fear that's still not the case.

> [Goal] The MathInlineLexer handles parsing the semantics of the expression.
> This means that we have only one location that is doing all of our parsing.

Actually, you clearly want to do the opposite: make the block parser result get parsed *again* by an inline parser. Your goal is to join all the "HTML generation" stuff together in a single method, but I don't see how that could help: the inline math token and the actual math "block" are indeed distinct things (unless you think on the actual math block as a "raw text block without text"), and there's still a bug with that inline parsing (non-math blocks that has a multiline math "token" that gets broken in multiple blocks by the mistune block grammar), something I don't know an easy way to fix in the general case.

Please forget about #611, it's not helping at all, on the contrary: until now, it just summoned a bug. The code should work fine, it doesn't matter if you minimize some abstract complexity measurement if the code don't do what it should do. Perhaps a general solution to this family of bugs where're talking about is to rewrite the whole mistune block grammar so that it won't ever break a math "token" (be it inline or multiline), but I don't see how joining/housekeeping regexes regarding a new [perhaps twice-parsed] "math block" can help on that.

> Does this example need to be so complicated?
> [...]
> What is this example adding over the previous examples?

The tests I've included are exactly the examples from #404 that used to fail. You agree that they test something that wasn't being tested before, don't you? Did you try to run these expressions on the previous version? These questions made me wonder if you're really understanding what's going on here.

> Whatever its purpose; it doesn't need to be a raw string so please use a standard literal string and escape anything that needs to be escaped.

I REALLY prefer raw strings for regexes and stuff with lots of `\`, like LaTeX code. Readability becomes really bad when you need to *escape the escape symbol* that appears twice in a row, why should one prefer avoiding that? The `r"^\$\$(.*?)\$\$|^\\\\\[(.*?)\\\\\]"` regex can be written without that leading `r`, but I don't think that would be any better.

You should see the [mistune contrib](https://github.com/lepture/mistune-contrib). It has a [seemingly incomplete] math block implementation example for mistune. Do you think mistune-contrib should be updated?

mpacer:
@danilobellini, I think that you should take some time to look at the Jupyter code of conduct: https://github.com/jupyter/governance/blob/master/conduct/code_of_conduct.md

I was originally hesitant to engage with this PR because of the tone you were taking. But it was within reason and I attempted to interpret your words generously. I attempted to be polite and explain where I was coming from. However your further response is even more extreme on a number of dimensions. In my opinion, your response is not respectful, considerate, understanding, friendly or patient. 

Indirectly, it's also not considerate of other people's time in the project. Part of the reason for the original changes is to make the code easier for other people in the project to read about and debug. You may think your change is simple, but if it makes it more complicated to reason about in the long run it is inappropriate for this code base.

I'm not going to close this PR yet, but if you do not wish to pursue the changes I have suggested I will open a new PR taking my preferred approach that will address this issue. If so, that will likely result in the closing of this PR.

In any case, would you please explain why each of those test cases adds something new on top of those that were previously there. I appreciate you taking the time to make this contribution and answering that question will be helpful in moving forward, regardless, on whether progress continues on this PR or elsewhere.

danilobellini:
As it won't be merged, I've deleted the branch. Please consider your tone, I was trying to help.

mpacer:
I had accidentally closed this while scrolling on my phone I'm impressed that you deleted your branch so quickly but if you would like to pursue a fix, I'm happy to reopen.

mpacer:
Could you point out what in my tone was problematic? If there is anything I would want to fix it.

mpacer:
Also, to  point out, you already *have* helped me understand this problem. If it weren't for you we still wouldn't know how to approach this bug. Your commits were good starts too!

takluyver:
It sounds like we do need a custom block lexer to get this right, one way or another. Do you want to prototype the two different proposed approaches and see how they compare?

Our math parsing doesn't do anything that complex - it just needs to preserve the contents of the math chunk unaltered so that mathjax can render it. In a situation like that, I think that a modest amount of duplication can make the code easier to understand than trying to apply DRY strictly. But it's quite a long time since I looked closely at this code, so I'd like to see what @mpacer is suggesting in another PR and compare it against this version.

danilobellini:
I'm not at home, even if I fork this project again, I can't recreate the branch without my computer, and I can't link this PR with a new branch of a new fork. This thread lost its sense. If you wish, I can make a new PR at home.

@mpacer, please try the examples in #404 and you'll have an answer you asked for. Actually, there's even a screenshot there, and there's also probably the right/best place for asking explanation about what the bug is all about.

takluyver:
The code's still there - once you've made a pull request, you can't entirely delete it, I think. ;-) I'll open a new PR with it so that I can look over it a bit more.

danilobellini:
> Could you point out what in my tone was problematic?

People trying to collaborate with a project aren't *puppets* that need to *obey* to everything, you should accept that collaborators/contributors sometimes have distinct ways of implementing stuff, and there's nothing wrong with that.

> However your further response is even more extreme on a number of dimensions

That bug *is* extreme. Not the people. As of today, I can't depend on nbviewer, that's extreme, I agree, but that's the truth. The bug appeared due to an aesthetic goal, and I strongly criticized that goal. The goal, not the people. I was addressing what you said in the context of trying to get this nasty bug fixed, not you as a person.

blink1073:
@danilobellini, I appreciate your position but statements like "These questions made me wonder if you're really understanding what's going on here" are what the code of conduct is intended to prevent.

takluyver:
@danilobellini any bug can be deeply annoying when it gets in the way of what you want to do, but I don't think it's reasonable to characterise this as 'extreme'. Apple's recent "log in as root with no password" bug was extreme. Certain kinds of math expressions not being rendered correctly is not.

Remember that strong criticism of a piece of work can spill over into implicitly criticising the person who did that work. If you say that a PR is the 'root of all evil', it's hard not to read that as a judgement of the person who made the PR. We're all more productive if we keep our criticism gentle and try to work together.

I've reopened the PR as #716. I hope we can draw a line under the argument and focus on fixing the bug. :-)

danilobellini:
@blink1073 maybe, but that's what I strongly believed, as the referenced questions were about the examples shown in the bug issue, not really about what I did. That wasn't intended as a personal criticism, as the question was about the bug itself, I wondered if the bug was being properly understood.

@takluyver, thanks. You're right. For me it's extreme as I strongly depend on math expression rendeding, but that's not the same for someone who don't require math expressions at all. I hope the same.

takluyver:
This might be a nuance that's less obvious if English isn't your first language, but saying "I wonder if you really understand what's going on" carries a more combative overtone than the literal meaning of the words.

sccolbert:
> People trying to collaborate with a project aren't puppets that need to obey to everything, you should accept that collaborators/contributors sometimes have distinct ways of implementing stuff, and there's nothing wrong with that.

There's absolutely nothing wrong with you implementing your *own* software in a way that you see fit. 

However, when you contribute to a project that you haven't written, it's not unreasonable for the maintainer of that project to have criteria for commits that maintains both the quality of the code base and the sanity of the maintainer. 

The maintainers of the Jupyter ecosystem are largely unpaid/underpaid volunteers. To be clear, you have paid nothing for the benefit they have provided you.

Notwithstanding the language barrier, I would suggest that you take a few steps back and re-evaluate your approach to engaging with the core contributors. As the saying goes: you'll catch more flies with honey than vinegar.

danilobellini:
@sccolbert, unnecessary. I'm not getting paid, as well.

Actually, I already contributed to IPython, and when @fperez came to Brazil as a keynote in a Python Brasil, he invited me personally to collaborate, and even to propose some research using Jupyter/IPython that could have been paid. He even paid an ice cream to me in that occasion. Money isn't the issue here, but it's mostly the other way around: I'm not getting paid, but some people in the Jupyter history got paid.

On the other hand, even people who actually get paid for a job aren't *puppets*. They have to do the job they proposed to do, not anything.

Conclusion: Getting paid is orthogonal to everything we're talking about.

danilobellini:
I should have said that it was strange to be reviewed by the person who made the PR with the goal that created the bug and couldn't be kept if we wish to fix it. But I tried to say that such goal was the problem, not the person himself.

I can wait many people to take a few steps back and re-evaluate what happened here, but that might never happen, at least not here, and I shouldn't ask nor depend on that. Unfortunately, the focus was never the bug. The focus here seem to be: how do we make this guy never try to collaborate again?

I didn't say anything about the skills of anyone, nor about any person himself (is there any traits?). I don't think I violated anything in the code of conduct, and every time I read what I wrote I'm getting more confident of that. I don't think I did anything wrong or personally offensive, but I'm certain that this is getting way too unfair against me. Personally, not just about what I wrote. And now I wonder if the code of conduct is protecting this behavior.

Is there a way to close this thread for new comments?

takluyver:
@danilobellini it's not our intention to stop you collaborating, but you approached this in what we see as an unnecessarily confrontrational way. It's entirely possible to say that a particular PR introduced a bug without saying that @mpacer's goal is 'the root of all evil' or describing it as 'aesthetic' (which is an implicit insult among engineers, because aesthetics are typically considered unimportant).

The thing about not annoying contributors goes both ways: we want you to be happy to contribute, but we also want @mpacer to feel happy contributing, and in this case that meant asking you to check your behaviour. Maybe it is partly due to cultural differences in how we express ourselves - I've experienced this before with collaborators from other cultures - but *impact* matters more than *intent*.

@blink1073 has beaten me to locking the conversation. Please don't take this as an opportunity to start it again on another issue. If you feel that we've treated you unfairly, I'd suggest you contact Fernando privately to try to resolve it.

fperez:
I was notified of this thread yesterday and at one point it appeared that it might be heading towards a calmer resolution, but I see now there's been more activity while I was offline. I'd like to point out a few things:

- we, as a team, absolutely want you to collaborate with the project.  Everything folks here have done is to encourage you to do that, including spending substantial time reviewing your code, and providing specific feedback from the perspective of the core maintainers.

- but we also, unambiguously, want those contributions to happen in a form that makes the project a welcoming and productive space for all involved, both newcomers and long-standing members for whom this is a full-time or nearly full-time activity.

- You may not see it that way, but calling a maintainer's work "the root of all evil" is most certainly offensive and an attack on their work.  A bug may well have been introduced, and that happens *all the time*.   We make mistakes, we fix them, the cycle of development repeats and continues.  But that does not, in any way, justify attacking someone

- Similarly, opening a comment with accusations of bureaucracy is similarly setting a confrontational tone towards a project team, who have an established workflow for code review and integration (and one that is in fact much lighter than that of other projects).

I recommend that you read [Bret Cannon's excellent post](https://nothingbutsnark.silvrback.com/why-i-took-october-off-from-oss-volunteering) on burnout among even paid OSS contributors. We're trying to create an atmosphere that makes it sustainable for all of us in the long run, and many of the issues Brett raises there are at the heart of this discussion.

I'm happy to talk to you privately over email if you'd like.  But it's important for me to convey these points also publicly.  I hope they will help you understand why multiple project maintainers (me included) are responding to you in this way, so that you can have productive interactions not only with Jupyter, but also with many other OSS projects, for a long time to come.

PR #716, opened 2017-12-06
Add Markdown block lexer/parser for LaTeX blocks

takluyver:
Resurrection of PR #715, because I want to be able to easily compare it with @mpacer's proposal.

See #404 for rationale.

CC @danilobellini


takluyver:
On the technical merits, it was a close call between this and #717, but I marginally preferred that one. It was based on this branch, however, so both pull requests now show as merged.

PR #717, opened 2017-12-06
Multiline math

mpacer:
This is a modification of the logic #715 (now #716) that takes the approach I suggested in https://github.com/jupyter/nbconvert/pull/715#issuecomment-349505394. 

In particular it uses the MathBlockLexer as a way to pass the multiline sections of math blocks through to the MathInlineLexer, which will do the actual parsing.

I don't want this to be merged until we can figure out how to simplify the test examples as they are making a set of already hard-to-read tests even more hard-to-read.

I also really want to know why our previous test with a LaTeX environment with a `*` in its name (the first instance of the test) didn't work to catch this. 
Specifically, that test case is:
https://github.com/jupyter/nbconvert/blob/3e203ce4c5989192908720e646f3c41cd2dea038/nbconvert/filters/tests/test_markdown.py#L124-L128

### DIFF
--- a/nbconvert/filters/markdown_mistune.py
+++ b/nbconvert/filters/markdown_mistune.py
@@ -60,7 +60,7 @@ class MathBlockLexer(mistune.BlockLexer):
         """Add token to pass through mutiline math."""
         self.tokens.append({
             "type": "multiline_math",
-            "text": m.group(1) or m.group(2) or m.group(3)
+            "text": m.group(0)
         })
 
 
--- a/nbconvert/filters/markdown_mistune.py
+++ b/nbconvert/filters/markdown_mistune.py
@@ -21,12 +21,43 @@ from pygments.util import ClassNotFound
 from nbconvert.filters.strings import add_anchor
 
 
+class MathBlockGrammar(mistune.BlockGrammar):
+    block_math = re.compile(r"^\$\$(.*?)\$\$|^\\\\\[(.*?)\\\\\]", re.DOTALL)
+    latex_environment = re.compile(r"^\\begin\{([a-z]*\*?)\}(.*?)\\end\{\1\}",
+                                   re.DOTALL)
+
+
 class MathInlineGrammar(mistune.InlineGrammar):
     inline_math = re.compile(r"^\$(.+?)\$|^\\\\\((.+?)\\\\\)", re.DOTALL)
     block_math = re.compile(r"^\$\$(.*?)\$\$|^\\\\\[(.*?)\\\\\]", re.DOTALL)
     latex_environment = re.compile(r"^\\begin\{([a-z]*\*?)\}(.*?)\\end\{\1\}",
                                    re.DOTALL)
     text = re.compile(r'^[\s\S]+?(?=[\\<!\[_*`~$]|https?://| {2,}\n|$)')
+
+
+class MathBlockLexer(mistune.BlockLexer):
+    default_rules = (['block_math', 'latex_environment']
+                     + mistune.BlockLexer.default_rules)
+
+    def __init__(self, rules=None, **kwargs):
+        if rules is None:
+            rules = MathBlockGrammar()
+        super(MathBlockLexer, self).__init__(rules, **kwargs)
+
+    def parse_block_math(self, m):
+        """Add token for a $$math$$ block"""
+        self.tokens.append({
+            'type': 'block_math',
+            'text': m.group(1) or m.group(2)
+        })
+
+    def parse_latex_environment(self, m):
+        """Add token for a \begin{.} ...LaTeX stuff... \end{.} block"""
+        self.tokens.append({
+            'type': 'latex_environment',
+            'name': m.group(1),
+            'text': m.group(2)
+        })
 
 
 class MathInlineLexer(mistune.InlineLexer):
@@ -53,7 +84,16 @@ class MarkdownWithMath(mistune.Markdown):
     def __init__(self, renderer, **kwargs):
         if 'inline' not in kwargs:
             kwargs['inline'] = MathInlineLexer
+        if 'block' not in kwargs:
+            kwargs['block'] = MathBlockLexer
         super(MarkdownWithMath, self).__init__(renderer, **kwargs)
+
+    def output_block_math(self):
+        return self.renderer.block_math(self.token["text"])
+
+    def output_latex_environment(self):
+        return self.renderer.latex_environment(self.token["name"],
+                                               self.token["text"])
 
 
 class IPythonRenderer(mistune.Renderer):
--- a/nbconvert/filters/markdown_mistune.py
+++ b/nbconvert/filters/markdown_mistune.py
@@ -22,9 +22,10 @@ from nbconvert.filters.strings import add_anchor
 
 
 class MathBlockGrammar(mistune.BlockGrammar):
-    block_math = re.compile(r"^\$\$(.*?)\$\$|^\\\\\[(.*?)\\\\\]", re.DOTALL)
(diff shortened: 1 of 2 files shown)


nbconvert/filters/markdown_mistune.py
```diff
@@ -21,15 +21,58 @@
 from nbconvert.filters.strings import add_anchor
 
 
+class MathBlockGrammar(mistune.BlockGrammar):
+    """This defines a single regex comprised of the different patterns that 
+    identify math content spanning multiple lines. These are used by the 
+    MathBlockLexer.
+    """
+    multi_math_str = "|".join([r"(^\$\$.*?\$\$)",
```
takluyver: Do you know about the `re.VERBOSE` option? It lets you split a regex over several lines and have comments, without needing to assemble the string yourself.

nbconvert/filters/markdown_mistune.py
```diff
@@ -53,8 +96,14 @@ class MarkdownWithMath(mistune.Markdown):
     def __init__(self, renderer, **kwargs):
         if 'inline' not in kwargs:
             kwargs['inline'] = MathInlineLexer
+        if 'block' not in kwargs:
+            kwargs['block'] = MathBlockLexer
         super(MarkdownWithMath, self).__init__(renderer, **kwargs)
 
+    
+    def output_multiline_math(self):
+        return self.inline(self.token["text"])
```
takluyver: Does this mean we're parsing it again with the inline parser? Can we skip that, given that we just want to put it in the output unmodified?

nbconvert/filters/markdown_mistune.py
mpacer: Could you give an example in this case? 

Because I genuinely prefer reading individual regexes joined with a pipe to indicate alternations to one long regex that is individually difficult to parse.

nbconvert/filters/markdown_mistune.py
mpacer: We're parsing it semantically for the first time in the inline parser. We actually aren't putting it in the output unmodified, because we need to escape the individual parts of it separately which occurs in the rendering step.

nbconvert/filters/markdown_mistune.py
takluyver: Something like this:

```python
multiline_math = re.compile(r"""(^\$\$.*?\$\$)|
                                (^\\\\\[.*?\\\\\])|
                                (^\\begin\{([a-z]*\*?)\}(.*?)\\end\{\4\})""",
                            re.DOTALL | re.VERBOSE)
```

Personally I prefer that to the string-joining approach, but you're the one writing it, so you get to decide on code style.

nbconvert/filters/markdown_mistune.py
```diff
@@ -21,15 +21,58 @@
...
+    def __init__(self, rules=None, **kwargs):
+        if rules is None:
+            rules = MathBlockGrammar()
+        super(MathBlockLexer, self).__init__(rules, **kwargs)
+
+    def parse_multiline_math(self, m):
+        """Add token to pass through mutiline math."""
+        self.tokens.append({
+            "type": "multiline_math",
+            "text": m.group(1) or m.group(2) or m.group(3)
```
takluyver: Since the three groups are each the whole of their alternative, this is equivalent to `m.group(0)`, right?

nbconvert/filters/markdown_mistune.py
takluyver: Gotcha, thanks.

nbconvert/filters/markdown_mistune.py
mpacer: Nice! Much cleaner, grazie.

nbconvert/filters/markdown_mistune.py
mpacer: Interesting, in this case I find them equally easy to read. it's more in cases like this in mistune https://github.com/lepture/mistune/blob/92b7f32664bad1a4b3740ee81eda47e5246e780f/mistune.py#L120-L134 where individual chunks need to split multiple lines, and accordingly detecting that an alternation is happening is a lot harder.

mpacer: Since I find these to be equally readable, I'm going to go with your style in this case. But I also figured out a way to simplify it further (since we don't really need the groups for grabbing the content, as you pointed out).

mpacer: can I ask your rationale for disliking the string joining approach?

mpacer: Ok… actually I'm going to keep using the join method, but with simpler component strings. 

I ended up not finding the pipe's use for alternations and the bitwise OR to be equally readable as not needing the second flag. I had skimmed over that detail at a first glance.

nbconvert/filters/markdown_mistune.py
```diff
@@ -21,7 +21,42 @@
...
+    def __init__(self, rules=None, **kwargs):
+        if rules is None:
+            rules = MathBlockGrammar()
+        super(MathBlockLexer, self).__init__(rules, **kwargs)
+
+    def parse_multiline_math(self, m):
+        """Add token to pass through mutiline math."""
+        self.tokens.append({
+            "type": "multiline_math",
+            "text": m.string
```
takluyver: I think this should be `m.group(0)`, which is the full match, not `m.string`, which is the string it tried to match against.

```
 3│ re.match(r'\d+-', '12-AB').string
 3> '12-AB'

 4│ re.match(r'\d+-', '12-AB').group(0)
 4> '12-'
```

I haven't checked how mistune works, but `m.string` could be wrong, whereas `m.group(0)` can't be. So let's go with the more robust option.

mpacer:
Ok simplified the tests, escaped them and fixed the m.string thing.

nbconvert/filters/tests/test_markdown.py
```diff
@@ -137,6 +137,18 @@ def test_markdown2html_math(self):
             "$$a<b&b<lt$$",
             "$$a<b&lt;b>a;a-b<0$$",
             "$$<k'>$$",
+            ("$$x\\n"
```
takluyver: Shouldn't these be actual newlines, rather than making a string with a backslash followed by a lowercase n? You don't type `\n` in markdown, as far as I'm aware. What am I missing?

mpacer:
These are encoded as strings now, not raw strings. Additionally they are
implicitly joined, not in triple quotes. If you look at the other examples,
except for the last example, this style is more in keeping with the rest of
our tests. This should be functionally identical in practice to the raw
triple quoted string with literal new lines.

Honestly, I was doing it to keep the style consistent with what was already
there.

There were reasons that I originally went along with that style long ago,
because of some 3rd party regex tools (e.g., regex101.com). I'd need to
reevaluate the state of the art to determine if those issues were still
constraints & applicable.

If you'd prefer them to be multi line we should make all the old multi-line
strings into literal multi-line strings and make that our uniform style for
those tests.

takluyver:
I'm happy with implicitly joined strings like this. But I think they should be e.g. `$$x\n` rather than `$$x\\n`. The former makes a string which ends with a newline (because Python's parser interprets `\n` as such), whereas the latter makes a string which ends with the characters `\n`, because the backslash is escaped.

See e.g. line 129 in the same file for an example of what I mean.

mpacer:
Oops! That's what I get for "s/\\/\\\\/g"ing will fix in a bit.

mpacer:
Ok… force pushed onto the last bit since this conceptually seemed like part of the immediately previous commit.

PR #724, opened 2017-12-26
Added Ubuntu Linux Instructions

yaaqov:
I've linked to a TeX Stackexchange article which has a straightforward answer on installing Xelatex on Ubuntu Linux (essential for saving Jupyter notebooks as PDFs).

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -57,6 +57,7 @@ Fortunately, there are packages that make this much easier. These
 packages are specific to different operating systems: 
 
 * Linux: `TeX Live <http://tug.org/texlive/>`_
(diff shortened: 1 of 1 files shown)


mpacer:
I think it’d be better to just mention the content directly; though it shouldn’t need sudo(I think?).

takluyver:
`apt` commands do indeed need sudo, because they act systemwide. Without sudo, you can usually only modify your home directory, `/tmp` and $XDG_RUNTIME_DIR.

takluyver:
But since the answer is basically one command, I'd agree that we can inline it rather than linking to stackexchange.

mpacer:
@yaaqov Could you change this to give the instructions directly rather than linking to the stack exchange answer?

mpacer:
I’m going to push this change into this branch in a bit.

takluyver:
Pushed the change, merging now. Thanks!

PR #725, opened 2017-12-28
Update Math Jax compatibility definitions

### LINKED ISSUE #719: Plain Text \LaTeX converted to logo.
Similar to issue #262 and its fix #687. However this problem predates that fix.

"\LaTeX" does not get converted into the logo in the markdown of the notebook, but the latex compiler interprets it upon conversion to pdf.

I already have an idea for a solution. I propose to update the mathjax compatibility definitions defined in #687 by replacing:
```
\renewcommand{\TeX}{\textrm{\Oldtex}} 
\renewcommand{\LaTeX}{\textrm{\Oldlatex}} 
```
with:
```
\renewcommand{\TeX}{\ifmmode \textrm{\Oldtex} \else \textbackslash TeX \fi}
\renewcommand{\LaTeX}{\ifmmode \Oldlatex \else \textbackslash LaTeX \fi}
```
This exactly replicates the use of "\LaTeX", "\TeX", "$\LaTeX$", and "$\TeX$" in the notebook. So plain text and the logo can be written in both the notebook and the converted pdf without any conversion error. And no the fact that "\textrm{}" only wraps one of those commands is not an error, since by default "\LaTeX" calls "\TeX".

t-makaro:
Fixes #719

### DIFF
--- a/nbconvert/templates/latex/base.tplx
+++ b/nbconvert/templates/latex/base.tplx
@@ -135,13 +135,13 @@ This template does not define a docclass, the inheriting class must define this.
     % Define a nice break command that doesn't care if a line doesn't already
     % exist.
     \def\br{\hspace*{\fill} \\* }
-    % Math Jax compatability definitions
+    % Math Jax compatibility definitions
     \def\gt{>}
     \def\lt{<}
     \let\Oldtex\TeX
     \let\Oldlatex\LaTeX
-    \renewcommand{\TeX}{\textrm{\Oldtex}}
-    \renewcommand{\LaTeX}{\textrm{\Oldlatex}}
+    \renewcommand{\TeX}{\ifmmode \textrm{\Oldtex} \else \textbackslash TeX \fi}
+    \renewcommand{\LaTeX}{\ifmmode \Oldlatex \else \textbackslash LaTeX \fi}
     % Document parameters
     % Document title
     ((* block title -*))


mpacer:
I do not want to reproduce notebook behavior in this regard. I’d prefer to get to notebook to behave like nbconvert if we are to make it consistent.

mpacer:
Thank you though for taking the time to solve this in this way. Happy to help figure out how to fix it on the notebook side of things.

t-makaro:
My main worry with changing the behaviour on the notebook's side to match nbconvert was losing the ability to right "\LaTeX" in plain text at all. However, upon testing writing "\\\LaTeX" renders "\LaTeX" in markdown as plain text, and properly converts to "\textbackslash LaTeX" in LaTeX which renders as plain text too. So long as a solution on the notebook side preserves this I'll be happy.

mpacer:
That’s exactly what I was going to suggest, so I’m happy that it works correctly!

eldad-a:
I believe the following issue is related, though may be too minor to merit its own issue (and Devs time).

In JupyterLab, the \LaTeX sign would not render correctly as part of a header.

The following code in a Markdown cell  does show up as the LaTeX symbol: 

```markdown
# SymPy $\to \LaTeX$
```

If however, a character is added in the preceding line, it does

```markdown
.

# SymPy $\to \LaTeX$
```

Issue #726, opened 2017-12-29
AttributeError with an empty math block 


pacahon:
With command `jupyter nbconvert --to html ./test.ipynb`  and `test.ipynb` source:

```
{
 "cells": [
  {
   "cell_type": "markdown",
   "metadata": {},
   "source": [
    "$$$$"
   ]
  }
 ],
 "metadata": {
  "kernelspec": {
   "display_name": "Python 3",
   "language": "python",
   "name": "python3"
  },
  "language_info": {
   "codemirror_mode": {
    "name": "ipython",
    "version": 3
   },
   "file_extension": ".py",
   "mimetype": "text/x-python",
   "name": "python",
   "nbconvert_exporter": "python",
   "pygments_lexer": "ipython3",
   "version": "3.6.1"
  }
 },
 "nbformat": 4,
 "nbformat_minor": 2
}
```
we get an exception
```
...
  File "/Users/pacahon/.pyenv/versions/3.5.3/lib/python3.5/cgi.py", line 1056, in escape
    s = s.replace("&", "&amp;") # Must be done first!
AttributeError: 'NoneType' object has no attribute 'replace'
```
The reason in `block_math` regex https://github.com/jupyter/nbconvert/blob/f8498361cb3d8ae7d12549761148f4a5a413f7b5/nbconvert/filters/markdown_mistune.py#L60-L61

I guess `IPythonRenderer.block_math` is the most appropriate place to check for `None` if we don't want to have `$$$$` in output.

mpacer:
I think what would be better would be to have it return an empty string I’ll need to look into why it’s not doing that.

pacahon:
Check this, please https://github.com/jupyter/nbconvert/blob/f8498361cb3d8ae7d12549761148f4a5a413f7b5/nbconvert/filters/markdown_mistune.py#L88
In case of "$$$$" it returns the result of the second expression (which is None) since empty string is Falsy.

mpacer:
Perfect! That's easy to fix!

We can add on another `or ""` since with python, a sequence of `or`s will return the last element if the prior two are Falsy, e.g.,

```python
>>> print(None or "" or {})
{}
>>> print(None  or {} or "")

>>> print({} or "" or None)
None
```

pacahon:
@mpacer Hello! Can you take a look at https://github.com/jupyter/nbconvert/pull/729?

mpacer:
Yep!

PR #729, opened 2018-01-09
fix for an issue with empty math block

pacahon:
https://github.com/jupyter/nbconvert/issues/726

### DIFF
--- a/nbconvert/exporters/tests/files/notebook2.ipynb
+++ b/nbconvert/exporters/tests/files/notebook2.ipynb
@@ -172,6 +172,16 @@
     "\n",
     "one *test* two *tests*. three *tests*"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Make sure markdown parser doesn't crash with empty Latex formulas blocks\n",
+    "$$$$\n",
+    "\\[\\]\n",
+    "$$"
+   ]
   }
  ],
  "metadata": {
--- a/nbconvert/exporters/tests/files/notebook2.ipynb
+++ b/nbconvert/exporters/tests/files/notebook2.ipynb
@@ -172,6 +172,13 @@
     "\n",
     "one *test* two *tests*. three *tests*"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Make sure markdown parser doesn't crash with empty Latex formulas blocks\n$$$$\n$$"
+   ]
   }
  ],
  "metadata": {
--- a/nbconvert/exporters/tests/files/notebook2.ipynb
+++ b/nbconvert/exporters/tests/files/notebook2.ipynb
@@ -172,13 +172,6 @@
     "\n",
     "one *test* two *tests*. three *tests*"
    ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "Make sure markdown parser doesn't crash with empty Latex formulas blocks\n$$$$\n$$"
-   ]
   }
  ],
  "metadata": {
--- a/nbconvert/filters/markdown_mistune.py
+++ b/nbconvert/filters/markdown_mistune.py
@@ -134,7 +134,7 @@ class IPythonRenderer(mistune.Renderer):
         return cgi.escape(text)
 
     def block_math(self, text):
-        return '$$%s$$' % self.escape_html(text) if text else ''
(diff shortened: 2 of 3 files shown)


mpacer:
So, it looks like your expected output for the modified test does not match the output that we should expect — which makes me think we need to reëvaluate that test, because it's not raising the error you mentioned in #726.

pacahon:
@mpacer Not sure I got it. I added smoke test to make sure `AttributeError` doesn't occur with `$$$$` in markdown cell. What should I try to do?

pacahon:
@mpacer I guess it's better now, but still little bit odd. With `$$$$` input we get `""` while with `$$ $$` we get the same string `$$ $$`. Maybe trim text inside math block?

mpacer:
Ok, new thought, i don't know why it wasn't succeeding in the first place when it ran into a `$$$$` since the pattern it's matching is `re.compile(r"^\$\$(.*?)\$\$|^\\\\\[(.*?)\\\\\]", re.DOTALL)` which should return the empty string when we get the groups from that.

mpacer:
Ok, thinking this through a bit more… at the point that we're returning the value, the `$$$$` is actually being normalized by us, so at the _least_ we can do `if text else "$$$$"`. But I'm still a little confused as to why we even need to do this in the first place.

mpacer:
Ok, did a couple of quick tests to make sure that I was following this, we need the "" as a fallback in the case where _either_ `$$$$` or `\\[\\]` fails. Right now, `\\[\\]` will work and if we swapped the order (e.g., `m.group(2) or m.group(1)` then `$$$$` would work and `\\[\\]` would fail). 

Since it was first triggered by the grammar match, we don't have to worry about returning an empty string in an inappropriate context.

So we can safely always return a fallback empty string, and then if we pass the empty string, it should correctly be interpreted as `$$$$`, which is what would be returned by return `'$$%s$$' % self.escape_html(text)`  where `text==""`.

So, the question remains: why did your the test you introduced in 4f23786 fail — it should have been passed through without issue.

mpacer:
@pacahon 

Ok, just double checked it; the issue with your initial commit was that you represented the multiline string as a single line. My guess is you were editing the file as a `JSON` object in a text editor as opposed to creating a new notebook. For the purposes of our nbconvert test files, we are expecting to see a list of strings as the on-disk representation of notebook cells (which is why you had the error you ran into). 

Anyway I created an example of how a notebook's cell should be formatted to pass the tests based on just opening and saving the notebook you modified in 4f23786.

https://github.com/mpacer/nbconvert/commit/9afbb097a76a77e938d741a14a572c10caf1f553

pacahon:
@mpacer Sorry, misunderstood your comment here https://github.com/jupyter/nbconvert/issues/726#issuecomment-354498727 Thought it was about `block_math` method, but looks like you told only about `output_block_math`. In this case I shouldn't edit `block_math`, of course. Anyway, since the problem related to renderer, isn't `test_markdown` is better place for this test?

pacahon:
@mpacer Travis tests were failed with 

```
The command "git clone --quiet --depth 1 https://github.com/minrk/travis-wheels travis-wheels" failed and exited with 128 during .
```
Could you please restart them? I didn't figure out how to do it.

takluyver:
I've restarted the failing jobs (@mpacer is probably asleep at the moment).

pacahon:
Thanks a lot!

mpacer:
Apologies @pacahon, @takluyver  was correct, I was asleep. 

So… if our previous behaviour was to drop them, then we could continue to do that to be consistent with that. I'm not sure which is the desired behaviour….

pacahon:
@mpacer By "them" you mean "$$$$"?
I think it's ok to leave "$$$$" in rendered markdown as is. Later it will be rendered as empty line by mathjax and user should decide how to deal with it since he added empty math block for some purpose.

Btw previous behaviour was `AttributeError` if I'm not wrong. Or you mean something different?

mpacer:
I meant the previous behaviour as in removing `$$$$` since you needed to modify the test file to get it to pass. Somehow it wasn't triggering the error… 

Anyway I'm going to merge this and say it's fixed, and we'll iterate on this again if it comes up.

pacahon:
Thanks. Since I made a mess in commit history, I little bit curious is it possible to squash commits when merging a PR?

takluyver:
It is, though in this case @mpacer did a conventional merge instead. Here's what the interface looks like:

screenshot from 2018-01-15 11-44-25

pacahon:
@takluyver Thanks for the clarification.

Issue #702, opened 2017-11-08
Locally hosted reveal.js leads to blank page when serving without `--post`


fperez:
When converting to slides and serving with `--post serve`, things work fine. But if the same exact HTML file is then served locally, even with a local copy of reveal.js available:

```
alpamayo[jslides]> d
/Users/fperez/dev/jslides
total 704
drwxr-xr-x@ 11 fperez     374 Nov 10  2012 reveal.js/
-rw-r--r--   1 fperez   51552 Oct 27 16:26 simple-test.ipynb
-rw-r--r--   1 fperez  304586 Nov  7 18:12 simple-test.slides.html
```

the resulting HTML page is blank, with the following error on the JS console:

![image](https://user-images.githubusercontent.com/57394/32533027-13c2af7e-c403-11e7-875f-7eeb1667a4d2.png)

In a comment to [issue 91](https://github.com/jupyter/nbconvert/issues/91#issuecomment-283736634), an explicit solution to this problem is provided, and I tested it works: adding the option `--reveal-prefix "https://cdnjs.cloudflare.com/ajax/libs/reveal.js/3.1.0"`.  But as mentioned in #296, we want to make it possible to easily serve slides fully offline too, so this workaround (while useful for some cases) isn't a full solution.

Additionally, the [current instructions](http://nbconvert.readthedocs.io/en/latest/usage.html#convert-revealjs) say that having reveal.js in the current directory works, which at least in my case (and confirmed by @Carreau and @pbstark, who originally reported the problem), it doesn't.

So I think we need to:

- Improve [the instructions](http://nbconvert.readthedocs.io/en/latest/usage.html#convert-revealjs) to clarify exactly what to do for a locally hosted reveal.js, that actually works (I tried multiple flavors of `--reveal-prefix` with `.`, `./`, `./reveal.js`, and none worked).

- Also provide the explicit tip of using the CDN path (`--reveal-prefix "https://cdnjs.cloudflare.com/ajax/libs/reveal.js/3.1.0"`), which is confirmed to work and for some cases works.  I can make a PR for this if it's confirmed it's the right thing to do.

- Figure out fully offline hosting with clear instructions (essentially #296).

mpacer:
@damianavila do you have any thoughts on how to best address this?

fperez:
[...]

PR #732, opened 2018-01-10
change default for slides to direct to the reveal cdn rather than locally

mpacer:
This is the quick fix suggested by @fperez (to change the default) for #702. 

My intuition is that we should provide a way to to create slides in a "stand-alone" mode.  That mode will actually solve the _actual_ problem with local copy of reveal that we're running into in #702.
That would probably also download the relevant copy of the reveal library into the output directory and maybe a bunch more.

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -92,31 +92,66 @@ Reveal.js HTML slideshow
 * ``--to slides``
 
   This generates a Reveal.js HTML slideshow.
-  It must be served by an HTTP server. The easiest way to do this is adding
-  ``--post serve`` on the command-line. The ``serve`` post-processor proxies
-  Reveal.js requests to a CDN if no local Reveal.js library is present.
-  To make slides that don't require an internet connection, just place the
-  Reveal.js library (version 3.x) in the same directory where ``your_talk.slides.html`` is
-  located, or point to another directory using the ``--reveal-prefix`` alias.
+  
+Running this slideshow requires a copy of reveal.js (version 3.x).
+  
+By default, this will include a script tag in the html that will directly load 
+reveal.js from a CDN.
 
-  To get a compatible version of revealjs in the current folder:
+However, some features (specifically, speaker notes) are only available if you
+use a local copy of reveal.js. This requires that first you have a local copy 
+of reveal.js and then that you redirect the script away from your CDN to your 
+local copy.
 
-  .. code-block:: shell
+To make this clearer, let's look at an example. 
 
-    git clone https://github.com/hakimel/reveal.js.git
-    cd reveal.js
-    git checkout 3.5.0
-    cd ..
+.. note:: 
 
-  .. note:: 
+  In order to designate a mapping from notebook cells to Reveal.js slides, 
+  from within the Jupyter notebook, select menu item 
+  View --> Cell Toolbar --> Slideshow. That will reveal a drop-down menu 
+  on the upper-right of each cell.  From it, one may choose from 
+  "Slide," "Sub-Slide", "Fragment", "Skip", and "Notes."  On conversion, 
+  cells designated as "skip" will not be included, "notes" will be included 
+  only in presenter notes, etc.
 
-    In order to designate a mapping from notebook cells to Reveal.js slides, 
-    from within the Jupyter notebook, select menu item 
-    View --> Cell Toolbar --> Slideshow. That will reveal a drop-down menu 
-    on the upper-right of each cell.  From it, one may choose from 
-    "Slide," "Sub-Slide", "Fragment", "Skip", and "Notes."  On conversion, 
-    cells designated as "skip" will not be included, "notes" will be included 
-    only in presenter notes, etc.
+Example: creating slides w/ speaker notes
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Let's suppose you have a notebook ``your_talk.ipynb`` that you want to convert
+to slides. For this example, we'll assume that you are working in the same
+directory as the notebook you want to convert (i.e., when you run ``ls .``,
+``your_talk.ipynb`` shows up amongst the list of files). 
+
+First, we need a compatible version of reveal.js in the current folder run the
+following commands inside the directory:
+
+.. code-block:: shell
+
+  git clone https://github.com/hakimel/reveal.js.git
+  cd reveal.js
+  git checkout 3.5.0
+  cd ..
+
+Then we need to tell nbconvert to point to this local copy. To do that we use 
+the ``--reveal-prefix`` command line flag to point to the local copy.
+
+.. code-block:: shell 
+
+  jupyter nbconvert your_talk.ipynb --to slides --reveal-prefix reveal.js
+
+This will create file ``your_talk.slides.html``, which you should be able to 
+access with ``open your_talk.slides.html``. To access the speaker notes, press 
+``s`` after the slides load and they should open in a new window.
+
+This should also allow you to use your slides without an internet connection.
+
+If this does not work, you can also try start a server as part of your nbconvert
+command. To do this we use the ``ServePostProcessor``, which we activate by
+appending the command line flag ``--post serve`` to the above command. This
+will not allow you to use speaker notes if you do not have a local copy of
+reveal.js. 
+
 
 .. _convert_markdown:
 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -105,7 +105,7 @@ Reveal.js HTML slideshow
 
     git clone https://github.com/hakimel/reveal.js.git
     cd reveal.js
-    git checkout 3.6.0
+    git checkout 3.5.0
     cd ..
 
   .. note:: 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -96,7 +96,7 @@ Reveal.js HTML slideshow
   ``--post serve`` on the command-line. The ``serve`` post-processor proxies
   Reveal.js requests to a CDN if no local Reveal.js library is present.
   To make slides that don't require an internet connection, just place the
-  Reveal.js library in the same directory where your_talk.slides.html is
+  Reveal.js library (version 3.x) in the same directory where ``your_talk.slides.html`` is
   located, or point to another directory using the ``--reveal-prefix`` alias.
 
   .. note:: 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -99,6 +99,15 @@ Reveal.js HTML slideshow
   Reveal.js library (version 3.x) in the same directory where ``your_talk.slides.html`` is
   located, or point to another directory using the ``--reveal-prefix`` alias.
 
+  To get a compatible version of revealjs in the current folder:
+
+  .. code-block:: shell
+
+    git clone https://github.com/hakimel/reveal.js.git
+    cd reveal.js
+    git checkout 3.6.0
+    cd ..
+
   .. note:: 
 
     In order to designate a mapping from notebook cells to Reveal.js slides, 
--- a/nbconvert/exporters/slides.py
+++ b/nbconvert/exporters/slides.py
@@ -76,11 +76,16 @@ class SlidesExporter(HTMLExporter):
     """Exports HTML slides with reveal.js"""
 
     reveal_url_prefix = Unicode(
-        help="""The URL prefix for reveal.js.
-        This can be a a relative URL for a local copy of reveal.js (version 3.x),
-        or point to a CDN.
-
-        For speaker notes to work, a local reveal.js prefix must be used.
+        help="""The URL prefix for reveal.js (version 3.x).
+        This defaults to the reveal CDN, but can be any url pointing to a copy 
+        of reveal.js. 
+        
+        For speaker notes to work, this must be a relative path to a local 
+        copy of reveal.js: e.g., "reveal.js".
+        
+        See the usage documentation
+        (https://nbconvert.readthedocs.io/en/latest/usage.html#reveal-js-html-slideshow)
+        for more details.
         """
     ).tag(config=True)
 
--- a/nbconvert/exporters/slides.py
+++ b/nbconvert/exporters/slides.py
@@ -77,7 +77,7 @@ class SlidesExporter(HTMLExporter):
 
     reveal_url_prefix = Unicode(
         help="""The URL prefix for reveal.js.
-        This can be a a relative URL for a local copy of reveal.js,
+        This can be a a relative URL for a local copy of reveal.js (version 3.x),
         or point to a CDN.
 
         For speaker notes to work, a local reveal.js prefix must be used.
--- a/nbconvert/exporters/slides.py
+++ b/nbconvert/exporters/slides.py
@@ -90,7 +90,7 @@ class SlidesExporter(HTMLExporter):
             warn("Please update RevealHelpPreprocessor.url_prefix to "
                  "SlidesExporter.reveal_url_prefix in config files.")
             return self.config.RevealHelpPreprocessor.url_prefix
-        return 'https://cdnjs.cloudflare.com/ajax/libs/reveal.js/3.1.0'
+        return 'https://cdnjs.cloudflare.com/ajax/libs/reveal.js/3.5.0'
 
     reveal_theme = Unicode('simple',
         help="""
--- a/nbconvert/exporters/slides.py
+++ b/nbconvert/exporters/slides.py
@@ -101,7 +101,8 @@ class SlidesExporter(HTMLExporter):
         help="""
         Name of the reveal.js theme to use.
 
-        We look for a file with this name under `reveal_url_prefix`/css/theme/`reveal_theme`.css.
+        We look for a file with this name under
+        ``reveal_url_prefix``/css/theme/``reveal_theme``.css.
 
         https://github.com/hakimel/reveal.js/tree/master/css/theme has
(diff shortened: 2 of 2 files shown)


fperez:
This looks good to me, but a 2nd pair of eyes would be useful, as I've been away from the code for too long to trust that I won't miss something...

fperez:
Thx a lot, btw! Having this work out of the box (even if it doesn't support 100% offline use) will help a lot of regular users.

mpacer:
@takluyver @damianavila could either of you give this a look before we merge it? I'd like to get out a release soon and I'd like this to be part of it.

takluyver:
I'm wary of slapping a band-aid fix on for a problem that, based on the discussion on #702, we don't seem to properly understand. Damian suggested that this change risks breaking some other people's workflow, and I'd rather leave the problem where it is for another release - where people have likely found some kind of workaround - rather than creating a new problem for a different set of users and causing more confusion.

(The users experiencing the problem at the moment will likely protest that theirs is the typical use case, and we should just fix their bug, because the things it breaks are bound to be less important. We're all prone to assuming that we're typical. I'm not trying to make a judgment on whose use case is more important, just that a consistent problem is better than switching between problems.)

Clearly that answer isn't very satisfactory, so I'm going to try to investigate what's going on with slides.

damianavila:
Sorry perople for being late here, I was sick the last few days.

In general agree with @takluyver here. I am more than open to discuss the default change, but if we take that path, we need to make it in a more general PR changing not only the default value but also other parts of the slides architecture. Changing just this value I think will bring us more problems than solutions. This is why +1 @takluyver PR on docs and I would be -1 into merging this one as is.

mpacer:
I’ll merge the docs… but I do think this is a default that benefits users “in-the-know” better than users who don’t get these details. It is not just that this is a typical use case but is the major reason people have quoted as their problems with using the slides export. Having a local copy of reveal is much further from “just working” than relying on the cdn. 

We should definitely provide a way to handle the local filesystem use case and I’m ok including that in this PR if that’s what is needed to address this default issue… but I genuinely think it should be changed to support easier new users and to support webpages for slides that work by default when opened without needing to pull from a relative directory (which is often undesirable, e.g., in the case of github.io pages).

damianavila:
If we change the current default, we need to tell people who use the speaker notes (which I presume is a big percentage) to change their workflow from:

`jupyter nbconvert my_slide.ipynb --to slides --post serve`

to

`jupyter nbconvert my_slide.ipynb -- to slides --reveal-prefix="reveal.js" --post serve`

Otherwise their speaker notes will be busted. Do you think that is acceptable for the users? It is pretty difficult to me to realize if this change is annoying enough to make people angry... it is not annoying for me but I am sure I am biased in my response to this change.
@takluyver @mpacer @fperez thoughts?

takluyver:
My thoughts (as above) are that I don't want to try to judge which course creates the bigger inconvenience for the larger group. Rather, I'd apply the status quo argument: I'd prefer to let people live with the current inconvenience for another release than change it and create a different inconvenience. But as the main nbconvert maintainer, @mpacer has the final call on whether this goes in or not.

Longer term, I think we all agree that we need to look more closely at how our slides export deals with external resources. We certainly want a way to produce a single file, and a way to have slides that work completely offline, not pulling anything from a CDN. I'd also like to end the dependence on the serve post-processor (and get rid of the notion of post-processors in nbconvert altogether).

What are the limitations on the speaker notes feature? Do I recall that it only works when served over HTTP (i.e. not from a `file://` URL)? Do you know why that is the case?

damianavila:
> What are the limitations on the speaker notes feature? Do I recall that it only works when served over HTTP (i.e. not from a file:// URL)? Do you know why that is the case?

The CDN is not serving the `notes.html` file: https://github.com/cdnjs/cdnjs/issues/3317
I guess we can serve that specific file (alonside with notes.js) from another place?

takluyver:
Do you want to open a separate issue about the speaker notes thing so this discussion doesn't get sidetracked (more than I already sidetracked it ;-)?

damianavila:
> Do you want to open a separate issue about the speaker notes thing so this discussion doesn't get sidetracked (more than I already sidetracked it ;-)?

Good idea!

takluyver:
Actually, the more I think about this, the more I agree with M - I think it would be nicer to have slides export produce a file you can use without a special server, even if that breaks speaker notes.

But if the speaker notes just needs the CDN to serve HTML, could we use [rawgit](https://rawgit.com/) as the prefix? I.e. this URL works:

https://cdn.rawgit.com/hakimel/reveal.js/3.5.0/plugin/notes/notes.html

damianavila:
> Actually, the more I think about this, the more I agree with M - I think it would be nicer to have slides export produce a file you can use without a special server, even if that breaks speaker notes.

Yep, this is starting to be my feeling as well... despite the change in the `status quo`.

> But if the speaker notes just needs the CDN to serve HTML, could we use rawgit as the prefix? I.e. this URL works:

I quickly tried on the fly but there is still some other pieces needed, I guess, because I get errors about Reveal.js not defined and Mixed content...

takluyver:
Would this work as a way to deal with the offline use case:

When nbconvert produces the file, let all the URLs point to CDNs, so it's a small file that works standalone. Then have a separate tool that can read the HTML file, fetch the necessary JS and CSS, and either place them alongside the file, or stuff them into `<script>` and `<style>` tags inside it to produce an offline version. I've just had some fun [writing a very simple version](https://gist.github.com/takluyver/152cacdac9ac855ad554b0f91e67fb9c) of this tool, but it feels like something that lots of other people must have written already.

takluyver:
> I quickly tried on the fly but there is still some other pieces needed, I guess, because I get errors about Reveal.js not defined and Mixed content...

I see those too. The problem is that the speaker notes page is loaded over HTTPS, but then it tries to connect to the local server showing the presentation, which is HTTP. I can work around it for now by changing the rawgit CDN address to `http://...`, but that [will stop working next month](https://rawgit.com/news/https-only).

takluyver:
You can also work around it in Firefox by temporarily disabling the mixed content protection - click on the padlock icon and find the button - but I don't know if I'd rely on that working either.

damianavila:
> Would this work as a way to deal with the offline use case:

Interesting idea. I think it should work.

> I see those too. The problem is that the speaker notes page is loaded over HTTPS, but then it tries to connect to the local server showing the presentation, which is HTTP. I can work around it for now by changing the rawgit CDN address to http://..., but that will stop working next month.

As a quick turnaround we can serve these files, `notes.js` and `notes.html`, from the jupyter cdn, right?
We are already serving css from there in https and http: http://cdn.jupyter.org/notebook/5.1.0/style/style.min.css

takluyver:
I think the speaker notes should work from localhost if we can get revealjs from any domain where we can rely on it being plain HTTP for some time yet. CC @minrk @rgbkrk

nbconvert/exporters/slides.py
```diff
@@ -90,7 +90,7 @@ def _reveal_url_prefix_default(self):
             warn("Please update RevealHelpPreprocessor.url_prefix to "
                  "SlidesExporter.reveal_url_prefix in config files.")
             return self.config.RevealHelpPreprocessor.url_prefix
-        return 'reveal.js'
+        return 'https://cdnjs.cloudflare.com/ajax/libs/reveal.js/3.1.0'
```
rgbkrk: Switch this to:


```
return '//cdnjs.cloudflare.com/ajax/libs/reveal.js/3.1.0'
```

to provide a URL that the browser will adhere to HTTPS or HTTP for the user's page.

takluyver: I'm not sure if it will work, because the CDN redirects HTTP requests to HTTPS. Worth a try, though - thanks @rgbkrk

damianavila: I used:

```javascript
{ src: "//cdn.rawgit.com/hakimel/reveal.js/3.5.0/plugin/notes/notes.js",
```

And it seems to work for me... @takluyver can you check that as well?

takluyver: Nope, it still opens the speaker notes view with an https URL, unfortunately.

Additionally, using `//` breaks viewing the slides from a `file://` URL (i.e. opening the file without running a local http server), whereas that works if we specify the protocol.

takluyver: @damianavila the rawgit URL works for now, but I think it will stop working next month, when rawgit starts redirecting all HTTP requests to HTTPS.

I tested with the URL @rgbkrk suggested - the `notes.html` file for speaker view returns a 403, but you can still see that it's opening with an https URL, which is the cause of the 'mixed content' problem.

rgbkrk: Your only other choice is to put in specific logic to pick which one is best based on how the jupyter server is hosted, looking at `location.origin`. ¯\\\_(ツ)\_/¯

takluyver: Hmmm. Well, thanks for the info!

nbconvert/exporters/slides.py
```diff
@@ -90,7 +90,7 @@ def _reveal_url_prefix_default(self):
             warn("Please update RevealHelpPreprocessor.url_prefix to "
                  "SlidesExporter.reveal_url_prefix in config files.")
             return self.config.RevealHelpPreprocessor.url_prefix
-        return 'reveal.js'
```
damianavila: If we finally accept this change, @mpacer you need to point to reveal 3.5.0 which is the version we already have in the serve postprocessor.

nbconvert/exporters/slides.py
damianavila: > Nope, it still opens the speaker notes view with an https URL, unfortunately.

I only modified the piece I referenced above (notes.js) keeping in place all the other url references. And it seems to work for me, but...

> @damianavila the rawgit URL works for now, but I think it will stop working next month, when rawgit starts redirecting all HTTP requests to HTTPS.

Yep, it will probably work until next month. Again, can we serve it from the jupyter CDN?

In any case, having to use `jupyter nbconvert my_slide.ipynb -- to slides --reveal-prefix="reveal.js" --post serve` to have working speaker notes does not seems a big change for our users if we document it properly and explain why we change the default option (maybe with a tiny blog post).

takluyver: Right, I forgot - that was what I was asking @rgbkrk and @minrk about: would it be OK to host a copy of reveal.js on cdn.jupyter.org so that we can access it with plain HTTP?

minrk: I'd prefer to pare down the use of cdn.jupyter.org, especially for things that aren't our own in favor of things like [unpkg](https://unpkg.com/#/). The only reason for cdn.jupyter I can see is the non-npm packages we own (basically just the notebook-classic js/css). But pointing to unpkg or another CDN for reveal.js ought to work.

BTW, isn't it fine to load https resources on an http page, just not the other way around? Plus, https will work for a `file://` page, whereas `//` will not, so I think explicit https is preferable to `//`, unless I'm missing something.

takluyver: The `https://` URL works except for the speaker notes. Speaker notes open an HTML file at a URL based on the one reveal is loaded from (e.g. https://cdn.rawgit.com/hakimel/reveal.js/3.5.0/plugin/notes/notes.html ). That page connects *back* to the server your slides are served from. 

So if the notes page is an HTTP URL, it works. But if the notes page is an HTTPS URL and you're serving the slides with plain HTTP, the browser doesn't like the insecure connection.

But it's a good point that it would also be a problem the other way around - we can't load reveal with HTTP if you serve the slides using HTTPS.

So I think we just have to use HTTPS, and say that you have to use a local copy of reveal.js if you want to use speaker notes.

minrk: > So I think we just have to use HTTPS, and say that you have to use a local copy of reveal.js if you want to use speaker notes.

👍

damianavila: > So I think we just have to use HTTPS, and say that you have to use a local copy of reveal.js if you want to use speaker notes.

I am OK with that as well after the whole discussion.

takluyver:
OK, thanks everyone. @mpacer I think the remaining thing to do is to update the URL to 3.5.0,  which is the version we refer to elsewhere for reveal.js.

damianavila:
> I think the remaining thing to do is to update the URL to 3.5.0, which is the version we refer to elsewhere for reveal.js.

I would say we should update the docs PR #735 to add the following incantation for those who want to use the speaker notes:

```
jupyter nbconvert my_slide.ipynb -- to slides --reveal-prefix="reveal.js" --post serve
```

Cheers.

takluyver:
Good point, let's not forget to update the docs!

mpacer:
Ok I’ll make the change to update the version number later today.

mpacer:
Thanks everyone for hashing this out!

mpacer:
Do you want me to merge the docs PR into this one? so that it's all changed at once?

damianavila:
> Do you want me to merge the docs PR into this one? so that it's all changed at once?

I think it is good idea, but it is @takluyver 's PR so let's wait for his feedback, I think.

takluyver:
Yup, it's probably a good idea to merge #735 in here, because then the docs need another change as part of this PR, since putting revealjs beside the HTML file is no longer sufficient.

docs/source/usage.rst
```diff
@@ -92,31 +92,66 @@ Reveal.js HTML slideshow
...
+  
+By default, this will include a script tag in the html that will directly load 
+reveal.js from a CDN.
+
+However, some features (specifically, speaker notes) are only available if you
+use a local copy of reveal.js. This requires that first you have a local copy 
+of reveal.js and then that you redirect the script away from your CDN to your 
+local copy.
+
+To make this clearer, let's look at an example.
```
rgbkrk: Excellent

docs/source/usage.rst
```diff
@@ -92,22 +92,66 @@ Reveal.js HTML slideshow
...
+``s`` after the slides load and they should open in a new window.
+
+This should also allow you to use your slides without an internet connection.
+
+If this does not work, you can also try start a server as part of your nbconvert
+command. To do this we use the ``ServePostProcessor``, which we activate by
+appending the command line flag ``--post serve`` to the above command. This
+will not allow you to use speaker notes if you do not have a local copy of
+reveal.js. 
+
```
rgbkrk: Solid prose here

nbconvert/exporters/slides.py
```diff
@@ -90,13 +95,14 @@ def _reveal_url_prefix_default(self):
...
+        return 'https://cdnjs.cloudflare.com/ajax/libs/reveal.js/3.5.0'
```
rgbkrk: All set!

takluyver:
Alright, time to press the button on this. Thanks everyone for the discussion - it's a small technical change, but I'm pleased we've taken the time to think through it and make sure the docs are clear. :-)

docs/source/usage.rst
```diff
@@ -92,22 +92,66 @@ Reveal.js HTML slideshow
...
+Then we need to tell nbconvert to point to this local copy. To do that we use 
+the ``--reveal-prefix`` command line flag to point to the local copy.
+
+.. code-block:: shell 
+
+  jupyter nbconvert your_talk.ipynb --to slides --reveal-prefix reveal.js
+
+This will create file ``your_talk.slides.html``, which you should be able to 
+access with ``open your_talk.slides.html``. To access the speaker notes, press 
+``s`` after the slides load and they should open in a new window.
```
damianavila: Should we mention that the timer inside the speaker notes will only work if you are serving the slides?

docs/source/usage.rst
```diff
@@ -92,22 +92,66 @@ Reveal.js HTML slideshow
...
+
+This should also allow you to use your slides without an internet connection.
```
damianavila: Well, this is not true... we are loading other things from the internet, such as font-awesome, jquery, require, mathjax... and reveal.js itself load google fonts (although they have a fallback in case of lack of connectivity.
This statement as is can cause confussion, I would prefer to delete it.

damianavila:
@takluyver @mpacer sorry for the post-merge comments, but I think they are enough important to change some things and prevent users confusion.

takluyver:
@damianavila no problem, there's always room for a bit more review. Do you want to make a PR yourself?

damianavila:
> @damianavila no problem, there's always room for a bit more review. Do you want to make a PR yourself?

Sure, will submit in a few minutes.

docs/source/usage.rst
mpacer: Ok I wasn’t sure why that was I thought I had a weird bug and that the serving was required to coordinate 2 windows… why does the timer only work if you serve the slides but it still allows you to communicate between 2 windows?

docs/source/usage.rst
mpacer: I was making that claim based on the previous docs: 

> To make slides that don't require an internet connection, just place the Reveal.js library in the same directory where your_talk.slides.html is.

So at least this isn’t a regression… 

Makes me think we should have a way to package something that we can guarantee works in an offline context (e.g., where we download the relevant files to the directory in question since this is a more people would like to use).

damianavila: > Makes me think we should have a way to package something that we can guarantee works in an offline context (e.g., where we download the relevant files to the directory in question since this is a more people would like to use).

Yep, I was thinking the same, maybe some nbconvert script to download all the files needed?

docs/source/usage.rst
damianavila: > why does the timer only work if you serve the slides but it still allows you to communicate between 2 windows?

Good question! I have not researched the cause yet. I just verified the behavior.

PR #734, opened 2018-01-14
Better content hiding; template & testing improvements

### LINKED ISSUE #733: --no-prompt removes code cells from LaTeX/pdf docs
There is a bug where removing prompts will remove the code cells if you output to LaTeX or PDF.

I'll have a PR to address this shortly.

mpacer:
Closes #733. 

It adds a test to make sure this behaviour does not occur again; I also tested for this explicitly in the html and AsciiDoc exporters (since their templates have additional prompt hiding logic).

I also improved indenting in some of the templates and fixed a bug in the logic for hiding input cells in the basic html template.

I removed some unnecessary conditions for hiding prompts in the html template.

NB: I made some changes to the AsciiDoc template in order to simplify its logic as well… but they ended up being substantial enough that I don't think they should be included in this context (especially since I don't have a functioning version of AsciiDoc on my comp; with it relying on python 2.6… and no later!). I'll make a PR with those changes at some point near in the future…but it definitely seems like the current template is broken considering how it strips space from output areas.


nbconvert/exporters/tests/test_asciidoc.py
```diff
@@ -39,3 +43,25 @@ def test_export(self):
...
+        c_no_prompt = Config(no_prompt)
+        exporter = ASCIIDocExporter(config=c_no_prompt)
+        (output, resources) = exporter.from_filename(
+            self._get_notebook(nb_name="prompt_numbers.ipynb"))
+            
+        in_regex = r"In&nbsp;\[(.*)\]:"
+        out_regex = r"Out\[(.*)\]:"
+
+        assert not re.findall(in_regex, output)
+        assert not re.findall(out_regex, output)
```
takluyver: Do you want to add a corresponding test that the in and out prompts are produced with the default config? It looks like we already have such tests for HTML and Latex, but not for asciidoc.

mpacer: Sure!

mpacer: done!

takluyver:
Great :-)

PR #735, opened 2018-01-15
Clarify that reveal.js 3.x is needed

takluyver:
Closes gh-702

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -105,7 +105,7 @@ Reveal.js HTML slideshow
 
     git clone https://github.com/hakimel/reveal.js.git
     cd reveal.js
-    git checkout 3.6.0
+    git checkout 3.5.0
     cd ..
 
   .. note:: 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -96,7 +96,7 @@ Reveal.js HTML slideshow
   ``--post serve`` on the command-line. The ``serve`` post-processor proxies
   Reveal.js requests to a CDN if no local Reveal.js library is present.
   To make slides that don't require an internet connection, just place the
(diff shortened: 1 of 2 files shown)


damianavila:
+ 1 as I said in the issue.

mpacer:
Looks good but I still think that the default of using the cdn is a better default for most people unless we want to explicitly help people download an up-to-date version of reveal.js.

mpacer:
Actually can you add a reference to the latest example of reveal.js that works? Also a test to make sure that the version mentioned in the docs works? That way, we’ll know if it ever breaks with the a nbconvert change?

takluyver:
I've added a note to the docs showing how to get reveal 3.6. Adding a test that reveal.js works would be a big job, however - it involves executing Javascript in a browser, which the nbconvert tests are not set up to do. And we know from the notebook tests what a pain it is.

docs/source/usage.rst
```diff
@@ -96,9 +96,18 @@ Reveal.js HTML slideshow
...
+  Reveal.js library (version 3.x) in the same directory where ``your_talk.slides.html`` is
   located, or point to another directory using the ``--reveal-prefix`` alias.
 
+  To get a compatible version of revealjs in the current folder:
+
+  .. code-block:: shell
+
+    git clone https://github.com/hakimel/reveal.js.git
+    cd reveal.js
+    git checkout 3.6.0
```
damianavila: We are using reveal.js 3.5.0 in the serve post-processor, we should keep it consistent and checkout 3.5.0 until we decide to upgrade the whole thing to 3.6.0, I think.

takluyver: Thanks Damian, updated.

mpacer:
Closing since we now have 732 merged.

PR #738, opened 2018-01-17
Fix Jinja syntax in custom template example.

mdboom:
The custom template example doesn't work as intended: it needs double `{{ }}` to call a Jinja2 function, otherwise it's just treated as literal content.


takluyver:
Thanks :-)

PR #739, opened 2018-01-18
Skip executing empty code cells

### LINKED ISSUE #657: Execution of empty code cell produces duplicate `execution_count`
When an empty code cell is executed in the Jupyter Notebook Editor, the `execution_count` remains empty (`null`): `In[ ]`.

However, with `jupyter nbconvert --execute`, an empty code cell will receive a non-`null` execution count; in fact, it will get the same value as the execution count of the next non-empty code cell. This way, execution counts are not unique, which can be problematic (or at least confusing) when inspecting output history.

It is also an unfortunate discrepancy in the behaviors of these two execution mechanisms.

takluyver:
This matches the behaviour of the notebook itself: empty code cells are not executed.

Closes #657.

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -290,7 +290,7 @@ class ExecutePreprocessor(Preprocessor):
 
         To execute all cells see :meth:`preprocess`.
         """
-        if cell.cell_type != 'code':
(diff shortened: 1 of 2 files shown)


mpacer:
That turned out to be a really nice fix! well done, merging.

PR #743, opened 2018-01-26
Remove offline statement and add some clarifications in slides docs

damianavila:
Some more changes after PR #732.

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -152,9 +152,10 @@ This will create file ``your_talk.slides.html``, which you should be able to
 access with ``open your_talk.slides.html``. To access the speaker notes, press 
 ``s`` after the slides load and they should open in a new window.
 
-Note: a local copy of reveal.js does allow slides that will run completely
-offline, because the slides rely on other resources (e.g., mathjax and jquery)
-that still require access to public CDNs (by default).
+Note: This does not enable slides that run completely offline. While you have a
+local copy of reveal.js, by default, the slides need to access mathjax, require,
+and jquery via a public CDN. Addressing this use case is an open issue and `PRs
+<https://github.com/jupyter/nbconvert/pulls>`_ are always encouraged.
 
 .. _ServePostProcessorExample:
 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -152,6 +152,9 @@ This will create file ``your_talk.slides.html``, which you should be able to
 access with ``open your_talk.slides.html``. To access the speaker notes, press 
 ``s`` after the slides load and they should open in a new window.
 
+Note: a local copy of reveal.js does allow slides that will run completely
+offline, because the slides rely on other resources (e.g., mathjax and jquery)
+that still require access to public CDNs (by default).
 
 .. _ServePostProcessorExample:
 
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -133,6 +133,10 @@ following commands inside the directory:
   git checkout 3.5.0
   cd ..
 
+Alternative, you can download a zip (or tar.gz) file containing reveal.js from
+https://github.com/hakimel/reveal.js/releases/tag/3.5.0, but be sure to unzip
+(or untar) the file to a directory named reveal.js.
+
 Then we need to tell nbconvert to point to this local copy. To do that we use 
 the ``--reveal-prefix`` command line flag to point to the local copy.
 
@@ -142,9 +146,9 @@ the ``--reveal-prefix`` command line flag to point to the local copy.
 
 This will create file ``your_talk.slides.html``, which you should be able to 
 access with ``open your_talk.slides.html``. To access the speaker notes, press 
-``s`` after the slides load and they should open in a new window.
-
-This should also allow you to use your slides without an internet connection.
+``s`` after the slides load and they should open in a new window. Keep in mind
+that if you want a functional timer inside the speaker notes, you need to serve
+the slides (see next paragraph forß details).
 
 If this does not work, you can also try start a server as part of your nbconvert
 command. To do this we use the ``ServePostProcessor``, which we activate by
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -146,16 +146,27 @@ the ``--reveal-prefix`` command line flag to point to the local copy.
 
 This will create file ``your_talk.slides.html``, which you should be able to 
 access with ``open your_talk.slides.html``. To access the speaker notes, press 
-``s`` after the slides load and they should open in a new window. Keep in mind
-that if you want a functional timer inside the speaker notes, you need to serve
-the slides (see next paragraph for details).
+``s`` after the slides load and they should open in a new window. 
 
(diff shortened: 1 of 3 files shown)


docs/source/usage.rst
```diff
@@ -133,6 +133,10 @@ following commands inside the directory:
   git checkout 3.5.0
   cd ..
 
+Alternative, you can download a zip (or tar.gz) file containing reveal.js from
```
mpacer: Honestly I’d prefer to keep this simpler in this location. If we really want to cover a lot of options on how to do this I think we should have a separate page. But otherwise I’d like this to be a linear flow.

docs/source/usage.rst
```diff
@@ -142,9 +146,9 @@ the ``--reveal-prefix`` command line flag to point to the local copy.
 
 This will create file ``your_talk.slides.html``, which you should be able to 
 access with ``open your_talk.slides.html``. To access the speaker notes, press 
-``s`` after the slides load and they should open in a new window.
-
-This should also allow you to use your slides without an internet connection.
+``s`` after the slides load and they should open in a new window. Keep in mind
+that if you want a functional timer inside the speaker notes, you need to serve
```
mpacer: This is somewhat complicated syntax. I also think I’d prefer a caveat with no explanation of how to fix it to come earlier in the paragraph. The explanation would be given by reference with a hyperlink to a new section header that lives above the statement about the https server which should explicitly mention the bit about the timer.

docs/source/usage.rst
mpacer: To explain why: the people who benefit most from this are going to be those who don’t quite understand why they need to do this. They’ll have the easiest time copying and pasting some commands, even if they don’t fully understand those commands. Telling someone to find a resource that they may or may not understand and then to conditionally manipulate it (`tar -xfd` or `unzip`) using tools they may or may not already know introduces unnecessary complication and uncertainty.  We should aim to be as simple as possible, which ideally implies a workflow uncluttered with as few decisions as possible.

mpacer:
I’ll push some changes in a bit that will bring this closer to what I imagine. 

It’s really important to me that we write these docs in a way that novices can use this utility without getting needlessly frustrated. I’ve seen it a lot and I want to avoid it going forward. So… that probably means I’m being more strict on the narrative structure that I normally would.

docs/source/usage.rst
damianavila: Fair enough, feel free to push the removal :wink:

docs/source/usage.rst
damianavila: That's OK for me.

damianavila:
> I’ll push some changes in a bit that will bring this closer to what I imagine.

Thank you for doing this!

> So… that probably means I’m being more strict on the narrative structure that I normally would.

I can understand this need and I am more than open to the changes you proposed. As you know, English is not my mother language so I can be a little bit cryptic sometimes...

mpacer:
> I can understand this need and I am more than open to the changes you proposed. As you know, English is not my mother language so I can be a little bit cryptic sometimes...

I thought your prose was fine, just better fit for a different purpose (e.g., a page that described in greater detail how to use the notebook for slides — which I think may make sense to include inside the nbconvert docs).

docs/source/usage.rst
```diff
@@ -146,16 +150,32 @@ the ``--reveal-prefix`` command line flag to point to the local copy.
 
 This will create file ``your_talk.slides.html``, which you should be able to 
 access with ``open your_talk.slides.html``. To access the speaker notes, press 
-``s`` after the slides load and they should open in a new window. Keep in mind
-that if you want a functional timer inside the speaker notes, you need to serve
-the slides (see next paragraph for details).
+``s`` after the slides load and they should open in a new window.
+
+Note: a local copy of reveal.js does allow slides that will run completely
```
takluyver: does *not* ?

docs/source/usage.rst
```diff
@@ -146,16 +150,32 @@ the ``--reveal-prefix`` command line flag to point to the local copy.
...
+Serving slides with an https server: ``--post serve``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Once you have speaker notes working you may notice that your timers don't work.
+Timers require a bit more infrastructure; you need to serve your local copy of
+reveal.js from a local https server.
+
+Fortunately, ``nbconvert`` makes this fairly straightforward through the use of
+the ``ServePostProcessor``. To activate this server, we append the command line
+flag ``--post serve`` to our call to nbconvert.
```
takluyver: Once you've got reveal.js set up next to the HTML file, I don't think you need the post processor - we could recommend `python3 -m http.server` to start a server.

If we can recommend this, it also gets us one step closer to getting rid of the postprocessor entirely.

docs/source/usage.rst
damianavila: I think so...

docs/source/usage.rst
damianavila: As you said, you don't *need* the post-processor, but it makes things easier for you.

I am +1 about the removal of post-processors, but I would keep the `--post serve` concept (maybe just calling it `serve` and implemented w/o post-processors) instead of recommending the `python3 -m http.server`.

docs/source/usage.rst
mpacer: oops

docs/source/usage.rst
mpacer: doesn't the simple http.server only serve http and not https? isn't it required to be an https server?

mpacer: Ok… well I just tested it and there is one tiny issue with `python3 -m http.server`, which is that it doesn't auto-open the file in the browser… meaning the person needs to navigate to `localhost:8000/your_talk.slides.html` in their browser which is both a context switch (cli to browser) and introduces a new layer of complexity (correctly arriving at the slides page relative to an implicit port statement). 

We can add the port declaration to the command `python3 -m http.server 8000` even if we keep it the default, but I still think that this is an additional opportunity for users to become confused.

damianavila: > meaning the person needs to navigate to localhost:8000/your_talk.slides.html in their browser which is both a context switch (cli to browser) and introduces a new layer of complexity (correctly arriving at the slides page relative to an implicit port statement).

I agree, this is why I would keep the `--post serve` until we provide a `--serve` option that is not based in a postprocessor.

mpacer: So… for this docs PR… I think we should still use `--post serve` @takluyver are you ok with that or do you feel strongly that we should include the slightly more cumbersome `python3 -m http.server` style instructions?

mpacer:
@takluyver I think you’re the only one who should merge this at this point, as both @damianavila and I have contributed to the PR and we’re on the same page regarding keeping the bit about `--post serve`. If you want me to remove it, I will but I also think that we could hash that out in a later PR.

takluyver:
Sorry, my reply seems to have got lost somewhere. I tried a couple of days ago to say something like "I don't feel strongly, go for it."

Issue #742, opened 2018-01-25
UnicodeDecodeError when converting to slides on Windows


HagaiHargil:
Hi,

I'm having issues converting my notebooks to slides. The error is the following:

```
(general) PS C:\Users\Hagai\Documents\Classes\python-course\Classes> jupyter nbconvert class3.ipynb --to slides --post serve
[NbConvertApp] Converting notebook class3.ipynb to slides
Traceback (most recent call last):
  File "C:\Users\Hagai\Miniconda3\envs\general\Scripts\jupyter-nbconvert-script.py", line 5, in <module>
    exit(main())
  File "C:\Users\Hagai\Miniconda3\envs\general\lib\site-packages\jupyter_core\application.py", line 266, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "C:\Users\Hagai\Miniconda3\envs\general\lib\site-packages\traitlets\config\application.py", line 658, in launch_instance
    app.start()
  File "C:\Users\Hagai\Miniconda3\envs\general\lib\site-packages\nbconvert\nbconvertapp.py", line 325, in start
    self.convert_notebooks()
  File "C:\Users\Hagai\Miniconda3\envs\general\lib\site-packages\nbconvert\nbconvertapp.py", line 493, in convert_notebooks
    self.convert_single_notebook(notebook_filename)
  File "C:\Users\Hagai\Miniconda3\envs\general\lib\site-packages\nbconvert\nbconvertapp.py", line 464, in convert_single_notebook
    output, resources = self.export_single_notebook(notebook_filename, resources, input_buffer=input_buffer)
  File "C:\Users\Hagai\Miniconda3\envs\general\lib\site-packages\nbconvert\nbconvertapp.py", line 393, in export_single_notebook
    output, resources = self.exporter.from_filename(notebook_filename, resources=resources)
  File "C:\Users\Hagai\Miniconda3\envs\general\lib\site-packages\nbconvert\exporters\exporter.py", line 174, in from_filename
    return self.from_file(f, resources=resources, **kw)
  File "C:\Users\Hagai\Miniconda3\envs\general\lib\site-packages\nbconvert\exporters\exporter.py", line 192, in from_file
    return self.from_notebook_node(nbformat.read(file_stream, as_version=4), resources=resources, **kw)
  File "C:\Users\Hagai\Miniconda3\envs\general\lib\site-packages\nbconvert\exporters\slides.py", line 172, in from_notebook_node
[...]

PR #757, opened 2018-02-07
Windows unicode error fixed, nosetest added to setup.py

HagaiHargil:
@takluyver Your suggestion at #742 was spot on.

I've also added `nose` to the test dependencies, since running `py.test --pyargs nbconvert` doesn't work when `nose` isn't installed (seems like iPython has this dependency).

### DIFF
--- a/nbconvert/preprocessors/csshtmlheader.py
+++ b/nbconvert/preprocessors/csshtmlheader.py
@@ -10,7 +10,6 @@ import hashlib
 import nbconvert.resources
 
 from traitlets import Unicode
-from ipython_genutils.py3compat import str_to_bytes
 from .base import Preprocessor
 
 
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks, the actual fix looks good to me, but it should be necessary to require nose for tests - PRs #595 and #660 got rid of the imports. What errors were you getting without nose?

mpacer:
Apparently, we still have some imports inside the `test_svg2pdf.py` and `test_io.py`. I don't know how I missed those before since I was grepping for all instances of nose.

mpacer:
ah… it's in the ipython_genutils decorators… hrm… I'll see if I can fix that.

HagaiHargil:
Would you like me to resubmit the PR without the `setup.py` change? Or will you change it back in a later commit?

takluyver:
I'd say we merge this and change it back when we can figure out the remaining nose dependencies. This is correct for now, after all. But I'll let @mpacer make the final call.

mpacer:
Yea… I agree, though I really don't like it. :grumble: 

merging.

PR #758, opened 2018-02-08
Removing more nose remnants via dependencies.

mpacer:
It was revealed in #757 super helpfully by @HagaiHargil that we were implicitly relying on nose because we were using ipython_genutils test decorators. 

This removes those `ipython_genutils` versions by porting the `@onlyif_cmds_exist` decorator and using `pytest.mark` inside of it instead of a bunch of different `nose` bits.

### DIFF
--- a/nbconvert/exporters/tests/test_pdf.py
+++ b/nbconvert/exporters/tests/test_pdf.py
@@ -28,8 +28,7 @@ class TestPDF(ExportersTestsBase):
         self.exporter_class()
 
 
-    @onlyif_cmds_exist('xelatex')
-    @onlyif_cmds_exist('pandoc')
+    @onlyif_cmds_exist('xelatex', 'pandoc')
     def test_export(self):
         """Smoke test PDFExporter"""
         with tempdir.TemporaryDirectory() as td:
--- a/nbconvert/exporters/tests/test_pdf.py
+++ b/nbconvert/exporters/tests/test_pdf.py
@@ -7,11 +7,11 @@ import logging
 import os
 import shutil
 
-from ipython_genutils.testing import decorators as dec
 from testpath import tempdir
 
 from .base import ExportersTestsBase
 from ..pdf import PDFExporter
+from ...utils.io import onlyif_cmds_exist
 
 
 #-----------------------------------------------------------------------------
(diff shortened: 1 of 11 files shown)


nbconvert/exporters/tests/test_pdf.py
```diff
@@ -28,8 +28,8 @@ def test_constructor(self):
         self.exporter_class()
 
 
-    @dec.onlyif_cmds_exist('xelatex')
-    @dec.onlyif_cmds_exist('pandoc')
+    @onlyif_cmds_exist('xelatex')
+    @onlyif_cmds_exist('pandoc')
```
takluyver: This could be combined into one decorator, right?

nbconvert/tests/test_nbconvertapp.py
```diff
@@ -117,8 +118,8 @@ def test_relative_template_file(self):
                 text = f.read()
             assert text == test_output
 
-    @dec.onlyif_cmds_exist('xelatex')
-    @dec.onlyif_cmds_exist('pandoc')
+    @onlyif_cmds_exist('xelatex')
+    @onlyif_cmds_exist('pandoc')
```
takluyver: Combine (and another one just below)

nbconvert/tests/test_nbconvertapp.py
```diff
@@ -408,8 +409,8 @@ def test_convert_from_stdin(self):
                 assert '```python' not in output1 # shouldn't have language
                 assert "```" in output1 # but should have fenced blocks
 
-    @dec.onlyif_cmds_exist('xelatex')
-    @dec.onlyif_cmds_exist('pandoc')
+    @onlyif_cmds_exist('xelatex')
+    @onlyif_cmds_exist('pandoc')
```
takluyver: Combine

nbconvert/tests/test_nbconvertapp.py
mpacer: yep! good call!

nbconvert/exporters/tests/test_pdf.py
mpacer: yep! good call!

nbconvert/tests/test_nbconvertapp.py
mpacer: yep! good call!

minrk:
Hooray for shrinking dependencies on genutils!

Issue #718, opened 2017-12-06
Save widget models with Execute preprocessor?


ricklupton:
I'm using a notebook & nbsphinx to document my custom widget. This works fine if I run the notebook by hand and explicitly "save widget state". But it doesn't work if I use nbsphinx to execute the notebook, because the execute preprocessor doesn't save widget states (comm messages are discarded [here](https://github.com/jupyter/nbconvert/blob/master/nbconvert/preprocessors/execute.py#L411)).

From a quick look it doesn't look like it would be too hard to dump these comm messages and save the `data` of the message into the `widgets` part of the notebook document. Is that a good idea?

mpacer:
That seems reasonable but I'd need to see the implementation. I'd happily review a PR if you have a working model.

ricklupton:
Thanks. I haven't finished testing with nbsphinx but this works with the unit test I wrote:
https://github.com/ricklupton/nbconvert/commit/12cff0657e41adf607962bd365b29f66b5748894

It doesn't remove default values like the proper notebook does. Do you know if there's some existing logic for doing that somewhere that can be reused? Or maybe it doesn't matter.

ricklupton:
@mpacer I've finished testing this and it's working in practice for building my documentation using nbsphinx -- I've made a PR if you have a chance to look.

scottdraves:
dup? https://github.com/jupyter/nbconvert/issues/751

t-makaro:
Most of the discussion is in #751, so I am closing this issue as the duplicate.

PR #760, opened 2018-02-09
Save widget model state to notebook metadata

ricklupton:
As proposed in #718.

Known limitations:
1. Default values are saved (the JS widget code only saves non-default values)
2. Binary buffers are not dealt with

For (1), is there any way of getting the default values of the widgets in the Execute Preprocessor? When running in the notebook the JS default values are used...

For (2) I didn't know what was meant to happen for binary buffers, but if you can point me towards what should be done with them I'm happy to try to add that.

I'm [currently using this to build documentation from Jupyter notebooks including widgets](https://github.com/ricklupton/floweaver/blob/master/docs/requirements.txt#L3) and [it's working](https://floweaver.readthedocs.io/en/latest/tutorials/quickstart.html)

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -273,12 +273,24 @@ class ExecutePreprocessor(Preprocessor):
             cwd=path)
         self.kc.allow_stdin = False
         self.nb = nb
+        self.widget_state = {}
 
         try:
             nb, resources = super(ExecutePreprocessor, self).preprocess(nb, resources)
         finally:
             self.kc.stop_channels()
             self.km.shutdown_kernel(now=self.shutdown_kernel == 'immediate')
+            if self.widget_state:
+                self.nb.metadata.widgets = {
+                    'application/vnd.jupyter.widget-state+json': {
+                        'state': {
+                            model_id: _serialize_widget_state(state)
+                            for model_id, state in self.widget_state.items()
+                        },
+                        'version_major': 2,
+                        'version_minor': 0,
+                    }
+                }
 
         delattr(self, 'nb')
 
@@ -409,6 +421,7 @@ class ExecutePreprocessor(Preprocessor):
                         cell_map[cell_index] = []
                 continue
             elif msg_type.startswith('comm'):
+                self.widget_state[content['comm_id']] = content['data']['state']
                 continue
 
             display_id = None
@@ -457,3 +470,16 @@ def executenb(nb, cwd=None, **kwargs):
         resources['metadata'] = {'path': cwd}
     ep = ExecutePreprocessor(**kwargs)
     return ep.preprocess(nb, resources)[0]
+
+
+def _serialize_widget_state(state):
+    """Serialize a widget state, following format in @jupyter-widgets/schema.
+
+    TODO: Does not currently split binary buffers or remove default values.
+    """
+    return {
+        'model_name': state['_model_name'],
+        'model_module': state['_model_module'],
+        'model_module_version': state.get('_model_module_version'),
+        'state': state,
+    }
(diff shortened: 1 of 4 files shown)


mpacer:
You need to add ipywidgets to the test dependencies in setup.py for the tests to pass.

ricklupton:
Sorry -- added now.

The `test_run_notebooks` test is still failing on Python 2.7, because the `test/plain` display data of a widget is it's `repr` which is different on Python 2 and 3:

Python 2.7: `Label(value=u'Hello World')`
Python 3.6: `Label(value='Hello World')`

I couldn't see any existing logic for dealing with this in the tests. Do you have any suggestions for how best to deal with this? My thoughts were
1. Have a Python 2 and a Python 3 version of the test notebook file, and run the appropriate one only
2. Normalise `u'...'` to `'...'` with a regexp

ricklupton:
Separately, I think this base64 normalisation code is confusing things as it's applied to non-base64 outputs such as `text/plain` and `text/html` -- perhaps it should be whitelisted to only apply to `image/png` etc?

https://github.com/jupyter/nbconvert/blob/master/nbconvert/preprocessors/tests/test_execute.py#L57-L59

(this means that `Label(value='Hello World')` is normalised to `LabelvalueHelloWorld` for example)

mpacer:
@takluyver I'd appreciate your thoughts on this.

mpacer:
@jasongrout @SylvainCorlay How does this differ from the approach that you have been taking with widgets previously?

mpacer:
Pinging @jasongrout again as requested in the Jupyter dev meeting February 27, 2018.

jasongrout:
If you have access to the kernel, you can get the serialized widget manager state using the embed module (see also http://ipywidgets.readthedocs.io/en/latest/embedding.html#python-interface), in particular:

```python
from ipywidgets import IntSlider
import ipywidgets.embed

x = IntSlider()
data = ipywidgets.embed.embed_data(None)
state = data['manager_state']
state
``` 
gives the state of the widget manager.

ricklupton:
@jasongrout thanks, but I thought code to save widgets state as part of nbexecute should be independent of the kernel? Here I'm capturing the state from the comm messages, is there anything wrong with that?

jasongrout:
That could work. You'll need to understand how to convert binary blobs to base64 binary blobs, for example, to properly serialize them to the notebook.

ricklupton:
Is the binary serialization on-the-wire in comm messages the same as in the notebook document? I thought it would be "just" a matter of capturing the binary and non-binary parts of comm messages and saving them in the right place?

scottdraves:
Hi, we need this fix too.  @ricklupton do you mind if we pick up your branch and continue work on it?  Happy to finish it for you.

ricklupton:
Fine by me!

t-makaro:
I believe this is now completed via #900 -- please reopen and rebase if you think there's anything that got missed.

timkpaine:
@t-makaro @jasongrout does this work for layout widgets? When I export to html I can see things like `IntSlider`, but `HBox` goes away (I.e. the outputs appear vertically)

SylvainCorlay:
@timkpaine I am not reproducing. HBox appears to work fine in the HTML export.

timkpaine:
@SylvainCorlay yep you're right, looks like my template is wrong.

certain properties aren't propagated, however:

Lab:
Screen Shot 2020-05-23 at 08 09 25
Nbconvert Master:
Screen Shot 2020-05-23 at 08 09 10

Code:
```python
from ipywidgets import Box, Output
from IPython.display import HTML

children = []

for text in ("1", "2", "3"):
    out = Output(layout={"flex": text})
    with out:
        display(HTML(text))
    children.append(out)

Box(children, layout={'display': 'flex', 'flex-direction': 'row'})
```

timkpaine:
@SylvainCorlay actually on second pass, I see the problem when `--execute`

SylvainCorlay:
Ok I will look into it. There is much morr in your example than an HBox, with the output widget.

Issue #767, opened 2018-02-26
Cleaning up `notebook.tex` file after creating pdf


saulshanabrook:
Is it possible to remove the `notebook.tex` file after converting a notebook to a PDF? I am trying to see if I can get this to work on JupyterLab(https://github.com/jupyterlab/jupyterlab/issues/4018). The `temp_file_exts` config looks promising, but I am not having much luck with it. I have tried adding `.tex` and `'/../notebook.tex'` to it to no avail.

When I was looking into debugging that, I also couldn't find where the [`clean_temp_files` function](https://github.com/jupyter/nbconvert/blob/53f3c86183abd3fc3a5c007c7146aca09d0785f6/nbconvert/exporters/pdf.py#L159) is called. Is it dead code? Or does it have to be enabled in some way?

takluyver:
At present, the `notebook.tex` file should be generated in a temporary folder, and the whole folder cleaned up when it finishes. That's this context manager:

https://github.com/jupyter/nbconvert/blob/53f3c86183abd3fc3a5c007c7146aca09d0785f6/nbconvert/exporters/pdf.py#L180

I think when we switched to that approach `clean_temp_files` became dead code.

PR #768, opened 2018-02-26
Cleanup notebook.tex during PDF generation

saulshanabrook:
Currently, when I generate a PDF from a notebook it leaves behind the `notebook.tex` file in the directory that the notebook is in. This fixes that by generating the tex file in the same temporary directory as the rest of the latex files. It also removes some related dead code. Closes https://github.com/jupyter/nbconvert/issues/767

### DIFF
--- a/nbconvert/exporters/tests/test_pdf.py
+++ b/nbconvert/exporters/tests/test_pdf.py
@@ -37,4 +37,6 @@ class TestPDF(ExportersTestsBase):
             (output, resources) = self.exporter_class(latex_count=1).from_filename(newpath)
             self.assertIsInstance(output, bytes)
             assert len(output) > 0
+            # tex file should be cleaned up
+            assert 'notebook.tex' not in os.listdir(td)
 
--- a/nbconvert/exporters/tests/test_pdf.py
+++ b/nbconvert/exporters/tests/test_pdf.py
@@ -32,11 +32,11 @@ class TestPDF(ExportersTestsBase):
     def test_export(self):
         """Smoke test PDFExporter"""
         with tempdir.TemporaryDirectory() as td:
-            newpath = os.path.join(td, os.path.basename(self._get_notebook()))
+            file_name = os.path.basename(self._get_notebook())
+            newpath = os.path.join(td, file_name)
             shutil.copy(self._get_notebook(), newpath)
             (output, resources) = self.exporter_class(latex_count=1).from_filename(newpath)
             self.assertIsInstance(output, bytes)
             assert len(output) > 0
-            # tex file should be cleaned up
-            assert 'notebook.tex' not in os.listdir(td)
-
+            # all temporary file should be cleaned up
+            assert {file_name} == set(os.listdir(td))
--- a/nbconvert/exporters/pdf.py
+++ b/nbconvert/exporters/pdf.py
@@ -61,10 +61,6 @@ class PDFExporter(LatexExporter):
         help="Whether to display the output of latex commands."
     ).tag(config=True)
 
-    temp_file_exts = List(['.aux', '.bbl', '.blg', '.idx', '.log', '.out'],
-        help="File extensions of temp files to remove after running."
-    ).tag(config=True)
-    
     texinputs = Unicode(help="texinputs dir. A notebook's directory is added")
     writer = Instance("nbconvert.writers.FilesWriter", args=(), kw={'build_directory': '.'})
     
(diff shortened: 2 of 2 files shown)


takluyver:
It looks like the only change is in a test?

saulshanabrook:
@takluyver It now includes the fix. I added that as a separate commit so that we could verify it did, in fact, fail and that the next commit fixed it.

takluyver:
Gotcha, thanks.

nbconvert/exporters/tests/test_pdf.py
```diff
@@ -37,4 +37,6 @@ def test_export(self):
             (output, resources) = self.exporter_class(latex_count=1).from_filename(newpath)
             self.assertIsInstance(output, bytes)
             assert len(output) > 0
+            # tex file should be cleaned up
+            assert 'notebook.tex' not in os.listdir(td)
```
Naereen: That's a great test!

mpacer: Given that this is also removing the `clean_temp_files` command, I'd appreciate a check that ensured any of the companion files would also be removed.

nbconvert/exporters/tests/test_pdf.py
```diff
@@ -37,4 +37,6 @@ def test_export(self):
             (output, resources) = self.exporter_class(latex_count=1).from_filename(newpath)
```
mpacer: I'd feel better if this test would run LaTeX with a bibfile since that can also introduce artifacts and should be tested as part of the smoke test. If this turns out to be too big of a change, then we can move this into a new PR & make a new issue. But, could you give that a shot?

saulshanabrook: I can try to do that. Do you think it makes sense to copy these examples files into the repo for the test? https://github.com/jupyter/nbconvert-examples/tree/master/citations I can't seem to find an existing example notebook in this repo.

nbconvert/exporters/tests/test_pdf.py
saulshanabrook: Sure. I updated it to make sure only the initial notebook exists in the directory.

nbconvert/exporters/tests/test_pdf.py
mpacer: I think that those might be out of date — given that they were last updated over 3 years ago. 

I'm now thinking that there may be no built-in way to do this… which is… odd. 

We may need to look into the LaTeX templates to see if we need to include a new bibliography area and traitlet to get that to work. That's starting to sound like a separate PR though.

Could you try to figure out if any of the other code related to this is dead? If you just include `assert False` in places it should trigger failures in the tests if those parts of the code are actually being tested.

nbconvert/exporters/tests/test_pdf.py
```diff
@@ -32,9 +32,11 @@ def test_constructor(self):
     def test_export(self):
         """Smoke test PDFExporter"""
         with tempdir.TemporaryDirectory() as td:
-            newpath = os.path.join(td, os.path.basename(self._get_notebook()))
+            file_name = os.path.basename(self._get_notebook())
```
mpacer: Nice — that is cleaner.

nbconvert/exporters/tests/test_pdf.py
saulshanabrook: Looking at the [code coverage](https://codecov.io/gh/jupyter/nbconvert/src/2210f10d96295906c330718c60b8189d8cded77a/nbconvert/exporters/pdf.py), it seems like the only lines not executed in the PDF exporting are the win32 related, error related, or for generating the bib.

I am not sure exactly when the code became dead, if there are other files that are affected.

mpacer: Code coverage actually doesn't get exactly what I was looking for. My ultimate aim was to know which tests were using the bibliography functionality to see _how_ it was being used. I can just do this myself though…

mpacer: Alright… so I'm fairly sure that right now that this code is only being run in the tests for the nbconvertapp, which are a gigantic pain to debug and investigate because they're hidden behind a `subprocess.Popen` call to `nbconvert` as part of the test… which means that pdb just causes everything to fail.

mpacer: Regardless my investigations left me believing that currently there is nothing that's properly testing the bibliography support… so we should look into that, but that's a separate PR.

saulshanabrook: Sounds good! I don't know much about bib support and haven't used it.

takluyver:
OK, it sounds like this is good to go. Thanks @saulshanabrook !

Issue #636, opened 2017-08-01
Update widget export to 7.0


jasongrout:
In order to keep live widget support in the html export, we'll need to figure out a way to support both 6.x and 7.x widget formats. At the very minimum, the script tag at https://github.com/jupyter/nbconvert/blob/db3036303237d45db9886c44e31132f90ef8d653/nbconvert/templates/html/full.tpl#L14 needs to be changed.

mpacer:
How does it need to be changed?

jasongrout:
@mpacer - I'm working on nailing down exactly what it changes to right now.

mpacer:
Adding this as a soon-to-be released patch release since I don't want to wait on this to get 5.3 out the door if at all possible.

jasongrout:
The embedding changes needed are noted at http://ipywidgets.readthedocs.io/en/stable/migration_guides.html?highlight=migrate#updating-embedded-widgets

As noted in https://github.com/jupyter/nbviewer/issues/717, it would be great if we could examine the metadata and see which widgets version we needed to embed.

Also, this ties into a larger conversation about how we provide custom renderers (in some sort of pluggable system).

vidartf:
Considering this was now fixed for nbviewer, is the solution transferable to nbconvert?

jasongrout:
> Considering this was now fixed for nbviewer, is the solution transferable to nbconvert?

It's probably the same or very similar.

vidartf:
(nbviewer fixed this in https://github.com/jupyter/nbviewer/pull/738)

vidartf:
A proposed fix based on jupyter/nbviewer#738 is available in #792.

Juanlu001:
If I understand correctly, when this is solved we will have interactive widgets in static HTML export, is that correct? Among the two linked PRs (https://github.com/jupyter/nbconvert/pull/778 and https://github.com/jupyter/nbconvert/pull/792) is there any preferred approach? Anything we can do to help moving this forward? Thanks!

cornhundred:
[...]

PR #778, opened 2018-03-07
#636 fixed widget display for both formats

scottdraves:
Fix https://github.com/jupyter/nbconvert/issues/636 using the same approach as we used on https://github.com/jupyter/nbviewer/issues/717

### DIFF
--- a/nbconvert/templates/html/full.tpl
+++ b/nbconvert/templates/html/full.tpl
@@ -12,7 +12,29 @@
 <title>{{nb_title}}</title>
 
 {%- if "widgets" in nb.metadata -%}
-<script src="https://unpkg.com/jupyter-js-widgets@2.0.*/dist/embed.js"></script>
+<script>
+  (function() {
+    function addWidgetsRenderer() {
+      var mimeElement = document.querySelector('script[type="application/vnd.jupyter.widget-view+json"]');
+      var scriptElement = document.createElement('script');
+      var widgetRendererSrc = 'https://unpkg.com/@jupyter-widgets/html-manager@*/dist/embed-amd.js';
+      var widgetState;
+
+      try {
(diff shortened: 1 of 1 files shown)


scottdraves:
@jasongrout @minrk

nbconvert/templates/html/full.tpl
```diff
@@ -12,7 +12,29 @@
 <title>{{nb_title}}</title>
 
 {%- if "widgets" in nb.metadata -%}
-<script src="https://unpkg.com/jupyter-js-widgets@2.0.*/dist/embed.js"></script>
+<script>
+  (function() {
+    function addWidgetsRenderer() {
+      var mimeElement = document.querySelector('script[type="application/vnd.jupyter.widget-view+json"]');
+      var scriptElement = document.createElement('script');
+      var widgetRendererSrc = 'https://unpkg.com/@jupyter-widgets/html-manager@*/dist/embed-amd.js';
```
minrk: Should this always be unpinned, or should we be picking a version?

scottdraves: This is identical to nbviewer and it's been working well there: https://github.com/jupyter/nbviewer/blob/67ee47e55fa697614e03aa060a86019f75f2ddca/nbviewer/templates/notebook.html#L90
and that code was taken from jason's documentation:
http://ipywidgets.readthedocs.io/en/stable/migration_guides.html?highlight=migrate#updating-embedded-widgets

mpacer:
Could you add a test for this new functionality?

nbconvert/templates/html/full.tpl
mpacer: Can you add some comments to explain the reasoning behind it this entire function? Also please include a pointer to jason's documentation in the comments?

PR #779, opened 2018-03-18
save widgets after execute

scottdraves:
finished @ricklupton's PR

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -3,7 +3,7 @@ and updates outputs"""
 
 # Copyright (c) IPython Development Team.
 # Distributed under the terms of the Modified BSD License.
-
+import base64
 from textwrap import dedent
 
 try:
@@ -274,6 +274,7 @@ class ExecutePreprocessor(Preprocessor):
         self.kc.allow_stdin = False
         self.nb = nb
         self.widget_state = {}
+        self.widget_buffers = {}
 
         try:
             nb, resources = super(ExecutePreprocessor, self).preprocess(nb, resources)
@@ -291,6 +292,10 @@ class ExecutePreprocessor(Preprocessor):
                         'version_minor': 0,
                     }
                 }
+                for key, widget in self.nb.metadata.widgets['application/vnd.jupyter.widget-state+json']['state'].items():
+                    buffers = self.widget_buffers.get(key)
+                    if buffers:
+                        widget['buffers'] = buffers
 
         delattr(self, 'nb')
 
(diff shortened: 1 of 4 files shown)


scottdraves:
Looking into fixing on Python 2.7...

mpacer:
@jasongrout So is this your preferred mechanism for having widgets in exported contexts? Rather than with the custom mimetype and using the cdn approach?

nbconvert/preprocessors/execute.py
```diff
@@ -273,12 +273,29 @@ def start_new_kernel(startup_timeout=60, kernel_name='python', **kwargs):
...
         self.nb = nb
+        self.widget_state = {}
+        self.widget_buffers = {}
 
         try:
             nb, resources = super(ExecutePreprocessor, self).preprocess(nb, resources)
         finally:
             self.kc.stop_channels()
             self.km.shutdown_kernel(now=self.shutdown_kernel == 'immediate')
+            if self.widget_state:
```
mpacer: Can you isolate this functionality in a separate method?

setup.py
```diff
@@ -199,7 +199,7 @@ def run(self):
 ]
 
 extra_requirements = {
-    'test': ['pytest', 'pytest-cov', 'ipykernel', 'jupyter_client>=4.2'],
+    'test': ['pytest', 'pytest-cov', 'ipykernel', 'jupyter_client>=4.2', 'ipywidgets'],
```
mpacer: Does this need to be a specific version?

nbconvert/preprocessors/tests/test_execute.py
```diff
@@ -270,3 +273,33 @@ def test_execute_function(self):
...
+
+        model_ids = [
+            data['application/vnd.jupyter.widget-view+json']['model_id']
+            for data in output_data
+            if 'application/vnd.jupyter.widget-view+json' in data
+        ]
+
+        wdata = output_nb['metadata']['widgets'] \
+                ['application/vnd.jupyter.widget-state+json']
+        for k in model_ids:
```
mpacer: Nice tests!

setup.py
jasongrout: should probably be at least >=7

nbconvert/preprocessors/tests/files/widget-hello-world.ipynb
```diff
@@ -0,0 +1,47 @@
...
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "import ipywidgets\n",
+    "ipywidgets.Label('Hello World')"
+   ]
+  }
+ ],
+ "metadata": {},
```
jasongrout: Where is the widget state for the view above?

jasongrout:
> @jasongrout So is this your preferred mechanism for having widgets in exported contexts? Rather than with the custom mimetype and using the cdn approach?

I glanced through it, and it looks all right for storing widget state in a notebook. For a notebook, the widget views are stored as custom mimetypes, and the widget state is stored in the notebook metadata. The cdn is used by an html output, for example, to actually get the js code to render the widget views.

blink1073:
@jasongrout, I'm about to cut 5.4, is this okay in your mind?

Juanlu001:
This needs a rebase, thanks for tackling it @scottdraves!

MSeal:
I believe this is now completed via https://github.com/jupyter/nbconvert/pull/900 -- please reopen and rebase if you think there's anything that got missed.

PR #780, opened 2018-03-18
handle embedded images in html converter

scottdraves:


mpacer:
Can you write tests for this functionality? I think it will require adding a new notebook

blink1073:
Thanks!

Issue #720, opened 2017-12-15
Why is nbconvert removing empty code cells?


mgeier:
At first I thought this is a bug, but this seems to be the intended behavior.

I just don't get it.
If the notebook contains an empty cell, I think it would be appropriate that the file created by `nbconvert` also contains that empty cell.

This behavior was introduced in #616.

I can of course disable this for my own use if I have to, but I'm wondering why removing stuff that's visible in the Jupyter Notebook should be the default.

mgeier:
Ping?

takluyver:
cc @mpacer @tillahoffmann ?

mgeier:
I've created #784 to fix this.

mgeier:
This is still very much broken, but since nobody seems to want to fix this, I've created a work-around for my problem: https://github.com/spatialaudio/nbsphinx/pull/200.

PR #784, opened 2018-03-21
Don't remove empty cells by default

mgeier:
Since there was no real response to #720, I thought I'll try it with a pull request ...

This PR restores the sane behavior of not removing stuff without the user wanting to remove stuff.

It also fixes the docstring so if people actually want to remove stuff, they can learn how to do that.


mgeier:
Ping?

mgeier:
Ping?

nbconvert/preprocessors/regexremove.py
```diff
@@ -20,16 +20,14 @@ class RegexRemovePreprocessor(Preprocessor):
...
     is removed from the notebook.
 
-    By default, `patterns = [r'\Z']` which matches the empty string such that
-    strictly empty cells are removed. To modify the list of matched patterns,
+    To modify the list of matched patterns,
     modify the patterns traitlet. For example, execute the following command
-    to convert a notebook to html and remove cells containing only whitespace:
+    to convert a notebook to html and remove cells containing only whitespace::
 
-    > jupyter nbconvert --RegexRemovePreprocessor.enabled=True \
```
tillahoffmann: The [default value](https://github.com/jupyter/nbconvert/blob/308fb2914a6528e44cffb9a697bac9d185e14fc3/nbconvert/preprocessors/base.py#L27) for the `enabled` flag appears to be `False`. We should either override the default value for the `RegexRemovePreprocessor` or keep the `--RegexRemovePreprocessor.enabled=True` flag. Or is a preprocessor enabled automatically if one of its attributes (`patterns`) is set?

nbconvert/preprocessors/regexremove.py
```diff
@@ -20,16 +20,14 @@ class RegexRemovePreprocessor(Preprocessor):
     of unicode strings. If the contents match any of the patterns, the cell
     is removed from the notebook.
 
-    By default, `patterns = [r'\Z']` which matches the empty string such that
-    strictly empty cells are removed. To modify the list of matched patterns,
+    To modify the list of matched patterns,
     modify the patterns traitlet. For example, execute the following command
-    to convert a notebook to html and remove cells containing only whitespace:
+    to convert a notebook to html and remove cells containing only whitespace::
```
tillahoffmann: Small typo of an extra `:` here.

nbconvert/preprocessors/regexremove.py
```diff
@@ -20,16 +20,14 @@ class RegexRemovePreprocessor(Preprocessor):
...
 
-    > jupyter nbconvert --RegexRemovePreprocessor.enabled=True \
-      --RegexRemovePreprocessor.patterns="['\\s*\\Z']" mynotebook.ipynb
+      jupyter nbconvert --RegexRemovePreprocessor.patterns="['\\s*\\Z']" mynotebook.ipynb
 
-    The first command line argument enables the preprocessor and the second
-    sets the list of patterns to '\\s*\\Z' which matches an arbitrary number
+    The command line argument
+    sets the list of patterns to ``'\\s*\\Z'`` which matches an arbitrary number
```
tillahoffmann: Are single back ticks sufficient here?

nbconvert/preprocessors/regexremove.py
mgeier: @tillahoffmann Thanks for the review!

I'm not quite sure, but I think the preprocessor is enabled by default because of this:

https://github.com/jupyter/nbconvert/blob/9f5c402b85cffdc3599a426fa13e9163946c6e2c/nbconvert/exporters/templateexporter.py#L126-L135

I don't care if it is enabled or not, as long as it doesn't remove stuff by default.

nbconvert/preprocessors/regexremove.py
mgeier: This is actually reST syntax, see http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks

Without that, the formatting doesn't work, see the currently broken docs:
https://nbconvert.readthedocs.io/en/5.3.1/api/preprocessors.html#nbconvert.preprocessors.RegexRemovePreprocessor

nbconvert/preprocessors/regexremove.py
mgeier: I tried it, and apparently not.

Sphinx allows to customize the formatting of single backticks, but it is apparently not configured for typewriter text (and I also don't think it would be a good idea to do so).

mgeier:
@tillahoffmann Is there anything else you need me to do?

tillahoffmann:
Happy for this PR to be merged. As an aside, thank you for the detailed Sphinx explanation.

blink1073:
Thanks!

PR #789, opened 2018-03-27
Latex ipython template: draw a vertical line left of ipython code

timhoffm:
This PR adds vertical lines to the left of code blocks in Latex output. The lines are colored in the respective input/output color.

IMHO this greatly improves readability as it's easier to distinguish between code and text and between input and output. Also it helps to see which code was originally grouped in the cells.

See the following example:

grafik

grafik

### DIFF
--- a/nbconvert/templates/latex/document_contents.tplx
+++ b/nbconvert/templates/latex/document_contents.tplx
@@ -13,7 +13,7 @@
 
 % Display python error text as-is
 ((* block error *))
-    \begin{Verbatim}[commandchars=\\\{\}]
+    \begin{Verbatim}[commandchars=\\\{\},frame=leftline,framerule=1pt,rulecolor=\color{red}]
 ((( super() )))
     \end{Verbatim}
 ((* endblock error *))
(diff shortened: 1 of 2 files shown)


blink1073:
Thanks, @timhoffm!  @tgeorgeux, does this look good to you?

Carreau:
I like it.

t-makaro:
Can we reconsider? I don't like that this occurs at the ```document_contents.tplx``` level. I have a template that's designed to mimic the notebook more closely [here](https://github.com/t-makaro/nb_pdf_template).

image

The extra line is super out of place in my templates, and I'll have to overwrite this change.

t-makaro:
I also just realised that this change will mess with people who use nbconvert to produce reports by hiding the code from the converted document since the lines will again be out of place.

MSeal:
I'm going to remove the 5.5 milestone on this PR, as we're about to release and there hasn't been a resolution to concerns raised sufficient to warrant merging, changing, or closing the PR imho.

mgeier:
There is an IMHO more promising approach in #992.

timhoffm:
Closing in favor of #992.

PR #793, opened 2018-04-05
update font-awesome version for slides

kikocorreoso:
Update font-awesome version when converting to slides in order to use that used by the notebook itself:
https://github.com/jupyter/notebook/blob/master/bower.json#L10


mpacer:
@kikocorreoso In the future can you include your github name in your metadata for committing? This allows us to automatically give you credit when we're making a new release.

PR #811, opened 2018-05-10
Moved super call into wrapper method for easier subclassing of ExecutePreprocessor

MSeal:
Needed for papermill repository to be able to safely inherit this class and manipulate it's functionality without copying it's preprocess code.

No functionality changes.

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -217,6 +217,31 @@ class ExecutePreprocessor(Preprocessor):
     # }
     _display_id_map = Dict()
 
+    def base_preprocess(self, nb, resources):
+        """
+        Applies base preprocessing of notebooks. Separated from preprocess for
+        easy access for overriding and inheritance capabilities.
+
+        The input argument `nb` is modified in-place.
+
+        Parameters
+        ----------
+        nb : NotebookNode
+            Notebook being executed.
+        resources : dictionary
+            Additional resources used in the conversion process. For example,
+            passing ``{'metadata': {'path': run_path}}`` sets the
+            execution path to ``run_path``.
+
+        Returns
+        -------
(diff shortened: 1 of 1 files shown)


MSeal:
Test failure appears completely unrelated to the change, but I don't appear to have permissions to retrigger it

mpacer:
I think that we I would prefer to move the stuff that occurs before the preprocess call into its own method (maybe called `setup_preprocess`?) and to have it available in all preprocessors, rather than to have this hook to override the base preprocess method specifically inside `ExecutePreprocessor` and no others. In general, I'd prefer the preprocessors to have as much of a shared control flow as possible so that someone's knowledge about one preprocessor transfers as easily as possible to other preprocessors (it's hard enough teaching people how to work with this code-base as it is :( ). Also that way we can count on the preprocessors to have as common a signature on `preprocess` as is possible.

**NB**: This still approach still helps in the `papermill` case because it allows "overwriting preprocess" to not need to copy much code, since it can reuse that `setup_preprocess` method. It also has the nice property of cleaning up the internal logic of `preprocess`. Also… it'll probably mean  moving the closure out of the function (since as far as I can tell … it's not using anything from the local namespace… which should make debugging kernel stuff much easier).

mpacer:
@MSeal If you want I can take a stab at implementing how I'm thinking about this.

MSeal:
Go for it if you like. I'll get to working on it next week otherwise. I gave you permissions to my fork so feel free to push directly over this branch.

mpacer:
@MSeal rather than work off of this — I found it more convenient to build off of #816 since I needed to make these changes to more easily test the changes that I needed. If those are merged, do they suffice to simplify the implementation where it's not too much of an overhead? If not we can keep working on this.

MSeal:
Sure, I'll close this one.

Issue #610, opened 2017-06-23
Mathjax tests for markdown are messy and complicated, could be improved


mpacer:
We should be able to make that set of tests much nicer. 

For example, the  [`test_markdown2html_math_mixed`](https://github.com/jupyter/nbconvert/blob/master/nbconvert/filters/tests/test_markdown.py#L163) is dependent on whether there is a space after the final period or not, which is an issue with checking for whether content is "in" the resulting html, and the html paragraph tag closes before a new line if there is such a space.

I don't know a generally good approach to this. Maybe it must be messy but that sort of finicky-ness seems undesirable.

mpacer:
Note: our current approach to parsing text allows us to treat _everything_ as inline text, which seems bizarre, but – technically – less complicated! 

https://github.com/mpacer/nbconvert/tree/mjax_all_inline

I don't know _why_ this is possible. But I think it points toward problems with our current mathjax handling.

mpacer:
note with #611 all the logic is now inline.

@pxhanus Are you interested in tackling this?

jzf2101:
@mpacer could we add this to our list of issues for hacktoberfest?

jzf2101:
@fenwickipedia also could be interesting?

PR #814, opened 2018-05-16
Check for too recent of pandoc version

hfuhrhurr:
In researching how I can contribute to #610 at the PyCon Sprints, I noticed 2 of the 9 tests in unmodified `test_markdown.py` failed.

@ivanov suggested I make a trivial change and submit a PR so the code would run through Travis to see if the failures are replicated.

### DIFF
--- a/nbconvert/filters/tests/test_markdown.py
+++ b/nbconvert/filters/tests/test_markdown.py
@@ -1,3 +1,6 @@
+# trivial line added to run travis on this file
+# because, locally, 2 of the 9 tests fail
+
 # coding: utf-8
 """Tests for conversions from markdown to other formats"""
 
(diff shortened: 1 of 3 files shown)


hfuhrhurr:
My local tests were failing because my version of pandoc was too recent (2.2.1).  Using 1.19.1 locally was met with success.  The PR now incorporates logic to notify a user of an unsupported version of pandoc (if it's 2.0.0 or newer).  This logic won't cause a failure; it will simply inform the user an unsupported version of pandoc is being used.

@ivanov

nbconvert/filters/tests/test_markdown.py
```diff
@@ -1,3 +1,6 @@
+# trivial line added to run travis on this file
+# because, locally, 2 of the 9 tests fail
```
ivanov: Could we take this out?

ivanov:
Thanks for your contribution @hfuhrhurr ! Merging.

mpacer:
Btw just as a heads up, this broke our docs by introducing an API incompatibility (`max_ver` shouldn't be a positional argument, but a keyword argument).

PR #818, opened 2018-05-21
[WIP] Export markdown cells as block strings instead of comments

tonyfast:
This pull request changes how markdown cells are written from the Python exporter.  Currently, nbconvert creates Python comments which are not captured by the ast.  This change wraps markdown cells in block quotes when exported to python.

The advantage of this change is that the first markdown cell may be used as both a module docstring and doctest.

screen shot 2018-05-21 at 1 58 22 pm

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -40,6 +40,7 @@ default_filters = {
         'markdown2latex': filters.markdown2latex,
         'markdown2rst': filters.markdown2rst,
         'comment_lines': filters.comment_lines,
+        'block_string_lines': filters.block_string_lines,
         'strip_ansi': filters.strip_ansi,
         'strip_dollars': filters.strip_dollars,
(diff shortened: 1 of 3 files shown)


mpacer:
This is neat. @takluyver do you see any side effects with this?

Could you explain why your example includes `assert True`?

tonyfast:
Thanks for having a look at this.

> Could you explain why your example includes assert True?

`>>> assert True` is a doctest, I was just using it  aslogic to prove that a doctest was indeed captured from the first markdown cell.

> Could you explain why your example includes jupyter nbconvert --version?

This is just a sanity check for example.

takluyver:
I don't see anything obviously wrong with that technically, though I imagine that there'll be a fairly even split of people who like it and people who don't.

What might be neat is to do it with just the first cell (if that's markdown), and convert the rest as ordinary comments. Unfortunately I think that's trickier to do in the template system.

PR #825, opened 2018-06-05
No input flag

mpacer:
This adds a new `--no-input` as talked about in few places. This removes prompts and input areas, leaving only the outputs.

This also adds a new test for this feature and removes some of the warnings and extraneous printing that was being done in our tests.

### DIFF
--- a/nbconvert/filters/tests/test_highlight.py
+++ b/nbconvert/filters/tests/test_highlight.py
@@ -75,8 +75,6 @@ class TestHighlight(TestsBase):
                 ( ht, ('def', )),
                 ( rb, ('def','end'  ) )
                 ]:
-            print(tkns)
-            print(lang)
             root = xml.etree.ElementTree.fromstring(lang)
             self.assertEqual(self._extract_tokens(root,'k'), set(tkns))
 
--- a/nbconvert/filters/tests/test_markdown.py
+++ b/nbconvert/filters/tests/test_markdown.py
@@ -83,7 +83,7 @@ class TestMarkdown(TestsBase):
     def test_pandoc_extra_args(self):
         # pass --no-wrap
         s = '\n'.join([
-            "#latex {{long_line | md2l(['--no-wrap'])}}",
+            "#latex {{long_line | md2l(['--wrap=none'])}}",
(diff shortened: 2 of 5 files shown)


t-makaro:
I believe ```exclude_input=True``` does take out the input prompt with it since it removes the whole block, so having ```exclude_input_prompt=True``` seems redundant. I also don't see how a custom template could overwrite that behaviour. I'm not complaining. Really, I just want to make sure that [my templates](https://github.com/t-makaro/nb_pdf_template) have the expected behaviour. 

Also, for symmetry, should there be a ```--no-output```?

mpacer:
@t-makaro 

Yes you're right it is redundant.
A custom template would need to override the `null.tpl` itself… which is definitely a bizarre anti-pattern that we shouldn't be coding for.

In theory we could include a `--no-output` flag… but I'm somewhat opposed. Our CLI is already pretty bloated, and I don't think anyone has actually asked for that (since there is always the `ClearOutputPreprocessor` that can be enabled). 

For historical context: 
A lot of this functionality was introduced because you couldn't implement a Preprocessor for removing input because then it would not be a valid notebook file. This comes from our constraint that Preprocessors are notebook → notebook transforms… which is one of the cleanest constraints we've got in nbconvert.

t-makaro:
I was under the impression that the ```ClearOutputPreprocessor``` actually overwrote the file. Though I think that only the ```--clear-output``` flag does that (since it defaults to inplace if I'm not wrong). So, long as there is a way to convert to another format without outputs and without clearing the original notebook file of outputs then ```--no-output``` really isn't needed.

I've definitely [seen people ask](https://www.reddit.com/r/Python/comments/8vl9fi/in_jupyter_notebook_how_to_convert_ipynb_file/) for no input for generating reports. I personally don't see myself using either ```no-input``` or ```no-output```.

MSeal:
`--no-input` has a lot of value for reporting purposes. If someone wants outputs hidden for this we cna always discuss it in a new issue / pull request

mpacer:
So I'd like to see this in 5.4 if we're going to be doing the release and there seems to be agreement that this is a good feature that people have been asking for for some time.

PR #830, opened 2018-06-18
Support for \mathscr

igrmk:
MathJax supports \mathscr command according to http://docs.mathjax.org/en/latest/tex.html
It works in notebook however exporting to PDF reports "! Undefined control sequence".
I added commonly used package "mathrsfs" to base template to support it.

This command provides a script font for capital Latin letters.
Description of font itself is here https://ctan.org/pkg/rsfs


igrmk:
This command provides a script font for capital Latin letters.
Description of font itself is here https://ctan.org/pkg/rsfs

PR #836, opened 2018-07-09
Expose pygments styles

### LINKED ISSUE #832: [feature request] Add option/settings to change the pygments style
By modifying [this line](https://github.com/jupyter/nbconvert/blob/474fe3c080211b5eefdd6c3e7c313a1ddba2fa9d/nbconvert/preprocessors/csshtmlheader.py#L73) and [this line](https://github.com/jupyter/nbconvert/blob/474fe3c080211b5eefdd6c3e7c313a1ddba2fa9d/nbconvert/preprocessors/latex.py#L46), to include ```style="some string or pygments style object"``` within the calls to ```HtmlFormatter``` and ```LatexFormatter```, we can open a new realm for new possibilities for document customization.

For example, the first link is equivalent to:
```
formatter = HtmlFormatter(style="default")
```
But pygments offers many styles:
```
>>> from pygments.styles import get_all_styles
>>> styles = list(get_all_styles())
```
and a plugin structure for styles.

If kernels define their own style, then this could be used to change the syntax highlighting given the style from the kernel. Currently, the lexer is switched depending on the language of the notebook, but there is no way to change the actual syntax highlighting style without subclassing the preprocessor. 

I'd imagine adding a
```
c.Preprocessor.style = "mystyle"
```
to the config file or the flag
```
--Preprocessor.style="mystyle"
```
from the command line to control the style.

I am currently working on a pygments style that faithfully recreates jupyter notebook's codemirror style for ipython, and it would be nice not to have to subclass the preprocessor to use it.

t-makaro:
closes #832  

Some discussion is needed, and I'm super willing to make changes to get this PR accepted. This will let html, and latex exporting to use any pygments style, which means one can set the colors that are used to highlight code!

I also have code to let the pygments style be automatically selected from the language of the notebook. This would mean that a kernel could define a style and use the pygments plugin system allowing nbconvert to automatically select the right color scheme for the language to match what the notebook looks like. Sort of like how the appropriate lexer is already automatically chosen based on notebook metadata. That should probably be left to a future PR though.


t-makaro:
https://github.com/JuliaLang/IJulia.jl/issues/105

t-makaro:
I don't know why. The build only failed on python 3.3, and it did so before getting to any tests. @mpacer @takluyver

takluyver:
Looks like it's getting a version of setuptools that doesn't support Python 3.3. There are probably ways to work around that, but I'd say we can just drop support for Python 3.3 as well.

mpacer:
This looks great! I want to kick it's tires a bit but I'm really liking this! 😃

t-makaro:
Well, I've been doing some work.

Since, for people to use this feature it requires useful styles to be defined. I created the first example over [here](https://github.com/t-makaro/nb_hightlight_styles). Ideally, kernels could define styles like this, but it may be hard for some kernels to use the entry points, so I'd be willing to compile a bunch of community created styles into a package.

t-makaro:
I've been thinking about this a lot and I'm not sure that the name ```style``` makes much sense to nbconvert. I choose to name it "style" because that's what pygments calls it, but now I think that ```highlight``` might be more descriptive, so the flag to change the highlight colors might be something like:
```
--latexpreprocessor.highlight=pygments_style_name
```
instead of:
```
--latexpreprocessor.style=pygments_style_name
```
since style can be confused with the style that the templates are used to change.

mpacer:
Could you write some tests for this feature? Also some docs explaining a bit about pygments styles? 

For example you could give people the ability to quickly see the different ways their code would be rendered with the following snippet:

```
from pygments.styles import get_all_styles
from pygments.formatters import Terminal256Formatter
from pygments.lexers import HtmlLexer
from pygments import highlight

code = "<tag attrib=\"value\"> content </tag>"
for style in get_all_styles():
    highlighted_code = highlight(code, HtmlLexer(), Terminal256Formatter(style=style))
    print(f"{style}: {highlighted_code}")
```

t-makaro:
Ok. I will right some docs (maybe a section under https://nbconvert.readthedocs.io/en/latest/customizing.html).

I suppose to test I can check that changing the style config actually changes the style definitions that are returned. 

I'll find the time in the next couple days to do this.

takluyver:
Naming: maybe call it something like `highlighting_style` - there are various aspects of highlighting that might be configured, and I don't think a setting just called `highlight` would obvious point to styles.

PR #842, opened 2018-07-18
Add --save-on-error flag, closes #626

### LINKED ISSUE #626: ExecutePreprocessor: stop at error but save notebook (and stack trace)
Hi,

I am trying to use `nbconvert` to run notebooks from the commandline. I would like to stop if cell execution fails anywhere (as it is the default), but in this case I see only the raw exception at the command line. Since the notebook is not saved in this case, I have no access to the full stack trace of the exception.

is it possible to either
* display the full stack trace in the console or
* save the notebook after the first error? #120 suggests to me that that should be the default. But for me the notebook is never saved unless I specify `allow_errors`, in which case all cells are executed

Thankyou for your help!

berleon:
Save the notebook even if an error occurs.

### DIFF
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -72,6 +72,13 @@ nbconvert_flags.update({
          "(the default behaviour is to abort conversion). This flag "
          "is only relevant if '--execute' was specified, too.")
         ),
+    'save-on-error' : (
+        {'ExecutePreprocessor' : {'save_on_error' : True}},
+        ("Saves the notebook even if one cell throws "
+         "an error. The error message is shown in the cell output."
+         "(the default behaviour is to not save the notebook on an error). This flag "
+         "is only relevant if '--execute' was specified, too.")
+    ),
     'stdin' : (
         {'NbConvertApp' : {
             'from_stdin' : True,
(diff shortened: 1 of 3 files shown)


takluyver:
So there's already an option `--allow-errors` which will continue if an error occurs. Is it important to you that you stop executing the notebook but still convert/save it afterwards? I'm slightly concerned by how many error-handling options we're growing here - it feels like there should be a simpler way.

With your new option, is the error shown in the terminal at all?

berleon:
My use-case is that I am running some experiments in the notebook. If an error occurs, I want to see the previous outputs. However, continue running (i.e. `--allow-errors`) can cost unnecessary time (multiple hours). 

I agree that the  additional flag might make it unnecessary complicated.  What would you say to unify the error handling behaviour as `--on-error=[abort|continue|stop]` where `abort` would be the current default behaviour, `continue` would correspond to `--allow-errors` and `stop` would mean to still save/convert the notebook but not to continue with the execution.

As with `--allow-errors`, the error is not shown in the terminal. I could print the error to stdout and add a message that the notebook will still be converted/saved.

ciaochiaociao:
Any update?

PR #843, opened 2018-07-18
Drop support for python 3.3

berleon:
As mention in #836, this PR drops support for python 3.3.


takluyver:
Thanks, I think this makes sense.

PR #844, opened 2018-07-24
Add note about local reveal_url_prefix

OliverEvans96:
I noticed that absolute paths do not work since a URL is constructed from the directory, but this was not clear from the documentation. If someone has a better wording, I'll be glad to change it. 

Also, this should updated in https://nbconvert.readthedocs.io/en/latest/config_options.html but the 'Edit on Github' button is broken.


minrk:
Thanks!

PR #847, opened 2018-07-26
properly treat JSON data

### LINKED ISSUE #846: Unable to use ExtractOutputPreprocessor with json output
To reproduce:

1. create a notebook where one of the outputs has a JSON mimetype.
2. Add the mimetype of the output to the filetypes of `ExtractOutputPreprocessor`
3. Run preprocessing

Preprocessor fails with `AttributeError: encode`.

---

The origin of the failure is relatively straightforward to trace back to [these lines](https://github.com/jupyter/nbconvert/blob/master/nbconvert/preprocessors/extractoutput.py#L77-L84).

There we're checking if the data is base64 encoded, and otherwise we're calling `data.encode('utf-8')`.
This relies on an assumption that data is always a string. However if data is a legitimate JSON, at this moment is has a type `nbformat.notebooknode.NotebookNode` (I figured that running a debugger).
Therefore we're now trying to access "encode" attribute of that object.

Instead we should treat the case when data is not a string specially.

akhmerov:
closes #846 

Still TODO: tests.

I would additionally appreciate an opinion whether this fix is clean enough.

### DIFF
--- a/nbconvert/preprocessors/extractoutput.py
+++ b/nbconvert/preprocessors/extractoutput.py
@@ -8,10 +8,17 @@ notebook file.  The extracted outputs are returned in the 'resources' dictionary
 from binascii import a2b_base64
 import sys
 import os
+import json
 from mimetypes import guess_extension
 
 from traitlets import Unicode, Set
 from .base import Preprocessor
+
+if sys.version_info < (3,):
+    text_type = basestring
+else:
+    text_type = str
+
 
 def guess_extension_without_jpe(mimetype):
     """
@@ -72,6 +79,17 @@ class ExtractOutputPreprocessor(Preprocessor):
             for mime_type in self.extract_output_types:
                 if mime_type in out.data:
                     data = out.data[mime_type]
+
+                    if (
+                        not isinstance(data, text_type)
+                        or mime_type == 'application/json'
+                    ):
+                        # Data is either JSON-like and was parsed into a Python
+                        # object according to the spec, or data is for sure
+                        # JSON. In the latter case we want to go extra sure that
+                        # we enclose a scalar string value into extra quotes by
+                        # serializing it properly.
+                        data = json.dumps(data)
 
                     #Binary files are base64-encoded, SVG is already XML
                     if mime_type in {'image/png', 'image/jpeg', 'application/pdf'}:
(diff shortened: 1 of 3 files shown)


nbconvert/preprocessors/extractoutput.py
```diff
@@ -73,6 +75,10 @@ def preprocess_cell(self, cell, resources, cell_index):
                 if mime_type in out.data:
                     data = out.data[mime_type]
 
+                    # If data was JSON, it will become a NotebookNode at this point.
+                    if isinstance(data, NotebookNode):
+                        data = json.dumps(data)
```
minrk: This shouldn't actually be turning it to text, I think. That would do incorrect things for `--to notebook`, for example. I think `ExtractOutputPreprocessor` needs to be taught how to handle when this is a non-string (i.e. json serialize as part of writing to a file), rather than casting it to a string here.

Test ideas:

- convert to notebook and verify that json outputs are preserved as objects
- extract output with json
- make sure to include a variety of json outputs which can be dicts, lists, scalars, etc., not just dicts.

akhmerov:
@minrk I think this is good now. Thanks for the suggestion to check multiple data types of JSON, this was very useful.

On a side-note: is a string scalar value allowed by `nbformat`? In the [spec itself](http://nbformat.readthedocs.io/en/latest/format_description.html#cell-types) I see nothing against it, but `IPython.display.JSON` only allows to represent container and not scalar data.

akhmerov:
OK, tests from other preprocessors are failing. What's the better strategy here? Update other tests or separate the test case for this preprocessor?

nbconvert/preprocessors/extractoutput.py
akhmerov: > convert to notebook and verify that json outputs are preserved as objects

I believe this was already the case: we're dealing with a preprocessor that doesn't modify the cell outputs.

I have added a test to verify that it is indeed the case.

akhmerov:
@minrk I think this is ready. Travis failure on python 2.7 seems unrelated, and a glitch.

minrk:
> is a string scalar value allowed by nbformat?

I believe it should be. Initially, JSON *had* to be a dict, but now I believe any JSON type should work.

PR #854, opened 2018-08-07
Move onlyif_cmds_exist decorator to test-specific utils

### LINKED ISSUE #853: pytest appears to be a runtime dependency
Running `nbconvert` from source, it seems that `nbconvert/utils/io.py` imports `pytest` since PR https://github.com/jupyter/nbconvert/pull/758

SylvainCorlay:
Fixes #853 

cc @Carreau @minrk


SylvainCorlay:
@Carreau that one is very mechanical and makes the master branch usable without pytest.

Carreau:
That still move fonction import. So may be breaking API

Carreau:
> This is moving something that was introduced after the last release.

Ok, that was one of the things I was going to check. 

Merging then.

PR #864, opened 2018-08-25
Add export_from_notebook

starcruiseromega:
I was working on https://github.com/jupyter/notebook/pull/3879 and it
looks like the intended way to determine whether the exporter should
show up in the list generated by the notebook server was by checking
`export_from_notebook`, but it isn't defined for any of the builtin
exporters.

The docs also say this specifies a friendly name for the exporter. In
the PR mentioned above, I used the name defined by the entrypoint to key
the exporter. It sounds like maybe we should use the value in
`export_from_notebook` instead, so I've made them match, but perhaps
it's confusing to have a "name" for the entrypoint in two places.

### DIFF
--- a/nbconvert/exporters/asciidoc.py
+++ b/nbconvert/exporters/asciidoc.py
@@ -23,6 +23,7 @@ class ASCIIDocExporter(TemplateExporter):
         return 'asciidoc'
 
     output_mimetype = 'text/asciidoc'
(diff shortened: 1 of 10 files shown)


starcruiseromega:
@ivanov @mpacer @dharmaquark

blink1073:
Thanks!

PR #872, opened 2018-08-31
Fix api break

mpacer:
Our documentation hasn't built since #814 was merged. 

This is because it broke our check_version API which is used by nbsphinx to convert notebooks to sphinx compatible documentation. 

This updates that to maintain backwards compatibility (by making it an optional argument) while maintaining the new functionality. 

This also adds a test for the expected functionality so we don't break APIs again. 

Also this updates the docs on how to build docs and release (and makes it easier to get the doc-building dependencies with a new extra requirements flag in `setup.py`).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,12 +9,33 @@ and the [IPython Contributing Guide](https://github.com/ipython/ipython/blob/mas
 In order to test all the features of nbconvert you need to have `pandoc` and
 `TexLive` installed. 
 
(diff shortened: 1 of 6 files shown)


mpacer:
@MSeal we should merge this before releasing.

ivanov:
sorry about introducing a breaking change @mpacer - PyCon sprints were a bit of a blur, I had `git grep`ed to ensure that function wasn't being used anywhere else in the code base, but I was wrong to not consider others using this API.  Thank you for fixing and adding tests so this doesn't happen again.

PR #877, opened 2018-09-06
Add option to overwrite the highlight_code filter

danielfrg:
Currently the `highlight_code` filter is created on the exporter `from_notebook_node` function so it cannot be overwritten the `filters` trait.

### DIFF
--- a/nbconvert/exporters/html.py
+++ b/nbconvert/exporters/html.py
@@ -90,8 +90,11 @@ class HTMLExporter(TemplateExporter):
     def from_notebook_node(self, nb, resources=None, **kw):
         langinfo = nb.metadata.get('language_info', {})
         lexer = langinfo.get('pygments_lexer', langinfo.get('name', None))
-        self.register_filter('highlight_code',
-                             Highlight2HTML(pygments_lexer=lexer, parent=self))
+        highlight_code = Highlight2HTML(pygments_lexer=lexer, parent=self)
+        if "highlight_code" in self.filters:
+            highlight_code = self.filters["highlight_code"]
+        self.register_filter('highlight_code', highlight_code)
+        
         output, resources = super(HTMLExporter, self).from_notebook_node(nb, resources, **kw)
         att_output = self.process_attachments(nb, output)
         return att_output, resources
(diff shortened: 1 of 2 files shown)


nbconvert/exporters/html.py
```diff
@@ -90,8 +90,11 @@ def process_attachments(self, nb, output):
     def from_notebook_node(self, nb, resources=None, **kw):
         langinfo = nb.metadata.get('language_info', {})
         lexer = langinfo.get('pygments_lexer', langinfo.get('name', None))
-        self.register_filter('highlight_code',
-                             Highlight2HTML(pygments_lexer=lexer, parent=self))
+        highlight_code = Highlight2HTML(pygments_lexer=lexer, parent=self)
```
MSeal: Can consolidate some code here 
```suggestion
        highlight_code = self.filters.get('highlight_code',
                                                           Highlight2HTML(pygments_lexer=lexer, parent=self))
```

nbconvert/exporters/latex.py
```diff
@@ -78,8 +78,11 @@ def default_config(self):
     def from_notebook_node(self, nb, resources=None, **kw):
         langinfo = nb.metadata.get('language_info', {})
         lexer = langinfo.get('pygments_lexer', langinfo.get('name', None))
-        self.register_filter('highlight_code',
-                             Highlight2Latex(pygments_lexer=lexer, parent=self))
+        highlight_code = Highlight2Latex(pygments_lexer=lexer, parent=self)
```
MSeal: Same idea here
```suggestion
        highlight_code = self.filters.get('highlight_code',
                                                           Highlight2HTML(pygments_lexer=lexer, parent=self))
```

nbconvert/exporters/html.py
danielfrg: Good idea, done.

nbconvert/exporters/latex.py
danielfrg: Good idea, done.

danielfrg:
Thanks for the review Mathew. I made the change you requested and added a small test.

danielfrg:
Ok, I decided to not pin the `ipython` version because it will break the tests for Python 2. Unfortunately then it required to duplicate some test files to test in IPython 6 and 7.

I ran the tests on master locally and they are failing because of this same reason so i bet if they get executed right now on travis it will fail but this will fix it.

PR #882, opened 2018-09-10
DOC: Add missing language specification to code-block

mgeier:
See http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block

I'm actually wondering why this is only a Sphinx warning (rather than an error), because it seems to silently remove the code block from the resulting page: https://nbconvert.readthedocs.io/en/5.4/changelog.html#raw-templates

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -59,7 +59,7 @@ Raw Templates
 
 Template exporters can now be assigned raw templates as string attributes by setting the ``raw_template`` variable.
 
-.. code-block::
+.. code-block:: python
 
   class AttrExporter(TemplateExporter):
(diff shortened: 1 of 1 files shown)


mgeier:
I've created an issue about the Sphinx warning vs. error situation: https://github.com/sphinx-doc/sphinx/issues/5407.

minrk:
That's surprising! I thought code-block had a default syntax-highlighting language if not specified for a given directive, or at least fallback on no highlighting.

mgeier:
@minrk Thanks for merging! And yes, it is strange, see https://github.com/sphinx-doc/sphinx/issues/5407 and https://github.com/sphinx-doc/sphinx/issues/1851.

PR #886, opened 2018-09-11
WIP Recreate old behaviour of ExecutePreprocessor(kernel_name="")

### LINKED ISSUE #878: 'NoneType' object has no attribute 'argv' with nbconvert 5.4 and nbsphinx
Recent nbconvert release broke nbsphinx:

```
$ sphinx-build -vW -b html docs/source docs-html
Running Sphinx v1.7.9
loading pickled environment... failed: build environment version not current
loading intersphinx inventory from https://docs.python.org/3/objects.inv...
loading intersphinx inventory from http://docs.astropy.org/en/stable/objects.inv...
loading intersphinx inventory from https://docs.scipy.org/doc/numpy/objects.inv...
loading intersphinx inventory from https://docs.scipy.org/doc/scipy/reference/objects.inv...
loading intersphinx inventory from http://matplotlib.org/objects.inv...
intersphinx inventory has moved: http://matplotlib.org/objects.inv -> https://matplotlib.org/objects.inv
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 26 source files that are out of date
updating environment: 26 added, 0 changed, 0 removed
reading sources... [  3%] about
reading sources... [  7%] api
reading sources... [ 11%] changelog
reading sources... [ 15%] core
reading sources... [ 19%] dastcom5 parameters
reading sources... [ 23%] examples/Catch that asteroid!
reading sources... [ 26%] examples/Comparing Hohmann and bielliptic transfers

  File "/home/juanlu/.miniconda36/envs/_test36_pol/lib/python3.6/site-packages/nbsphinx.py", line 653, in parse
    rststring, resources = exporter.from_notebook_node(nb, resources)
  File "/home/juanlu/.miniconda36/envs/_test36_pol/lib/python3.6/site-packages/nbsphinx.py", line 575, in from_notebook_node
    nb, resources = pp.preprocess(nb, resources)
  File "/home/juanlu/.miniconda36/envs/_test36_pol/lib/python3.6/site-packages/nbconvert/preprocessors/execute.py", line 359, in preprocess
    with self.setup_preprocessor(nb, resources, km=km):
  File "/home/juanlu/.miniconda36/envs/_test36_pol/lib/python3.6/contextlib.py", line 81, in __enter__
    return next(self.gen)
  File "/home/juanlu/.miniconda36/envs/_test36_pol/lib/python3.6/site-packages/nbconvert/preprocessors/execute.py", line 304, in setup_preprocessor
    self.km, self.kc = self.start_new_kernel(cwd=path)
  File "/home/juanlu/.miniconda36/envs/_test36_pol/lib/python3.6/site-packages/nbconvert/preprocessors/execute.py", line 253, in start_new_kernel
    km.start_kernel(extra_arguments=self.extra_arguments, **kwargs)
  File "/home/juanlu/.miniconda36/envs/_test36_pol/lib/python3.6/site-packages/jupyter_client/manager.py", line 246, in start_kernel
    kernel_cmd = self.format_kernel_cmd(extra_arguments=extra_arguments)
  File "/home/juanlu/.miniconda36/envs/_test36_pol/lib/python3.6/site-packages/jupyter_client/manager.py", line 170, in format_kernel_cmd
    cmd = self.kernel_spec.argv + extra_arguments
AttributeError: 'NoneType' object has no attribute 'argv'

During handling of the above exception, another exception occurred:

  File "/home/juanlu/.miniconda36/envs/_test36_pol/lib/python3.6/site-packages/sphinx/cmdline.py", line 304, in main
    app.build(args.force_all, filenames)
  File "/home/juanlu/.miniconda36/envs/_test36_pol/lib/python3.6/site-packages/sphinx/application.py", line 331, in build
    self.builder.build_update()
  File "/home/juanlu/.miniconda36/envs/_test36_pol/lib/python3.6/site-packages/sphinx/builders/__init__.py", line 342, in build_update
    'out of date' % len(to_build))
  File "/home/juanlu/.miniconda36/envs/_test36_pol/lib/python3.6/site-packages/sphinx/builders/__init__.py", line 355, in build
    updated_docnames = set(self.env.update(self.config, self.srcdir, self.doctreedir))
  File "/home/juanlu/.miniconda36/envs/_test36_pol/lib/python3.6/site-packages/sphinx/environment/__init__.py", line 568, in update
    self._read_serial(docnames, self.app)
  File "/home/juanlu/.miniconda36/envs/_test36_pol/lib/python3.6/site-packages/sphinx/environment/__init__.py", line 587, in _read_serial
    self.read_doc(docname, app)
  File "/home/juanlu/.miniconda36/envs/_test36_pol/lib/python3.6/site-packages/sphinx/environment/__init__.py", line 662, in read_doc
    doctree = read_doc(self.app, self, self.doc2path(docname))
  File "/home/juanlu/.miniconda36/envs/_test36_pol/lib/python3.6/site-packages/sphinx/io.py", line 302, in read_doc
    pub.publish()
  File "/home/juanlu/.miniconda36/envs/_test36_pol/lib/python3.6/site-packages/docutils/core.py", line 217, in publish
    self.settings)
  File "/home/juanlu/.miniconda36/envs/_test36_pol/lib/python3.6/site-packages/docutils/readers/__init__.py", line 72, in read
    self.parse()
  File "/home/juanlu/.miniconda36/envs/_test36_pol/lib/python3.6/site-packages/docutils/readers/__init__.py", line 78, in parse
    self.parser.parse(self.input, document)
  File "/home/juanlu/.miniconda36/envs/_test36_pol/lib/python3.6/site-packages/nbsphinx.py", line 664, in parse
    str(e))
nbsphinx.NotebookError: AttributeError in examples/Comparing Hohmann and bielliptic transfers.ipynb:
'NoneType' object has no attribute 'argv'

Notebook error:
AttributeError in examples/Comparing Hohmann and bielliptic transfers.ipynb:
'NoneType' object has no attribute 'argv'
```

If I remove the notebooks and the references to nbsphinx everything works fine. If I downgrade to 5.3.1 and keep the notebooks, everything works too.

I report the issue here because I understand that projects that depend on nbconvert should not break from 5.3 to 5.4, but perhaps it's indeed an nbsphinx issue. cc @spatialaudio

Environment details below:

<details>

```
$ conda info -a

     active environment : _test36_pol
    active env location : /home/juanlu/.miniconda36/envs/_test36_pol
            shell level : 1
       user config file : /home/juanlu/.condarc
 populated config files : /home/juanlu/.miniconda36/.condarc
          conda version : 4.5.11
    conda-build version : 3.12.1
         python version : 3.6.5.final.0
       base environment : /home/juanlu/.miniconda36  (writable)
           channel URLs : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
                          https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/free/linux-64
                          https://repo.anaconda.com/pkgs/free/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
                          https://repo.anaconda.com/pkgs/pro/linux-64
                          https://repo.anaconda.com/pkgs/pro/noarch
          package cache : /home/juanlu/.miniconda36/pkgs
                          /home/juanlu/.conda/pkgs
       envs directories : /home/juanlu/.miniconda36/envs
                          /home/juanlu/.conda/envs
               platform : linux-64
             user-agent : conda/4.5.11 requests/2.18.4 CPython/3.6.5 Linux/4.15.0-33-generic linuxmint/18.3 glibc/2.23
                UID:GID : 1000:1000
             netrc file : /home/juanlu/.netrc
           offline mode : False

# conda environments:
#
base                     /home/juanlu/.miniconda36
_numba36_test            /home/juanlu/.miniconda36/envs/_numba36_test
_test27                  /home/juanlu/.miniconda36/envs/_test27
_test36                  /home/juanlu/.miniconda36/envs/_test36
_test36_2                /home/juanlu/.miniconda36/envs/_test36_2
_test36_pol           *  /home/juanlu/.miniconda36/envs/_test36_pol
_testpl27                /home/juanlu/.miniconda36/envs/_testpl27
_testpl36                /home/juanlu/.miniconda36/envs/_testpl36
api36                    /home/juanlu/.miniconda36/envs/api36
astropy36_dev            /home/juanlu/.miniconda36/envs/astropy36_dev
brte36                   /home/juanlu/.miniconda36/envs/brte36
brte36dsk18              /home/juanlu/.miniconda36/envs/brte36dsk18
geostore36               /home/juanlu/.miniconda36/envs/geostore36
gmat36                   /home/juanlu/.miniconda36/envs/gmat36
marvin34                 /home/juanlu/.miniconda36/envs/marvin34
mbd36                    /home/juanlu/.miniconda36/envs/mbd36
pgadmin4                 /home/juanlu/.miniconda36/envs/pgadmin4
poliastro36              /home/juanlu/.miniconda36/envs/poliastro36
poliastro37              /home/juanlu/.miniconda36/envs/poliastro37
py34                     /home/juanlu/.miniconda36/envs/py34
py36                     /home/juanlu/.miniconda36/envs/py36
                         /home/juanlu/.miniconda36/envs/py36/conda-bld/cspice_1529866334362/_build_env
                         /home/juanlu/.miniconda36/envs/py36/conda-bld/cspice_1529866449168/_build_env
                         /home/juanlu/.miniconda36/envs/py36/conda-bld/cspice_1529866543542/_build_env
py37                     /home/juanlu/.miniconda36/envs/py37
py37_def                 /home/juanlu/.miniconda36/envs/py37_def
pypiserver36             /home/juanlu/.miniconda36/envs/pypiserver36
pyrenfe36                /home/juanlu/.miniconda36/envs/pyrenfe36
spice37                  /home/juanlu/.miniconda36/envs/spice37
telluric36               /home/juanlu/.miniconda36/envs/telluric36
unitshack36              /home/juanlu/.miniconda36/envs/unitshack36
                         /home/juanlu/Development/poliastro/benchmarks/env/1e6a43cbb25dbca4e803dce4b31b80e5
                         /home/juanlu/Development/poliastro/benchmarks/env/4ef9678914ff6e01972901ec0de8ad50

sys.version: 3.6.5 | packaged by conda-forge | (defau...
sys.prefix: /home/juanlu/.miniconda36
sys.executable: /home/juanlu/.miniconda36/bin/python
conda location: /home/juanlu/.miniconda36/lib/python3.6/site-packages/conda
conda-build: /home/juanlu/.miniconda36/bin/conda-build
conda-convert: /home/juanlu/.miniconda36/bin/conda-convert
conda-develop: /home/juanlu/.miniconda36/bin/conda-develop
conda-env: /home/juanlu/.miniconda36/bin/conda-env
conda-index: /home/juanlu/.miniconda36/bin/conda-index
conda-inspect: /home/juanlu/.miniconda36/bin/conda-inspect
conda-metapackage: /home/juanlu/.miniconda36/bin/conda-metapackage
conda-render: /home/juanlu/.miniconda36/bin/conda-render
conda-skeleton: /home/juanlu/.miniconda36/bin/conda-skeleton
conda-smithy: /home/juanlu/.miniconda36/bin/conda-smithy
conda-verify: /home/juanlu/.miniconda36/bin/conda-verify
user site dirs: 

CIO_TEST: <not set>
CONDA_DEFAULT_ENV: _test36_pol
CONDA_EXE: /home/juanlu/.miniconda36/bin/conda
CONDA_PREFIX: /home/juanlu/.miniconda36/envs/_test36_pol
CONDA_PROMPT_MODIFIER: (_test36_pol) 
CONDA_PYTHON_EXE: /home/juanlu/.miniconda36/bin/python
CONDA_ROOT: /home/juanlu/.miniconda36
CONDA_SHLVL: 1
DEFAULTS_PATH: /usr/share/gconf/cinnamon.default.path
MANDATORY_PATH: /usr/share/gconf/cinnamon.mandatory.path
PATH: /home/juanlu/.miniconda36/envs/_test36_pol/bin:/home/juanlu/.rbenv/bin:$/home/juanlu/.rbenv/shims:/home/juanlu/.gem/ruby/2.3.0/bin:/home/juanlu/.local/share/umake/bin:/home/juanlu/bin:/home/juanlu/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
REQUESTS_CA_BUNDLE: <not set>
SSL_CERT_FILE: <not set>
WINDOWPATH: 8


WARNING: could not import _license.show_info
# try:
# $ conda install -n root _license
$ conda list
# packages in environment at /home/juanlu/.miniconda36/envs/_test36_pol:
#
# Name                    Version                   Build  Channel
alabaster                 0.7.11                    <pip>
asn1crypto                0.24.0                    <pip>
astropy                   3.0.4                     <pip>
astroquery                0.3.8                     <pip>
atomicwrites              1.2.1                     <pip>
attrs                     18.2.0                    <pip>
Babel                     2.6.0                     <pip>
backcall                  0.1.0                     <pip>
beautifulsoup4            4.6.3                     <pip>
bleach                    2.1.4                     <pip>
bzip2                     1.0.6                h470a237_2    conda-forge
ca-certificates           2018.8.24            ha4d7672_0    conda-forge
certifi                   2018.8.24                py36_1    conda-forge
cffi                      1.11.5                    <pip>
chardet                   3.0.4                     <pip>
coverage                  4.5.1                     <pip>
cryptography              2.3.1                     <pip>
cycler                    0.10.0                    <pip>
decorator                 4.3.0                     <pip>
defusedxml                0.5.0                     <pip>
docutils                  0.14                      <pip>
entrypoints               0.2.3                     <pip>
gmp                       6.1.2                hfc679d8_0    conda-forge
html5lib                  1.0.1                     <pip>
idna                      2.7                       <pip>
imagesize                 1.1.0                     <pip>
ipykernel                 4.9.0                     <pip>
ipython                   6.5.0                     <pip>
ipython-genutils          0.2.0                     <pip>
ipywidgets                7.4.1                     <pip>
jedi                      0.12.1                    <pip>
jeepney                   0.3.1                     <pip>
Jinja2                    2.10                      <pip>
jplephem                  2.8                       <pip>
jsonschema                2.6.0                     <pip>
jupyter-client            5.2.3                     <pip>
jupyter-core              4.4.0                     <pip>
keyring                   15.0.0                    <pip>
kiwisolver                1.0.1                     <pip>
libffi                    3.2.1                hfc679d8_5    conda-forge
libgcc-ng                 7.2.0                hdf63c60_3    conda-forge
libstdcxx-ng              7.2.0                hdf63c60_3    conda-forge
llvmlite                  0.24.0                    <pip>
MarkupSafe                1.0                       <pip>
matplotlib                2.2.3                     <pip>
mistune                   0.8.3                     <pip>
more-itertools            4.3.0                     <pip>
nbconvert                 5.4.0                     <pip>
nbformat                  4.4.0                     <pip>
ncurses                   6.1                  hfc679d8_1    conda-forge
notebook                  5.6.0                     <pip>
numba                     0.39.0                    <pip>
numpy                     1.15.1                    <pip>
openssl                   1.0.2p               h470a237_0    conda-forge
packaging                 17.1                      <pip>
pandas                    0.23.4                    <pip>
pandoc                    1.19.2                        0    conda-forge
pandocfilters             1.4.2                     <pip>
parso                     0.3.1                     <pip>
pexpect                   4.6.0                     <pip>
pickleshare               0.7.4                     <pip>
pip                       18.0                     py36_1    conda-forge
plotly                    3.2.0                     <pip>
pluggy                    0.7.1                     <pip>
poliastro                 0.11.dev0                 <pip>
prometheus-client         0.3.1                     <pip>
prompt-toolkit            1.0.15                    <pip>
ptyprocess                0.6.0                     <pip>
py                        1.6.0                     <pip>
pycodestyle               2.4.0                     <pip>
pycparser                 2.18                      <pip>
Pygments                  2.2.0                     <pip>
pyparsing                 2.2.0                     <pip>
pytest                    3.8.0                     <pip>
pytest-cov                2.6.0                     <pip>
python                    3.6.6                h5001a0f_0    conda-forge
python-dateutil           2.7.3                     <pip>
pytz                      2018.5                    <pip>
pyzmq                     17.1.2                    <pip>
readline                  7.0                  haf1bffa_1    conda-forge
requests                  2.19.1                    <pip>
retrying                  1.3.3                     <pip>
scipy                     1.1.0                     <pip>
SecretStorage             3.1.0                     <pip>
Send2Trash                1.5.0                     <pip>
setuptools                40.2.0                   py36_0    conda-forge
simplegeneric             0.8.1                     <pip>
six                       1.11.0                    <pip>
snowballstemmer           1.2.1                     <pip>
Sphinx                    1.7.9                     <pip>
sphinx-rtd-theme          0.4.1                     <pip>
sphinxcontrib-websupport  1.1.0                     <pip>
sqlite                    3.24.0               h2f33b56_1    conda-forge
terminado                 0.8.1                     <pip>
testpath                  0.3.1                     <pip>
tk                        8.6.8                         0    conda-forge
tornado                   5.1                       <pip>
traitlets                 4.3.2                     <pip>
urllib3                   1.23                      <pip>
wcwidth                   0.1.7                     <pip>
webencodings              0.5.1                     <pip>
wheel                     0.31.1                   py36_1    conda-forge
widgetsnbextension        3.4.1                     <pip>
xz                        5.2.4                h470a237_1    conda-forge
zlib                      1.2.11               h470a237_3    conda-forge

```
</details>

mpacer:
This fixes #878 and https://github.com/spatialaudio/nbsphinx/issues/202.

We had an API before that allowed passing an empty string as your kernel name, which it would then fall back on getting the kernel name from the notebook metadata.

This is now the actual default behaviour, and before this PR setting `kernel_name=""` would overwrite this default. This reintroduces the fall-back as well as a warning that this behaviour shouldn't be used (you can pass no value to `kernel_name` and you will get the same behaviour). 

To ensure the same behaviour in both cases, it introduces the `find_kernel_name` API, which is then used in both the default and fallback mechanisms. 

This also introduces 2 tests and a new notebook file to be tested. In order to ensure that the functionality works I'm specifying a python3 kernel, which cannot be guaranteed to exist, so we have an expected fail in the case where there is no python3 kernel.

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -4,6 +4,8 @@ and updates outputs"""
 # Copyright (c) IPython Development Team.
 # Distributed under the terms of the Modified BSD License.
 
+import warnings
+
 from textwrap import dedent
 from contextlib import contextmanager
 
@@ -161,7 +163,7 @@ class ExecutePreprocessor(Preprocessor):
     @default('kernel_name')
     def _kernel_name_default(self):
         try:
-            return self.nb.metadata.get('kernelspec', {}).get('name', 'python')
+            return find_kernel_name(self.nb)
         except AttributeError:
             raise AttributeError('You did not specify a kernel_name for '
                                  'the ExecutePreprocessor and you have not set '
@@ -248,6 +250,16 @@ class ExecutePreprocessor(Preprocessor):
         kc : KernelClient
             Kernel client as created by the kernel manager `km`.
         """
+        
+        # Because of an old API, an empty kernel string should be interpreted as indicating
+        # that you want to use the notebook's metadata specified kernel.
+        # We use find_kernel_name to do this.
+        if self.kernel_name == u"":
+            self.kernel_name = find_kernel_name(self.nb)
+            warnings.warn("Setting kernel_name to '' as a way to use the kernel in a notebook metadata's is deprecated as of 5.4. \n"
+                          "Instead, do not set kernel_name, this is now default behaviour.", DeprecationWarning, stacklevel=2)
+            
+            
         km = self.kernel_manager_class(kernel_name=self.kernel_name,
                                        config=self.config)
         km.start_kernel(extra_arguments=self.extra_arguments, **kwargs)
@@ -516,6 +528,20 @@ class ExecutePreprocessor(Preprocessor):
 
         return exec_reply, outs
 
+def find_kernel_name(nb):
+    """This is a utility that finds kernel names from notebook metadata.
+    
+    Parameters
+    ----------
+    nb : NotebookNode
+        The notebook that has a kernelspec defined in its metadata.
+
+    Returns
+    -------
+    kernel_name: str
+        The string representing the kernel name found in the metadata.
+    """
+    return nb.metadata.get('kernelspec', {}).get('name', 'python')
 
 def executenb(nb, cwd=None, km=None, **kwargs):
     """Execute a notebook's code, updating outputs within the notebook object.
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -254,13 +254,15 @@ class ExecutePreprocessor(Preprocessor):
         # Because of an old API, an empty kernel string should be interpreted as indicating
         # that you want to use the notebook's metadata specified kernel.
         # We use find_kernel_name to do this.
-        if self.kernel_name == u"":
-            self.kernel_name = find_kernel_name(self.nb)
+        if not self.kernel_name:
+            kernel_name = find_kernel_name(self.nb)
             warnings.warn("Setting kernel_name to '' as a way to use the kernel in a notebook metadata's is deprecated as of 5.4. \n"
                           "Instead, do not set kernel_name, this is now default behaviour.", DeprecationWarning, stacklevel=2)
+        else:
+            kernel_name = self.kernel_name
             
             
-        km = self.kernel_manager_class(kernel_name=self.kernel_name,
+        km = self.kernel_manager_class(kernel_name=kernel_name,
                                        config=self.config)
         km.start_kernel(extra_arguments=self.extra_arguments, **kwargs)
 
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -14,7 +14,7 @@ try:
 except ImportError:
     from Queue import Empty  # Py 2
 
-from traitlets import List, Unicode, Bool, Enum, Any, Type, Dict, Integer, default
+from traitlets import List, Unicode, Bool, Enum, Any, Type, Dict, Integer, default, validate
 
 from nbformat.v4 import output_from_msg
 
@@ -163,13 +163,33 @@ class ExecutePreprocessor(Preprocessor):
     @default('kernel_name')
     def _kernel_name_default(self):
         try:
-            return find_kernel_name(self.nb)
+            return self.nb.metadata.get('kernelspec', {}).get('name', 'python')
         except AttributeError:
             raise AttributeError('You did not specify a kernel_name for '
                                  'the ExecutePreprocessor and you have not set '
                                  'self.nb to be able to use that to infer the '
                                  'kernel_name.')
+        
 
+    @validate('kernel_name')
+    def _valid_kernel_name(self, proposal):
+        # import pdb;pdb.set_trace()
+        if proposal.value ==u"" and hasattr(self, 'nb'):
+            response = self.nb.metadata.get('kernelspec', {}).get('name', 'python') 
+            if response not in list('python', *self.kernelspec_manager_class(parent=self).find_kernel_specs()):
+                raise AttributeError("""
+                    You specified an empty kernel_name `''`. We inferred the kernel name `{}` 
+                    from your notebook metadata. The {} was unable to find 
+                    a kernel by that name.""".format(response, self.kernelspec_manager_class))
+        elif proposal.value==u"":
+            response = 'python'
+        else:
+            response = proposal.value
+            if response not in list('python', *self.kernelspec_manager_class(parent=self).find_kernel_specs()):
+                raise AttributeError('You specified the kernel_name `{}`. The {} was unable '
+                                     'to find a kernel by that name'.format(response, self.kernelspec_manager_class))
+        return response 
+        
     raise_on_iopub_timeout = Bool(False,
         help=dedent(
             """
@@ -219,6 +239,19 @@ class ExecutePreprocessor(Preprocessor):
         except ImportError:
             raise ImportError("`nbconvert --execute` requires the jupyter_client package: `pip install jupyter_client`")
         return KernelManager
+        
+    kernelspec_manager_class = Type(
+        config=True,
+        help='The kernelspec manager class to use.'
+    )
+    @default('kernelspec_manager_class')
+    def kernelspec_manager_class_default(self):
+        """Use a dynamic default to avoid importing jupyter_client at startup"""
+        try:
+            from jupyter_client.kernelspec import KernelSpecManager
+        except ImportError:
+            raise ImportError("`nbconvert --execute` requires the jupyter_client package: `pip install jupyter_client`")
+        return KernelSpecManager
 
     _display_id_map = Dict(
         help=dedent(
@@ -254,16 +287,9 @@ class ExecutePreprocessor(Preprocessor):
         # Because of an old API, an empty kernel string should be interpreted as indicating
         # that you want to use the notebook's metadata specified kernel.
         # We use find_kernel_name to do this.
-        if not self.kernel_name:
-            kernel_name = find_kernel_name(self.nb)
-            warnings.warn("Setting kernel_name to '' as a way to use the kernel in a notebook metadata's is deprecated as of 5.4. \n"
-                          "Instead, do not set kernel_name, this is now default behaviour.", DeprecationWarning, stacklevel=2)
-        else:
-            kernel_name = self.kernel_name
             
-            
-        km = self.kernel_manager_class(kernel_name=kernel_name,
-                                       config=self.config)
+        km = self.kernel_manager_class(kernel_name=self.kernel_name,
+                                       parent=self)
         km.start_kernel(extra_arguments=self.extra_arguments, **kwargs)
 
         kc = km.client()
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -556,20 +556,6 @@ class ExecutePreprocessor(Preprocessor):
 
         return exec_reply, outs
 
-def find_kernel_name(nb):
-    """This is a utility that finds kernel names from notebook metadata.
-    
-    Parameters
-    ----------
-    nb : NotebookNode
-        The notebook that has a kernelspec defined in its metadata.
-
-    Returns
-    -------
-    kernel_name: str
-        The string representing the kernel name found in the metadata.
-    """
-    return nb.metadata.get('kernelspec', {}).get('name', 'python')
 
 def executenb(nb, cwd=None, km=None, **kwargs):
     """Execute a notebook's code, updating outputs within the notebook object.
--- a/nbconvert/preprocessors/tests/files/UnicodePy3.ipynb
+++ b/nbconvert/preprocessors/tests/files/UnicodePy3.ipynb
@@ -1,0 +1,32 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "\u2603\n"
+     ]
+    }
+   ],
+   "source": [
+    "print('\u2603')"
+   ]
+  }
+ ],
+ "metadata": {
+  "kernelspec": {
+   "display_name": "Python 3",
+   "language": "python",
+   "name": "python3"
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 0
+}
--- a/nbconvert/preprocessors/tests/test_execute.py
+++ b/nbconvert/preprocessors/tests/test_execute.py
@@ -21,6 +21,8 @@ import pytest
 from .base import PreprocessorTestsBase
 from ..execute import ExecutePreprocessor, CellExecutionError, executenb, find_kernel_name
 
+from jupyter_client.kernelspec import KernelSpecManager
+
 from nbconvert.filters import strip_ansi
 from testpath import modified_env
 from ipython_genutils.py3compat import string_types
@@ -150,6 +152,19 @@ class TestExecute(PreprocessorTestsBase):
         res = self.build_resources()
         res['metadata']['path'] = ''
         input_nb, output_nb = self.run_notebook(filename, {}, res)
+        self.assert_notebooks_equal(input_nb, output_nb)
+    
+    @pytest.mark.xfail("python3" not in KernelSpecManager().find_kernel_specs(),
+                        reason="requires a python3 kernelspec")
+    def test_empty_kernel_name(self):
+        """Can kernel in nb metadata be found when an empty string is passed?
+        
+        Note: this pattern should be discouraged in practice.
+        Passing in no kernel_name to ExecutePreprocessor is preferable.
+        """
+        filename = os.path.join(current_dir, 'files', 'UnicodePy3.ipynb')
+        res = self.build_resources()
+        input_nb, output_nb = self.run_notebook(filename, {"kernel_name": ""}, res)
         self.assert_notebooks_equal(input_nb, output_nb)
 
     def test_disable_stdin(self):
--- a/nbconvert/preprocessors/tests/test_execute.py
+++ b/nbconvert/preprocessors/tests/test_execute.py
@@ -19,7 +19,7 @@ import sys
 import pytest
 
 from .base import PreprocessorTestsBase
-from ..execute import ExecutePreprocessor, CellExecutionError, executenb
+from ..execute import ExecutePreprocessor, CellExecutionError, executenb, find_kernel_name
 
 from nbconvert.filters import strip_ansi
 from testpath import modified_env
@@ -265,6 +265,15 @@ class TestExecute(PreprocessorTestsBase):
         for method, call_count in expected:
             self.assertNotEqual(call_count, 0, '{} was called'.format(method))
 
+    def test_find_kernel_name(self):
+        current_dir = os.path.dirname(__file__)
+        filename = os.path.join(current_dir, 'files', 'UnicodePy3.ipynb')
+
+        with io.open(filename) as f:
+            input_nb = nbformat.read(f, 4)
+        
+        assert find_kernel_name(input_nb) == "python3"
+        
     def test_execute_function(self):
         # Test the executenb() convenience API
         current_dir = os.path.dirname(__file__)
--- a/nbconvert/preprocessors/tests/test_execute.py
+++ b/nbconvert/preprocessors/tests/test_execute.py
@@ -19,7 +19,7 @@ import sys
 import pytest
 
 from .base import PreprocessorTestsBase
-from ..execute import ExecutePreprocessor, CellExecutionError, executenb, find_kernel_name
+from ..execute import ExecutePreprocessor, CellExecutionError, executenb
 
 from testpath import modified_env
 from traitlets import TraitError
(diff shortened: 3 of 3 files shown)


mpacer:
Note: https://github.com/spatialaudio/nbsphinx/pull/208 updates nbsphinx to use the preferred means of getting a kernel from notebook metadata.

mpacer:
@mgeier @akhmerov @Juanlu001 This should address your concerns & test to avoid further regressions.

nbconvert/preprocessors/execute.py
```diff
@@ -248,6 +250,16 @@ def start_new_kernel(self, **kwargs):
         kc : KernelClient
             Kernel client as created by the kernel manager `km`.
         """
+        
+        # Because of an old API, an empty kernel string should be interpreted as indicating
+        # that you want to use the notebook's metadata specified kernel.
+        # We use find_kernel_name to do this.
+        if self.kernel_name == u"":
```
MSeal: Would it be better to do `if not self.kernel_name:`? As a python developer I would expect that if "" is treated as falsy for a condition that None would be too. Or does the kernel_manager_class already handle None separately?

```
if not self.kernel_name:
  self.kernel_name = find_kernel_name(self.nb)
  if isinstance(self.kernel_name, string_types):
    warnings.warn(...)
```

mpacer: After looking at it, I'm pretty sure `None` would have worked previously… but I still don't want to encourage passing people to do this regardless. So we'll want to warn no matter what type the value was.

MSeal: So would checking for false kernel_name and declaring it deprecated in all cases make more sense from a function perspective then? I'm ok with merging if you feel strongly to just check for unicode string, it just had some code smell for something that might fail under similar situations -- like python version differences

mgeier:
This is all really confusing!

I don't see the necessity for dynamically changing the default value for `kernel_name`.

Why don't you just completely get rid of `_kernel_name_default()`?
The helper function `find_kernel_name()` isn't (currently) necessary either.

Then, in `start_new_kernel()` you could simply do something like this (untested!):

```python
        if self.kernel_name:
            kernel_name = self.kernel_name
        else:
            kernel_name = self.nb.metadata.get('kernelspec', {}).get('name', 'python')
```

MSeal:
@mgeier I think I follow the suggestion. I'm still learning the nbconvert config system, so I don't know if there's side-effects that are important in assigning the default in the config and raising there for the AttributeError. But if there's not a secondary reason for this pattern, and nothing else would reuse this config then it would probably be simpler to move all the checks into `start_new_kernel`.

minrk:
I would recommend one of two approaches:

- remove dynamic default for kernel_name, so that the `if not self.kernel_name` logic in this PR is the behavior for both the default
- go "full traitlets" and add, in addition to the dynamic default, a `@validate('kernel_name')` that coerces `kernel_name = ''` into `find_kernel_name()`. That way `self.kernel_name` is always defined as the correct kernel name, whether config specified it as `empty==auto` or didn't specify it at all.

I don't have a strong preference, but the advantage of the first is that it relies less on traitlets magic, while the advantage of the second uses traitlets to ensure that `self.kernel_name` is always available as the kernel name being used. An argument for case 1 is if it makes sense to re-use a Preprocessor instance across multiple exports, since the dynamic default will be constructed only on the first notebook it sees, not on each call. I don't recall if this use case makes sense, though.

mgeier:
Using `@validate('kernel_name')` sounds slightly more reasonable than using `@default('kernel_name')`. But still, it doesn't really make sense. The notebook is passed to the `ExecutePreprocessor` in the method `preprocess(self, nb, resources, km=None)`. Validating the `kernel_name` before this method is called doesn't make sense, because at that point the notebook isn't known yet.

It makes much more sense to keep `kernel_name=''` and then validate it within the `preprocess()` method.

Or am I missing something?

MSeal:
FYI telling someone their design doesn't make sense comes off a bit aggressive, even when the suggested changes are valid suggestions.

I would argue passing an empty string as a signal/default coercion behavior is not usually a good practice. `None` or absence from kwargs entirely would be more Pythonic, but for compatibility to capture `u""` falsey is a reasonable fallback. Having a validate that checks that you passed something we can cooerce into a kernel, or is a valid kernel name as @mpacer described sounds reasonable to me. Having a common function to convert placeholders (or empty strings) to actual kernel names JIT or as a function wrapper also seems the most robust and reusable since there is more than one place that would likely want this behavior.

Given this is just a backwards compatibility item, I wouldn't stress _too_ much on exact details as we I believe we want to drop explicit `u""` matching in the future anyway.

mgeier:
> I would argue passing an empty string as a signal/default coercion behavior is not usually a good practice.

There are certainly cases where it is not a good practice, but this case is not one of them.

There is no reasonable interpretation of different intent whether a user passes an empty string or nothing at all for `kernel_name`. Those two cases are for all intents and purposes equal.

Sure, the original author could have chosen `None` as default value, that would have been fine, too.

> Having a common function to convert placeholders (or empty strings) to actual kernel names JIT or as a function wrapper also seems the most robust and reusable since there is more than one place that would likely want this behavior.

That would be totally fine and probably useful, as long as this function takes the notebook in question as an argument or if this function is a method of an object that represents a notebook.

> I wouldn't stress too much on exact details as we I believe we want to drop explicit `u""` matching in the future anyway.

Please don't.

> FYI telling someone their design doesn't make sense comes off a bit aggressive, even when the suggested changes are valid suggestions.

I'm sorry if I sounded aggressive, I didn't intend to.

I'm not a native speaker, I didn't know of another way to say it and I didn't know that somebody could interpret this as aggressive.
I just wanted to state the fact that the discussed change would be logically inconsistent and simply bad software engineering, I thought colloquially using "doesn't make sense" would transport this sentiment quite succinctly.

I've looked a bit more into this and found that the logical error was committed much earlier than this pull request or the commit it tries to fix.
The real problem is that the notebook should not be assigned to `self.nb`.
The `ExecutePreprocessor` should only hold state based on its initial configuration. It shouldn't hold state about the currently processed notebook nor about the currently running kernel.
The current notebook should only be known (and the kernel should only be running) during the runtime of  the `preprocess()` method. If some runtime information is needed in other methods, it should be passed there via function arguments.

It is also quite confusing that there is a method named `preprocess_cell()` and a very similarly named method `run_cell()` as well.

The method `preprocess_cell()` is normally (in the other preprocessors) run repeatedly from within the method `preprocess()`. So here it should either be removed (or deactivated somehow since it would also be inherited from the base class) or it should be combined with `run_cell()`.

`preprocess_cell()` has a `resources` argument, so the obvious place to put the runtime state (running kernel and reference to the notebook if needed) would be this `resources` argument.
But it's probably simpler to disable `preprocess_cell()` and pass all the necessary information to `run_cell()` as arguments.

Once this problem is resolved, our discussion here will become moot. But in the meantime, can we please keep `kernel_name=''` as a default that can also be passed by a user?

mgeier:
Update: I just realized that `preprocess_cell()` is still called by the base class, as it is supposed to be.

In this case, all necessary state about the current notebook and the currently running kernel should be transported in the `resources` argument.

akhmerov:
> Once this problem is resolved, our discussion here will become moot. But in the meantime, can we please keep `kernel_name=''` as a default that can also be passed by a user?

What is your motivation for this preference? `None` is a much more common none-value. While I imagine that there wouldn't be kernels that have an empty string for a name, having `None` for a default is a more standard interface.

mgeier:
@akhmerov 

> What is your motivation for this preference? None is a much more common none-value.

Sorry I wasn't clear enough.

In general I don't have a strong preference. As I said, I think both `''` and `None` would be fine as default values in this case.

The reason why I mentioned `kernel_name=''` is for backwards compatibility. The default was an empty string before and I don't see a reason to change this.

If the default were changed to `None` (and the empty string were disallowed), this would break my `nbsphinx` module (again).

mgeier:
@akhmerov No, using `kernel_name=None` doesn't currently work, as @mpacer said in https://github.com/jupyter/nbconvert/pull/886#discussion_r216873529. Trying to pass `None` leads to

```
traitlets.traitlets.TraitError: The 'kernel_name' trait of an ExecutePreprocessor instance must be a unicode string, but a value of None <class 'NoneType'> was specified.
```

MSeal:
I do agree with @akhmerov that None is the better pattern here and that from a python developer perspective `""` is not expected to have special behavior independent of None. But given history and all the discussions here, treating anything pythonic falsey solves the issue either way and is a reasonable interface choice.

Let's get M's change finalized to respect the pattern (even if it's not perfect for everyone) and redirect some of this energy into improving Preprocessor design overall. I think that's a more valuable use of time.

mgeier:
It looks like #809 would also solve the issue that this PR is solving. However, #809 has milestone `6.0`, and it would be great if we could solve this issue (namely #878) in release `5.5`!

Can someone please tag this with milestone `5.5`?

mgeier:
Since there doesn't seem to be any progress here, I've created an alternative (simpler) PR here: #924.

I'm fine with either solution, as long as the problem is fixed before the next `nbconvert` release.

MSeal:
Closing as request has been satisfied by related PRs

PR #889, opened 2018-09-23
Expose pygments styles

### LINKED ISSUE #832: [feature request] Add option/settings to change the pygments style

t-makaro:
closes #832 

This will let html, and latex exporting use any pygments style, which means one can set the colors that are used to highlight code!

This is my reattempt at #836, because I messed up the commit history.

- [x] Expose Pygments Styles
- [x] Write docs explaining Pygments Styles
- [x] Write LaTeX test
- [ ] Write HTML test

### DIFF
--- a/docs/source/highlighting.rst
+++ b/docs/source/highlighting.rst
@@ -1,0 +1,62 @@
+Customizing Syntax Highlighting
+===============================
+
+Under the hood, nbconvert uses pygments to highlight code. Both pdf and html exporting support 
+changing the highlighting style.
+
+Using Builtin styles
+--------------------
+Pygments has a number of builtin styles available. To use them, we just need to set the style setting 
+in the relevant preprocessor.
+
+To change the html highlighting export with:
+
+.. code-block:: bash
+
+    jupyter nbconvert --to html notebook.ipynb --CSSHTMLHeaderPreprocessor.style=<name>
+
+To change pdf and latex highlighting export with:
+
+.. code-block:: bash
+
+    jupyter nbconvert --to pdf notebook.ipynb --LatexPreprocessor.style=<name>
+
+where ``<name>`` is the name of the pygments style. Available styles may vary from system to system.
+You can find all available styles with:
+
+.. code-block:: bash
+
+    pygmentize -L styles
+
+from a terminal or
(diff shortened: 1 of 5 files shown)


t-makaro:
@mpacer I added some docs explaining pygments styles a bit. I also added a test for the latex preprocessor. I was going to build a similar test for the html preprocessor, but the header is built a little differently.

For the latex preprocessor, I'd like to add a resource that allows passing of the name of the style to the template. This would be useful since I have a template that uses latex's minted to highlight the code instead. Should I add a similar resource to the html preprocessor as well?

MSeal:
@t-makaro That sounds like a reasonable and useful change. I'm not super familiar with that section of the code but I'd be willing to review and get it in if you make the change.

MSeal:
Fair enough on testing -- change looks good from my perspective.

PR #892, opened 2018-10-02
LaTeX: Better placement of non-text code cell output

mgeier:
I did the same thing for `nbsphinx` in https://github.com/spatialaudio/nbsphinx/pull/212/commits/fc4fe32aa2931d5f40e2fa1efa99d090ab215129, so I thought I'll just apply the same trick here.

Before:

grafik

After:

grafik

Obviously, the vertical distance to the next paragraph is still too short, but I didn't know what's the best way to fix this. It's not obvious to me, any ideas?

### DIFF
--- a/nbconvert/templates/latex/document_contents.tplx
+++ b/nbconvert/templates/latex/document_contents.tplx
@@ -58,7 +58,7 @@
 % Redirect execute_result to display data priority.
 ((* block execute_result scoped *))
     ((* block data_priority scoped *))
-    ((( super() )))
+    \noindent ((( super() )))
     ((* endblock *))
 ((* endblock execute_result *))
 
--- a/nbconvert/templates/latex/style_ipython.tplx
+++ b/nbconvert/templates/latex/style_ipython.tplx
@@ -38,7 +38,7 @@
             ((*- if type in ['text/plain'] *))
 ((( add_prompt(output.data['text/plain'] | escape_latex, cell, 'Out', 'outcolor') )))
             ((* else -*))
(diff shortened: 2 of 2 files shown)


mgeier:
OK, I tried to improve the distance a bit (1f617c38c53823320efec415a211440458ccedbd), but the spacing is still off.

IMHO there is too much space between code cell inputs and outputs and too little space after the outputs.

grafik

mgeier:
BTW, I forgot to mention that I made above screenshots with https://github.com/ipython/ipython/pull/11357 in place. Without that, the math stuff would be centered.

mgeier:
Just a little update: Please note that my PR here adds a `minipage` around the output, therefore making page breaks within the output impossible. That's not good!

In the meantime I've come up with a way to change this in `nbsphinx` (https://github.com/spatialaudio/nbsphinx/pull/218), but this uses some Sphinx machinery, so sadly this isn't applicable here.

Hopefully somebody else comes up with a proper solution for `nbconvert`.

t-makaro:
This could probably be redone within the new #992 template. Right now, text outputs are placed within a breakable tcolorbox. Image outputs could be placed within an unbreakable tcolorbox which would fix the issue with prompts appearing on a different page then the image. What outputs, other than text, need to be in a breakable box?

mgeier:
@t-makaro 

> What outputs, other than text, need to be in a breakable box?

I guess there is plain text, Markdown and "stream" text (potentially with ANSI escape sequences).

Anyway, after the recent LaTeX improvements, this PR is obsolete and can be closed.

PR #900, opened 2018-10-16
Embed widget state in notebook on execute

maartenbreddels:
Rebase of #779 which is a followup of #760

PR #905, opened 2018-10-28
Refactored execute preprocessor to have a process_message function

MSeal:
Built ontop of https://github.com/jupyter/nbconvert/pull/982 which added additional requested testing. We can merge that PR first if it's desired.

Broke apart the large run_cell function to enable intercepting messages easier. This will enable deleting code from papermill where we have to reimplement the entire `run_cell` method.

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -588,12 +588,12 @@ class ExecutePreprocessor(Preprocessor):
         except ValueError:
             self.log.error("unhandled iopub msg: " + msg_type)
             return
-        else:
-            if self.clear_before_next_output:
-                self.log.debug('Executing delayed clear_output')
-                outs[:] = []
-                self.clear_display_id_mapping(cell_index)
-                self.clear_before_next_output = False
+
+        if self.clear_before_next_output:
+            self.log.debug('Executing delayed clear_output')
+            outs[:] = []
+            self.clear_display_id_mapping(cell_index)
+            self.clear_before_next_output = False
 
         if display_id:
             # record output index in:
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -409,7 +409,9 @@ class ExecutePreprocessor(Preprocessor):
         if cell.cell_type != 'code' or not cell.source.strip():
             return cell, resources
 
-        reply = self.run_cell(cell, cell_index)
+        reply, outputs = self.run_cell(cell, cell_index)
+        # Backwards compatability for processes that wrap run_cell
+        cell.outputs = outputs
 
         cell_allows_errors = (self.allow_errors or "raises-exception"
                               in cell.metadata.get("tags", []))
@@ -523,7 +525,8 @@ class ExecutePreprocessor(Preprocessor):
             except CellExecutionComplete:
                 break
 
-        return exec_reply
+        # Return cell.outputs still for backwards compatability
+        return exec_reply, cell.outputs
 
     def process_message(self, msg, cell, cell_index):
         """
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -25,6 +25,8 @@ from .base import Preprocessor
 from ..utils.exceptions import ConversionException
 
 
+class CellExecutionComplete(Exception): pass  # Used as a control signal
+
 class CellExecutionError(ConversionException):
     """
     Custom exception to propagate exceptions that are raised during
@@ -400,14 +402,13 @@ class ExecutePreprocessor(Preprocessor):
         if cell.cell_type != 'code' or not cell.source.strip():
             return cell, resources
 
-        reply, outputs = self.run_cell(cell, cell_index)
-        cell.outputs = outputs
+        reply = self.run_cell(cell, cell_index)
 
         cell_allows_errors = (self.allow_errors or "raises-exception"
                               in cell.metadata.get("tags", []))
 
         if self.force_raise_errors or not cell_allows_errors:
-            for out in outputs:
+            for out in cell.outputs:
                 if out.output_type == 'error':
                     raise CellExecutionError.from_cell_and_msg(cell, out)
             if (reply is not None) and reply['content']['status'] == 'error':
@@ -509,22 +510,45 @@ class ExecutePreprocessor(Preprocessor):
                 # not an output from our execution
                 continue
 
-            if not self.process_message(msg, cell, cell_index):
+            # Will raise CellExecutionComplete when completed
+            try:
+                self.process_message(msg, cell, cell_index)
+            except CellExecutionComplete:
                 break
 
-        return exec_reply, cell.outputs
+        return exec_reply
 
     def process_message(self, msg, cell, cell_index):
-        '''
-        Returns None if execution should be halted.
-        '''
+        """
+        Processes a kernel message, updates cell state, and returns the
+        resulting output object that was appended to cell.outputs.
+
+        The input argument `cell` is modified in-place.
+
+        Parameters
+        ----------
+        msg : dict
+            The kernel message being processed.
+        cell : nbformat.NotebookNode
+            The cell which is currently being processed.
+        cell_index : int
+            The position of the cell within the notebook object.
+
+        Returns
+        -------
+        output : dict
+            The execution output payload (or None for no output).
+
+        Raises
+        ------
+        CellExecutionComplete
+          Once a message arrives which indicates computation completeness.
+
+        """
         msg_type = msg['msg_type']
         self.log.debug("msg_type: %s", msg_type)
         content = msg['content']
         self.log.debug("content: %s", content)
-
-        # Default to our input as the "result" of processing the message
-        result = msg
 
         display_id = content.get('transient', {}).get('display_id', None)
         if display_id and msg_type in {'execute_result', 'display_data', 'update_display_data'}:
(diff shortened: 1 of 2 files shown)


MSeal:
Hmmm seeings this test fail against master so I don't think it's related to the PR

maartenbreddels:
I see the same in #900

mpacer:
@Carreau I think you mentioned something about changing the error formatting in IPython a while back — shall I assume that the errors that are cropping up are related to a recent release of IPython?

nbconvert/preprocessors/execute.py
```diff
@@ -466,55 +466,59 @@ def run_cell(self, cell, cell_index=0):
...
+            else:
+                if display_id:
+                    # record output index in:
+                    #   _display_id_map[display_id][cell_idx]
+                    cell_map = self._display_id_map.setdefault(display_id, {})
+                    output_idx_list = cell_map.setdefault(cell_index, [])
+                    output_idx_list.append(len(cell.outputs))
 
-            outs.append(out)
+                cell.outputs.append(result)
```
mpacer: I don't really like the idea of having an API that works primarily via side-effects — that's a difficult API to document and test for. 

Can you think of a way that we could do this as a pure function? E.g., could we adhere to the pattern for we use with preprocessors? I think that would involve returning and reassigning the modified cell object rather than returning the result itself while treating the cell.outputs as a location where we can append changes?

nbconvert/preprocessors/execute.py
```diff
@@ -466,55 +466,59 @@ def run_cell(self, cell, cell_index=0):
...
+                # Set result to None to halt execution
+                result = None
+        elif msg_type == 'clear_output':
+            cell.outputs[:] = []
+            # clear display_id mapping for this cell
+            for display_id, cell_map in self._display_id_map.items():
+                if cell_index in cell_map:
+                    cell_map[cell_index] = []
+        # Check for remaining messages we don't process
+        elif not (msg_type in ['execute_input', 'update_display_data'] or msg_type.startswith('comm')):
```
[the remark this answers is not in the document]
MSeal: It's actually identical logically, by not returning it will continue here

nbconvert/preprocessors/execute.py
MSeal: I can take a look. Mostly I tried to keep the logic the same and not change the design, just break it into smaller parts.

nbconvert/preprocessors/execute.py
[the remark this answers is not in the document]
MSeal: Might have missed the not?

MSeal:
@mpacer I added a new test. From reading the code I think we'd need to do a semi-significant update to add in the message handler we talked about. Because of that I'd prefer to split the implementation into two parts and do that refactor post-merge of this PR.

nbconvert/preprocessors/execute.py
```diff
@@ -466,55 +466,59 @@ def run_cell(self, cell, cell_index=0):
...
+        self.log.debug("msg_type: %s", msg_type)
+        content = msg['content']
+        self.log.debug("content: %s", content)
+
+        # Default to our input as the "result" of processing the message
+        result = msg
+
+        display_id = content.get('transient', {}).get('display_id', None)
+        if display_id and msg_type in {'execute_result', 'display_data', 'update_display_data'}:
+            self._update_display_id(display_id, msg)
```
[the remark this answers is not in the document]
mpacer: @mseal could you respond to this comment?

nbconvert/preprocessors/execute.py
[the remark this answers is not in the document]
mpacer: Additionally now if something is an `update_display_data` message, it will return something (specifically the content of the msg). 

I'm starting to wonder if this all might be fixed if we default to returning `None`…

nbconvert/preprocessors/tests/test_execute.py
```diff
@@ -277,6 +277,30 @@ def test_custom_kernel_manager(self):
...
+
+        current_dir = os.path.dirname(__file__)
+        filename = os.path.join(current_dir, 'files', 'HelloWorld.ipynb')
+
+        with io.open(filename) as f:
+            input_nb = nbformat.read(f, 4)
+
+        original = copy.deepcopy(input_nb)
+        wpp = WrappedPreProc()
+        executed = wpp.preprocess(input_nb, {})[0]
```
mpacer: I would prefer it if this processed more kinds of messages since my main concerns have to do with maintaining compatibility with the current logic across all of the different message types (including `comm`s,  `update_display_data`, etc.).

nbconvert/preprocessors/tests/test_execute.py
```diff
@@ -277,6 +277,30 @@ def test_custom_kernel_manager(self):
...
+        self.assertEqual(outputs, [
```
mpacer: Elsewhere in the library we tend to use direct asserts as we are using pytest to run the tests. In addition to stylistic consistency, I think it's a lot easier to read.

nbconvert/preprocessors/tests/test_execute.py
```diff
@@ -277,6 +277,30 @@ def test_custom_kernel_manager(self):
...
+            {'name': 'stdout', 'output_type': 'stream', 'text': 'Hello World\n'}
+        ])
+        self.assert_notebooks_equal(original, executed)
```
mpacer: I **love** that we can test individual outputs now. I think I would prefer if we could make it so that we only test the message handling here and not the entire preprocess chain, it will make it clearer what the purpose of the test is and will make it easier to tell that we're covering all the cases we care about.

nbconvert/preprocessors/tests/test_execute.py
```diff
@@ -277,6 +277,30 @@ def test_custom_kernel_manager(self):
         for method, call_count in expected:
             self.assertNotEqual(call_count, 0, '{} was called'.format(method))
 
+    def test_process_message_wrapper(self):
+        outputs = []
+
+        class WrappedPreProc(ExecutePreprocessor):
```
mpacer: I would prefer if we directly created the context needed so that `process_message` can be tested functionally, iterating over the cell, msg and cell_index as directly as possible.

I'd prefer if we weren't concatenating all of the outputs into a single object for the whole notebook, I'm thinking it should follow the cell structure.

I'd also like to make sure that message types that are supposed to return None are returning None, which means keeping track of at least the msg type in relation to the output if not some of the more predictive fields (e.g., a display_id that we know is set to a particular value).

nbconvert/preprocessors/execute.py
```diff
@@ -466,55 +466,59 @@ def run_cell(self, cell, cell_index=0):
...
+            cell.outputs[:] = []
+            # clear display_id mapping for this cell
+            for display_id, cell_map in self._display_id_map.items():
+                if cell_index in cell_map:
+                    cell_map[cell_index] = []
+        # Check for remaining messages we don't process
+        elif not (msg_type in ['execute_input', 'update_display_data'] or msg_type.startswith('comm')):
             try:
-                out = output_from_msg(msg)
+                # Assign output as our processed "result"
```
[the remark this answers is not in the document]
mpacer: The problem is that display_id should have been dealt with in the `cell_map` logic _regardless of the message_ type previously, except in the cases where a continue was followed. That `try` is not going to be hit by all message types and so the `else` also won't be hit by all message types.

nbconvert/preprocessors/tests/test_execute.py
MSeal: Noted, will fix

nbconvert/preprocessors/execute.py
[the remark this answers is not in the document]
MSeal: Late response (post rebase), but yes it's implicit by not raising CellExecutionComplete here

nbconvert/preprocessors/tests/test_execute.py
MSeal: Done -- this is rebased off https://github.com/jupyter/nbconvert/pull/982 now

nbconvert/preprocessors/tests/test_execute.py
MSeal: Done

nbconvert/preprocessors/execute.py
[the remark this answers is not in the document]
MSeal: Was working fine post-rebase, but I changed this to a raise control flow pattern which I think is a lot easier to read.

MSeal:
@mpacer PR should now be ready for re-review

nbconvert/preprocessors/execute.py
```diff
@@ -25,6 +25,8 @@
 from ..utils.exceptions import ConversionException
 
 
+class CellExecutionComplete(Exception): pass  # Used as a control signal
```
mpacer: can you make this an actual docstring?

nbconvert/preprocessors/tests/test_execute.py
```diff
@@ -185,18 +185,18 @@ def normalize_output(output):
     def assert_notebooks_equal(self, expected, actual):
         expected_cells = expected['cells']
         actual_cells = actual['cells']
-        self.assertEqual(len(expected_cells), len(actual_cells))
+        assert len(expected_cells) == len(actual_cells)
```
mpacer: Huzzah! Many thanks for switching to using the pytest `assert` idiom.

nbconvert/preprocessors/tests/test_execute.py
```diff
@@ -395,6 +395,30 @@ def test_custom_kernel_manager(self):
...
             self.assertNotEqual(call_count, 0, '{} was called'.format(method))
 
+    def test_process_message_wrapper(self):
+        outputs = []
+
+        class WrappedPreProc(ExecutePreprocessor):
+            def process_message(self, msg, cell, cell_index):
+                result = super(WrappedPreProc, self).process_message(msg, cell, cell_index)
+                if result:
+                    outputs.append(result)
```
mpacer: why does this still need special logic distinct from the standard `process_message`?

MSeal: It's purely to test that one could wrap an retrieve outputs to process on it's own. It has no real purpose for doing so beyond showing how to wrap the function in inherited classes. Namely to prove we can wrap the function safely in papermill cleanly.

mpacer:
This looks great! Thanks for working through all the issues!

PR #910, opened 2018-11-06
do not pipe non-json data through json.dumps

### LINKED ISSUE #904: Use of SVG generates TypeError: Object of type 'bytes' is not JSON serializable
When using SVG plots
```python
%config InlineBackend.figure_format = 'svg'
```

the conversion to latex generate the error:

`TypeError: Object of type 'bytes' is not JSON serializable`

If I use PDF or PNG

```python
%config InlineBackend.figure_format = 'pdf'
```

or

```python
%config InlineBackend.figure_format = 'png'
```

No error is detected. I understand that this is an issue associated with the svg to json conversion. The log is:

```
jupyter-nbconvert-3.6 --to pdf notebook.ipynb

[NbConvertApp] Converting notebook ProbeArmResults.ipynb to pdf
  File "/opt/local/bin/jupyter-nbconvert-3.6", line 11, in <module>
    load_entry_point('nbconvert==5.4.0', 'console_scripts', 'jupyter-nbconvert')()
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/jupyter_core/application.py", line 266, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/traitlets/config/application.py", line 658, in launch_instance
    app.start()
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/nbconvert/nbconvertapp.py", line 337, in start
    self.convert_notebooks()
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/nbconvert/nbconvertapp.py", line 507, in convert_notebooks
    self.convert_single_notebook(notebook_filename)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/nbconvert/nbconvertapp.py", line 478, in convert_single_notebook
    output, resources = self.export_single_notebook(notebook_filename, resources, input_buffer=input_buffer)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/nbconvert/nbconvertapp.py", line 407, in export_single_notebook
    output, resources = self.exporter.from_filename(notebook_filename, resources=resources)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/nbconvert/exporters/exporter.py", line 178, in from_filename
    return self.from_file(f, resources=resources, **kw)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/nbconvert/exporters/exporter.py", line 196, in from_file
    return self.from_notebook_node(nbformat.read(file_stream, as_version=4), resources=resources, **kw)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/nbconvert/exporters/pdf.py", line 158, in from_notebook_node
    nb, resources=resources, **kw
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/nbconvert/exporters/latex.py", line 83, in from_notebook_node
    return super(LatexExporter, self).from_notebook_node(nb, resources, **kw)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/nbconvert/exporters/templateexporter.py", line 300, in from_notebook_node
    nb_copy, resources = super(TemplateExporter, self).from_notebook_node(nb, resources, **kw)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/nbconvert/exporters/exporter.py", line 138, in from_notebook_node
    nb_copy, resources = self._preprocess(nb_copy, resources)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/nbconvert/exporters/exporter.py", line 315, in _preprocess
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/nbconvert/preprocessors/base.py", line 47, in __call__
    return self.preprocess(nb, resources)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/nbconvert/preprocessors/base.py", line 69, in preprocess
    nb.cells[index], resources = self.preprocess_cell(cell, resources, index)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/nbconvert/preprocessors/extractoutput.py", line 92, in preprocess_cell
    data = json.dumps(data)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/encoder.py", line 180, in default
    o.__class__.__name__)
TypeError: Object of type 'bytes' is not JSON serializable
```

sgsaenger:
fixes #904 and possible similar errors by treating json data explicitely

### DIFF
--- a/nbconvert/preprocessors/extractoutput.py
+++ b/nbconvert/preprocessors/extractoutput.py
@@ -43,7 +43,8 @@ class ExtractOutputPreprocessor(Preprocessor):
     ).tag(config=True)
 
     extract_output_types = Set(
-        {'image/png', 'image/jpeg', 'image/svg+xml', 'application/pdf'}
+        {'image/png', 'image/jpeg', 'image/svg+xml',
+         'application/pdf', 'application/json'}
     ).tag(config=True)
 
     def preprocess_cell(self, cell, resources, cell_index):
@@ -80,10 +81,7 @@ class ExtractOutputPreprocessor(Preprocessor):
                 if mime_type in out.data:
                     data = out.data[mime_type]
 
-                    if (
-                        not isinstance(data, text_type)
-                        or mime_type == 'application/json'
-                    ):
+                    if mime_type == 'application/json':
                         # Data is either JSON-like and was parsed into a Python
                         # object according to the spec, or data is for sure
                         # JSON. In the latter case we want to go extra sure that


mgeier:
I'm not sure if this makes sense here, but instead of `mime_type == 'application/json'` you might want to consider using this regular expression:

    ^application/(.*\+)?json$

https://github.com/jupyter/nbformat/blob/5c18151bab30c4b6c81da279812bf1434ec54271/nbformat/v4/nbformat.v4.schema.json#L407

sgsaenger:
I don't even know how i would end up with plain json-data in my notebooks, therefore I'd rather not add any complexity I can't judge myself.

mgeier:
AFAICT, Jupyter widgets use JSON to store their data in the notebook file. They use MIME types like `application/vnd.jupyter.widget-view+json`, which are matched by the regular expression mentioned above.

An example notebook: https://github.com/jupyter-widgets/ipywidgets/blob/master/docs/source/examples/Widget%20List.ipynb

I don't know if those output data should be ignored by the `ExtractOutputPreprocessor`, though.

If not, they should be added `extract_output_types`, too.

sgsaenger:
At least in this example, I'd wager it should explicitly _not_ be extracted. (What can the user want with the model_id that seems to be for communication with the widget-handler?)
But my intention was just fixing #904, maybe @akhmerov and @minrk who handled the PR #847 that introduced the json-handling can have a look at this?

akhmerov:
I'm quite confused why a byte string ended up in the notebook data to be honest. I thought [according to the nbformat](https://nbformat.readthedocs.io/en/latest/format_description.html#display-data) everything is a JSON data type, so either a string or another JSON. In that case the fix would be to ensure that the data is read as a string instead.

The problem with the spec is that it's a bit ambiguous: JSON output *may* consist of a single string. That's why the logic I applied in determining the output type is to check whether it's not a string, and therefore must be JSON, or a string, in which case it *may* be a JSON and we have to check the mime type.

If for some reason we do expect that byte arrays may appear in cell outputs, I believe a better fix would be to include those in the check `if isinstance(data, (text_type, bytes)): ...`.

In addition to ipywidgets storing their outputs as JSON, so does plotly (its mime type is application/vnd.plotly.v1+json) and `ipython.display.JSON` (jupyterlab even has a nice rendering for it).

sgsaenger:
I guess you're right about the specs. The file itself even says that it expects images to be enclosed in a string: https://github.com/jupyter/nbconvert/blob/061543f3599678acf07e40b7bbbbe3a997f818db/nbconvert/preprocessors/extractoutput.py#L94

Still, i'm not sold on your solution.
I don't know about the problem you tried to solve, but now you're piping exclusively non-JSON data through ´json.dumps´, which is unnecessary and imposes limits that are nowhere else agreed upon (at least until now).
If these limits are needed than let's change the contract and make sure that no non-string data is here,
but even then, comparing the mime_type should suffice here. This function is based upon trusting the mime_type, isn't it?

akhmerov:
> Note that you omitted including json in the list of accepted mime types).

The list of mimetypes accepted by default is a separate issue from extracting the output correctly, and therefore I did not touch it. That traitlet isn't a complete list of mime types that should ever be treated by the extract output preprocessor, but merely what the extract output treats by default. Several preprocessors override that list. We therefore may not assume that we are never going to encounter other mimetypes. Specifically your PR would break nbconvert for those who want to extract widget information as pointed by @mgeier, or a plotly plot, which is also JSON.

> We know the mime type before we reach this part of the code. By the contract of this function, we trust the mime type (else we wouldn't be here. 

That is certainly true, but this bit of code is not about the *meaning* of the data (represented as mime type), but rather about the *storage format* of the data. Unfortunately right now the storage format is not explicitly specified in the metadata (see also #858). Roughly speaking, there are 3 options currently:

* any JSON-like data is stored as JSON without double serialization
* any other plaintext-ish mime type is stored as a string
* anything else is base64 encoded

The second and third option would become strings at this point in the pipeline, therefore the logic I am applying is the following;

* If we encounter any python object other than a string (something like a list/dict), we're dealing with JSON-encoded data (that's the reason for what you called `isinstance`-dance)
* Otherwise if we are either dealing with either a single JSON string or the other two options. Storing a single JSON string is rather uncommon, but allowed by the spec, and therefore only in that case I do rely on the explicit mimetype.

---

That's my understanding of what is happening in the preprocessor, largely based on discussions with @minrk.

What I really do not understand is: why do we ever encounter a byte string! This seems like a bug: JSON doesn't have a way to specify the type of string data, svg may contain arbitrary unicode symbols, and nothing in the spec tells about byte strings. Instead we're base64 encoding anything that isn't unicode.

sgsaenger:
> That traitlet isn't a complete list of mime types that should ever be treated by the extract output preprocessor, but merely what the extract output treats by default. Several preprocessors override that list.

Ah, totally didn't catch that usage, but fair point.

> If we encounter any python object other than a string (something like a list/dict), we're dealing with JSON-encoded data (that's the reason for what you called `isinstance`-dance)

Ok, I now understand your assumptions. But still, is this the right place to (implicitly) introduce strong typing? I have two issues with this:
- We're implicitly introducing strong type-dependencies on data we don't care about. If it is important that we more closely match some specs, shouldn't this be done explicitly, and possibly elsewhere?
- Isn't there a way to do the json-conversion while being both more explicit in our intent and/or being more robust?
We could e.g. check for `json in mime_type` to match the widget information, or wrap the dump in `try` and ignore possible exceptions when the data is not wanted.

akhmerov:
Some remarks:

* The best way to avoid guessing would be #858.
* For now if we let something that doesn't look like a string go further down the pipeline, the follow-up code will break. This would happen with this PR and a mime-type that we don't explicitly catch. I don't like strong typing just likewise, but here I don't see a good reason to avoid it. (Obviously there's something subtle happening with `bytes`, so I'm not claiming that the current code is correct.)
* I don't know enough about mimetypes combined with how jupyter decides which data storage format to apply to a datatype. I am therefore not certain that a check `'json' in mimetype` is safe. Maybe some json-related mimetypes would somehow be still string-serialized and we would end up double-serializing them? Maybe some mimetypes that don't contain JSON in their name are stored as a JSON data structure?

MSeal:
Related to this topic, I ended up tackling the `application.json` vs other data type paths in https://github.com/jupyter/nbconvert/pull/985. It doesn't expand the scope of data transformation but fixed a bug where latex exports ended up in the data.decode path.

PR #919, opened 2018-11-22
exclude header

ocefpaf:
I'm not sure if I'm altering the right templates here but the goal is to add an option to exclude the header.

The idea is that not all Python scripts need the shebang, added in the header recently, and sometimes one may want to get parts of the latex document to concatenate with another header.

I understand that one can do the same with a custom template but sometimes it is easier to just set a flag.

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -189,6 +189,10 @@ class TemplateExporter(Exporter):
     #Extension that the template files use.
     template_extension = Unicode(".tpl").tag(config=True, affects_environment=True)
 
+    exclude_header = Bool(False,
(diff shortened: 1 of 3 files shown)


takluyver:
It's almost always easier to set a flag, but templates are a flexible system that let you do lots of different things. I push back on adding options for things that templates can already do, because that way lies an untamed sea of hundreds of config options with lots of unforeseen interactions.

So is this use case something so common that it's too much to write a tiny custom template to cut out the header?

ocefpaf:
> So is this use case something so common that it's too much to write a tiny custom template to cut out the header?

IMO? Yes, that is why I sent the PR :grimacing:

meeseeksmachine:

ocefpaf:
The project that required this moved away from nbconvert. Thanks.

PR #921, opened 2018-11-28
Use utf-8 encoding in execute_api example

JarnoRFB:
As not setting the utf-8 encoding explicitly might produce broken notebooks on Windows, the encoding is set explicitly in the example. Moreover, the mode is set to 'w' only as text mode is standard for open and the 't' flag is not commonly used.


takluyver:
Closing/reopening to rerun tests with updated merge.

PR #922, opened 2018-12-03
Add LGTM code quality badges

xcorail:
Hi there!

I thought you might be interested in adding these LGTM [code quality badges](https://lgtm.com/projects/g/jupyter/nbconvert/ci/#badges) to your project. They indicate how you care about code quality and encourage your future contributors to do the same.
To get an idea of the analyses reflected by these grades, check the [alerts discovered by LGTM](https://lgtm.com/projects/g/jupyter/nbconvert).

N.B.: I am on the team behind LGTM.com, I'd appreciate your feedback on this initiative, whether you're interested or not, if you find time to drop me a line. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@
 [![Documentation Status](https://readthedocs.org/projects/nbconvert/badge/?version=latest)](https://nbconvert.readthedocs.io/en/latest/?badge=latest)
 [![Documentation Status](https://readthedocs.org/projects/nbconvert/badge/?version=stable)](http://nbconvert.readthedocs.io/en/stable/?badge=stable)
 [![codecov.io](https://codecov.io/github/jupyter/nbconvert/coverage.svg?branch=master)](https://codecov.io/github/jupyter/nbconvert?branch=master)
+[![Code Quality: Python](https://img.shields.io/lgtm/grade/python/g/jupyter/nbconvert.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/jupyter/nbconvert/context:python)
+[![Total Alerts](https://img.shields.io/lgtm/alerts/g/jupyter/nbconvert.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/jupyter/nbconvert/alerts)
 
 
 The **nbconvert** tool, `jupyter nbconvert`, converts notebooks to various other


takluyver:
I could go either way on this. I think static analysis can be useful for spotting problems, but I don't like the presentation of a 'grade' based on how many things their analysis flags. Code quality is much more complicated than that. And I have some concern that putting a badge saying '13 alerts' in the README will lead people to go and try to 'fix' them, even if they don't really need fixing.

xcorail:
Hello @takluyver 

Thanks for the feedback.
> I don't like the presentation of a 'grade' based on how many things their analysis flags. Code quality is much more complicated than that.

I definitely agree with that. But the grade is not just based on the number of alerts. You can read the study of our data science team [here](https://lgtm.com/blog/code_quality_1) to get where these grades come from.

> And I have some concern that putting a badge saying '13 alerts' in the README will lead people to go and try to 'fix' them, even if they don't really need fixing

`even if they don't really need fixing` makes me think that you might have spotted false positives in the alerts ... If this is the case, I would be super grateful if you could report them to me please, it it's not too much of your time, of course.

Finally, even if you don't put these badges, note that the LGTM alerts will always be there for your project, it's a free service for public projects. The alerts are refreshed once a day, but if you activate[ automated code review](https://lgtm.com/projects/g/jupyter/nbconvert/ci/), the PRs will be analyzed automatically.  

Cheers and thanks again for the feedback.

takluyver:
Well, currently the first alert in the list is a false positive:

> Call to method ExecutePreprocessor._wait_for_reply with too few arguments; should be no fewer than 2.

The second parameter is optional, so calling it with one argument is OK.

Some of the other errors about overridden method signatures are probably too severe; they may not be great coding practice, but they're not actually causing problems at runtime. But it is also usefully picking up mistakes; I made PR #949 after looking at the alerts.

Thanks for engaging with us, but as you're an LGTM employee, I'm going to say no to this for now. I'm open to reconsidering that if one of the other maintainers thinks a badge would be useful, but I feel like you're taking advantage of our "default to yes" instincts to promote a product.

xcorail:
Thanks for the FP report, I'll raise it to my team

> But it is also usefully picking up mistakes

I am happy to hear that.

>  I feel like you're taking advantage of our "default to yes" instincts to promote a product

I am sorry, this was not my intention:  As said in my previous comment, even without accepting these badges, I am happy if the service is useful to your projects. But I understand what you say and thank you for this feedback. 

Cheers

PR #924, opened 2018-12-04
ExecutePreprocessor: Get kernel name when executing, and not during initialization

mgeier:
Since there doesn't seem to be any progress in PR #886, I'm proposing this more minimalist approach to fixing #878 (that I've already suggested in https://github.com/jupyter/nbconvert/pull/886#issuecomment-420734298).

The underlying design problems (as mentioned in https://github.com/jupyter/nbconvert/pull/886#issuecomment-421959025) are of course still there, but this PR at least fixes the regression introduced in #816. Hopefully this can be incorporated in the next `nbconvert` release!

If desired, I can make another PR at a later point to fix those other problems, but I think that is not as urgent as the fix in this PR.

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -158,16 +158,6 @@ class ExecutePreprocessor(Preprocessor):
         )
     ).tag(config=True)
 
-    @default('kernel_name')
-    def _kernel_name_default(self):
-        try:
-            return self.nb.metadata.get('kernelspec', {}).get('name', 'python')
-        except AttributeError:
-            raise AttributeError('You did not specify a kernel_name for '
-                                 'the ExecutePreprocessor and you have not set '
-                                 'self.nb to be able to use that to infer the '
-                                 'kernel_name.')
-
     raise_on_iopub_timeout = Bool(False,
         help=dedent(
             """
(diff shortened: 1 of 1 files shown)


mpacer:
This has some issues from a traitlets perspective since it will no longer be a traitlet if set in this way. That shouldn't be an issue per se since it's primarily acting as a default value and there isn't any `@observe` logic here. 

I would prefer a test of all expected behaviour if this is to be merged.

SylvainCorlay:
> This has some issues from a traitlets perspective since it will no longer be a traitlet if set in this way. 

I think that it would still be a trait. This change merely removes the default initialization and sets a value later. 

> `self._trait_values['kernel_name']`

I would recommend not to use private functions of traitlets!

mgeier:
Oh, looks like @SylvainCorlay is right:

```python
import traitlets

class Foo(traitlets.HasTraits):
    bar = traitlets.Int()

x = Foo(bar=55)
x.bar = 99
print(x._trait_values)
```

This prints:

    {'bar': 99}

So it indeed seems to be still a trait(let), otherwise the `_trait_values` dict would not reflect the latest change, right?

So this means my PR can stay unchanged?

About the tests y'all were mentioning: can someone else please write them?

MSeal:
I can copy M's test over and get it merged independently if it helps.

MSeal:
Got a soft +1 from M in person. Will merge and will pop up a unittest here in a minute.

SylvainCorlay:
:+1:

mgeier:
Thanks a lot @MSeal and all others that were involved, it's great to have a working `master` branch again!

Are there any plans yet for a new `nbconvert` release?

MSeal:
I was going to see what should finish getting merged this week and get a smaller release going.

mgeier:
@MSeal That sounds great, thanks!

Issue #925, opened 2018-12-07
Latex formulas not rendering in html.


matthieu-pa:
Hi everyone, first a huge thanks for all the amazing job you do in nbconvert, which is really super convenient to share results to somebody who doesn't have jupyter installed :)

## Encountered Problem
When I use Jupyter Lab 0.35.2 export notebook or nbconvert (5.4.0) CLI to html, my equations don't render at all and raw latex is displayed instead.

## Possibly related issues
I thought #404 could be related to my problem but even extremely simple latex formulas don't render on my side.

## Temporary Fix
https://stackoverflow.com/questions/25655350/ipython-latex-math-not-rendered-in-html-using-nbconvert leaded me to try editing my html file.

I could successfully fix my problem replacing

```html
<script src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-AMS_HTML"></script>
```
by
```html
<script src='https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/latest.js?config=TeX-AMS_HTML'></script>
```

## Possible Solution
Although I haven't tried,  I assume changing https://github.com/jupyter/nbconvert/blob/master/nbconvert/templates/html/mathjax.tpl#L1 would also fix my problem.

precabal:
I'm facing the same issue. Did you change that line in your generated html or did you recompiled nbconvert?

matthieu-pa:
Good to see I'm not alone :)

For now I only tried changing the line in the generated html.

matthieu-pa:
@precabal 

I finally did try out the change in the mathjax.tpl url. It workded well so I opened the following pull request: #932

PR #932, opened 2018-12-25
Mathjax.tpl url update

matthieu-pa:
Hi dear nbconvert maintainers, thank you again taking care of such a valuable tool!

As reported in #925 the current MathJax url doesn't allow proper rendering of latex math formulas when exporting to html. The commit in this PR fixes this problem.


MSeal:
Thanks (sorry for the delay on review)

Issue #928, opened 2018-12-08
What are the plans regarding usage of pandoc?


mgeier:
Currently, we get a warning if we are using `pandoc` >= 2.0:

https://github.com/jupyter/nbconvert/blob/02c5cd1424b65da018f244e0f033a3dde55f76ca/nbconvert/utils/pandoc.py#L18

The warnings look like this:

    RuntimeWarning: You are using an unsupported version of pandoc (2.2.1).
    Your version must be at least (1.12.1) but less than (2.0.0).

Do you really want users to keep using old `pandoc` versions?
Right now that's probably still quite easy, but at some point, older versions will be harder and harder to get (or will have to be compiled from source, but that might also get harder with newer versions of Haskell compilers).

Currently, we are at `pandoc` release 2.5, and it is evolving quite rapidly: https://pandoc.org/releases.html

The warning is not only annoying for me, it's also confusing for others, e.g. #897.

There's actually an open issue: #815, but it doesn't look very active ...

I'm not sure what's the actual problem (#815 doesn't mention it), but for my use cases (using `nbsphinx`) I didn't see any problems, I'm just getting those annoying warnings (and many of them!).

Could this warning probably be limited to the use cases where there are actually problems?

paultopia:
Related to this issue, the warning that is issued comes out very spammy.  When using nbconvert as a library to convert a single notebook to pdf, I get the pandoc warning after calling the `.from_notebook_node()` method at REPL no less than... actually, I stopped counting at 40.  Seriously, it just filled multiple terminal screens with the same damn warning over and over again, apparenting calling `check_pandoc_version()` dozens and dozens of times.  

What would it take to get this down to 1?  If it requires a check for the pandoc version lots of times, can there be some kind of state flag in the conversion process that gets set once the warning has been delivered once, so it doesn't spam people to death?

minrk:
#964 supports pandoc 2 and fixes the warning being shown more than once when an unrecognized pandoc is found.

PR #933, opened 2018-12-29
Bump pandoc upper bound to 2.5

tk3369:
Let's keep it up to date unless there's any breaking issue.

Reference: https://stackoverflow.com/questions/53966433/jupyter-nbconvert-cannot-work-with-latest-version-of-pandoc/53967931#53967931

### DIFF
--- a/nbconvert/utils/pandoc.py
+++ b/nbconvert/utils/pandoc.py
@@ -15,7 +15,7 @@ from ipython_genutils.py3compat import cast_bytes, which
 from .exceptions import ConversionException
 
 _minimal_version = "1.12.1"
-_maximal_version = "2.0.0"
+_maximal_version = "2.5.0"
 
 def pandoc(source, fmt, to, extra_args=None, encoding='utf-8'):
     """Convert an input string using pandoc.


mgeier:
Related: #928

choldgraf:
the error in travis on 3.4 ends with this: 

```
  s.bind("tcp://%s:%i" % (self.ip, port))
  File "zmq/backend/cython/socket.pyx", line 547, in zmq.backend.cython.socket.Socket.bind
  File "zmq/backend/cython/checkrc.pxd", line 25, in zmq.backend.cython.checkrc._check_rc
zmq.error.ZMQError: Address already in use
==================== 1 failed, 253 passed in 150.85 seconds ====================
The command "py.test --cov nbconvert -v --pyargs nbconvert" exited with 1.
```

perhaps that's unrelated to the pandoc bump? In which case somebody could try restarting to see if that fixes it. I think a good way to answer the question "should newer versions of pandoc be supported" is to see if the tests all pass w/ newer pandoc

choldgraf:
I know that @ivanov has been looking into pandoc/notebook/etc stuff, maybe he has experience w/ pandoc 2.X and knows if there would be compatibility issues

tk3369:
The ZMQ error feels like a system issue rather than a code problem... Can anyone trigger the build again?

The nightly also failed but it seems to have a more _tangible_ reason:

```
/home/travis/virtualenv/python3.7-dev/lib/python3.7/site-packages/nbconvert/preprocessors/tests/test_execute.py:83: in assert_notebooks_equal
    self.assertEqual(normalized_expected_outputs, normalized_actual_outputs)
E   AssertionError: Lists differ: [{'data': {'text/plain': "'Code 1'"}, 'exe[46 chars]lt'}] != []
E   
E   First list contains 1 additional elements.
E   First extra element 0:
E   {'data': {'text/plain': "'Code 1'"}, 'execution_count': 1, 'output_type': 'execute_result'}
E   
E   + []
E   - [{'data': {'text/plain': "'Code 1'"},
E   -   'execution_count': 1,
E   -   'output_type': 'execute_result'}]
```

takluyver:
FWIW, you can usually rerun tests by closing and reopening the PR. At the moment, though, there's an unrelated problem with incompatible testing packages, so it will fail if we re-run it now. I'm trying to fix that in #941.

The most fragile bit for pandoc versions is [this filter](https://github.com/jupyter/nbconvert/blob/master/nbconvert/filters/filter_links.py) which converts links within a notebook to Latex references. It's used here:

https://github.com/jupyter/nbconvert/blob/8b47cf1ebfde7465e21927db5a699a54e4fc1c1a/nbconvert/templates/latex/report.tplx#L25

That converts the markdown to pandoc's JSON document format, applies the link converter to it, and converts the updated JSON to Latex. So it depends on the structure of pandoc's document format.

So before we expand the constraint, please double check that the latest version of Latex you want to be supported still performs that conversion correctly.

chrisjsewell:
Yes restricting the pandoc version is particularly a pain if you are using conda, which currently prohibits installing the latest versions of nbconvert and pandoc together.

Personally I would suggest you swap out pandocfilters for [panflute](http://scorreia.com/software/panflute/), which is more 'pythonic' and will deal with changes in the pandoc AST. 

Somemthing along the lines of:

```python 
import io
import panflute as pf

def resolve_references(source):
   input_stream = io.StringIO(source)
   with io.StringIO() as output_stream:
       pf.run_filter(resolve_one_reference, 
                          input_stream=input_stream, 
                          output_stream=outut_stream)
       return output_stream..getvalue()

def resolve_one_reference(link, doc):
    # type: (Link, Doc) -> Element
    """extract links that point to internal items, e.g. [text](#label)"""
    if not isinstance(link, pf.Link):
        return None
    match = re.match(r'#(.+)$', link.url)
    if not match:
        return None
    # pandoc automatically makes labels for headings.
    label = m.group(1).lower()
    label = re.sub(r'[^\w-]+', '', label) # Strip HTML entities
    return pf.RawInline(r'Section \ref{%s}' % label, format="tex")
```

choldgraf:
Just a note that another reason to bump pandoc at some point might be to support 2.6, which supports ipynb files and might make it possible to remove some code in nbconvert after a few depreciation cycles

PR #937, opened 2019-01-02
Convert proxy handler from callback to coroutine

takluyver:
According to the tornado docs, the `web.asynchronous` decorator is deprecated in 5.1 and will be removed in 6.0. There's no direct suggested replacement; it just says to use coroutines instead.

### DIFF
--- a/nbconvert/postprocessors/serve.py
+++ b/nbconvert/postprocessors/serve.py
@@ -9,7 +9,7 @@ import os
 import webbrowser
 import threading
 
-from tornado import web, ioloop, httpserver, log
+from tornado import web, ioloop, httpserver, log, gen
 from tornado.httpclient import AsyncHTTPClient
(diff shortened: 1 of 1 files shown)


takluyver:
I'll merge this soon unless anyone wants more time to look at it (@mpacer @MSeal )

MSeal:
Looks good. That reminds me that we need to get the patch release going as well.

alejopijuan:
@takluyver hey Thomas, I'm still getting the 500 Internal Server Error on the jupyter notebook. Any idea of what I should do?

takluyver:
If you're seeing the error this PR fixed, it's because you're using tornado 6.0a1 - a as in alpha. In that case, what you should do is go back to the stable version: `pip install tornado<6`. Hopefully any issues will be fixed well by the time Tornado 6 is stable.

PR #940, opened 2019-01-07
Fix LaTeX base template name in docs

clayadavis:
The docs currently specify `basic` as the name of the base LaTeX template, which is incorrect, resulting in the following error when used on the command line: `jinja2.exceptions.TemplateNotFound: basic`. In fact `basic` is the name of the base HTML template; the base LaTeX template is named `base`. These names are correct in the module help.


takluyver:
Thanks

PR #945, opened 2019-01-26
remove deprecated `export_by_name`

WrRan:
Through `exporter_locator` was deprecated since nbconvert 5.0,
https://github.com/jupyter/nbconvert/blob/2a7419e3784d71bca0c3566dda7f29b2fd13b7e7/nbconvert/exporters/exporter_locator.py#L27-L28
there still exists an import from it:
https://github.com/jupyter/nbconvert/blob/947055120e2268cca2672653490c294939123844/nbconvert/exporters/__init__.py#L3

This PR trys to fix it. :)

### DIFF
--- a/nbconvert/exporters/__init__.py
+++ b/nbconvert/exporters/__init__.py
@@ -1,6 +1,5 @@
 from .base import (export, get_exporter, 
                    ExporterNameError, get_export_names)
-from .exporter_locator import export_by_name
 from .html import HTMLExporter
 from .slides import SlidesExporter
 from .templateexporter import TemplateExporter


takluyver:
I'm OK with this. The semantic versioning scheme would say you can only remove things in a major release (like 6.0), but I think it's reasonable to allow rolling removals with a suitable deprecation period. It's been 2 years since 5.0 was released.

MSeal:
That seems reasonable to me. I was planning on shipping the 5.4.1 release -- should we merge this after or before the patch release you think @takluyver ?

takluyver:
I wouldn't remove something in a patch release - leave it for 5.5.

MSeal:
Sounds like we'll need to make some parallel fixes in the notebook server before 5.5 can release with this change fyi :)

Issue #896, opened 2018-10-10
Nbconvert silences errors if a pdf is successfully produced


t-makaro:
In the event that LaTeX throws an error but still produces a pdf, no error is displayed to the user. This makes diagnosing what caused LaTeX to stop part way very difficult. See #893 and #797

Currently, to find this error, a user must export to LaTeX and compile it themselves so LaTeX will display the error.

t-makaro:
[Here is the line](https://github.com/jupyter/nbconvert/blob/6dd08eb1590c0f6339f21f2ad2332f4cb4ede36d/nbconvert/exporters/pdf.py#L178). The error is only captured if the pdf file is not created.

PR #947, opened 2019-01-29
Return an error if latex does.

ryanlovett:
LaTeX can error, but still produce a PDF. This ensures the user is aware
of the error.

### DIFF
--- a/nbconvert/exporters/pdf.py
+++ b/nbconvert/exporters/pdf.py
@@ -172,8 +172,7 @@ class PDFExporter(LatexExporter):
             self.log.info("Building PDF")
             rc = self.run_latex(tex_file)
             if rc:
-                rc = self.run_bib(tex_file)
-            if rc:
+                self.run_bib(tex_file)
                 rc = self.run_latex(tex_file)
             
             pdf_file = notebook_name + '.pdf'
--- a/nbconvert/exporters/pdf.py
+++ b/nbconvert/exporters/pdf.py
@@ -175,7 +175,7 @@ class PDFExporter(LatexExporter):
                 rc = self.run_latex(tex_file)
             
             pdf_file = notebook_name + '.pdf'
-            if not os.path.isfile(pdf_file):
+            if not rc or not os.path.isfile(pdf_file):
                 raise LatexFailed('\n'.join(self._captured_output))
             self.log.info('PDF successfully created')
             with open(pdf_file, 'rb') as f:


yuvipanda:
Test failures seem to indicate either we already had failures trying to export to PDF, or latex's return code of non-zero means nothing.

nbconvert/exporters/pdf.py
```diff
@@ -175,7 +175,7 @@ def from_notebook_node(self, nb, resources=None, **kw):
                 rc = self.run_latex(tex_file)
             
             pdf_file = notebook_name + '.pdf'
-            if not os.path.isfile(pdf_file):
+            if not rc or not os.path.isfile(pdf_file):
```
yuvipanda: It would be clearer to write `rc != 0` since it's an integer for sure.

yuvipanda: Won't help with the test failures though.

ryanlovett: Looks like run_command does return a boolean,
https://github.com/jupyter/nbconvert/blob/master/nbconvert/exporters/pdf.py#L85

t-makaro:
Relevant open issue: #896

nbconvert/exporters/pdf.py
t-makaro: I think I see the issue. Look at the code a few lines above. The first ```rc = self.run_latex(tex_file)``` succeeds. Then:
```python
if rc:
     rc = self.run_bib(tex_file)
```
runs and fails which with your change causes tests to think they are failing.

I think saving the first return code as ```latex_rc = rc = self.run_latex(tex_file)``` and then doing
```python
if not latex_rc or …
```
should work.

ryanlovett: @t-makaro I think optimally `self.run_bib(tex_file)` shouldn't be run if there is no bibliographic data, but I agree that your approach works in those cases. As I understand it, wouldn't it ignore legitimate bibtex failures too?

t-makaro: Yes, the bibtex should not be run if there is no bibliographic data. I'm not sure how the bibtex integration works. Detecting the existence of a bibtex file may not be enough.

I think that just checking for latex failures would catch/display >95% of errors, and it would not harm any future endeavour to catch legitimate bibtex errors which are not caught now anyway. IMHO, catching only latex errors should be merged for the 5.5 release as it would solve a very long standing issue.

**CC** @MSeal

MSeal: I think it's fair to solve for latex issues first and tackle bibtex later. I too am not familiar with the bibtex integration enough to say how it should work here without some research into the problem.

The plan is to get 5.5 out here in April. If you want to get tests passing and then try tackling the biliographic failure catches in a follow-up thread we can get at least the minimal improvement. If you think you see a clear path to catching both that your confident in I'm ok with that too. Up to you on what you can fix before we release in a week or so. Tests will need to be passing for a merge either way.

MSeal: Also thank you for digging in to find the underlying issue. Been needing some eyes on these issues for a bit now.

nbconvert/exporters/pdf.py
```diff
@@ -172,12 +172,11 @@ def from_notebook_node(self, nb, resources=None, **kw):
             self.log.info("Building PDF")
             rc = self.run_latex(tex_file)
             if rc:
-                rc = self.run_bib(tex_file)
-            if rc:
+                self.run_bib(tex_file)
```
t-makaro: The only problem that I see with removing the if statement is that it means running latex an extra 3 times even if we don't need it. Before the second latex would only run if bibtex was successful.

ryanlovett: @t-makaro I agree.

However, it seems that manually running xelatex on the nbconverted-converted notebook2.ipynb fails (returns non-zero exit code) though it does produce a .pdf file. To me this seems like a bad .tex file to run tests on -- so maybe the conversion from ipynb to tex is incorrect somehow? As I indicated above I think a non-zero return from xelatex should return a LatexFailed exception.

t-makaro: I think it is reasonable to change the ```.ipynb``` used for the test, so that it does not produce a latex error.

MSeal: +1

t-makaro: The error occurs in [this notebook](https://github.com/jupyter/nbconvert/blob/master/nbconvert/exporters/tests/files/notebook2.ipynb). The error is in the last markdown cell.

Change:
```
$$$$
\[\]
$$
```
to
```
$$ $$
\[ \]
$ $
```
and the conversion will be successful. Pandoc butchers the empty math blocks without the space in the middle.

MSeal:
I believe this is now solved by https://github.com/jupyter/nbconvert/pull/1053 -- going to close this issue and get the release ball rolling now.

PR #956, opened 2019-02-15
add ascii output type to command line docs page, mention image folder output

discdiver:


discdiver:
I'm not sure why this failed. It passed when I ran it the first time. I then fixed a typo and it failed on Python 3.4.

discdiver:
And we're all green. Seems like a bug that Travis failed that one time on Python 3.4 with no substantive changes and then passed the other two times.

MSeal:
Thanks for the PR!

PR #959, opened 2019-02-17
Solves hanging indefinitely when kernel dies

dwjbosman:
I use Jupyter notebooks for automatic regression testing. I use nbconvert to execute notebooks containing tests:

 `jupyter nbconvert ..... --allow-errors ....  --ExecutePreprocessor.timeout=-1 .... --execute test.ipynb`

In one of the tests the kernel dies. Unfortunately nbconvert will hang on this. I can't use timeouts as some cells in the test notebooks take a long time to execute. If the kernel dies I want nbconvert to detect this almost immediately.

The patch checks that the kernel is alive while waiting for a message.

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -414,6 +414,24 @@ class ExecutePreprocessor(Preprocessor):
                 if not timeout or timeout < 0:
                     timeout = None
                 msg = self.kc.shell_channel.get_msg(timeout=timeout)
+                
+                if timeout is not None:
+                    # timeout specified
+                    msg = self.kc.shell_channel.get_msg(timeout=timeout)
+                else:
+                    #no timeout specified, if kernel dies still handle this correctly
+                    while True:
+                        try:
+                            #check every few seconds if kernel is still alive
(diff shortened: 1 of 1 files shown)


dwjbosman:
Travis CI reports a test failed on Python 3.5.  From the logs I can't see how my modifications impact this test.

MSeal:
Thanks for submitting this -- I was planning on tackling it soon as well (see https://github.com/nteract/papermill/issues/125). I've started seeing this error message occasionally during tests with papermill and nbconvert and will take a look if I can figure out why. I'll do a more thorough testing for this PR before merging as it's really critical that this operate correctly for papermill.

MSeal:
I dug through all of the client code and this reads as the correct fix for the issue.

I will say though that given how critical this piece section of code is for the library it should have a unittest for each branch possible here. This area may be undertested today as well. If you can add unittests for this pattern I'm willing to go dig deeper in to the unrelated build failure and trace back what's causing that.

I will be out of town for a bit, so might be slow on responding to this thread. But when I get back I'll want to get this and a few other changes merged for a new release.

MSeal:
Tested this out locally and will add tests in a separate PR given how important this is to get added.

PR #963, opened 2019-03-06
address deprecation warning in cgi.escape

minrk:
use html.escape when available


MSeal:
Thanks for getting that in. The number of direction warnings has been getting unmanageable.

PR #969, opened 2019-03-20
FIX: respect wait for clear_output

maartenbreddels:
clear_output was not respecting wait. This caused, for instance, ipywidgets' interact not to work https://github.com/QuantStack/voila/issues/85

The follow code:
```python
print("Hello world")
clear_output(wait=True)  # no output after this
```

did not output 'Hello world'. This is fixed with this PR. More complicated scenarios are tested in the notebook.

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -440,6 +440,7 @@ class ExecutePreprocessor(Preprocessor):
         exec_reply = self._wait_for_reply(msg_id, cell)
 
         outs = cell.outputs = []
+        clear_before_next_output = False
 
         while True:
             try:
@@ -475,11 +476,13 @@ class ExecutePreprocessor(Preprocessor):
             elif msg_type == 'execute_input':
                 continue
             elif msg_type == 'clear_output':
-                outs[:] = []
-                # clear display_id mapping for this cell
-                for display_id, cell_map in self._display_id_map.items():
-                    if cell_index in cell_map:
-                        cell_map[cell_index] = []
+                if content.get('wait'):
+                    self.log.debug('Wait to clear output')
+                    clear_before_next_output = True
+                else:
+                    self.log.debug('Immediate clear output')
+                    outs[:] = []
+                    self.clear_display_id_mapping(cell_index)
                 continue
             elif msg_type.startswith('comm'):
                 continue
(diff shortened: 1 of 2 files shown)


MSeal:
This looks like a nice fix ping me when the WIP ascpect is done and I'll +1 to merge.

maartenbreddels:
@MSeal thanks. I was done but thought I missed something. There are issues with [voila](https://github.com/QuantStack/voila/pull/91), but I don't think they should be solved in nbconvert. I did refactored it a bit so it's easier to reuse the `run_cell`. It can rebase #900 when merged, that will  also be clearer to implement.

The CI failure seems similar to other PR's, I'm assuming it is unrelated.

nbconvert/preprocessors/execute.py
```diff
@@ -440,6 +440,7 @@ def run_cell(self, cell, cell_index=0):
         exec_reply = self._wait_for_reply(msg_id, cell)
 
         outs = cell.outputs = []
+        self.clear_before_next_output = False
```
MSeal: Should we check if this is True before exiting (if it were the final message) and clear just before?

nbconvert/preprocessors/execute.py
```diff
@@ -493,22 +491,48 @@ def run_cell(self, cell, cell_index=0):
...
+            self.log.debug('Immediate clear output')
+            outs[:] = []
+            self.clear_display_id_mapping(cell_index)
+
+    def clear_display_id_mapping(self, cell_index):
+        for display_id, cell_map in self._display_id_map.items():
+            if cell_index in cell_map:
+                cell_map[cell_index] = []
+
+    def handle_comm_msg(self, outs, msg, cell_index):
```
MSeal: good idea for extensions to overwrite :)

nbconvert/preprocessors/execute.py
maartenbreddels: No, if nothing is outputted, it should not clear.

maartenbreddels: This should be covered by the notebook btw.

MSeal: I more meant if the final line of the final cell were `clear_output(wait=True)` wouldn't this not apply clear before exiting `run_cell`? I might be missing a behavioral aspect this morning before my coffee kicks in :D

maartenbreddels:
Yes, it would not clear, and it should not. One of the cells in the
notebook has this.

(from mobile phone)

nbconvert/preprocessors/execute.py
maartenbreddels: Yes, it would not clear, and it should not. One of the cells in the
notebook has this. Unless you found a scenario I did not test. (apparently, reply by email does not work well with threads).

MSeal: Ok thanks, looks good to me them. I'll get to looking at your other PRs related to widgets this next week (been super busy)

Issue #970, opened 2019-03-21
PDFExporter needs to have output_mimetype attribute set


Madhu94:
The [notebook](https://github.com/jupyter/notebook/blob/b26221ad44037f16610db0234ef109d7879ab437/notebook/nbconvert/handlers.py#L182) uses this attribute as the `Content-Type` header, and currently this is being picked up from its superclass, ``LatexExporter`` - [here](https://github.com/jupyter/nbconvert/blob/0ef92837822da09821b96955074002726b08a99f/nbconvert/exporters/latex.py#L46).

Let me know if you have any issues if I set this to ``application/pdf`` in the subclass

PR #972, opened 2019-03-25
PDFExporter currently has its "output_mimetype" traitlet to be 'text/latex'

Madhu94:
Closes https://github.com/jupyter/nbconvert/issues/970 

This causes issues when using "Export to pdf" option in the notebook,
as the notebook web application uses this traitlet's value as the
mimetype of the attached file.


MSeal:
LGTM, thanks for making the fix.

PR #979, opened 2019-04-04
Removed python 3.4 support, added 3.7 to tests

MSeal:
Discussed in person with @mpacer -- given our several underlying libraries no longer support 3.4 and we see some consistent test failures on PRs with 3.4 we thought removing python 3.4 from nbconvert 5.5 would make sense.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,15 @@ sudo: required
 dist: trusty
 
 language: python
-python:
-    - "nightly"
-    - 3.6
-    - 3.5
-    - 3.4
-    - 2.7
+matrix:
+  include:
+    - python: 2.7
+    - python: 3.5
+    - python: 3.6
+    - python: 3.7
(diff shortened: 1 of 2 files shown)


mpacer:
huzzah LGTM!

mpacer:
@Carreau I wanted to hear from you before I merged. But my finger is hovering over the button :D.

Carreau:
Sounds good !

Carreau:
Sounds good. If we are moving support for 3.4 do you want to bump the major version number ? I don't see any needs to do that, but it can also be a good time to maybe drop some 3.4 compatibility APIs if we have some.

MSeal:
I was planning on doing a 5.5 release in a week or two -- there's a lot of good work built up. Would you rather it jump to 6.0?

Also totally down to go scrub 3.4 specific branches of code if you know where any are.

Carreau:
There does not seem to be any 3.4 specific code, so I guess a 5.5 is fine. 

Thanks for handling all of that !

PR #980, opened 2019-04-05
Improved attachments +fix supporting non-unique names

maartenbreddels:
Via https://github.com/QuantStack/voila/issues/16 we found out that nbconvert does not handle attachments correctly when they have the same name.
With the new attachment notebook, the old result gave:
image
While this fix gives:
image

The fix works as following:
 * for every attachment, give them a globally unique name, by prefixing with `globally_unique_{cell_index}`
 * replace markdown source by replacing e.g. `](attachment:image.png)` by `](attachment:globally_unique_2_image.png)`

The last point might lead to unexpected results if somehow the markdown source include that 'magic' string. However, that would currently already break, because https://github.com/jupyter/nbconvert/blob/1fc39684d8d2fd57e261777f704ab923916a20b6/nbconvert/exporters/html.py#L85 would do an even less restricted replacement.


nbconvert/exporters/html.py
```diff
@@ -87,7 +89,29 @@ def process_attachments(self, nb, output):
                             'data:' + att_type + ';base64, ' + attachment[att_type])
         return output
 
+    def _make_attachments_globally_unique(self, nb):
+        nb = copy.deepcopy(nb)
```
maartenbreddels: Do we need to do a copy, or can we modify in place?

maartenbreddels:
I think this can be done more elegantly, for backtracking, the initial start was in https://github.com/jupyter/nbconvert/pull/780

maartenbreddels:
I think this implementation for attachments is much cleaner, since it know about it's context (the cell), so the attachments refer to the cell's attachment, avoiding name collisions (bug exposed by one of the untitests).

This also will be less error prone, since no string replacement or regex-fu is used. This approach also makes embedding images much more easy compared to [this approach](https://github.com/ipython-contrib/jupyter_contrib_nbextensions/blob/master/src/jupyter_contrib_nbextensions/nbconvert_support/pre_embedimages.py) which is something we need for voila.

@MSeal would love to see this in the next release, so we can push voila forward, happy to iterate on this quickly.

SylvainCorlay:
Looking forward with a release including that fix!

MSeal:
Got a few hanging PRs, then we should be set to release. Some reviews of the PRs I have open would help speed it up :)

PR #982, opened 2019-04-07
Added tests for each branch in execute's run_cell method

MSeal:
Refactoring run_cell wasn't really feasible to do safely without these tests. This should now allow https://github.com/jupyter/nbconvert/pull/905 to be tested (and updated) with all code paths that are currently supported.

### DIFF
--- a/nbconvert/preprocessors/tests/test_execute.py
+++ b/nbconvert/preprocessors/tests/test_execute.py
@@ -17,12 +17,13 @@ import re
 import nbformat
 import sys
 import pytest
+import functools
 
 from .base import PreprocessorTestsBase
 from ..execute import ExecutePreprocessor, CellExecutionError, executenb
 
 import IPython
-from mock import patch
+from mock import patch, MagicMock
 from traitlets import TraitError
 from jupyter_client.kernelspec import KernelSpecManager
 from nbconvert.filters import strip_ansi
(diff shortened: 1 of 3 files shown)


nbconvert/preprocessors/tests/test_execute.py
```diff
@@ -40,7 +42,64 @@ def _normalize_base64(b64_text):
     except (ValueError, TypeError):
         return b64_text
 
-class TestExecute(PreprocessorTestsBase):
+
+def merge_dicts(first, second):
```
mpacer: You could make this a *args and then change the code below to be 
```
outcome = {}
for d in args:
    outcome.update(d)
return outcome
```

mpacer: Also since this is a general pattern you might want to put it in `nbconvert/preprocessors/tests/base.py` and import it here.

nbconvert/preprocessors/tests/test_execute.py
```diff
@@ -40,7 +42,64 @@ def _normalize_base64(b64_text):
...
+
+
+class ExecuteTestBase(PreprocessorTestsBase):
+    def build_preprocessor(self, opts):
+        """Make an instance of a preprocessor"""
+        preprocessor = ExecutePreprocessor()
+        preprocessor.enabled = True
+        for opt in opts:
+            setattr(preprocessor, opt, opts[opt])
+        # Perform some state setup that should probably be in the init
```
mpacer: Is this a suggestion for a change that should be made in the ExecutePreprocessor `__init__` or a test specific init?

nbconvert/preprocessors/tests/test_execute.py
```diff
@@ -17,11 +17,13 @@
 import nbformat
 import sys
 import pytest
+import functools
 
 from .base import PreprocessorTestsBase
 from ..execute import ExecutePreprocessor, CellExecutionError, executenb
 
 import IPython
+from mock import MagicMock
```
mpacer: in Python 3 isn't it from `unittest.mock import MagicMock`?

nbconvert/preprocessors/tests/test_execute.py
```diff
@@ -40,7 +42,64 @@ def _normalize_base64(b64_text):
...
+    @staticmethod
+    def prepare_cell_mocks(*messages):
+        messages = list(messages)
+        def prepared_wrapper(func):
+            @functools.wraps(func)
+            def test_mock_wrapper(self):
+                parent_id = 'fake_id'
+                cell_mock = MagicMock(source='"foo" = "bar"', outputs=[])
+                # Hack to help catch `.` and `[]` style access to outputs against the same mock object
+                cell_mock.__getitem__.side_effect = lambda n: cell_mock.outputs if n == 'outputs' else None
```
mpacer: is this to get IPython [`IPython.utils.ipstruct.Struct`](https://ipython.readthedocs.io/en/stable/api/generated/IPython.utils.ipstruct.html#IPython.utils.ipstruct.Struct) like functionality? Is this needed because the MagicMock isn't able to inherit from that? Could we create a MagicMock subclass that just builds this functionality in for any of its attributes if we're going to be treating it like a struct?

nbconvert/preprocessors/tests/test_execute.py
```diff
@@ -40,7 +42,64 @@ def _normalize_base64(b64_text):
...
+        for opt in opts:
+            setattr(preprocessor, opt, opts[opt])
+        # Perform some state setup that should probably be in the init
+        preprocessor._display_id_map = {}
+        preprocessor.widget_state = {}
+        preprocessor.widget_buffers = {}
+        return preprocessor
+
+    @staticmethod
+    def prepare_cell_mocks(*messages):
```
mpacer: I think this needs a good deal of explanation at a high level before this will be easy to follow. I don't think we've been using mock much up to this point and this makes all of these heavily dependent on understanding how that works. I'm going to stop here, not because I don't have reviews to give for code further down but because this is going to take a few times going over with greater understanding on each pass given the new approach this is introducing.

meeseeksmachine:

nbconvert/preprocessors/tests/test_execute.py
```diff
@@ -350,3 +435,464 @@ def test_widgets(self):
...
+
+
+class TestRunCell(ExecuteTestBase):
+    """Contains test functions for ExecutePreprocessor.run_cell"""
+
+    @ExecuteTestBase.prepare_cell_mocks()
+    def test_idle_message(self, preprocessor, cell_mock, message_mock):
+        preprocessor.run_cell(cell_mock)
+        # Just the exit message should be fetched
+        message_mock.assert_called_once()
```
mpacer: This seems to be raising an error in 3.5 — any thoughts?

mpacer:
This looks great! We should probably get the tests passing before merging though.

nbconvert/preprocessors/tests/test_execute.py
MSeal: That is weird... not sure why yet.

MSeal: Gah, since I changed to the built-in mock library for 3, this function isn't implemented in 3.5: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once

mpacer: I think it's good to switching to the built-in when possible — it sets us up for dropping the dependency when we drop support for python 2.

MSeal:
:confetti_ball:

PR #983, opened 2019-04-07
Added note about widget support to execute docs

MSeal:
Followup to https://github.com/jupyter/nbconvert/pull/900, where we discussed what docs to add.


MSeal:
Just going to merge this one since it's docs only

Issue #981, opened 2019-04-06
Failure converting to LaTeX


boffi:
HTML conversion is fine, latex fails in a way that I find obscure.
All my notebooks (oh well, the five I've tested) trigger the same behaviour, so I guess it's not a problem of a particular notebook.

I have an up to date anaconda distribution on Debian Sid.

Here follows the backtrace from nbconvert:

```
23:39 boffi@debian:~/.../hw02 $ jupyter nbconvert Homework2.ipynb 
[NbConvertApp] Converting notebook Homework2.ipynb to html
[NbConvertApp] Writing 627605 bytes to Homework2.html
23:40 boffi@debian:~/.../hw02 $ jupyter nbconvert --to latex Homework2.ipynb
[NbConvertApp] Converting notebook Homework2.ipynb to latex
Traceback (most recent call last):
  File "/home/boffi/lib/miniconda3/bin/jupyter-nbconvert", line 11, in <module>
    sys.exit(main())
  File "/home/boffi/lib/miniconda3/lib/python3.7/site-packages/jupyter_core/application.py", line 266, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "/home/boffi/lib/miniconda3/lib/python3.7/site-packages/traitlets/config/application.py", line 658, in launch_instance
    app.start()
  File "/home/boffi/lib/miniconda3/lib/python3.7/site-packages/nbconvert/nbconvertapp.py", line 337, in start
    self.convert_notebooks()
  File "/home/boffi/lib/miniconda3/lib/python3.7/site-packages/nbconvert/nbconvertapp.py", line 507, in convert_notebooks
    self.convert_single_notebook(notebook_filename)
  File "/home/boffi/lib/miniconda3/lib/python3.7/site-packages/nbconvert/nbconvertapp.py", line 478, in convert_single_notebook
    output, resources = self.export_single_notebook(notebook_filename, resources, input_buffer=input_buffer)
  File "/home/boffi/lib/miniconda3/lib/python3.7/site-packages/nbconvert/nbconvertapp.py", line 407, in export_single_notebook
    output, resources = self.exporter.from_filename(notebook_filename, resources=resources)
  File "/home/boffi/lib/miniconda3/lib/python3.7/site-packages/nbconvert/exporters/exporter.py", line 178, in from_filename
    return self.from_file(f, resources=resources, **kw)
[...]

PR #985, opened 2019-04-07
Fixed latex export for svg data in python 3

MSeal:
Fixes https://github.com/jupyter/nbconvert/issues/981

Verified that the rendered latex output matches python 2 output visually. This will probably fix other python exports outside of latex svg given the code path involved in the fix.


nbconvert/preprocessors/extractoutput.py
```diff
@@ -80,27 +80,28 @@ def preprocess_cell(self, cell, resources, cell_index):
                 if mime_type in out.data:
                     data = out.data[mime_type]
 
-                    if (
-                        not isinstance(data, text_type)
-                        or mime_type == 'application/json'
-                    ):
+                    if mime_type == 'application/json':
```
akhmerov: This is too strict and will break some data types; compare with even a partial listing here: https://jupyter.readthedocs.io/en/latest/reference/mimetype.html

Also this change invalidates the follow-up comment, please keep them in sync.

akhmerov:
In the original implementation I've relied on the nbformat spec that guaranteed that json data is parsed directly and not serialized.

I don't actually understand why that bit requires changing to support svg: IIUC svg should be of text-type and shouldn't be treated as json by that code.

nbconvert/preprocessors/extractoutput.py
MSeal: Fair point on the strictness.

MSeal:
If you only add the svg test I added you'll see if fails for python 3 in the same way the original issue describes. The `data` object coming back from `out.data[mime_type]` is definitely `bytes` and adding `data = data.decode('utf-8')`, or even the original `data = json.dumps(data)`, generically to all byte data before any other processing seemed wrong.

I'll look into the test case I added and see if there's a less strict way to solve the issue.

akhmerov:
At the very least lists/dicts should be detected as definitely json

MSeal:
I looked into the failing pattern, it's when the mimetype is `application/pdf` that it fails in the test case originally.

I changed the implementation to mirror the original behavior more closely, with just logically moving the base64 data type checks in front of the json / text checks. Can you take a look again @akhmerov for any issue you can see?

MSeal:
Thanks for the review!

Issue #965, opened 2019-03-09
convert partial if name got space inside


SeuperHakkerJa:
When using the command line to convert the Ipython notebook, the conversion tends to fail. The symptom is either conversion failed directly or a part of the fail left undone making the converted file incomplete.

SeuperHakkerJa:
```jupyter nbconvert --to pdf long\ name.ipynb ``` provide partially converted pdf

**Comment by MSeal** [comment]
I am on vacation but can look into this when I get back unless someone else
beats me to investigating.

**Comment by SeuperHakkerJa** [comment]
No problem! Thanks

**Comment by MSeal** [comment]
Uhg, ok reproduced. This is a weird issue and one of us will need to dig a bit to figure out why. I've got a few things on my plate but I'll work towards it when I get some more critical items fixed elsewhere if no one else gets to this.

**Comment by mpacer** [comment]
Could you provide a complete minimal working example? i'm having a little confusion around what exactly that partial failure is and when I try to do this on my notebooks with spaces in the names it works fine.

**Comment by MSeal** [comment]
Failed Outcome (`jupyter nbconvert --to pdf "highlight date.ipynb"`)
[highlight date.pdf](https://github.com/jupyter/nbconvert/files/2998550/highlight.date.pdf)
Successful Outcome (`jupyter nbconvert --to pdf highlight_dates_run_cli.ipynb`)
[highlight_dates_run_cli.pdf](https://github.com/jupyter/nbconvert/files/2998551/highlight_dates_run_cli.pdf)
Source notebook (with extension changed to .txt so github will load it):
[highlight_dates_run_cli.txt](https://github.com/jupyter/nbconvert/files/2998716/highlight_dates_run_cli.txt)

**Comment by mpacer** [comment]
Thank you! 

To be clear, this only applies to pdf conversion, correct?

From looking at this output: 

```
! Missing $ inserted.
<inserted text>
                $
l.384 ...light date_files/highlight date_12_1.png}

?
! Emergency stop.
<inserted text>
                $
l.384 ...light date_files/highlight date_12_1.png}
```

I'm pretty sure the solution to this will relate to how we create links to anchors in https://github.com/jupyter/nbconvert/blob/master/nbconvert/filters/filter_links.py#L33-L35. 

Specifically, we may need to do something like that when we're making the image include statements.

However, I think the template that will need to change is probably this part

[...]

Issue #974, opened 2019-03-27
invalid filename for PDFExporter.latex_command


GorgiAstro:
Dear all,

I am not able to generate a PDF via pdflatex in Windows. The same code with the same configuration works on Ubuntu 18.04 x64 but fails on Windows 10 x64, with nbconvert 5.4.1 installed from conda-forge.

My `jupyter_nbconvert_config.py` has the following definition for latex_command:

`c.PDFExporter.latex_command = ['pdflatex', '{filename}']`

In Windows, it looks like the notebook.tex filename gets an extra backslash: 
```
[NbConvertApp] Running pdflatex 3 times: ['pdflatex', '.\\notebook.tex']

[NbConvertApp] CRITICAL | p failed: pdflatex .\notebook.tex

This is pdfTeX, Version 3.14159265-2.6-1.40.19 (TeX Live 2018/W32TeX) (preloaded format=pdflatex)

restricted \write18 enabled.

entering extended mode

! Undefined control sequence.

<*> .\notebook

              .tex

?

! Emergency stop.

<*> .\notebook

              .tex

!  ==> Fatal error occurred, no output PDF file produced!
```

Whereas it works on Linux:
`[NbConvertApp] Running pdflatex 3 times: ['pdflatex', './notebook.tex']`

Thank you,
Cheers
GorgiAstro

MSeal:
There's been a few filename issues in the pdf exporter, I've asked for some for some of the other contributors to try and take a look at that code.

t-makaro:
I cannot replicate this on windows. I see the double backslash too, so I don't think that's the issue. At first, I thought it was pdflatex, but I switched my config to use pdflatex instead of the default XeLaTeX and it still worked fine. The ```! Undefined control sequence.``` makes it look like ```.\notebook.tex``` is being placed inside the ```.tex``` file itself, and not any error with finding the file.

@GorgiAstro If you send me a ```.tex``` file (and other output files) from the output of ```Jupyter nbconvert --to latex notebook.ipynb```. I'd be happy to attempt to search the file ```.tex``` file for the cause.

Additional question: Does your notebook have a space in the filename? If so, try removing the space from the filename.

DevinCharles:
I think this is the same issue, accidentally started in the Jupyter Core repo.

https://github.com/jupyter/jupyter_core/issues/144

t-makaro:
[...]

Issue #633, opened 2017-07-26
adding a space to notebook title breaks conversion to pdf


clancia:
Hi there,

not sure if it qualifies as an issue.
I've just noticed that if the notebook title contains a space, then
```
jupyter nbconvert --to pdf note\ book.ipnb
```
fails if the notebook contains image output.

The error given by xelatex is

```
! Missing $ inserted.
<inserted text> 
                $
```
The issue is not present with Markdown-embedded images, but only with output images (e.g. generated by matplotlib.pyplot).

Please note that exporting to pdf via File > Download as > PDF via LaTeX works fine. Perhaps this is because spaces are replaced by +'s?
Anyway, a very simple solution is to avoid spaces in the notebook name and use a template for having a nicely formatted title when exporting to pdf.

Jupyter v. 4.3.0
Python 3.6.2 :: Anaconda custom (x86_64)
macOS 10.12.6

mpacer:
If you go to tex directly and then convert it does this work?

jpgrayson:
I am also observing this issue with jupyter-core 4.3.0. Running:

    jupyter-nbconvert --to pdf Notebook\ With\ Spaces.ipynb

produces a truncated, single-page pdf whereas `File > Download as > PDF via LaTeX` produces the correct (six-page) pdf.

> If you go to tex directly and then convert it does this work?

Yes, this workflow seems to produce the same output as the browser export:

    jupyter-nbconvert --to latex Notebook\ With\ Spaces.ipynb
    pdflatex Notebook\ With\ Spaces.tex

clancia:
Seems to be xelatex related, as far as I understand

```
jupyter nbconvert --to latex note\ book.ipynb
xelatex note\ book.ipynb
```

throws the same error as `jupyter nbconvert --to pdf note\ book.ipynb`

liamim:
this is related to spaces in the filename when calling the `\adjustimage` command.

t-makaro:
See my closing comment [here](https://github.com/jupyter/nbconvert/issues/965#issuecomment-518875394).

ytomar:
Work perfectly for me if we enclose the file name in double quotes

jupyter nbconvert --to pdf --TemplateExporter.exclude_input=True "Note book with spaces.ipynb"

PR #986, opened 2019-04-08
Fixed pdf export path and naming failures

MSeal:
The fix involves isolating all file paths to a temp directory with safe characters. I tried doing something less intrusive but the underlying latext command can't handle a variety of cases and I had to wrap just about all pathing for the exporter in a temp_dir wrapper to get all cases covered.

Fixes https://github.com/jupyter/nbconvert/issues/965 and https://github.com/jupyter/nbconvert/issues/633 

Might address https://github.com/jupyter/nbconvert/issues/974 given it sounded link complication by path characters.

Also I unintentionally fixed https://github.com/jupyter/nbconvert/pull/947 in digging into the original problem I believe.


nbconvert/exporters/pdf.py
```diff
@@ -154,41 +157,78 @@ def log_error(command, out):
...
             notebook_name = 'notebook'
             tex_file = self.writer.write(latex, resources, notebook_name=notebook_name)
             self.log.info("Building PDF")
             rc = self.run_latex(tex_file)
             if rc:
-                rc = self.run_bib(tex_file)
-            if rc:
+                # This can fail for non-conversion blocking reasons
+                self.run_bib(tex_file)
                 rc = self.run_latex(tex_file)
```
t-makaro: This second latex should only be run if the bibtex command is successful. Otherwise, the conversion will take almost twice as long as necessary. Maybe nest an ```if``` statement.

nbconvert/exporters/pdf.py
```diff
@@ -154,41 +157,78 @@ def log_error(command, out):
...
-            
+            if not rc:
+                raise RuntimeError('Failed to run "{}" commands'.format(
+                    [cmd.format(filename=tex_file) for cmd in self.latex_command]))
```
t-makaro: This won't display the captured output, will it? I think displaying the captured output is necessary if this is to fix #896 / #947. Otherwise, the user will still have zero idea why latex failed.

nbconvert/exporters/pdf.py
MSeal: Good catch, will change that.

nbconvert/exporters/pdf.py
MSeal: It does emit a critical log that goes to stderr by default. I didn't put the full output in the error message because it can be 100's or 1000's or lines long when I was testing. I'll look at truncating the output and putting a piece of it here so there's some visibility if log messages aren't being streamed to the client for some reason.

MSeal:
Only test with an issue still is the `test_linked_images` one. This is a bit tricky to fix -- need to see if I can just map the relative paths to a quoted absolute path when building the tex.

nbconvert/exporters/pdf.py
t-makaro: #988 Will fix latex error messages being far too long and not user friendly. I think that along with the fix to the notebook2.ipynb (which I see you did here), and the check for latex's return code beside the check for the file (like in #947) should nicely return any latex error and resolve #896.

MSeal: I'll rip out the message trimming I put in after #988 merges. I think this PR now implements #947 -- by raising on latex failure inside the `run_command` -- which the author hasn't been updating for the past week anyway.

t-makaro: Ah, I didn't see the fact that you moved the latex failure into ```run_command```. I like that.

MSeal:
Fixed the relative pathing (and spaces in paths) issues for latex/pdf extraction, I did a lot of testing locally to try different combinations of pathing issues and tried to update or add tests to capture those cases most concisely.

`xelatex` is annoyingly unable to handle spaces or other special characters in paths. Even with `grffile` on it only sorta works (depending on characters) but then will render all the parts of the path before the space character in a weird way alongside the actual image. To fix this I pass down path information to trigger a copy to the temp build directory during conversion. This could be extended to make non-local pathed files usable in convertion with a small follow-up PR.

nbconvert/exporters/tests/files/notebook2.ipynb
```diff
@@ -178,7 +178,7 @@
    "metadata": {},
    "source": [
     "Make sure markdown parser doesn't crash with empty Latex formulas blocks\n",
-    "$$$$\n",
+    "$$ $$\n",
```
MSeal: The rest of the file changes were just rerunning the notebook -- this was the only real change

nbconvert/templates/latex/base.tplx
```diff
@@ -20,6 +20,8 @@ This template does not define a docclass, the inheriting class must define this.
     % Basic figure setup, for now with no caption control since it's done
     % automatically by Pandoc (which extracts ![](path) syntax from Markdown).
     \usepackage{graphicx}
+    % protect against spaces in asset names
+    \usepackage[space]{grffile}
```
MSeal: Only sort of works, but it keeps things from hard crashing and just introduces rendering idiosyncrasies for when any code paths might leave a space in a rendered asset path.

nbconvert/templates/latex/document_contents.tplx
```diff
@@ -35,7 +35,7 @@
 
 % Display markdown
 ((* block data_markdown -*))
-    ((( output.data['text/markdown'] | citation2latex | strip_files_prefix | convert_pandoc('markdown+tex_math_double_backslash', 'latex'))))
+    ((( output.data['text/markdown'] | citation2latex | strip_files_prefix | convert_pandoc('markdown+tex_math_double_backslash', 'latex', extra_args=[], relative_path_replacement=resources.working_directory, build_path_replacement=resources.output_files_dir))))
```
MSeal: These values are set in `_fake_output_files_dir` in `exporters/pdf.py`. This was the easiest way for the wrapping process to inform the conversion that we're operating in a different directory than the file, and to provide a latex-safe path for conversions to take place within.

nbconvert/tests/test_nbconvertapp.py
```diff
@@ -444,7 +444,7 @@ def test_linked_images(self):
         """
         Generate PDFs with an image linked in a markdown cell
         """
-        with self.create_temp_cwd(['latex-linked-image.ipynb', 'testimage.png']):
+        with self.create_temp_cwd(['latex-linked-image.ipynb', 'test image.png']):
```
MSeal: Make's the test cover more edges by having a space in the asset.

meeseeksmachine:

nbconvert/utils/pandoc.py
```diff
@@ -118,6 +139,77 @@ def check_pandoc_version():
...
+    maybe_caption_groups = r'( )?(\"[^)\"]+\")?'
+    img_close_group = r'(\))'
+
+    if relative_path_replacement:
+        # We can't quote the path for xetex downstream if it has spaces in it
+        # See https://tug.org/pipermail/xetex/2009-December/015313.html
+        # However the will still render but will double render some path
+        # characters alongside the image. To avoid this also use build_path_replacement.
+
+        # => (!\[(?:[iI]mage|[aA]lt [tT]ext|)\]\()((?![^:\.]+:\/\/)[^\/][^)\"]*\.[^)\"]*[^ ])( )?(\"[^)\"]+\")?(\))
```
MSeal: add a link to regex 101 here

nbconvert/filters/tests/test_markdown.py
```diff
@@ -57,6 +59,88 @@ def test_markdown2latex(self):
...
+        l = 'Some \\includegraphics{/abs/path/to/reference.png}'
+        # Relative path replacement should be ignored
+        self.assertEqual(
+            convert_pandoc(s, 'markdown_strict', 'latex', [], '/path/with spaces'),
+            l)
+
+    @onlyif_cmds_exist('pandoc')
+    def test_markdown2latex_path_reference(self):
+        """markdown2latex replace_relative_path against referenced path test"""
+        s = 'Some ![image](reference)\n[reference]: ref.png'
```
MSeal: todo try with 2 newlines and see if it gets fixed

nbconvert/exporters/pdf.py
```diff
@@ -153,42 +163,64 @@ def log_error(command, out):
...
-            rc = self.run_latex(tex_file)
-            if rc:
-                rc = self.run_bib(tex_file)
-            if rc:
-                rc = self.run_latex(tex_file)
-            
+            self.run_latex(tex_file)
+            # This can fail for non-conversion blocking reasons
+            if self.run_bib(tex_file):
+                self.run_latex(tex_file)
```
mpacer: Technically — if we're running bibTeX, we should be running LaTeX one more time. That is it should look like
```
xelatex file.tex
bibtex file.tex
xelatex file.tex
xelatex file.tex
```

For those who want to know why — https://tex.stackexchange.com/questions/53235/why-does-latex-bibtex-need-three-passes-to-clear-up-all-warnings explains it. But basically because LaTeX was designed in a way that it kept little in memory, it needs to separately create references (LaTeX  run 1) , create a file to house all the bibliography references (bibTeX), add the reference file contents to the page (LaTeX run 2), resolve all references (LaTeX run 3).

t-makaro: I believe the ```run_latex``` command actually runs latex 3 times by default, so it should be fine.

MSeal:
```

File: nbconvert/filters/markdown.py [hunk 10]
```diff
@@ -54,7 +55,10 @@ def markdown2latex(source, markup='markdown', extra_args=None):
     out : string
       Output as returned by pandoc.
     """
-    return convert_pandoc(source, markup, 'latex', extra_args=extra_args)
+    return convert_pandoc(source, markup, 'latex',
+        extra_args=extra_args,
+        relative_path_replacement=relative_path_replacement,
```
mpacer: I would not add these keyword arguments to `convert_pandoc`. I think it would be cleaner to do it in either: 
1. a jinja filter that is run prior to calling `convert_pandoc` 
2. a preprocessor that changes these links before jinja templating is applied 

going into the details of this would be too long for this little comment, but I'll add more details about the implications of both later.

The same point should be made for `build_path_replacement=build_path_replacement` below.

nbconvert/filters/markdown.py
```diff
@@ -33,7 +33,8 @@ def markdown2html_mistune(source):
 ]
 
 
-def markdown2latex(source, markup='markdown', extra_args=None):
+def markdown2latex(source, markup='markdown', extra_args=None,
+                   relative_path_replacement=None, build_path_replacement=None):
```
mpacer: I also wouldn't add these keyword arguments here.

nbconvert/filters/pandoc.py
```diff
@@ -1,7 +1,8 @@
 from nbconvert.utils.pandoc import pandoc
 
 
-def convert_pandoc(source, from_format, to_format, extra_args=None):
+def convert_pandoc(source, from_format, to_format, extra_args=None,
+                   relative_path_replacement=None, build_path_replacement=None):
```
mpacer: See other comment about these flags.

nbconvert/filters/pandoc.py
```diff
@@ -18,9 +19,21 @@ def convert_pandoc(source, from_format, to_format, extra_args=None):
     to_format : string
         Pandoc format for output.
 
+    extra_args : list (optional)
```
mpacer: Many thanks for adding this!

mpacer:
A high level point — this is doing a lot of cool things, but how much of the issues are fixed by adding the following to the preamble?

```
    \usepackage{grffile}

\makeatletter
\def\Gread@@xetex#1{%
\IfFileExists{"\Gin@base".bb}%
    {\Gread@eps{\Gin@base.bb}}%
    {\Gread@@xetex@aux#1}%
}
\makeatother
```

I got that from https://github.com/ho-tex/oberdiek/issues/31#issuecomment-441094438 which is a patch while maintainers of latex core push this fix out more broadly.

If you want to see a working example of this in a directory structure you can clone https://github.com/mpacer/tex_ex/ and run `xelatex` on it's `base.tex` and it should work for you.

nbconvert/templates/latex/base.tplx
mpacer: This seems like it's redundant with the addition of grffile below.

nbconvert/templates/latex/base.tplx
```diff
@@ -161,7 +163,7 @@ This template does not define a docclass, the inheriting class must define this.
 
     ((* block commands *))
     % Prevent overflowing lines due to hard-to-break entities
-    \sloppy 
+    \sloppy
```
mpacer: I would see if you can turn off your auto-formatter for .tplx and .tpl files. There are situations in which having spaces can be semantically meaningful in LaTeX.

mpacer:
So @MSeal and I were working through this today.

My major suggestion is to keep the pandoc APIs unchanged.  

Instead, switch the current code into its own jinja filter (in the `nbconvert/filters/` directory) and ensure that it is added to the list of filters available in the environment (you can follow along with https://github.com/jupyter/nbconvert/blob/master/nbconvert/exporters/latex.py#L48-L51).

Fortunately, the model you're already using here https://github.com/jupyter/nbconvert/blob/7cf79f82d5a890570fe4b55a0b2ede0e982c0727/nbconvert/utils/pandoc.py#L79-L81 is literally equivalent to adding a filter before the `convert_pandoc` call in the template here https://github.com/jupyter/nbconvert/blob/7cf79f82d5a890570fe4b55a0b2ede0e982c0727/nbconvert/templates/latex/document_contents.tplx#L67

What might be even cleaner would be to have two separate filters rather than one:  `replace_relative_paths` and `replace_build_paths`. These could reasonably live inside the existing [`filter_links.py`](https://github.com/jupyter/nbconvert/blob/master/nbconvert/filters/filter_links.py)

so then the code would look something like 

```
# inside filters/filter_links.py
def replace_relative_paths(source, working_directory):
    …
def replace_build_paths(source, output_files_dir):
    …
```
```
# Inside exporters/pdf.py
from nbconvert.filters.filter_links import resolve_references
from nbconvert.filters.filter_links import replace_relative_paths
from nbconvert.filters.filter_links import replace_build_paths

    def default_filters(self):
        for x in super(LatexExporter, self).default_filters():
            yield x 
        latex_filters = (('resolve_references', resolve_references), ('replace_relative_paths', replace_relative_paths), ('replace_build_paths', resolve_references))
        for filter_tuple in latex_filters: 
             yield filter_tuple
```

```
% Inside templates/latex/document_contents.tplx
    ((( cell.source | citation2latex | strip_files_prefix | replace_relative_paths(resources.working_directory)| replace_build_paths(resources.output_files_dir) | convert_pandoc('markdown+tex_math_double_backslash', 'json', extra_args=[]) | resolve_references | convert_pandoc('json','latex'))))
```

If you wanted to make a change that you couldn't make from manipulating the markdown itself but that you needed to solve before it was written out into the LaTeX, you can do that by manipulating the a JSON representation of the pandoc AST that pandoc knows how to produce and consume. This is what `resolve_references` is doing today. If you want to learn more about how to implement that you can check out: https://github.com/jgm/pandocfilters.

However, given that you haven't needed to do this for your fix, we can probably just get away with creating two new filters.

If you'd like, I could push these changes on top of your PR. Or we could merge this and then I could make a new one with these changes in place — your call.

nbconvert/exporters/pdf.py
```diff
@@ -153,42 +163,64 @@ def log_error(command, out):
...
-        if resources and resources.get('metadata', {}).get('path'):
-            self.texinputs = resources['metadata']['path']
-        else:
-            self.texinputs = getcwd()
-        
-        self._captured_outputs = []
-        with TemporaryWorkingDirectory():
+        if not resources:
+            resources = {}
+        with self._fake_output_files_dir(resources):
```
t-makaro: Modifying the path locations should occur at the level of the latex exporter. This set-up removes the ability to use the latex exporter + manually compiling with latex as a debugging tool, since the intermediate latex will be different.

nbconvert/templates/latex/base.tplx
t-makaro: grffile only protects against spaces in names when using pdflatex, but not XeLaTeX.

mpacer: In the example I provided at https://github.com/mpacer/tex_ex/ `grffile` protects against spaces in names when using XeLaTeX. But, it requires adding the fix from https://github.com/ho-tex/oberdiek/issues/31#issuecomment-441094438 to get it to work.

nbconvert/templates/latex/base.tplx
```diff
@@ -53,8 +55,8 @@ This template does not define a docclass, the inheriting class must define this.
     \usepackage[mathletters]{ucs} % Extended unicode (utf-8) support
     \usepackage[utf8x]{inputenc} % Allow utf-8 characters in the tex document
     \usepackage{fancyvrb} % verbatim replacement that allows latex
-    \usepackage{grffile} % extends the file name processing of package graphics 
-                         % to support a larger range 
+    \usepackage{grffile} % extends the file name processing of package graphics
```
mpacer: This is being added above in addition to existing here.

nbconvert/exporters/pdf.py
```diff
@@ -6,23 +6,25 @@
 import subprocess
 import os
 import sys
+import shutil
```
mpacer: It doesn't look like this import is being used.

willingc:
@MSeal Going to bump this to 6.0

MSeal:
@t-makaro I was going to take a stab at updating this PR for the 6.0 milestone, if you had thoughts about it since we last were working on this.

t-makaro:
@MSeal I'm not sure that this is needed anymore. My understanding of this PR was that it created a temp directory for output images that were extracted by the extractoutputpreprocessor. The extractoutputpreprocessor was sometimes creating file names that LaTeX could not handle.

Due to improvements in LaTeX's packages, and my fix in #1193 (which was a fix to my earlier fix), I think that this should work fine. In fact, the LaTeX package fixes should also improve file path problems for files referenced in markdown.

MSeal:
Yeah I think you're right! I tested out a few of the failing patterns and we're at a working place for the ones I was targeting with this PR. I'll close this out :)

Issue #552, opened 2017-03-20
Pdf doesn't include images


xNok:
When exporting to PDF it does not include images inserted via markdown

When using the markdown `![Description](pb1.png)` I got this error:
```
LaTeX Warning: File `pb1.png' not found on input line 320.

! Unable to load picture or PDF file 'pb1.png'.
<to be read again> 
                   }
l.320 \includegraphics{pb1.png}
                               
? 
! Emergency stop.
<to be read again> 
                   }
l.320 \includegraphics{pb1.png}
                               
No pages of output.
Transcript written on notebook.log.
```

When using the image html tag `<img  src=\"pb1-bfs.png\"/>` I got no image as an output

Everything is displayed correctly in the notebook for both case.

mpacer:
Could you try converting first to `.tex` (with the `--to latex` exporter invocation) and then manually converting to pdf using `xelatex`? 

Do you run into the same error? 

We don't support pdf export from general HTML markup in markdown cells. I'm not sure if this is within or outside the purview of our intended coverage area.

takluyver:
I think the issue here is that our PDF conversion produces latex in a temporary directory, and then runs xelatex on it there. It extracts images embedded in the notebook to a location where it can refer to them, but it doesn't copy images referenced from Markdown cells to a matching location in the temporary directory, so the references to them in the generated Latex break.

See also #52.

xNok:
@mpacer you are right once I get the latex file i can compile it manually with my latex compiler with no issues. I guess It confirms @takluyver explanation.

Why don't we have access to `download as latex` via the file menu? Is this intentional?

mpacer:
Great question! i'll make a PR and we'll find out.

mpacer:
@xNok  So made a PR, but even if that is merged, I'd still recommend using the LaTeX converter from the command line because right now the download buttons aren't exactly emulating the functionality of the command line LaTeXExporter. 

Specifically, the command line LaTeX exporter will create a directory of files to represent any of your outputs that need to be called from an external source (i.e., images which are included via `\begin{figure}…\end{figure}`). In order to do this in the browser, we need to return a zip file (not just a TeX file) that will contain both the .tex and the appropriate images. Right now that zipping utility doesn't seem to be being triggered for either the LaTeX or the Markdown exporter (didn't test others but I'm guessing its the same). There are utilities for creating that zip file, but they aren't being used right now; I'm not sure why. 

[...]

PR #989, opened 2019-04-11
Strict Placement of Images from Markdown

t-makaro:
When pandoc converts markdown images it places them inside a ```\begin{figure}…\end{figure}```. Since the ```figure``` environment is a float, latex will place the image where is seems fit and likely not the same location as in the notebook. This fixes that. (Issue mentioned [here](https://github.com/jupyter/nbconvert/issues/552#issuecomment-478138406).)


t-makaro:
Sorry, I made a rookie mistake and didn't make this branch off master and thus combining it with my other PR.

Issue #340, opened 2016-07-19
Hidden \includegraphics[]{} latex call


kapilsh:
I was trying to override the report.tplx template to add a cover page to it.

I had to use \includegraphics with width and height that equaled the page width and page height.

However, due to the code below:

```
\let\Oldincludegraphics\includegraphics
% Set max figure width to be 80% of text width, for now hardcoded.
\renewcommand{\includegraphics}[1]{\Oldincludegraphics[width=.8\maxwidth]{#1}}

```

the command was hidden.

I had to dig through the code to figure out that I could call \Oldincludegraphics.

michaelpacer:
This was introduced in https://github.com/jupyter/nbconvert/pull/178 as a way of making the default `\includegraphics` command print figures specifically that would not span the page. 

It may be possible to use [`\xparse`](https://www.ctan.org/pkg/xparse) to make this customisable, but it is nontrivial to make it the standard `\includegraphics` work with arguments while also having this nicer default. 

Perhaps some mention of this should be added to the documentation? @willingc where in the documentation would you think this would most appropriately fit?

JosefAssad:
This trips up quite a few other things one might be doing with a LaTeX nbconvert template. I just ran into this issue now while setting a background png for all pages. It looks to me like any package calling `\includegraphics` is keeling over and dying.

mpacer:
I'm still not sure how to resolve this. One possibility would be to create a custom filter for the pandoc to LaTeX conversion? @takluyver  do you have any thoughts?

takluyver:
Maybe we could put the conflicting code in our template (the code @kapilsh identified) into its own block, so that custom templates can override that specifically.

MSeal:
No one is pushing implementation here. Removing milestone until someone actively picks it up.

t-makaro:
Dev note: [This tex.stackexchange](https://tex.stackexchange.com/questions/83020/set-standard-default-scaling-of-includegraphics) might help. Apparently, ```\setkeys``` can be used to set defaults. I will try it out when I have time.

Issue #21, opened 2015-05-15
Missing file during install


rsdenijs:
Cloning and executing pip install -r requirements.txt -e . results into

----
     Running setup.py develop for nbconvert
        Complete output from command /home/xxx/.virtualenvs/jupyter-dev/bin/python -c "import setuptools, tokenize; __file__='/home/xxx/code/nbconvert/setup.py'; exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" develop --no-deps:
    /home/xxx/.virtualenvs/jupyter-dev/local/lib/python2.7/site-packages/setuptools/dist.py:282:     UserWarning: Normalizing '4.0.0.dev' to '4.0.0.dev0'
      normalized_version,
    running develop
    running css
    Failed to download css from https://cdn.jupyter.org/notebook/4.0.0-dev/style/style.min.css: [Errno socket error] [Errno 1] _ssl.c:510: error:14077438:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert internal error
    error: Need Notebook CSS to proceed: nbconvert/resources/style.min.css

----
Any ideas?

takluyver:
Is it reproducible? I've just tried, and it downloads OK from here.

minrk:
@rsdenijs when did you do this? It's possible the server was unavailable for a bit. Does it work now?

rsdenijs:
I tried just before creating the issue.
I just tried it on a different computer and it also happens. 

Steps that repro
Create a new virtualenv
pip install --upgrade pip setuptools
Clone jupyter noteobook 
pip install -r requirements -e . 

All goes fine until downloading the aforementioned css file.

damianavila:
I can see the CDN w/o problems... and I could install it successfully...

minrk:
@rsdenijs what happens when you visit that URL with a browser, or try to download the file with curl/wget?

takluyver:
@dongweiming reported the same error. I'm not sure what the common factor is. It looks like @rsdenijs is on Linux (going by the paths in the traceback), while @dongweiming appears to be on a Mac.

minrk:
And it installs correctly for me on Linux, FreeBSD, and OS X, on various VMs, networks, and cloud services. @rsdenijs @dongweiming what OS/distro version and Python version, and how did you install Python? I'm guessing it's something wrong with the Python's linked SSL/TLS support on the client.

[...]

PR #990, opened 2019-04-11
[WIP] Strict placement of images

t-makaro:
When pandoc converts markdown images to latex it places them inside a ```\begin{figure}…\end{figure}```. Since the ```figure``` environment is a float, latex will place the image where is seems fit which is likely not the same location as in the notebook. This fixes that. (Issue mentioned [here](https://github.com/jupyter/nbconvert/issues/552#issuecomment-478138406).)

### DIFF
--- a/nbconvert/templates/latex/base.tplx
+++ b/nbconvert/templates/latex/base.tplx
@@ -65,6 +65,8 @@ This template does not define a docclass, the inheriting class must define this.
     \usepackage[normalem]{ulem} % ulem is needed to support strikethroughs (\sout)
                                 % normalem makes italics be italics, not underlines
     \usepackage{mathrsfs}
+    \usepackage{float}
+    \floatplacement{figure}{H}
     ((* endblock packages *))
 
     ((* block definitions *))


mgeier:
Alternatively, you could also experiment with disabling the `implicit_figures` extension using the `pandoc` option `markdown-implicit_figures`, see http://pandoc.org/MANUAL.html#extension-implicit_figures.

This will create `\includegraphics{…}` instead.

t-makaro:
Interesting. I chose the method I did because that's what I did over in [nb_pdf_template #21](https://github.com/t-makaro/nb_pdf_template/pull/21) as suggested in some stack overflow answer I found but can't find again.

Presumably disabling implicit figures would occur [here](https://github.com/jupyter/nbconvert/blob/4756a41c29ee05c3d4c379e3363007bebf6b081e/nbconvert/templates/latex/document_contents.tplx#L38), and it would better replicate the original notebook since captions won't appear. I'll try it out.

t-makaro:
Ok, I've got both methods to work as expected. It's really a matter of opinion.

Method 1. (Currently in this PR)(```\usepackage{float}``` etc.)
Pros:
1. Leaves the image centered like in the notebook
2. Probably easier to fix other latex floats from doing the same if we encounter them.

Cons:
1. Uses the text in the markdown link as caption text which the notebook does not do. (Could be a pro?) It should be noted that if the image does not have caption text in markdown (ie ``````) then pandoc does not use the implicit figures anyway and none of this matters.

Method 2: (Disable implicit figures in pandoc)
This would add a ```-implicit_figures``` to the markdown to latex pandoc calls on the 2 lines mentioned in my other comment.
Pros:
1. Consistent behaviour for all markdown images. No caption ever, and always right aligned.

Cons:
1. Image is not centered.
2. No ability to have a caption on the figure.

IMHO: I think I prefer method 1 (the latex solution), since if no "caption" is given in markdown, then the behaviour of images appearing in the wrong spot does not occur. We can then consider the "caption" to be a feature of the output. This would also result is less change to the current behaviour, and more control for the user.

Thoughts?

MSeal:
I think I lean towards # 1 but I'm not an authority on how latex rendering should appear. If we need to remove the con from # 1 we could address that with a configuration option in the future to eliminate the caption on the nbconvert side. This is also a lower impact change to what's already there in rendering paths.

Would prefer another opinion that agrees with the trade-off before feeling comfortable with merging. @mpacer any opinion on this one?

mgeier:
As @t-makaro mentioned, we are talking about two different things here:

* centered images
* figures with captions

I think `nbconvert` should try to somewhat mimic the behavior of the Classic Notebook and/or JupyterLab.

If images are centered there, `nbconvert` should try to center them (if feasible).

But since there are no figure captions there, `nbconvert` shouldn't have those either.

While I'm not at all against centering images, I don't think `pandoc`'s `implicit_figures` extension should be used to implement this.

BTW, the text between the brackets in `alt text` is supposed to be used as the "alt" attribute of the `` tag in HTML. It is meant for the case where the image cannot be displayed of for accessibility reasons (e.g. for blind users). It is *not* meant as a mechanism for providing a figure caption!

t-makaro:
Added the [WIP] tag since #992 actually implements this (otherwise images can appear in the middle of code cells), but at least some refractor could be used. If the usepackage implementation is to be used, it should be moved into the lower level templates so that all cell styles my use it.

t-makaro:
Another option would be to modify the ```\includegraphics``` macro to include centering by default, and then disable ```implicit-figures```. The only downside is that this worsens #340.

However, I just noticed that Jupyter lab does not center the image. It appears to me that disabling the ```implicit-figures``` extension would accurately replicate jlab's behaviour. If we do this route, I would recommend also removing centering from image outputs for consistency. The thing to consider here though is that we are outputting to a page. Does is make sense to not center outputs on a printed page? I think no, images should be centered.

I did ask a question over on [pandoc's forum](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/pandoc-discuss/RCN-cGzE4h0/PgbBr4jMDAAJ) that hopefully might provide some answers.

mgeier:
> I just noticed that Jupyter lab does not center the image.

Interesting.

One important question is if you consider an image that appears in a paragraph on its own as a "block" element. When an image is used within some text, it should be an "inline" element and centering isn't really an option, right?

See also http://pandoc.org/MANUAL.html#extension-implicit_figures.

BTW, the HTML `` element is by default "inline", but you can switch to `display: block;` via CSS, see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#Styling_with_CSS.

t-makaro:
This PR is superceded by #1024. The current behaviour turns off all captions, centers images, and fixes their proper location.

Issue #952, opened 2019-02-05
Different results from exporting to "script" versus "python"


justmarkham:
With the notebooks I have tested, these produce the exact same output:

```
jupyter nbconvert example.ipynb --to script
jupyter nbconvert example.ipynb --to python
```

When I wanted to exclude Markdown cells, I tried the following, and only the second option (`python`) worked, whereas the first option (`script`) appear to ignore the configuration option:

```
jupyter nbconvert example.ipynb --to script --TemplateExporter.exclude_markdown=True
jupyter nbconvert example.ipynb --to python --TemplateExporter.exclude_markdown=True
```

Here are my questions:

1. I only found [one page](https://nbconvert.readthedocs.io/en/latest/customizing.html) in the documentation which advises using `python`. Everywhere else in the documentation advises using `script`. Is that because `script` is the officially endorsed way to export to a script, and `python` is just an alias for `script`?
2. Is it a bug that my `script` example ignored the configuration option (whereas `python` did not), or is that the expected behavior?

Thanks!

justmarkham:
Note: The behavior I described above holds true in both nbconvert 5.3.1 and 5.4.

MSeal:
Sorry this got missed. I'm not as familiar with this code path but it sounds like a bug and not expected behavior. I'll tag this as a good item to look into for 5.5.

t-makaro:
Dev note: I took a quick look at the script exporter, and it should call the python exporter. My guess is that the ```--TemplateExporter``` flag is being passed to the script exporter, but the script exporter is not passing it to the python exporter making this (possibly) a quirk in traitlets. 

(That's just my guess, as I'm more familiar with the pdf/latex exporter than the script exporter)

t-makaro:
**CC** @MSeal. Should be closed?

PR #993, opened 2019-04-20
Enabled configuration to be shared to exporters from script exporter

MSeal:
Solves https://github.com/jupyter/nbconvert/issues/952


mpacer:
LGTM

Issue #659, opened 2017-08-31
nbconvert vs. interactive Jupyter operation


ElectricGizmo:
I am running the same notebook interactively and using nbconvert on Windows, clean Python 3 kernel installation (using Anaconda).

The notebook works just fine from within Jupyter but I get the error when converting the notebook to html to automate generation of event reports for colleagues. When I convert the notebook to html I get  
`[NbConvertApp] WARNING | Timeout waiting for IOPub output`  
and the output is truncated for one of the cells, not always in the exact same position (for the same input data), never with any exception detail.  
I switched the _raise_on_iopub_timeout_ flag and apparently `execute.py line 348 in run_cell` raised `RuntimeError("Timeout waiting for IOPup output")` and announces `The system cannot find the file specified`  
The notebook makes heavy (heavy!) use of matplotlib to show the data and crunches for several minutes in every case, whether it works (Jupyter) or not (nbconvert).  
Also worth mentioning, the graph generation is done in object methods defined in an external set of scripts first run by the notebook (no problem there) then invoked at the data display time.  
Again, it all works fine when in an interactive Jupyter-served page.  

- Is there a way to display the callstack to identify what file is not found? I suspect it is a png generated by matplotlib, but which one?
- What difference between interactive and nbconvert processing could explain the issue?

Thanks!

For the records I invoke nbconvert from a Python script as follows:
> os.system(r"jupyter nbconvert " +
>           r"--ExecutePreprocessor.timeout=None "+
>           r"--ExecutePreprocessor.iopub_timeout=10 " +
>           r"--to html " +
>           # r"--allow-errors " +
>           r"--ExecutePreprocessor.raise_on_iopub_timeout=True  " +
>           r"--execute {}ARManalysis.ipynb".format(localPath))

takluyver:
An IOPub timeout means it has produced no output for a certain amount of time. You're setting that time to 10 seconds. Is it possible that your crunching code produces no output for 10 seconds.

I don't know where the "cannot find the file" error is coming from, but i suspect that's a red herring.

mpacer:
Would it be possible to trigger something like the `%debug` magic that will drop you into a debugger if an error is raised when running --execute from the command line?

ElectricGizmo:
@takluyver , thanks for your answer.
I would be surprised the processing takes more than twice longer when using nbconvert than when running from within Jupyter. Besides I tried with 600s with the same effect (except it took longer to announce the issue).
[...]

PR #994, opened 2019-04-20
Fixing #659 (output cutoff and iopub timeout)

djherbis:
@MSeal @mpacer This is probably a little rough but figured this might be a good place to start.

I still need to add a test showing that this fixes my example. #659 

I'll follow up with changing things towards this https://github.com/jupyter/nbconvert/issues/659#issuecomment-485063446, though the reason that this seems more complex than the suggestion there was I was trying to maintain the current behavior in as many cases as possible.

In particular, the test cases seemed to fail with a "The handle is invalid" error if exec_reply is polled too late, but I'll still give a shot at your suggestion and let you know how it goes.

I'm busy today but might try to play with this Sunday or when I get back to work on Monday.

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -537,7 +537,6 @@ class ExecutePreprocessor(Preprocessor):
         polling_exec_reply = True
 
         while more_output or polling_exec_reply:
-
             if polling_exec_reply:
                 if self._passed_deadline(deadline):
                     polling_exec_reply = False
@@ -559,7 +558,7 @@ class ExecutePreprocessor(Preprocessor):
                         continue
 
                     if self.raise_on_iopub_timeout:
-                        raise RuntimeError("Timeout waiting for IOPub output")
+                        raise TimeoutError("Timeout waiting for IOPub output")
                     else:
                         self.log.warning("Timeout waiting for IOPub output")
                         more_output = False
@@ -568,11 +567,11 @@ class ExecutePreprocessor(Preprocessor):
                 # not an output from our execution
                 continue
 
-            # Will raise CellExecutionComplete when completed
             try:
+                # Will raise CellExecutionComplete when completed
                 self.process_message(msg, cell, cell_index)
             except CellExecutionComplete:
-                break
+                more_output = False
 
         # Return cell.outputs still for backwards compatability
         return exec_reply, cell.outputs
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -534,7 +534,16 @@ class ExecutePreprocessor(Preprocessor):
         cell.outputs = []
         self.clear_before_next_output = False
 
+        # This loop resolves #659. By polling iopub_channel's and shell_channel's
+        # output we avoid dropping output and important signals (like idle) from
+        # iopub_channel. Prior to this change, iopub_channel wasn't polled until
+        # after exec_reply was obtained from shell_channel, leading to the
+        # aforementioned dropped data.
+
+        # These two variables are used to track what still needs polling:
+        # more_output=true => continue to poll the iopub_channel
         more_output = True
+        # polling_exec_reply=true => continue to poll the shell_channel
         polling_exec_reply = True
 
         while more_output or polling_exec_reply:
@@ -543,6 +552,8 @@ class ExecutePreprocessor(Preprocessor):
                     polling_exec_reply = False
                     continue
 
+                # Avoid exceeding the execution timeout (deadline), but stop
+                # after at most 1s so we can poll output from iopub_channel.
                 timeout = self._timeout_with_deadline(1, deadline)
                 exec_reply = self._poll_for_reply(parent_msg_id, cell, timeout)
                 if exec_reply is not None:
@@ -552,10 +563,14 @@ class ExecutePreprocessor(Preprocessor):
                 try:
                     timeout = self.iopub_timeout
                     if polling_exec_reply:
+                        # Avoid exceeding the execution timeout (deadline) while
+                        # polling for output.
                         timeout = self._timeout_with_deadline(timeout, deadline)
                     msg = self.kc.iopub_channel.get_msg(timeout=timeout)
                 except Empty:
                     if polling_exec_reply:
+                        # Still waiting for execution to finish so we expect that
+                        # output may not always be produced yet.
                         continue
 
                     if self.raise_on_iopub_timeout:
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -508,8 +508,8 @@ class ExecutePreprocessor(Preprocessor):
                 continue
 
     def _timeout_with_deadline(self, timeout, deadline):
-        if deadline is not None and deadline - time.time() < timeout:
-            timeout = deadline - time.time()
+        if deadline is not None and deadline - time.monotonic() < timeout:
+            timeout = deadline - time.monotonic()
 
         if timeout < 0:
             timeout = 0
@@ -517,7 +517,7 @@ class ExecutePreprocessor(Preprocessor):
         return timeout
 
     def _passed_deadline(self, deadline):
-        if deadline is not None and deadline - time.time() <= 0:
+        if deadline is not None and deadline - time.monotonic() <= 0:
             self._handle_timeout()
             return True
         return False
@@ -527,7 +527,7 @@ class ExecutePreprocessor(Preprocessor):
         self.log.debug("Executing cell:\n%s", cell.source)
         exec_timeout = self._get_timeout(cell)
         if exec_timeout is not None:
-            deadline = time.time() + exec_timeout # verify this is correct. (monotonic)
+            deadline = time.monotonic() + exec_timeout
 
         cell.outputs = []
         self.clear_before_next_output = False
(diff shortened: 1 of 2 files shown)


djherbis:
I seem to be running into some flakiness with the tests, everything looks to pass on my machine, but travis is randomly pass/fail for a few diff methods, but it does look related to my changes possibly (possible the test is inherently racy too?)

MSeal:
So the python 3 failure you're seeing is some race that only occurs on Travis. I've had to restart many Travis Builds to avoid it. I haven't been able to reproduce the race on my local machine yet to figure out why the kernel is dieing. Might be related to resources on travis or slow disks not responding before a timeout? For your PR we'll need to look closely at each Kernel Timeout since we're touching that logic, but likely if it passes on a restart build action it's a pre-existing test issue.

For python 2 you have some 2/3 compatibility issues in the code you'll need to resolve.

djherbis:
@MSeal I merged in latest and fixed the compat issues, what would you like next?

djherbis:
Gah, more failures, I'll fix them

nbconvert/preprocessors/execute.py
```diff
@@ -448,6 +452,37 @@ def _update_display_id(self, display_id, msg):
...
+
+        return timeout
+
+    def _handle_timeout(self):
+        self.log.error(
+            "Timeout waiting for execute reply (%is)." % self.timeout)
+        if self.interrupt_on_timeout:
+            self.log.error("Interrupting kernel")
+            self.km.interrupt_kernel()
+            return
```
MSeal: Don't need this return

MSeal:
Thanks a ton for getting the PR updated @djherbis -- and sorry for all the merge conflicts with the 5.5 release stuff.

I think the next steps are making some tests that mock the behavioral change. Specifically where we pass the deadline in both the exec_reply and get_msgs code paths, for both exception raising an non-exception raising cases.

MSeal:
Figured out that random test failure btw, it's unrelated to your changes.

Let me know if you need any help with getting tests added.

djherbis:
@MSeal running into a little bit of pain for tests, I was currently adding them under test_execute.py::TestRunCell. Main issues seems to be with controlling the timeout in there.

(no way to introduce a delay at the moment, and using timeout=0 gets converted to timeout=None internally).

Should I be adding these tests under test_execute.py::TestExecute::test_run_notebooks, or somewhere else?

MSeal:
Yea under TestRunCell is probably right. The way to achieve this test is to mock the `kc.shell_channel.get_msg` and `kc.iopub_channel.get_msg` for each pattern respectfully, which `prepare_cell_mocks` does for message receiving. You can replace those mocks responses via `preprocessor.kc.shell_channel.return_value = ...` or `preprocessor.kc.iopub_channel.side_effect = [...]`.

If you need help getting those test mocks setup I can probably help push a starting point onto your PR branch later today.

djherbis:
I tried directly mocking the get_msg calls for some exceptional cases, let me know if this is an alright way to accomplish it.

djherbis:
Thanks that makes more sense, wasn't super clear to me how to use the mock that way.

Anything else left?

nbconvert/preprocessors/execute.py
```diff
@@ -473,49 +501,77 @@ def _wait_for_reply(self, msg_id, cell=None):
...
+                    continue
+
+                timeout = self._timeout_with_deadline(1, deadline)
+                exec_reply = self._poll_for_reply(parent_msg_id, cell, timeout)
+                if exec_reply is not None:
+                    polling_exec_reply = False
+
+            if more_output:
+                try:
+                    timeout = self.iopub_timeout
```
MSeal: So now with all the functionality in place, we likely need to address the fact that `iopub_timeout` no longer acts as it used to. In particular, since we aren't waiting for exec reply to fully complete we'll expect to timeout here often. Which means that `raise_on_iopub_timeout` will fire when users wouldn't expect.

@mpacer I think we should perhaps remove `raise_on_iopub_timeout` and `iopub_timeout` with this change and use `timeout = self._timeout_with_deadline(1, deadline)` for both conditional branches. Those other configurations just don't make any sense any more and are problematic if defined as they were in earlier versions. What's your thoughts? Even if we do a less severe change we'd need to address the change in timeout behavior somehow with this PR merging.

djherbis: Actually I tried to preserve that behavior, it won't hit the raise_on_iopub_timeout unless polling_exec_reply is false, and it's only false if we've passed the deadline or if exec_reply is not None (completed).

If we've passed the deadline we hit handle_timeout, which will either interrupt or throw. If it throws it won't reach this section, if it interrupts I suspect we would get an idle message?

Let me know if I got anything wrong here.

MSeal: Ahh yes you're correct, my end-of-conference brain power failed to read properly. Then maybe we should deprecate those fields in the next release (5.5.1) so they can be removed in 6.0? That would help cleanup this code and make it simpler to control in the future imo.

We should add a test for this case if it's not covered yet to ensure we won't timeout unexpectantly. That might be tricky to write with Mocks, so I can probably make that test if it's not clear how to set that up. After that I think I'm good to merge once @mpacer approves / has any further concerns resolved.

djherbis: Hey sorry, I was also gone for a week long work even this week.
Looks like I've got more merge conflicts to fix, I'll take a look Monday.

MSeal: Sorry there's been so many merge conflicts. I've been more careful to review / get more opinions on this PR before merging. Shouldn't be too many more conflicts. I asked @mpacer in person to take a look again when she's available so we don't keep this PR open for much longer.

djherbis: Yeah I understand, I think I managed to merge it in correctly now.

mpacer: I am not quite following the reasoning behind why the fields need to be deprecated. Could you lay out the situation where these fields cease to make sense?

nbconvert/preprocessors/execute.py
```diff
@@ -474,49 +502,77 @@ def _wait_for_reply(self, msg_id, cell=None):
...
-            except Empty:
-                self.log.warning("Timeout waiting for IOPub output")
-                if self.raise_on_iopub_timeout:
-                    raise RuntimeError("Timeout waiting for IOPub output")
-                else:
-                    break
+        more_output = True
+        polling_exec_reply = True
+
+        while more_output or polling_exec_reply:
```
mpacer: `more_output` being used as a signaling mechanism feels like a pretty significant piece of internal logic. Could you add a comment explaining why we're taking this approach?

I would think a description about the intention of the separate roles of `more_output` and `polling_exec_reply` would be enough to address the curious individual.

nbconvert/preprocessors/tests/test_execute.py
```diff
@@ -510,6 +515,65 @@ def test_busy_message(self, preprocessor, cell_mock, message_mock):
         # Ensure no outputs were generated
         assert cell_mock.outputs == []
 
+    @prepare_cell_mocks({
```
mpacer: These all look great!

nbconvert/preprocessors/execute.py
djherbis: Great point, I added some detailed comments around this section to explain why it's written this way. Let me know if it's still not clear for future readers.

nbconvert/preprocessors/tests/test_execute.py
djherbis: All the tests are great work by @MSeal (thank you!) it's my first contribution so I was unfamiliar with how to use these test frameworks.

nbconvert/preprocessors/execute.py
djherbis: @MSeal probably has a better answer than me, my interpretation was that iopub_timeout exists to give a small window after receiving exec_reply (so we know all output should be produced) for the output to reach the iopub_channel (in case of network latency etc.). I think @MSeal might feel that since we are polling for this output more regularly (and not just after exec_reply) that the purpose of this flag has changed somewhat. Perhaps I shouldn't be using iopub_timeout at all unless exec_reply has finished, and use a constant timeout / deadline similar to _poll_for_reply. I could be misinterpreting @MSeal though so please correct me!

Even with all of that, I still think iopub_timeout is relevant post exec_reply being received, since you may still need to wait for output produced near the end of execution to be received.

MSeal: That's pretty much on point. It's odd the to have a post exec_reply response specific timeout in addition to an overall timeout when the two channels are being consumed in parallel.

I'll start a separate PR / Issue on the field independent of this PR to avoid leaving this open longer.

djherbis: @MSeal Thanks, what do you think about using a constant (like 1s) insead of the iopub_timeout in this PR for the poll to iopub_channel while polling_exec_reply=True, and then switching to iopub_timeout once exec_reply is received?

MSeal: Possibly that would be a way to approach it. In general I'm saying I think making it simpler and having 1 timeout would reduce confusion and be less nuanced.

MSeal:
@djherbis Ty so much for going through all the bumps with the PR. Really appreciate you driving the change despite the time it took.

meeseeksmachine:

PR #998, opened 2019-04-24
Fixup kernel death detection

mpacer:
This PR addresses issues that came up in discussion with MSeal around how we're testing kernel deaths (inconsistency between my machine and his + travis' on one test).

This unveiled a broader issue with how we were detecting the death of kernels.


nbconvert/preprocessors/execute.py
```diff
@@ -24,6 +24,8 @@
 from .base import Preprocessor
 from ..utils.exceptions import ConversionException
 
+class KernelIsDead(RuntimeError):
```
MSeal: `KernelDied` or `KernelDiedError`

PR #1000, opened 2019-04-25
Fix pdf exporter inheriting the wrong file extension

### LINKED ISSUE #999: Pdf exporter thinks it outputs to .tex
I went to look at fixing #938, and I found this:
image

I think this is due to inheriting the file extension from the latex exporter.

t-makaro:
Fixes #999


t-makaro:
Tests are passing, but I think this change makes the latex exporter write the .tex file as a .pdf file instead (which the latex compiler reads as text just fine).

MSeal:
Yep. Tested that the conversion was ok post change.

Issue #938, opened 2019-01-03
Duplicate Entries in Download as Menu


hansen-m:
After updating to the latest version of nbconvert I'm seeing duplicate entries for the exporters in the Download as menu. Any way I can remove or manage these? I'm running JupyterHub via Anaconda 3. 

<img width="339" alt="screen shot 2019-01-03 at 13 45 11" src="https://user-images.githubusercontent.com/1102305/50655552-8e48f800-0f5e-11e9-96f4-8238793baaa3.png">

justmarkham:
I experienced the same behavior when I upgraded from nbconvert 5.3.1 to 5.4 (and that is the only thing I changed in my environment).

t-makaro:
Dev note: Jupyter notebook added support for custom exporters in the menu [here](https://github.com/jupyter/nbconvert/pull/759). This added a ```export _from_notebook``` to allow custom exporters to be added to the menu. #759 Documented that change in nbconvert, and #864 added ```export_from_notebook``` to the built-in exporters (which then resulted in the built-in exporters being duplicated in the menu). I believe. This could be fixed, by reverting #864, or on the notebooks side.

MSeal:
Hmm @t-makaro is right about the cause, I'm not sure what the intended pattern for UIs is supposed to be here. @mpacer @SylvainCorlay do either of you know those expectations better / can suggest the best change to make here?

mpacer:
So this seems to be an issue on the notebook side rather than on nbconvert's side. Notebook has hardcoded values back in the days when nbconvert didn't make these as easily visible.


The template that has the hardcoding is here:

https://github.com/jupyter/notebook/blob/eea37a6b2110e0b8a5491f17e241fa3c8aecb5f7/notebook/templates/notebook.html#L113-L126

The logic for getting exporters lives in `get_custom_frontend_exporters` here: 

https://github.com/jupyter/notebook/blob/b8b66332e2023e83d2ee04f83d8814f567e01a4e/notebook/notebook/handlers.py#L18-L28

The jinja rendering that populates that template with the exporter names and display names is in the same file as the `get_custom_frontend_exporters` definition: 

https://github.com/jupyter/notebook/blob/b8b66332e2023e83d2ee04f83d8814f567e01a4e/notebook/notebook/handlers.py#L59

mpacer:
[...]

PR #1001, opened 2019-04-25
fix export_from_notebook names

t-makaro:
When looking at #938, I noticed that the template exporter has ```export_from_notebook``` defined which means it appears in the notebook's download as menu, and the html exporter was inheriting this.


t-makaro:
I don't know why the test is failed. Only one test failed on only one version of python, and it seems unrelated to my change. It also passes on my machine with python 3.6.

MSeal:
Test failure was a red harring. I tried a few conversion options and checked that nothing obvious was affected. I'll plop this in the release so the notebooks PR can merge post release.

PR #995, opened 2019-04-23
Release 5.5 Notes

MSeal:
This has all the change notes, along with a few fixes to release instructions. Please review and I will rebase this PR until everything is ready for the release.

PR #1002, opened 2019-04-25
fix export_form_notebook in script exporter

t-makaro:
**CC** @MSeal 

Quick. Found another one.


mpacer:
This looks good @MSeal this should be added to #995.

Issue #755, opened 2018-02-06
Support for custom mime-renderers


Madhu94:
From https://github.com/jupyter/nbconvert/issues/481, I don't think custom mime types are supported in nbviewer and you should send an additional payload of ``text/html`` or some standard mime to get outputs to render on nbviewer.

Could we not do something similar to what was done for embedding widgets ? Authors can publish their renderers on npm/unpkg and nbviewer can ``require`` those.   The name of the renderer module (say ``MyRenderer`` has to be stored in the notebook somewhere, we require that module and call a known method on it (say ``MyRenderer.render_mime``).

What do you think ?

https://github.com/jupyter/nbviewer/issues/263 is similar, but deals with the generic set of nbextensions and not just mime types.

mpacer:
Hrm, I have vague concerns about the security implications of this… but I am not able to articulate them well. 

@Carreau @rgbkrk @minrk do you see any obvious issue with this?

minrk:
There are security implications, depending on where the HTML is served. Not for nbviewer itself, because the nbviewer domain by design shouldn't have access to anything interesting (nbviewer hosted behind JupyterHub would be a very different story). Just like HTML/js output, this should be fine as long as there's a way for particular users to disable and/or whitelist renderers. I'm not sure what the defaults should be.

Madhu94:
> the nbviewer domain by design shouldn't have access to anything interesting 
Yes, makes sense. But don't the same security concerns apply for embedding widgets? The default is unpkg.com and I think even thats' configurable now.

minrk:
Yes, 100%. This ought to be in the same class as the custom widgets support.

rgbkrk:
I think there's two distinctions for custom mimerenderers here -- those that should be supported by nbconvert by default that are now default in jupyterlab and nteract (vdom, vega), and those which would require vetting.

Madhu94:
As I see it, the benefits are -  no need to save duplicate data in the notebook and no need to bother with a fallback to ``<script>`` tags and html 

> those which would require vetting.

Do you think loading renderers would be useful if this was configurable with a whitelist as proposed by @minrk ?

mpacer:
We also should consider if/how this should apply to non-html output. For example, for LaTeX we have a longstanding concern that we have no good way to convert raw-html to LaTeX; would this apply to the ePub exporter(which I believe we were going to base on the html exporter); how should it play with slides? 

[...]

PR #1005, opened 2019-04-30
WIP - Lab template

SylvainCorlay:
This PR adds an nbconvert template producing the same DOM structure as JupyterLab.

## Templates

The goal is to enable the styling of HTML produced with nbconvert with JupyterLab's themes.

- `basic.tpl` contains the main jinja2 block producing the JLab DOM structure.
- `full.tpl` includes the CSS. The CSS is generated with the https://github.com/SylvainCorlay/labcss_bundler JS project.

## Pygments theme

Besides, I am working on a `pygments` theme using the JupyterLab CSS variables used by the Jupyter CodeMirror theme.

For reference, the pygments stylesheet that produces the same colors as JupyterLab. (This pygments style requires the [Pygments PR  #814](https://bitbucket.org/birkenfeld/pygments-main/pull-requests/814/))

<details>

```python
from pygments.style import Style
from pygments.token import Keyword, Name, Comment, String, Error, Number, Operator, Generic, Whitespace

class JupyterStyle(Style):
    """
    A Jupyter style for pygments

  --jp-mirror-editor-atom-color
  --jp-mirror-editor-def-color
  --jp-mirror-editor-variable-color
  --jp-mirror-editor-variable-2-color
  --jp-mirror-editor-variable-3-color
  --jp-mirror-editor-punctuation-color
  --jp-mirror-editor-property-color
  --jp-mirror-editor-string-2-color
  --jp-mirror-editor-meta-color
  --jp-mirror-editor-qualifier-color
  --jp-mirror-editor-builtin-color
  --jp-mirror-editor-bracket-color
  --jp-mirror-editor-tag-color
  --jp-mirror-editor-attribute-color
  --jp-mirror-editor-header-color
  --jp-mirror-editor-quote-color
  --jp-mirror-editor-link-color
  --jp-mirror-editor-hr-color
    """

    default_style = ''
    background_color = 'var(--jp-cell-editor-background)'
    highlight_color = 'var(--jp-cell-editor-active-background)'

    styles = {
        Comment:                   'var(--jp-mirror-editor-comment-color)',
        Keyword:                   'bold var(--jp-mirror-editor-keyword-color)',
        Operator:                  'var(--jp-mirror-editor-operator-color)',
        String:                    'var(--jp-mirror-editor-string-color)',
        Number:                    'bold var(--jp-mirror-editor-number-color)',
        Error:                     'var(--jp-mirror-editor-error-color)',
        Name:                      '',
    }
```
</details>

## Light theme:

Light

## Dark theme:

Dark

cc @mpacer @MSeal @maartenbreddels @ivanov @ellisonbg


t-makaro:
Just a note on highlight style. I believe the style is limited by our use of the pygments lexers. I made a style over at [nb_pdf_template](https://github.com/t-makaro/nb_pdf_template/blob/master/nb_pdf_template/jupyter_python.py) that mimics the notebook highlight scheme the best it can.

SylvainCorlay:
Thanks for the pointer @t-makaro.  Another goal of my PR is to use jupyterlab's CSS variable so that we can switch themes dynamically and nbconverted HTML can be used within JupyterLab while being consistent with the current Jupyterlab Theme.

SylvainCorlay:
For reference, opened a PR to pygments to enable the use of CSS variables in pygments stylesheets.

SylvainCorlay:
Here is the pygments stylesheet.

```
from pygments.style import Style
from pygments.token import Keyword, Name, Comment, String, Error, \
     Number, Operator, Generic, Whitespace


class JupyterStyle(Style):
    """
    A Jupyter style for pygments

  --jp-mirror-editor-atom-color
  --jp-mirror-editor-def-color
  --jp-mirror-editor-variable-color
  --jp-mirror-editor-variable-2-color
  --jp-mirror-editor-variable-3-color
  --jp-mirror-editor-punctuation-color
  --jp-mirror-editor-property-color
  --jp-mirror-editor-string-2-color
  --jp-mirror-editor-meta-color
  --jp-mirror-editor-qualifier-color
  --jp-mirror-editor-builtin-color
  --jp-mirror-editor-bracket-color
  --jp-mirror-editor-tag-color
  --jp-mirror-editor-attribute-color
  --jp-mirror-editor-header-color
  --jp-mirror-editor-quote-color
  --jp-mirror-editor-link-color
  --jp-mirror-editor-hr-color
    """

    default_style = ""

    styles = {
        Comment:                   "var(--jp-mirror-editor-comment-color)",
        Keyword:                   "bold var(--jp-mirror-editor-keyword-color)",
        Operator:                  "var(--jp-mirror-editor-operator-color)",
        String:                    "var(--jp-mirror-editor-string-color)",
        Number:                    "bold var(--jp-mirror-editor-number-color)",
        Error:                     "var(--jp-mirror-editor-error-color)",
        Name:                      "",
    }
```

SylvainCorlay:
cc @mpacer @MSeal @maartenbreddels 

So this appears to work (and to support JupyterLab themes) but the way to include the JupyterLab CSS is still manual in `lab_full.tpl`

At the moment, the CSS from the classic notebook (style.min.css) is included in the resources for the jinja template [here](https://github.com/jupyter/nbconvert/blob/5.5/nbconvert/preprocessors/csshtmlheader.py#L66-L75):.

Either I replace `basic.tpl` and `full.tpl` with the new JupyterLab versions, or we find a means to make these resources pluggable.

SylvainCorlay:
In the last commit, I replace the style.min.css with the one generated for jlab. This can easily be tested now. This does not include the new pygment theme yet.

@ellisonbg you might be interested.

SylvainCorlay:
The `style.min.css` CSS bundle that I am using was produced by a webpack bundle. The bundler package may be added to jupyterlab and the bundle uploaded to unpkg.

SylvainCorlay:
The test failure is due to nbconvert test checking that `Out[` is not in the generated HTML but it is in the JupyterLab CSS.

SylvainCorlay:
https://github.com/SylvainCorlay/labcss_bundler

SylvainCorlay:
Hey @MSeal @mpacer I would love to hear from you on how I should move forward with this.

MSeal:
To compare the styles from before / after.

lab_theme

I'm not sure what the stance has been for style of nbconvert (in particular the default style) as that was established a bit before I joined. So I think I'd want opinions from devs that have more context on style. It might be surprising in classic notebook to not see the same style exported as you saw on screen for example.

@minrk @willingc I'm not sure if either of you had opinions on the topic or know how non-bug fix style decisions in nbconvert have been made in the past?

MSeal:
Codewise the changes look solid.

SylvainCorlay:
> To compare the styles from before / after.

Quick note: on your screenshot, "after" is on the left.

nbconvert/tests/test_nbconvertapp.py
```diff
@@ -311,7 +311,6 @@ def test_no_prompt(self):
             with open("notebook1.html",'r') as f:
                 text = f.read()
                 assert "In&nbsp;[" not in text
-                assert "Out[" not in text
```
SylvainCorlay: `Out[` is in fact in the jlab CSS bundle, so I had to remove that test.

ivanov:
This is looking great, @SylvainCorlay, thank you! Can we copy these files to make a new template so we keep the default as it is right now?

akhmerov:
Looks great!

Any idea if it would be possible to hook in the jupyterlab mime rendering machinery? That'd be the only way to display custom rich output iiuc (extensions aren't supposed to render themselves anymore).

SylvainCorlay:
> This is looking great, @SylvainCorlay, thank you! Can we copy these files to make a new template so we keep the default as it is right now?

Thanks. There is a small obstacle which is that beyond the jinja template files, the defaut `HTMLExporter` [provides the CSS of the classic notebook](https://github.com/jupyter/nbconvert/blob/master/nbconvert/preprocessors/csshtmlheader.py#L39-L92), and this cannot be super-easily overriden. Ideally, the resources required for a template should be specified as part of the template...

I am actually thinking of using a custom HTML exporter in `voila` so that we can specify such resources in the template directory. The voila configuration directory approach could be upstreamed into nbconvert.

SylvainCorlay:
> Looks great!
> 
> Any idea if it would be possible to hook in the jupyterlab mime rendering machinery? That'd be the only way to display custom rich output iiuc (extensions aren't supposed to render themselves anymore).
>

Actually, we already support rendering live widgets in nbconverted notebooks (see e.g. [here](https://nbviewer.jupyter.org/github/jupyter-widgets/ipywidgets/blob/master/docs/source/examples/Widget%20List.ipynb) or in the ipywidgets documentation, using the rst nbconvert exporter of nbsphinx) but this does not use generic rendermime. This is definitely something we should do, but it is somewhat orthogonal to the choice of template.

bollwyvl:
Re: mimerenderers: this is greatly desired for me, but at that point, I'd
just want all (or as much as I want) of lab... but definitely out of scope
of this PR to get us to the next level!

This is totally possible:

https://deathbeds.github.io/jyve/lab/
(Implementation
https://github.com/deathbeds/jyve/blob/master/jyve/exporter.py)

The number of patches is smaller than one would imagine, mainly around
session management and contents warnings.

I've been tinkering with the voila stack to make it work there, there but
no demo yet... Lots of url mangling will be required.

The biggest challenge would be in documenting and then building with which
extensions you want baked in... I guess I'm thinking each deployed lab
would be specified with a config file (or a "master" notebook) which could
specify extensions and a workspace.

Looking further out: for an nbviewer formatter this gets a little crazy....
We'd basically have to offer a requirejs escape hatch.

akhmerov:
Got it, sorry for the offtopic. I also now found #755, which is the correct place to track plans for supporting mimerender.

SylvainCorlay:
I opened https://github.com/jupyterlab/jupyterlab/pull/6314 into JupyterLab to produce the CSS bundle. As soon as the package is released, I will be able to reference the `unpkg` URL in setup.py instead of the current one.

SylvainCorlay:
@bollwyvl thanks for all the pointers! I was completely unaware of Jive. I would love to hear more about it!

SylvainCorlay:
> I opened jupyterlab/jupyterlab#6314 into JupyterLab to produce the CSS bundle. As soon as the package is released, I will be able to reference the unpkg URL in setup.py instead of the current one.

The JupyterLab CSS bundler has been merged and published to NPM. I have updated this PR to reflect this change. I think this is getting close to a mergeable state.

SylvainCorlay:
Note: you can now select the dark theme with

```
jupyter nbconvert --to html --CSSHTMLHeaderPreprocessor.theme='dark' foobar.ipynb
```

minrk:
@MSeal requested historical perspective: nbconvert HTML style has been defined as "what the notebook does" which is why we bundle the notebook css in nbconvert, and *intend* to reproduce the DOM of the live notebook.

In that context, with JupyterLab as a peer to nbclassic and its intended replacement, it makes sense to me to add jupyterlab as an output style, and even in the future replace the nbclassic style with jupyterlab as the default, especially if the differences prove uncontroversial.

MSeal:
Thanks for the context.

@SylvainCorlay could you preserve the default as the classic look and make lab an adjacent template then? We'd likely need classic template available even if the default changes in the future.

nbconvert/preprocessors/csshtmlheader.py
```diff
@@ -66,16 +70,69 @@ def _generate_header(self, resources):
...
+        )
+        
+        # Load style theme file.
+        with io.open(theme_filename, encoding='utf-8') as f:
+            header.append(f.read())
+
         # Add pygments CSS
-        formatter = HtmlFormatter(style=self.style)
-        pygments_css = formatter.get_style_defs(self.highlight_class)
+        # formatter = HtmlFormatter(style=self.style)
```
t-makaro: I know you are working on allowing pygments styles to include css, but I just want to emphasize that this should be changed back before merging because this breaks customizing highlighting style. Also, I don't think we can make the new style the default style without bumping our minimum supported pygments version (assuming that your PR over at pygments gets merged). If pygments styles can be made to include css, then it would be nice to have a style for the classic notebook too. If we separate the lab and classic notebook styles into different templates, we could write some logic to use the respective styles by default. (and checking if ```self.style=='default'``` to still allow overriding).

SylvainCorlay: Don't worry. I doubt this becomes the default template right away - and the intent is very much to keep the ability to select a pygments theme.

SylvainCorlay:
> @SylvainCorlay could you preserve the default as the classic look and make lab an adjacent template then? We'd likely need classic template available even if the default changes in the future.

This may not be possible merely as an "alternative" jinja template for the HTML exporter, but only as a completely new template exporter (to both change the way the CSS header is composed and the jinja template).

What I am thinking of is to enable the sort of template configuration that we have in voila, where a template is merely a directory in a conf.d, holding all the required resources.

nbconvert/preprocessors/csshtmlheader.py
SylvainCorlay: Quick update: the pygments [pull request](https://bitbucket.org/birkenfeld/pygments-main/pull-requests/814/allow-for-css-variable-in-pygments/) just got in.

Unfortunately it did not make it for pygments 2.4 but should be in 2.4.1 in a couple of weeks.

SylvainCorlay: @t-makaro pygments 2.4.1 was released including my patch.

This PR now makes use of the `style` attribute again, which can be a string (as before) or a type inheriting from `pygments.style.Style`.

SylvainCorlay: Resolving that part of the review.

MSeal:
That could work. I'd say use best judgement on how to achieve something mergable in this PR. If we need to hack the exporter a bit to avoid clobbering the current default view I'm ok with that if we have a longer term approach (this could be a change for 6.0?), but we do need a path to preserve the classic export look-n-feel as default for now imo.

SylvainCorlay:
> That could work. I'd say use best judgement on how to achieve something mergable in this PR. If we need to hack the exporter a bit to avoid clobbering the current default view I'm ok with that if we have a longer term approach (this could be a change for 6.0?), but we do need a path to preserve the classic export look-n-feel as default for now imo.

Absolutely. Let's hold this up for a bit more and see if we come up with a nicer approach than a completely new template exporter without too much change.

SylvainCorlay:
Closing as this was included through another PR.

PR #1008, opened 2019-05-01
Update svg2pdf.py

### LINKED ISSUE #302: raise TypeError("Inkscape svg to pdf conversion failed")
Not sure why I can not use jupyter nbconvert to get a pdf... and why there is a reference to a stand alone application Inkscape? I need to share my notebooks with collaborators in and easy way. Those folks think advanced programming is opening Excel. I can not save notebooks larger than 50mb, which eliminates all graphical output. So, I need a turnkey solution to output the entire NB, graphics and all, and send it to the other folks. What is the EASY-SIMPLE-SOLUTION_WORKABLE to share NBs that have embedded graphics and thus are 100mb or more?

terminal-w:
updated Inkscape Registry Location to:
`SOFTWARE\\WOW6432Node\\Microsoft\\Windows\\CurrentVersion\\App Paths\\inkscape.exe`
Which is consistent with current versions of Inkscape.

This combined with the current master version of `extractoutput.py` fixes #302 for me preferably requires testing on other Windows setups before being marked mature for release.

### DIFF
--- a/nbconvert/preprocessors/svg2pdf.py
+++ b/nbconvert/preprocessors/svg2pdf.py
@@ -1,4 +1,4 @@
-"""Module containing a preprocessor that converts outputs in the notebook from 
+"""Module containing a preprocessor that converts outputs in the notebook from
 one format to another.
 """
 
@@ -42,19 +42,20 @@ class SVG2PDFPreprocessor(ConvertFiguresPreprocessor):
 
     command = Unicode(
         help="""The command to use for converting SVG to PDF
-        
+
         This string is a template, which will be formatted with the keys
         to_filename and from_filename.
-        
+
         The conversion call must read the SVG from {from_flename},
         and write a PDF to {to_filename}.
         """).tag(config=True)
 
     @default('command')
     def _command_default(self):
+        print(self.inkscape)
         return self.inkscape + \
                ' --without-gui --export-pdf="{to_filename}" "{from_filename}"'
-    
+
     inkscape = Unicode(help="The path to Inkscape, if necessary").tag(config=True)
     @default('inkscape')
     def _inkscape_default(self):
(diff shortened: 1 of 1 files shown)


takluyver:
It should probably check two different registry paths, because the `WOW6432Node` is for 32-bit applications installed on 64-bit Windows. If someone installs it on 32-bit Windows, or a 64-bit build of Inkscape, it probably won't use that path. Unfortunately I can't tell you what the correct path should be in those cases.

juhasch:
I aggree, both paths should be checked. On my work Laptop it should be the same path, just without the "WOW6432Node" part.

nbconvert/preprocessors/svg2pdf.py
```diff
@@ -42,19 +42,20 @@ def _to_format_default(self):
...
         to_filename and from_filename.
-        
+
         The conversion call must read the SVG from {from_flename},
         and write a PDF to {to_filename}.
         """).tag(config=True)
 
     @default('command')
     def _command_default(self):
+        print(self.inkscape)
```
MSeal: Left-over print

nbconvert/preprocessors/svg2pdf.py
```diff
@@ -64,8 +65,9 @@ def _inkscape_default(self):
         if sys.platform == "win32":
             wr_handle = winreg.ConnectRegistry(None, winreg.HKEY_LOCAL_MACHINE)
             try:
-                rkey = winreg.OpenKey(wr_handle, "SOFTWARE\\Classes\\inkscape.svg\\DefaultIcon")
+                rkey = winreg.OpenKey(wr_handle, "SOFTWARE\\WOW6432Node\\Microsoft\\Windows\\CurrentVersion\\App Paths\\inkscape.exe")
```
MSeal: Add a `rkey = winreg.OpenKey(wr_handle, "SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\App Paths\\inkscape.exe")` to the except section and retry so we can fallback for older microsoft distros.

willingc:
Adding to 6.0 milestone.

nbconvert/preprocessors/svg2pdf.py
t-makaro: Checking my system: my registry key is at:
```Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Classes\inkscape.svg\DefaultIcon```
So the original line is needed for me.

MSeal:
Closing as #1115 resolved the issue

PR #1017, opened 2019-05-07
Disable IPython History in executing preprocessor

alexrudy:
IPython has a feature where command history is written to a SQLite database. When running multiple kernels in parallel, this feature runs into problems, as the SQLite database will be open from several processes, and sometimes hits a locking error.

To fix this, we have disabled SQLite history in IPython Kernels launched from the preprocessing executor, as recommended by ipython/ipython#11460.

This issue was originally brought up in the context of multiprocessing (see nteract/papermill#239) and this PR partially solves that problem, and adds a test to prove that two notebook kernels, *started from the same process* can be run simultaneously.

I am working on a separate PR to test and maintain multiprocessing support, as there is at least one more issue to be solved.

Thanks @MSeal for all of the help and guidance during PyCon sprints to figure this out.

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -220,6 +220,20 @@ class ExecutePreprocessor(Preprocessor):
             )
     ).tag(config=True)
 
+    ipython_hist_file = Unicode(
+        default_value=':memory:',
+        help="""Path to file to use for SQLite history database for an IPython kernel.
+        
+        The specific value `:memory:` (including the colon
+        at both end but not the back ticks), avoids creating a history file. Otherwise, IPython
+        will create a history file for each kernel. 
+        
+        When running kernels simultaneously (e.g. via multiprocessing) saving history a single
+        SQLite file can result in database errors, so using `:memory:` is recommended in non-interactive
+        contexts.
+        
+        """).tag(config=True)
+
     kernel_manager_class = Type(
         config=True,
         help='The kernel manager class to use.'
@@ -268,6 +282,8 @@ class ExecutePreprocessor(Preprocessor):
                 'kernelspec', {}).get('name', 'python')
         km = self.kernel_manager_class(kernel_name=self.kernel_name,
                                        config=self.config)
+        if km.ipykernel and self.ipython_hist_file:
+            self.extra_arguments += ['--HistoryManager.hist_file={}'.format(self.ipython_hist_file)]
         km.start_kernel(extra_arguments=self.extra_arguments, **kwargs)
 
         kc = km.client()
--- a/nbconvert/preprocessors/tests/files/Parallel Execute.ipynb
+++ b/nbconvert/preprocessors/tests/files/Parallel Execute.ipynb
@@ -1,6 +1,18 @@
 {
  "cells": [
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Ensure notebooks can execute in parallel\n",
+    "\n",
+    "This notebook uses a file system based \"lock\" to assert that two instances of the notebook kernel will run in parallel. Each instance writes to a file in a temporary directory, and then tries to read the other file from\n",
+    "the temporary directory, so that running them in sequence will fail, but running them in parallel will succed.\n",
+    "\n",
+    "Two notebooks are launched, each with an injected cell which sets the `this_notebook` variable. One notebook is set to `this_notebook = 'A'` and the other `this_notebook = 'B'`."
+   ]
+  },
+  {
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
@@ -18,6 +30,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "# the variable this_notebook is injectected in a cell above by the test framework.\n",
     "other_notebook = {'A':'B', 'B':'A'}[this_notebook]\n",
     "directory = os.environ['NBEXECUTE_TEST_PARALLEL_TMPDIR']\n",
     "with open(os.path.join(directory, 'test_file_{}.txt'.format(this_notebook)), 'w') as f:\n",
@@ -44,20 +57,6 @@
     "else:\n",
     "    assert False, \"Timed out – didn't get a message from {}\".format(other_notebook)"
    ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": []
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": []
   }
  ],
  "metadata": {
--- a/nbconvert/preprocessors/tests/files/Parallel Execute.ipynb
+++ b/nbconvert/preprocessors/tests/files/Parallel Execute.ipynb
@@ -1,0 +1,84 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "import os.path\n",
+    "import tempfile\n",
+    "import time"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "other_notebook = {'A':'B', 'B':'A'}[this_notebook]\n",
+    "directory = os.environ['NBEXECUTE_TEST_PARALLEL_TMPDIR']\n",
+    "with open(os.path.join(directory, 'test_file_{}.txt'.format(this_notebook)), 'w') as f:\n",
+    "    f.write('Hello from {}'.format(this_notebook))"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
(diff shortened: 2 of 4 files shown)


nbconvert/preprocessors/tests/test_execute.py
```diff
@@ -238,18 +250,66 @@ def assert_notebooks_equal(expected, actual):
...
+    return nb
+
+
+def test_parallel_notebooks(capfd, tmpdir):
+    """Two notebooks should be able to be run simultaneously without problems.
+    
+    The two notebooks spawned here use the filesystem to check that the other notebook
+    wrote to the filesystem."""
+
+    opts = dict(kernel_name="python")
```
MSeal: Did you end up needing to overwrite the kernel?

nbconvert/preprocessors/tests/files/Parallel Execute.ipynb
```diff
@@ -0,0 +1,84 @@
...
+    "import time"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "other_notebook = {'A':'B', 'B':'A'}[this_notebook]\n",
```
MSeal: Might be good to add a comment about how this works with the A waiting for B file and visa versa

MSeal: Also you probably want a comment that this_notebook is undefined and injected by the test

nbconvert/preprocessors/execute.py
```diff
@@ -220,6 +220,20 @@ class ExecutePreprocessor(Preprocessor):
             )
     ).tag(config=True)
 
+    ipython_hist_file = Unicode(
+        default_value=':memory:',
```
MSeal: :+1: We should have probably had this default a long time ago regardless of parallelism concerns.

nbconvert/preprocessors/tests/test_execute.py
MSeal: I meant that this link is replacing the kernel from the notebook, and was curious if the line was actually needed.

nbconvert/preprocessors/tests/files/Parallel Execute.ipynb
MSeal: Thanks

nbconvert/preprocessors/tests/test_execute.py
alexrudy: Ahh, yes, I remember discussing. It is probably unnecessary, I’ll double check.

alexrudy: Indeed, it was redundant. To better match the other tests here, I removed the kernel spec from the notebook file. Is that okay with you? It might make it run on python2, but I think this should work on both python2 and python3. Though my understanding kernel names is admittedly weak!

MSeal:
Bumped priority for getting other reviewers, but I'll merge this next week if we don't get more eyes on it.

nbconvert/preprocessors/execute.py
```diff
@@ -268,6 +282,8 @@ def start_new_kernel(self, **kwargs):
                 'kernelspec', {}).get('name', 'python')
         km = self.kernel_manager_class(kernel_name=self.kernel_name,
                                        config=self.config)
+        if km.ipykernel and self.ipython_hist_file:
```
mpacer: Is `ipykernel` an attribute of a KernelManager? It doesn't seem to be https://jupyter-client.readthedocs.io/en/stable/api/manager.html#jupyter_client.KernelManager

If not, I'd prefer to know where this is being assigned.

nbconvert/preprocessors/tests/test_execute.py
```diff
@@ -12,7 +12,10 @@
 import glob
 import io
 import os
+import logging
```
mpacer: Why are we adding logging?

nbconvert/preprocessors/tests/test_execute.py
```diff
@@ -218,6 +224,12 @@ def assert_notebooks_equal(expected, actual):
         actual_execution_count = actual_cell.get('execution_count', None)
         assert expected_execution_count == actual_execution_count
 
+def notebook_resources():
+    res = ResourcesDict()
+    res['metadata'] = ResourcesDict()
```
mpacer: Does this need to be a nested ResourcesDict()? I feel like we should have a convenience class for creating a ResourcesDict that is prepopulated a value for metadata rather than doing this.

The pattern is fine for right now, but it would be useful to have a docstring describing the purpose of this new function.

nbconvert/preprocessors/execute.py
alexrudy: It [is a property of the Kernel Manager](https://github.com/jupyter/jupyter_client/blob/master/jupyter_client/manager.py#L93) but it isn't documented. I used it to encapsulate the logic about using an `ipython` specific argument, rather than doing my own check against the kernel name before passing this argument in.

nbconvert/preprocessors/tests/test_execute.py
alexrudy: My mistake – will remove.

nbconvert/preprocessors/tests/test_execute.py
alexrudy: 🤔 I took this snippet from `test_run_all_notebooks` and extracted it. I'm not actually sure what is required in the metadata, but I didn't want to copy/paste this around the tests in this file.

I'll add a quick docstring, to the effect of "prepares a notebook resources dictionary"

nbconvert/preprocessors/tests/test_execute.py
```diff
@@ -238,18 +250,66 @@ def assert_notebooks_equal(expected, actual):
...
     input_file = os.path.join(current_dir, 'files', input_name)
-    res = ResourcesDict()
-    res['metadata'] = ResourcesDict()
-    res['metadata']['path'] = os.path.join(current_dir, 'files')
-    input_nb, output_nb = run_notebook(input_file, opts, res)
+    input_nb, output_nb = run_notebook(input_file, opts, notebook_resources())
     assert_notebooks_equal(input_nb, output_nb)
 
 
+def label_parallel_notebook(nb, label):
```
mpacer: I think this would more idiomatically be treated as a custom preprocessor. You would pass in via resources the label value and then overwrite the `preprocess` method:

https://github.com/jupyter/nbconvert/blob/8af6aac0fe9f9096c1886a90449cbd66e6a9f835/nbconvert/preprocessors/base.py#L51-L70

nbconvert/preprocessors/tests/test_execute.py
```diff
@@ -238,18 +250,66 @@ def assert_notebooks_equal(expected, actual):
...
+    """Insert a cell in a notebook which sets the variable `this_notebook` to the string `label`.
+    
+    Used for parallel testing to label two notebooks which are run simultaneously.
+    """
+    label_cell = nbformat.NotebookNode(
```
mpacer: I think this would be cleaner if it used `nbformat.v4`'s `new_code_cell` function:
https://github.com/jupyter/nbformat/blob/master/nbformat/v4/nbbase.py#L98-L110

mpacer: I think this would be cleaner if it used `nbformat.v4`'s `new_code_cell` function:
https://github.com/jupyter/nbformat/blob/11903688167d21af96c92f7f5bf0634ab51819f1/nbformat/v4/nbbase.py#L98-L110

nbconvert/preprocessors/tests/test_execute.py
alexrudy: That makes sense – do you think that is worthwhile for this set of tests?

I was hesitant to re-implement papermill's templating here, but happy to do so if that makes it more maintainable.

nbconvert/preprocessors/tests/test_execute.py
MSeal: I'm fine with the test doing what it does and not blocking a merge on test helpers. It's not critical path code and it's readable for anyone who wants to tweak it in the future imo.

nbconvert/preprocessors/tests/test_execute.py
MSeal: I share that hesitation just for testing purposes.

alexrudy:
Follow up!

I think the following are the outstanding questions from the review of this PR:

- [ ] Is it okay to use the `ipython` property of KernelManager – it is not a documented property, but seems like the best way to identify `ipython` kernels and avoid sending the history argument to other kernels.
- [ ] Do I need to refactor `notebook_resources()` in the test cases?
- [ ] Does the parallel test need to implement a full preprocessor for the test, and use that preprocessor in the notebook execution?

We've started the discussion for each of these questions above – I would love more feedback about all of them so that we can merge this PR (rather than having it get stale).

MSeal:
Sorry we've been swamped after PyCon catching up. I agree let's not let this go stale and get some merges. On that note I'll follow-up on getting a jupyter_client release for the other fix you made.

I checked the first box. I think that's fine and I've got some more permissions on that project now to help maintain it.

The other two I find maintainable enough as is. I'm going to merge now and if anyone feels strongly enough they can open an issue or PR to address it differently.

Thanks for the improvement!

meeseeksmachine:

PR #1021, opened 2019-05-11
kwargs is used when km is exist, but not defind in func.

00Kai0:


MSeal:
Seems like a helpful contract extension. Thanks.

Issue #1007, opened 2019-05-01
Backwards compatibility broken on #markdown2html()


paos:
Compatibility with fast.ai's librarys doc() method is broken. Using it now creates an error due to the latest update in markdown2html.

Commit 654d6a395650e403b41f16a235cf29f0fdb29a92 changes the method signature:

```
- def markdown2html(self, source):
+ def markdown2html(self, context, source):
```

fast.ai's librarys doc() method calls `output = HTMLExporter().markdown2html(md)`

Is it possible to make the change backwards compatible instead of introducing a mandatory extra argument in front of the old one?

erinspace:
Looking into this at the PyCon sprints!

ivanov:
The plan is to revert this backwards incompatible change and have the next release include `context` as the last argument, with a default value of `None` that will create an empty dictionary in the method body. That way we keep the older code from when this method only took one parameter of `source` working without making any changes.

tl;dr - flip it and reverse it!

maartenbreddels:
I'm not sure that works, since this method is called by jinja2, please check if the argument order matters.

maartenbreddels:
Reverting this would break voila: https://github.com/QuantStack/voila/blob/807f635d56fc741d1029dbe31b1a51841a8fb996/voila/html.py#L22

Looking at
 * https://github.com/pallets/jinja/blob/194c9d4135dba0fdd9357f9ce8b537c14c99aa36/jinja2/filters.py#L29
 * https://github.com/pallets/jinja/blob/fb7e12cce67b9849899f934e697f7e2a91d604c2/jinja2/environment.py#L440

It seems the order *does* matter. I don't think `markdown2html` should be a public API, it's a filter used for jinja2.
However, a possible solution (although ugly) would be to have the second argument default to None, and check if `is not instanceof(context, jinja2.nodes.EvalContext)`, and if so, it probably means the first argument was the source.

erinspace:
Seems like there's a bit more discussion that should take place, going to drop this and focus on another error for PyCon :)

ivanov:
[...]

PR #1022, opened 2019-05-12
Fix backwards incompatibility with markdown2html

jhamrick:
Recent changes to `html.py`'s handling of `markdown2html` also breaks nbgrader, which uses the `markdown2html` filter. The original filter only assumes the input is source code, not necessarily a cell, but the new version of the filter in this file assumes the context includes a `'cell'` field. This isn't the case for the way the filter is being used by nbgrader, but this change alleviates the problem by returning an empty dictionary by default for the cell contents.

Related to #1007

### DIFF
--- a/nbconvert/exporters/html.py
+++ b/nbconvert/exporters/html.py
@@ -71,7 +71,7 @@ class HTMLExporter(TemplateExporter):
     @contextfilter
     def markdown2html(self, context, source):
         """Markdown to HTML filter respecting the anchor_link_text setting"""
-        cell = context['cell']
+        cell = context.get('cell', {})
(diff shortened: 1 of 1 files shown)


MSeal:
That's a reasonable safeguard in the code.

MSeal:
There's a few items floating around for 5.5.1 or possible a 5.6 release in the not-too-distance future so it shouldn't linger outside a release for too long.

jhamrick:
Great, thanks!

maartenbreddels:
Hi Jessica, sorry for breaking that!

jhamrick:
No problem, it's part of the process 😄

meeseeksmachine:

PR #1024, opened 2019-05-13
Latex improvements

### LINKED ISSUE #340: Hidden \includegraphics[]{} latex call

t-makaro:
The follow up PR to #992.
- Slightly better title spacing. (The gap between title/author/date is smaller. Added ```usepackage{titling})```)
- Better spacing between cells. (Removed ```\tcbset{nobeforeafter}```)
- Paragraphs are no longer auto-indented like in markdown. (Added ``` \usepackage{parskip}```)
- Better prompt font, and spacing (using tcolorboxes lengths)
- Fixes #340. I was unable to use the ```\setkeys``` option to fix this, since it had a side effect of shrinking ```adjustimage``` boxes. However, by adding the ```Export``` flag to the adjustbox package it exports a new version of ```includegraphics``` that still allows use of all the keys whilst still setting a maximum size for images.
- Stop alt-text from being used as an image caption. Apparently, before, it was just removing the caption label but not the caption as the comments described. This PR now supersedes #990.

Before:
image

After:
image

### DIFF
--- a/nbconvert/templates/latex/base.tplx
+++ b/nbconvert/templates/latex/base.tplx
@@ -27,7 +27,7 @@ This template does not define a docclass, the inheriting class must define this.
     % in the conversion process - todo).
     \usepackage{caption}
     \DeclareCaptionLabelFormat{nolabel}{}
-    \captionsetup{labelformat=nolabel}
+    \captionsetup{format=nolabel}
 
     \usepackage[Export]{adjustbox} % Used to constrain images to a maximum size
     \adjustboxset{max size={0.9\linewidth}{0.9\paperheight}}
--- a/nbconvert/templates/latex/base.tplx
+++ b/nbconvert/templates/latex/base.tplx
@@ -26,8 +26,8 @@ This template does not define a docclass, the inheriting class must define this.
     % proper Figure object with a Caption API and a way to capture that
     % in the conversion process - todo).
     \usepackage{caption}
-    \DeclareCaptionFormat{nolabel}{}
-    \captionsetup{format=nolabel}
+    \DeclareCaptionFormat{nocaption}{}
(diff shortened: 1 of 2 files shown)


t-makaro:
Also: Spacing between cells:
Before:
image

After:
image

This applies to text before/after cells too.

t-makaro:
I believe it is the right approach.

The ```adjustbox``` package has an ability to add features to the original ```includegraphics``` (with a lower case ```export``` instead of ```Export```), but the documentation explicitly warns that not all features may work with that method. This should actually be more feature filled and customizable than the original ```includegraphics```. I tested setting the ```width``` key to larger than the 90% maximum that I've set, and it will ignore the maximum if asked too which is good for people using custom templates.

My bigger worry is if someone made a template that used ```\Oldincludegraphics```. This would break their template. I could re-add the ```\Oldincludegraphics``` definition for compatibility.

MSeal:
Ahh the `\Oldincludegraphics` might be a valid concern. I've seen a lot of threads over the past two releases where people reference that in relation to other issue.

We could keep the backwards compatibility for the next patch release but note in the release notes that we'll remove that compatibility in 6.x?

t-makaro:
@mseal I managed to squeeze in one more fix. It's good to merge once tests pass.

meeseeksmachine:

PR #1030, opened 2019-05-22
Fix Testing on Windows

t-makaro:
I develop primarily on windows, and I have been having problems running tests. Manually testing of features in failing tests shows that the issues seem to be the tests. Most of the problems seem to be due to ```shlex.split``` not understanding windows (even with ```posix=False```). This mostly fixes testing on windows.

Though I still seem to have some reliability problems. ```preprocessors\tests\test_execute.py```, ```exporters\tests\test_pdf.py```, and ```tests\test_nbconvertapp.py``` seem to have some odd failures but only sometimes. This appears to me like pytests caching issues because the issues tend to temporarily disappear when I restart my console. This also may be caused by sometimes using ```pytest``` instead of ```py.test```. Using ```pytests``` seems to cause errors, and then using ```py.test``` after using ```pytests``` does not resolve them without restarting the console. For reference, this required me to set vs code to use ```"python.testing.pyTestPath": "py.test"``` in my workspace settings.

For now, I believe my solution is sufficient, though ```py.test``` is considered legacy. Perhaps, we should consider investigating moving to ```pytest```. Also, It might be worth having CI, like appveyor, set up to test on windows.

### DIFF
--- a/nbconvert/preprocessors/tests/test_execute.py
+++ b/nbconvert/preprocessors/tests/test_execute.py
@@ -24,7 +24,6 @@ from ..execute import ExecutePreprocessor, CellExecutionError, executenb, DeadKe
 from ...exporters.exporter import ResourcesDict
 
 import IPython
-from mock import MagicMock
 from traitlets import TraitError
 from nbformat import NotebookNode
 from jupyter_client.kernelspec import KernelSpecManager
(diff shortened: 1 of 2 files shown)


MSeal:
I agree with pytest of py.test. Appveyor would be a good addition too. I had some interest from folks thinking about adding that to papermill. If that goes anywhere I'll look to extend it to some other jupyter projects like nbconvert.

meeseeksmachine:

Issue #786, opened 2018-03-21
UTF-8 problem during converting to PDF


theasder:
Hey there,

I have jupyter notebook from anaconda on Ubuntu 16.04 with installed xetex. I tried to convert it to PDF and it fetched successfully all english words and formulas, but there were some utf-8 symbols and it ignored it. Some logs here:

```
[W 18:37:28.242 NotebookApp] Notebook ЮраМолодец.ipynb is not trusted
[I 18:37:32.967 NotebookApp] Starting buffering for 005c099d-5a37-410e-812a-9b062b5744fc:e6a21be8811347f892517ca798014ee1
[I 18:37:33.290 NotebookApp] Kernel restarted: 005c099d-5a37-410e-812a-9b062b5744fc
[I 18:37:35.346 NotebookApp] Adapting to protocol v5.1 for kernel 005c099d-5a37-410e-812a-9b062b5744fc
[I 18:37:35.346 NotebookApp] Restoring connection for 005c099d-5a37-410e-812a-9b062b5744fc:e6a21be8811347f892517ca798014ee1
[I 18:37:35.347 NotebookApp] Replaying 6 buffered messages
[W 18:37:42.269 NotebookApp] Notebook ЮраМолодец.ipynb is not trusted
[I 18:37:43.054 NotebookApp] Support files will be in 
[I 18:37:43.054 NotebookApp] Making directory /root
[I 18:37:43.055 NotebookApp] Making directory /root
[I 18:37:43.055 NotebookApp] Making directory /root
[I 18:37:43.055 NotebookApp] Making directory /root
[I 18:37:43.056 NotebookApp] Writing 30590 bytes to /root/notebook.tex
[I 18:37:43.056 NotebookApp] Building PDF
[I 18:37:43.056 NotebookApp] Running xelatex 3 times: ['xelatex', '/root/notebook.tex']
[I 18:37:47.278 NotebookApp] Running bibtex 1 time: ['bibtex', '/root/notebook']
[W 18:37:47.302 NotebookApp] bibtex had problems, most likely because there were no citations
[I 18:37:47.303 NotebookApp] PDF successfully created
```

hycakir:
See my answer here: https://stackoverflow.com/a/49582428/2372611

[...]

PR #1039, opened 2019-05-30
Fix LaTeX exporting '?' for non-ascii title

imtsuki:
I have the same issue with #786 . When I tried to export a PDF file from a notebook that has a non-ascii title (which is the notebook's file name),  the output is just '????':

螢幕快照 2019-05-30 17 26 26

I discovered that it was not a XeLaTeX problem; I use `ctex` package to support non-ascii rendering (You can see them in the date under the title), but the title is still '???'. I tried to directly export `.tex` source file instead, and it turned out that `nbconvert` itself produces `'\title{????}'` for title while other non-ascii characters in the document body is fine.

In `nbconvert/templates/latex/base.tplx`, I found this line confusing:

`\title{((( nb_title | ascii_only | escape_latex )))}`

I have no idea what `ascii_only` is for, and deleting this is supposed to make sense.

螢幕快照 2019-05-30 17 29 46

### DIFF
--- a/nbconvert/templates/latex/base.tplx
+++ b/nbconvert/templates/latex/base.tplx
@@ -147,7 +147,7 @@ This template does not define a docclass, the inheriting class must define this.
     % Document title
     ((* block title -*))
     ((*- set nb_title = nb.metadata.get('title', '') or resources['metadata']['name'] -*))
-    \title{((( nb_title | ascii_only | escape_latex )))}
+    \title{((( nb_title | escape_latex )))}
     ((*- endblock title *))
     ((* block date *))((* endblock date *))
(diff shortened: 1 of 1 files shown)


t-makaro:
LGTM! I think ```ascii_only``` is a remnant from before we were using XeLaTeX. I will merge once tests pass.

Also, thanks for the tip on ```ctex```. Are you just using a custom template to add ```\usepackage{ctex}```?

imtsuki:
Yeah, I added that line to `article.tplx` for my own use, and it only supports CJK characters (as far as I know).

meeseeksmachine:

PR #1040, opened 2019-05-30
Set flag to not always stop kernel execution on errors

jhamrick:
There is some non-deterministic error in nbgrader related to the kernel stopping executing after it's received an error, see https://github.com/jupyter/nbgrader/issues/1056 for discussion. In particular, @samuelmanzer gave the following explanation:

> I think that in some subset of the runs, the ZeroMQ messages for multiple cells reside simultaneously on the execution stack, and stop_on_error=True results in the cell containing the assignment statements being thrown away as it is on the stack when the earlier (deliberate) AssertionError exception causes the stack to be discarded. The subsequent assertion which depends on that assignment statement executing successfully then fails.

This PR changes the `stop_on_error` flag to be set adaptively depending on whether `self.allow_errors` is configured.

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -540,7 +540,8 @@ class ExecutePreprocessor(Preprocessor):
         return False
 
     def run_cell(self, cell, cell_index=0):
-        parent_msg_id = self.kc.execute(cell.source)
+        parent_msg_id = self.kc.execute(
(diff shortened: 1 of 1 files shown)


MSeal:
Ahh I believe you are correct that this should be added. Thank you for investigating!

Hold off on restarting that test for a bit -- I think it's a race condition I've been trying to catch and debug around parallel executions of nbconvert.

meeseeksmachine:

PR #1053, opened 2019-06-18
Return error if latex does

### LINKED ISSUE #896: Nbconvert silences errors if a pdf is successfully produced

t-makaro:
My attempt to get latex to raise errors even if a PDF if produced. Fixes #896

### DIFF
--- a/nbconvert/preprocessors/extractoutput.py
+++ b/nbconvert/preprocessors/extractoutput.py
@@ -75,7 +75,6 @@ class ExtractOutputPreprocessor(Preprocessor):
         unique_key = resources.get('unique_key', 'output')
         output_files_dir = resources.get('output_files_dir', None)
 
-
         #Make sure outputs key exists
         if not isinstance(resources['outputs'], dict):
             resources['outputs'] = {}
--- a/nbconvert/preprocessors/extractoutput.py
+++ b/nbconvert/preprocessors/extractoutput.py
@@ -81,8 +81,11 @@ class ExtractOutputPreprocessor(Preprocessor):
         output_files_dir = resources.get('output_files_dir', None)
         if self.escape_dir_spaces:
             unique_key = unique_key.replace(' ', '_')
+            resources['unique_key'] = unique_key
             if output_files_dir is not None:
-                output_files_dir = unique_key.replace(' ', '_')
+                output_files_dir = output_files_dir.replace(' ', '_')
+                resources['output_files_dir'] = output_files_dir
+            
 
         #Make sure outputs key exists
         if not isinstance(resources['outputs'], dict):
--- a/nbconvert/preprocessors/extractoutput.py
+++ b/nbconvert/preprocessors/extractoutput.py
@@ -12,7 +12,7 @@ import os
 import json
 from mimetypes import guess_extension
 
-from traitlets import Unicode, Set
+from traitlets import Unicode, Set, Bool
 from .base import Preprocessor
 
 if sys.version_info < (3,):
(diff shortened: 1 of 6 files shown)


t-makaro:
```TestNbConvertApp.test_filename_spaces``` is failing on my system (windows) due to this change. The latex file spits an error if I use xelatex, but it is fine with pdflatex.

**Comment by t-makaro** [comment]
Ok, I've opted to modify the ExtractOutputPreprocessor to stop it from making spaces in the extra files directory when using this preprocessor with latex.

File: nbconvert/preprocessors/extractoutput.py [hunk 1]
```diff
@@ -54,6 +54,11 @@ class ExtractOutputPreprocessor(Preprocessor):
         {'image/png', 'image/jpeg', 'image/svg+xml', 'application/pdf'}
     ).tag(config=True)
 
+    escape_dir_spaces = Bool(
+        False, help="If True, spaces will be replaced with underscores in the "
+        "output directory."
+    ).tag(config=True)
```
MSeal: This is a slippery slope, which bends into the larger PR which maps everything to a tmp directory with safe file naming before copying back. It's also not clear if this is a legal operation change as a default.

MSeal: I'd maybe skip this or separate it into another PR for temporarily mapping files to safe paths before copying back to the actual output_dir

t-makaro: From my knowledge of the larger PR. It only fixes things for the pdf exporter. It someone converted to latex and then manually compiled with latex, then paths would create issues still. This is a partial fix, and IMHO is good to do for the pure latex.

Also, if a latex error is raised the pdf will not be saved. I don't know why, since latex should be saving the pdf. This prevents a regression of notebooks that used to convert (but with a visual glitch) don't anymore. I need to look further into why the pdf isn't saving.

t-makaro: Ah, I do see that in the larger PR there is this ```true_output_files_dir = resources.get('output_files_dir', cwd)``` If I pass the new escaped ```output_files_dir``` back to the resources then this would work.

nbconvert/preprocessors/extractoutput.py
```diff
@@ -74,6 +79,13 @@ def preprocess_cell(self, cell, resources, cell_index):
         #has been specified
         unique_key = resources.get('unique_key', 'output')
         output_files_dir = resources.get('output_files_dir', None)
+        if self.escape_dir_spaces:
+            unique_key = unique_key.replace(' ', '_')
+            resources['unique_key'] = unique_key
+            if output_files_dir is not None:
+                output_files_dir = output_files_dir.replace(' ', '_')
+                resources['output_files_dir'] = output_files_dir
```
t-makaro: I manually tested that ```resources.get('output_files_dir', cwd)``` returns the appropriate directory when called from within ```from_notebook_node```.

t-makaro:
This should now be compatible with the larger PR. I don't feel comfortable merging this into master without the change to ExtractOutputPreprocessor, since tests fails on my computer without this change.

If you'd like me to change how I get the latex exporter to tell the ExtractOutputPreprocessor to escape spaces, I can do that.

t-makaro:
I'm going to need to think more about this....

t-makaro:
```TestNbConvertApp.test_filename_spaces``` fails on my windows machine, but the error that I get is not a latexError, so I'm confused as to why it's failing.

t-makaro:
Ok. I believe I've fixed it. ```\usepackage{grffile}``` only fixes spaces in filenames on pdflatex. Using [this hack](https://github.com/ho-tex/oberdiek/issues/31#issuecomment-441094438), I was able to make grffile fix spaces in xelatex. The tests now pass on both windows and travis. This should fix most of our spaces in filenames issues.

**CC** MSeal. Ready for re-review.

MSeal:
Tested locally to double check change. Seems solid.

meeseeksmachine:

Issue #1032, opened 2019-05-25
Optional store_history for ExecutePreprocessor.preprocess_cell


wwitzel:
I would appreciate having the option to turn off history storing when calling preprocess_cell for an ExecutePreprocessor.  That is, accept an additional "store_history" argument with a default value of True in the preprocess_cell method, as well as the run_cell method.  This "store_history" would need to be passed from preprocess_cell to run_cell which will then pass it on to the "execute" method of the KernelClient object.  I would find this convenient when using nbconvert to execute a collection of notebooks for regression testing and don't have a need to store the history while doing so.  I have tested out that this suits my purposes by overriding "run_cell" and passing store_history=False to the KernelClient's "execute" method.

MSeal:
I believe https://github.com/jupyter/nbconvert/pull/1017 does exactly this for the ipython kernel and will go out in the next release. I agree that storing history should not be defaulted. Is there another instance you're aware of where history is saving that you're referencing?

wwitzel:
Thanks Matthew, but I don't think this is quite the same thing.  #1017 seems to be referring to storing the history in an SQL file, however, I am proposing an option to store no history in the memory either.  The reason this is desirable to me is that I am actually recycling the same Kernel to execute many notebooks (using %reset and deleting modules other than initially loaded ones).  It took a LOT of trial and error, but it seems to be behaving as desired.  Specifically, if I check the Python garbage collector, there are no new objects after resetting everything.  In order to accomplish this, however, the KernelClient needs to run "execute" with the "store_history" flag set to False.  Otherwise, it will record all of the inputs from all of the notebooks as it goes.  In order to do this, I had to override the "run_cell" method of the ExecutePreprocessor, copying the version in nbconvert with a minor modification.  That's not a good permanent solution, though.


Does this make sense?  If you'd like to see my code for doing this, see the RecyclingExecutePreprocessor class in

https://github.com/PyProveIt/Prove-It/blob/master/build.py


Thanks,

Wayne


________________________________

MSeal:
Hi Wayne,

Ahh I see what your referencing. We probably should be setting `store_history = False` by default. If you want to submit a PR to change the default for nbconvert please feel free to do so. I'll be a little more attentive to a PR if you open one up to get it merged.

wwitzel:
[...]

PR #1055, opened 2019-06-26
Added 'store_history' option to 'preprocess_cell' and 'run_cell'

wwitzel:
…methods of ExecutePreprocessor.

The default value is True to be backwards compatible since the default of KernelClient.execute has store_history=True.  Allowing "store_history" to be disabled can be valuable when "hacking" nbconvert to execute many notebooks with one Kernel session and not store memory along the way that is not needed.


MSeal:
Thanks for the change. For others looking see #1032 for context.

I'm going to make a new issue to change the default to False for nbconvert 6.0 release which will be a later release.

meeseeksmachine:

Issue #822, opened 2018-05-29
--clear-output flag is broken


spencermathews:
I appreciate the `--clear-output` convenience flag introduced in #619. However, unless I'm missing something it simply does not work.

These commands should both remove output, but the second does not:

```bash
jupyter nbconvert --ClearOutputPreprocessor.enabled=True --inplace example.ipynb
jupyter nbconvert --clear-output example.ipynb
```

The first form includes the following output (using the `--debug` option):

```bash
[NbConvertApp] Applying preprocessor: TagRemovePreprocessor
[NbConvertApp] Applying preprocessor: ClearOutputPreprocessor
[NbConvertApp] Applying preprocessor: coalesce_streams
```

However, the `ClearOutputProcessor` line is missing when using `--clear-output`.

See [this](https://stackoverflow.com/a/47774393) StackOverflow answer for another user who observed the same issue.

Using Anaconda Python 3.6.5 + Jupyter 4.4.0

t-makaro:
I can confirm that ```--clear-output``` is broken, but I can't for the life of me figure out why....

rawlins:
I have tracked down (at some level) the why, but this is bound up in deep traitlets behavior that I cannot figure out at all, so I'm not sure what the fix is. In [traitlets.Application.flatten_flags](https://github.com/ipython/traitlets/blob/master/traitlets/config/application.py#L603), which is called on `flags`, various processing happens. One of the things that this function apparently does, is replace all class names of classes that have a single subclass with the name of that subclass.  Since `ClearOutputProcessor` has only one subclass, namely `TagRemoveProcessor`, ever since that subclass was added, the `clear-output` flag gets changed from:

```
{'NbConvertApp': {'use_output_suffix': False, 'export_format': 'notebook'}, 'ExecutePreprocessor': {'enabled': False}, 'FilesWriter': {'build_directory': ''}, 'ClearOutputPreprocessor': {'enabled': True}}
```
into

```
{'NbConvertApp': {'use_output_suffix': False, 'export_format': 'notebook'}, 'ExecutePreprocessor': {'enabled': False}, 'FilesWriter': {'build_directory': ''}, 'TagRemovePreprocessor': {'enabled': True}}
```

This is the crucial traitlets code:

```
            for cls, subdict in flagdict.items():
                children = mro_tree[cls]
                # exactly one descendent, promote flag section
                if len(children) == 1:
                    cls = children[0]
```

[...]

PR #1056, opened 2019-06-28
Refactor template system to be easily extensible and voila compatible

maartenbreddels:
cc @SylvainCorlay @martinRenou @jtpio

This is the followup of what was discussed in https://github.com/jupyter/nbconvert/pull/1028, planned for voila in https://github.com/QuantStack/voila/pull/208.

The basic idea is that templates are a directory in either:
$prefix/nbconvert/templates/<template_name> where $prefix is determined by jupyter_paths.
On OSX it would be:
```
	/Users/maartenbreddels/Library/Jupyter/nbconvert/templates/default
	/Users/maartenbreddels/miniconda3/envs/voila/share/jupyter/nbconvert/templates/default
	/usr/local/share/jupyter/nbconvert/templates/default
	/usr/share/jupyter/nbconvert/templates/default
```

With this PR, https://github.com/jupyter/nbconvert/pull/1005 would simply be a directory share/jupyter/nbconvert/templates/lab and now the old full.tpl is a template called 'notebook'.

In combination with https://github.com/QuantStack/voila/pull/208 we can make all nbconvert template compatible by only overriding the `{% block ipywidgets %}`, and instead of inlining CSS, we can request it from the server.

# Loose ends
 * the latex templates clutter the default directory a bit, maybe we can put these in different template directories. `
 * The meaning of`TemplateExporter.template_file` changed a bit, it's cannot be an absolute path anymore, it now refers to the filename in the search path, resolved by jinja2 only. (also the reason unit tests fail).
 * We discussed using index.j2.html instead of filenames for templates, I think without the .j2 is looks cleaner.
 * Some systems may want to list template (eg voila), filtered by what kind of exporter is supported, we could list that in the conf.json file.
 * the html exporter should maybe use the notebook template by default, since now it's using the old 'basic.tpl' template.


MSeal:
@t-makaro should probably help review this as he's been more involved on the template side of things recently.

MSeal:
I need to do some testing this weekend to help make sure everything is in working order, but it looks like all the changes were what we discussed. I'm not sure if we agreed for this change to be in 6.0 release or in 5.x path, if you remember?

The test suite is quite angry and will need some attention for the new patterns.

> the latex templates clutter the default directory a bit, maybe we can put these in different template directories. `

MSeal:
> the latex templates clutter the default directory a bit, maybe we can put these in different template directories. `

Can also be a follow-up PR to avoid making this any larger

MSeal:
> Some systems may want to list template (eg voila), filtered by what kind of exporter is supported, we could list that in the conf.json file.

Probably needs a separate issue to discuss. We may want this to be a dynamic calculation.

MSeal:
> the html exporter should maybe use the notebook template by default, since now it's using the old 'basic.tpl' template.

Need to look at the differences here, but I think there was consensus to have exports be the classic notebook format by default and lab format to be an opt-in pattern, if that's what's being referenced by this question.

MSeal:
> I need to do some testing this weekend to help make sure everything is in working order, but it looks like all the changes were what we discussed. I'm not sure if we agreed for this change to be in 6.0 release or in 5.x path, if you remember?
> 
> The test suite is quite angry and will need some attention for the new patterns.

Missed the 6.0 tag on the PR makes sense. Let me know if you need any help with figuring out tests later on.

maartenbreddels:
I changed the target branch to 6.x, but CI does not run on this, can this be enabled?

> Need to look at the differences here, but I think there was consensus to have exports be the classic notebook format by default and lab format to be an opt-in pattern, if that's what's being referenced by this question.

I've changed it, such that lab is a  template: `$ nbconvert --template lab ...`

The old system of using a template file still exists, but now is available as `--template-file=`,

maartenbreddels:
I've seem to have hit a bug in traitlets however, like voila, I want this to work:
```
$ voila     --template gridstack --theme dark basics.ipynb
$ nbconvert --template gridstack --theme dark --to html --execute basics.ipynb
``` 
Such that arguments `--template` and `--theme` have the same meaning.

However, it seems the HTMLExporters does not pick this up, the alias `--theme` gets translated to `SlidesExporter.theme` (!) (from a debug session):
```
(Pdb) self.config  # <-- this is HTMLExporter.config
{'NbConvertBase': {'display_data_priority': ['application/vnd.jupyter.widget-state+json', 'application/vnd.jupyter.widget-view+json', 'application/javascript', 'text/html', 'text/markdown', 'image/svg+xml', 'text/latex', 'image/png', 'image/jpeg', 'text/plain']}, 'CSSHTMLHeaderPreprocessor': {'enabled': True}, 'HighlightMagicsPreprocessor': {'enabled': True}, 'RegexRemovePreprocessor': {'enabled': True}, 'TagRemovePreprocessor': {'enabled': True}, 'NbConvertApp': {'export_format': 'html'}, 'TemplateExporter': {'template_name': 'gridstack'}, 'SlidesExporter': {'theme': 'dark'}, 'ExecutePreprocessor': {'enabled': True}}
```
Maybe I am missing something, if not we should open an issue for traitlets.

nbconvert/exporters/tests/test_html.py
```diff
@@ -74,7 +74,7 @@ def test_prompt_number(self):
                 }
             }
         )
-        exporter = HTMLExporter(config=no_prompt_conf, template_file='full')    
+        exporter = HTMLExporter(config=no_prompt_conf, template_name='notebook')
```
SylvainCorlay: Maybe the name of the template should be `classic` instead of `notebook`

share/jupyter/nbconvert/templates/reveal/index.html
```diff
@@ -1,5 +1,5 @@
-{%- extends 'basic.tpl' -%}
```
SylvainCorlay: does `default` stand for the default html template?

nbconvert/exporters/tests/test_html.py
maartenbreddels: For now 'default' is the classic one, it may have been a good idea to have the default html template be a 'nothing' template, without css and all
The classic template would just add some css to the template, have some opinion on the dom structure etc. This way the lab template would not inherit from the classical template.

share/jupyter/nbconvert/templates/reveal/index.html
maartenbreddels: yes, I think we can make reveal actually a normal template, just like we want with voila, with just some extra resources. So you would do `--template reveal`.

maartenbreddels:
I don't understand check-manifest, I added:
```
+include share/jupyter/nbconvert/templates/default/static/style.css
+include share/jupyter/nbconvert/templates/lab/static
+include share/jupyter/nbconvert/templates/lab/static/index.css
+include share/jupyter/nbconvert/templates/lab/static/theme-dark.css
```
To MANIFEST.in, but it still gave a warning, so FTM I'm ignoring that directory.

t-makaro:
I will take a deeper look at this (hopefully) tomorrow, but I do have a few things on my mind that I'd like to put out there.

1. Do in-memory templates still work? (Given that template_file and template_name are too different things, I believe the answer is yes.)
2. Can we have templates of the same name for different file types? Perhaps by having multiple index files within a directory.
```
template_name/
    index.j2.html
    index.j2.markdown
    index.j2.py
```
(looking at the code I believe the answer is yes).
3. I do think that template files should have ```.j2``` within them.
4. Does ```template_path``` still add the cwd to the search path by default (the the user's ability to change it)? Again, I believe the answer is yes.
5. Can we add some sort of check to make absolute paths still work?
 Currently, [this line](https://github.com/jupyter/nbconvert/blob/8e3cfb4de6d3199a2fa7de02d48268c66a063c07/nbconvert/exporters/templateexporter.py#L162) ensures that an absolute path is added to the template search paths if needed. I suppose that ```template_file``` can be used with an absolute path?
6. Can ```index.j2.*``` files still reference other ```*.j2.*``` files inside the same directory? Again, I believe the answer is yes.

@maartenbreddels 
> However, it seems the HTMLExporters does not pick this up, the alias --theme gets translated to SlidesExporter.theme (!) (from a debug session):

Perhaps related to #822?

maartenbreddels:
> 1. Do in-memory templates still work? (Given that template_file and template_name are too different things, I believe the answer is yes.)

yes, convered in test



> 2\. Can we have templates of the same name for different file types? Perhaps by having multiple index files within a directory.

yes



> 3. I do think that template files should have `.j2` within them.

I'm 50/50 on it, maybe other may express their opinion, I'll leave it like this for the moment, we can always do a rename at the end.



> Does `template_path` still add the cwd to the search path by default (the the user's ability to change it)? Again, I believe the answer is yes.

yes



> Can we add some sort of check to make absolute paths still work?

covered in tests

> Can `index.j2.*` files still reference other `*.j2.*` files inside the same directory? Again, I believe the answer is yes.

yes, in the same, *and* other directories.

maartenbreddels:
After discussion with @SylvainCorlay, the plan is to:
move templates to their separate directory, e.g.:
```
templates/
          latex/
          markdown/
          python/
          classic/
          lab/
```

in conf.json, each template can list the mimetype it outputs to (so you can still have multiple mimetypes/formats in 1 template directory, e.g. latex+html). This makes it easier to list only templates of a specific mimetype.
Common template may still go into a directory (say base, or skeleton).

maartenbreddels:
the suggested change are in, I think this is ready for review.

t-makaro:
Why not rename ```article.tex``` to ```index.tex``` instead of creating a new file? I believe that the other files inside the latex template directory should be renamed to have ```.tex``` extensions instead of ```.tplx``` too. Also, the ```report.tex``` file could be it's own template.

t-makaro:
In number #1076, I made it possible to use ```style_*.tplx``` as their own top level template (making ```article.tplx``` redundant). This means in nbconvert 5.6 one can do one of the following:

```
Jupyter nbconvert --to pdf notebook.ipynb --template=article.tplx
Jupyter nbconvert --to pdf notebook.ipynb --template=style_jupyter.tplx
Jupyter nbconvert --to pdf notebook.ipynb --template=style_ipython.tplx
```
where the first 2 are equivalent (and the extension can in theory be dropped).

My understand as that after this PR I would do:
```
Jupyter nbconvert --to pdf notebook.ipynb --template=latex
Jupyter nbconvert --to pdf notebook.ipynb --template-file=style_jupyter.tplx
Jupyter nbconvert --to pdf notebook.ipynb --template-file=style_ipython.tplx
```

This is where my concern is. It may seem arbitrary to a user to have to use ```--template-file```. If I do:
```
Jupyter nbconvert --to pdf notebook.ipynb --template=style_jupyter.tplx
```
will it work? If not, I would refactor these latex templates a fair bit to make these ```style_*.tplx``` templates top level. Though I would be happy to do that in a separate PR.

nbconvert/nbconvertapp.py
```diff
@@ -51,7 +51,9 @@ def validate(self, obj, value):
 nbconvert_aliases.update(base_aliases)
 nbconvert_aliases.update({
     'to' : 'NbConvertApp.export_format',
-    'template' : 'TemplateExporter.template_file',
+    'template' : 'TemplateExporter.template_name',
+    'theme' : 'HTMLExporter.theme',
+    'template-file' : 'TemplateExporter.template_file',
```
t-makaro: Do we want ```template-file``` to be a part of the CLI API? I would think not. I also think that the ```theme``` alias isn't necessary and could be confused for having functionality with other exporters.

t-makaro:
The more that I look at the template directory the more that I think that the files should have ```.j2.``` in them. I think that the fact the files are partial files should be explicit. At the very least, we should replace ```.tpl``` with ```.j2``` and ```.tplx``` with ```.tex```.

nbconvert/preprocessors/csshtmlheader.py
```diff
@@ -28,8 +30,9 @@ class CSSHTMLHeaderPreprocessor(Preprocessor):
                               help="CSS highlight class identifier"
     ).tag(config=True)
 
-    style = Unicode('default',
-            help='Name of the pygments style to use'
+    style = Union([Unicode('default'), Type(klass=Style)],
```
t-makaro: This change should be done for the latexPreprocessor too.

maartenbreddels:
> Why not rename `article.tex` to `index.tex` instead of creating a new file?

To make it more clear that it was using the article class, but for many of the latex parts I was waiting for you feedback, so I'll happily do that.

> I believe that the other files inside the latex template directory should be renamed to have `.tex` extensions instead of `.tplx` too. 

I agree.

> Also, the `report.tex` file could be it's own template.

Does that make sense? It would mean you for a normal article template you do:
 * `jupyter nbconvert --to pdf notebook.ipynb`

And for a report template, you do:
 * `jupyter nbconvert --to pdf notebook.ipynb --template=latex-report`

Are you ok with that?


> Though I would like if we could detect if a user passed in a `template_name` with a file extension and use that directly as the `template_file` then.

That is what now the Exporter does, it sets the default template name.

> My understand as that after this PR I would do:
> 
> ```
> Jupyter nbconvert --to pdf notebook.ipynb --template=latex
> Jupyter nbconvert --to pdf notebook.ipynb --template-file=style_jupyter.tplx
> Jupyter nbconvert --to pdf notebook.ipynb --template-file=style_ipython.tplx
> ```

You can skip the template argument, since the `PDFExporter` inherits from `LatexExporter`, which has a default `template_name` of `'latex'`., so you can simply do:

```
jupyter nbconvert --to pdf notebook.ipynb
```
And indeed, it means your last two commands also work.







> The more that I look at the template directory the more that I think that the files should have `.j2.` in them. I think that the fact the files are partial files should be explicit. At the very least, we should replace `.tpl` with `.j2` and `.tplx` with `.tex`.

That last part if fully agree with (`.tpl` with `.j2` and `.tplx` with `.tex`.).

Regarding the extensions, I prefer `index.html.j2` now too.
Similar to how a zipped `file.tar.gz` uncompresses to a `file.tar` file, a `index.html.j2` template renders to a `index.html` file.

Renaming `*.html` to `*.html.j2` and `*.tex` to `*.tex.j2` I'd like to leave open, and have @SylvainCorlay and possibly @MSeal 's opinion on.

nbconvert/preprocessors/csshtmlheader.py
maartenbreddels: Is this good, or is there an issue with this that should be attacked in this PR?

nbconvert/nbconvertapp.py
maartenbreddels: Since the theme argument isn't working, i'm not even gonna argue :)

I think the template-file could be useful, as demonstrated in the examples below (with the latex template).

MSeal:
Just my 2 cents if it helps

> And for a report template, you do:
> - jupyter nbconvert --to pdf notebook.ipynb --template=latex-report
> Are you ok with that?

I think that's a good pattern. It's easy to remember and clear about what it's doing.

> Renaming *.html to *.html.j2 and *.tex to *.tex.j2 I'd like to leave open, and have @SylvainCorlay and possibly @MSeal 's opinion on.

In my opinion having the `.j2` in the extensions is useful. The [ansible best-practices](https://docs.ansible.com/ansible/latest/user_guide/playbooks_best_practices.html) list that pattern as well. I would note that `*.j2.html` loads and reads from editors better, but it looks like from scanning a few other projects `*.html.j2` is the pattern that's used externally.

nbconvert/exporters/templateexporter.py
```diff
@@ -58,6 +63,29 @@
         'json_dumps': json.dumps,
 }
 
+
+# copy of https://github.com/jupyter/jupyter_server/blob/b62458a7f5ad6b5246d2f142258dedaa409de5d9/jupyter_server/config_manager.py#L19
+def recursive_update(target, new):
```
mkcor: Why not `from jupyter_server.config_manager import recursive_update` instead? I think you taught me that one, @maartenbreddels!

It would add a dependency but, in this respect, would that be any different from the `from jupyter_core.paths import jupyter_path` addition line 14?

MSeal: Adding `jupyter_server` to nbconvert as a dependency would be substantial and unwanted as it would dramatically increase the width of overall dependencies needed for the project. If there's shared code we prefer it be added to `jupyter_core` over time to avoid pull adjacent dependency requirements.

MSeal:
@t-makaro @maartenbreddels What still needs deciding / changes? I'd like to get this change into the 6.x branch sooner rather than later so it doesn't get lost in memory and infinite review cycles if we can.

t-makaro:
Looks good to me. Maybe rebase the 6.x branch off master before merging this into the 6.x branch.

maartenbreddels:
> Renaming `*.html` to `*.html.j2` and `*.tex` to `*.tex.j2`

This is the only open thing, apart from that I think it it mergable/reviewable.

t-makaro:
While ```*.j2.html``` does have syntax highlighting, my editor displays lots of formatting errors without an extension to render jinja files properly. Displaying lots of errors when there aren't any is counter productive. For this reason, I vote for ```*.html.j2``` and ```*.tex.j2```. This will render properly with a plugin (Like "Better Jinja" for vs code), and it will not display syntax errors without an plugin. This also seems to be what some other projects do.

MSeal:
> Looks good to me. Maybe rebase the 6.x branch off master before merging this into the 6.x branch.

I can take care of that now.

> While *.j2.html does have syntax highlighting, my editor displays lots of formatting errors without an extension to render jinja files properly. Displaying lots of errors when there aren't any is counter productive. For this reason, I vote for *.html.j2 and *.tex.j2. This will render properly with a plugin (Like "Better Jinja" for vs code), and it will not display syntax errors without an plugin. This also seems to be what some other projects do.

Agreed :+1: .

MSeal:
Ok rebasing 6.x with master added some conflict files unfortunately.

nbconvert/exporters/templateexporter.py
mkcor: Ok, thanks for explaining @MSeal.

maartenbreddels:
squash and rebased.
Note: this should be split in 2 commits before merking to give @SylvainCorlay credits for the lab template, but this was impossible to rebase otherwise.

MSeal:
@maartenbreddels If you wish you can coauthor the rebase commit by adding https://help.github.com/en/articles/creating-a-commit-with-multiple-authors

I'm good to merge either way, I'll check on this tomorrow and merge it then either way. Thanks a ton for getting this prepped.

MSeal:
I can do the squash on merge if you want, or we can leave a few minor commits hanging around.

MSeal:
I'm gonna merge this -- we can make changes later if need be :D

choldgraf:
✨✨✨✨✨💪💪💪💪🤩🤩🤩🤩🤩

maartenbreddels:
Awesome!
Managed to get the following templates to work with nbconvert and voila
 * https://github.com/voila-dashboards/voila-material/pull/8/files
 * https://github.com/voila-dashboards/voila-gridstack/pull/3

And the voila templates contains almost no code, except for the spinner in the lab template:
 * https://github.com/voila-dashboards/voila/pull/208/files

bdklahn:
Is this issue related, at all, to why there is no "to gridstack"-like entry in the export menu item?
I have been using the Gridstack editor, successfully (with pop-out browser tab rendering), but wondered why there might not be a corresponding entry in the export menu.
I wondered if it were the Gridstack folks who would/could/should add that, but . . . 
[https://github.com/voila-dashboards/voila-gridstack/issues/180#issuecomment-1204376884](https://github.com/voila-dashboards/voila-gridstack/issues/180#issuecomment-1204376884)

I've been able to get the Voila-DejaVu-like (static page) behavior using the following config settings, in a script.

```python
    c = Config()
    c.TemplateExporter.exclude_input = True
    c.TemplateExporter.exclude_output_prompt = True
    c.TemplateExporter.exclude_input_prompt = True
    c.ExecutePreprocessor.enabled=True
    c.HTMLExporter.template_name = 'gridstack'
 ```
 
 Barring any potentially-related fundamental issues, possibly like this, I would think adding a Gridstack render export could be pretty straight-forward(?).
 
 I apologize if this is too unrelated to this and should be in it's own issue.

PR #1060, opened 2019-07-03
Require mock only on Python 2

kapsh:
Only usage of mock is inside "except ImportError" meant for running tests
under Python 2. So there is no need to install it for any Python 3.


meeseeksmachine:

PR #1061, opened 2019-07-05
Changed store_history default to False for run_cell

MSeal:
Default should be False here, but the PR submitter was worried about changing existing behavior for 5.x release.


mpacer:
This seems great for a 6.0 release.

t-makaro:
Just for reference: Closes #1032

PR #1064, opened 2019-07-19
Update base.tplx

KrokodileDandy:
\usepackage[T1]{fontenc} shouldn't be used with xelatex and throws errors like "!Text line contains an invalid character".
\usepackage{fontspec} recommended.

### DIFF
--- a/nbconvert/templates/latex/base.tplx
+++ b/nbconvert/templates/latex/base.tplx
@@ -14,8 +14,6 @@ This template does not define a docclass, the inheriting class must define this.
     
     ((* block packages *))
     \usepackage{fontspec}
-    % Nicer default font (+ math font) than Computer Modern for most use cases
-    \usepackage{mathpazo}
 
     % Basic figure setup, for now with no caption control since it's done
     % automatically by Pandoc (which extracts ![](path) syntax from Markdown).
--- a/nbconvert/templates/latex/base.tplx
+++ b/nbconvert/templates/latex/base.tplx
@@ -13,7 +13,7 @@ This template does not define a docclass, the inheriting class must define this.
     ((* block docclass *))((* endblock docclass *))
     
     ((* block packages *))
-    \usepackage[T1]{fontenc}
+    \usepackage{fontspec}
     % Nicer default font (+ math font) than Computer Modern for most use cases
     \usepackage{mathpazo}
 
--- a/nbconvert/templates/latex/base.tplx
+++ b/nbconvert/templates/latex/base.tplx
@@ -46,7 +46,6 @@ This template does not define a docclass, the inheriting class must define this.
     \usepackage{upquote} % Upright quotes for verbatim code
     \usepackage{eurosym} % defines \euro
     \usepackage[mathletters]{ucs} % Extended unicode (utf-8) support
-    \usepackage[utf8x]{inputenc} % Allow utf-8 characters in the tex document
     \usepackage{fancyvrb} % verbatim replacement that allows latex
     \usepackage{grffile} % extends the file name processing of package graphics 
                          % to support a larger range 
--- a/nbconvert/templates/latex/base.tplx
+++ b/nbconvert/templates/latex/base.tplx
@@ -13,7 +13,13 @@ This template does not define a docclass, the inheriting class must define this.
     ((* block docclass *))((* endblock docclass *))
     
     ((* block packages *))
-    \usepackage{fontspec}
+    \usepackage{iftex}
+    \ifPDFTeX
+    	\usepackage[T1]{fontenc}
+    	\usepackage{mathpazo}
+    \else
+    	\usepackage{fontspec}
+    \fi
 
     % Basic figure setup, for now with no caption control since it's done
     % automatically by Pandoc (which extracts ![](path) syntax from Markdown).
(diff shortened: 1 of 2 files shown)


MSeal:
Tests seem pretty angry at the change. Is that xelatex change version specific perhaps?

KrokodileDandy:
It shouldn't be version specific. Xelatex shouldn't even use the package fontenc I think. I'll look further into it.

njapke:
I don't know what is wrong with the tests, but since XeLaTeX is used for compilation, `fontspec` should be used instead of `fontenc` (see: [https://tex.stackexchange.com/a/3000](https://tex.stackexchange.com/a/3000)). `fontenc` breaks utf8 characters, since it loads old LaTeX fonts, and XeLaTeX uses utf8 fonts by default. I have implemented this exact fix on my system and suddenly German text started working when before it wouldn't. I would suggest someone should look more deeply into the test cases and font loading with regard to languages other than English and fix the test cases, or suggest a better fix.

njapke:
I just ran all the tests locally and the tests failing on TravisCI regarding pdf generation pass on my local environment. Skimming the logs generated by `xelatex` it seems to be missing at least one font, so this might be a problem with the texlive setup (I have a full texlive installation instead of a minimal one, which is obviously not a good fix for Travis). I could look into which packages might be needed to make the tests work.

Changes I applied to `base.tplx` for testing were the change provided by this PR and removing line 49 of `base.tplx` (`\usepackage[utf8x]{inputenc}`).

KrokodileDandy:
As suggested by @njapke I removed line 49 of `base.tplx`.

njapke:
Now that `lmodern` has been added the tests run perfectly. I noticed another thing, though. The package `mathpazo` provides `latex` fonts, but makes problems with `xelatex`. This might be a problem resulting from changing `fontenc` to `fontspec`, since the font loading is now different, but loading `mathpazo` incorrectly typesets text that should be bold in a normal typeface. Removing the import of `mathpazo` in line 18 fixes that problem.

MSeal:
@t-makaro @willingc Could one of your help review and either get this merged or see if the changes need more attention? Thanks

willingc:
This test is failing on Python 3.8dev: preprocessors/tests/test_execute.py::test_run_all_notebooks[SVG.ipynb-opts9] FAILED

@KrokodileDandy @t-makaro Any thoughts as to why this one is failing?

nbconvert/templates/latex/base.tplx
```diff
@@ -13,9 +13,7 @@ This template does not define a docclass, the inheriting class must define this.
     ((* block docclass *))((* endblock docclass *))
     
     ((* block packages *))
-    \usepackage[T1]{fontenc}
-    % Nicer default font (+ math font) than Computer Modern for most use cases
-    \usepackage{mathpazo}
+    \usepackage{fontspec}
```
t-makaro: Some nbconvert users still choose to use pdflatex, and fontenc does work on XeLaTeX (with warnings), but fontspec does not work at all on pdflatex.

Solution:
```latex
\usepackage{iftex}
\ifPDFTeX {
    \usepackage[T1]{fontenc}
    \usepackage{mathpazo}
} \else {
    \usepackage{fontspec}
} \fi
```
This will continue to use fontspec even on lualatex too.

nbconvert/templates/latex/base.tplx
```diff
@@ -46,7 +44,6 @@ This template does not define a docclass, the inheriting class must define this.
     \usepackage{upquote} % Upright quotes for verbatim code
     \usepackage{eurosym} % defines \euro
     \usepackage[mathletters]{ucs} % Extended unicode (utf-8) support
-    \usepackage[utf8x]{inputenc} % Allow utf-8 characters in the tex document
```
t-makaro: Good catch. Not only is this ignored on XeLaTeX, the default for PDFLaTeX is now ```\usepackage[utf8]{inputenc}``` (since April 2018) which has been recommended over ```utf8x``` for a while now anyway.

nbconvert/templates/latex/base.tplx
njapke: This seems like a very good idea! A similar if structure could then also be used for `\usepackage[utf8]{inputenc}` to ensure compatibility across latex compilers.

t-makaro: That won't be necessary for inputenc. Inputenc is ignored in XeLaTeX and it defaults to utf8 in recent PDFLaTeX. See https://www.latex-project.org/news/latex2e-news/ltnews28.pdf and https://tex.stackexchange.com/questions/13067/utf8x-vs-utf8-inputenc.

> This means that with new documents one can assume UTF-8 input and it is no longer required to always specify \usepackage[utf8]{inputenc}. But if this line is present it will not hurt either.

njapke:
It seems that `iftex` is part of the apt package `texlive-generic-extra`, so that has to be added to Travis for the tests to work again.

nbconvert/templates/latex/base.tplx
njapke: Ah, I did not know that yet. That makes it a lot simpler.

KrokodileDandy:
I implemented the requested changes from @t-makaro and added the package `texlive-generic-extra` to the `.travis.yml` to allow the usage of these if-statements.

PR #1050, opened 2019-06-15
Fix RTD

t-makaro:
Readthedocs was failing because it wasn't finding the right version of nbconvert (related to #1048). This will ensure that the module near the docs is always found.

PR #1069, opened 2019-07-27
Pip-install nbconvert on readthedocs.org

mgeier:
This reverts #1050 and tries to install `nbconvert` via `pip` on https://readthedocs.org/.

WARNING: I don't know if this actually works!

I'm using this method on several of my RTD projects, but I don't use `conda`, so I don't know if that also works with `conda`.

But if it *does* work, it seems cleaner than hacking `sys.path`.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -18,3 +18,6 @@ build:
 
 python:
   version: 3.7
+  install:
+    - method: pip
+      path: .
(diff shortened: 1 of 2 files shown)


MSeal:
Hmm I don't manage the doc builds much for nbconvert (and also don't use conda much) and it's not clear to me how to test this without merging to a branch or master. From what I can test I think this should work fine. I'm going to merge and check if the doc build has any issues and fix if it needs something more specific.

Thanks for the improvement!

MSeal:
Looks like the docs-build is still working so I think this change was good.

meeseeksmachine:

PR #1074, opened 2019-07-30
Lock mistune version

lepture:
Mistune will release a 2.0 version in the future, which will break the API. I'll submit a patch for 2.0 when it is released.


meeseeksmachine:

PR #1076, opened 2019-07-30
Improve template whitespacing

### LINKED ISSUE #376: Markdown convert adds empty row as first row
When converting an notebook to markdown, nbconvert adds a blank row as the first row of the file. Is there a reason for this?

To be honest my reasons for not wanting this first row are totally selfish. I convert many notebooks to markdown to use in Pelican and Pelican's md->html convert requires the first line be the title of the blog post (e.g. `Title: Beautiful Soup Basic HTML Scraping`) which is currently annoying via nbconvert because I have to manually delete the first row after nbconverting, before running pelican.

### LINKED ISSUE #1059: allowing shebang `#!` when converting to script
My first notebook cell starts with a shebang, e.g.:
```
#!/usr/bin/env julia
versioninfo()
```

When I convert this notebook to a `--to=script` the output starts with a new line followed by the she bang:

```

#!/usr/bin/env julia
versioninfo()
```

However the newline prevents the shebang to be interpreted as such. The first new line (0x0a) should be omitted.

t-makaro:
Fixes #1059, and fixes #376. I chose this method because I noticed that the markdown exporter has the same issue which is problematic for people that need a title as the first thing in a markdown export.

I also made some improves to enhance readability of the latex exporter, but nothing will render differently due to this.

I also added a default documentclass to the ```base.tplx``` so that all of the ```style_*.tplx``` files can now be used on there own. This technically makes ```article.tplx``` redundant.

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -319,6 +319,8 @@ class TemplateExporter(Exporter):
 
         # Top level variables are passed to the template_exporter here.
         output = self.template.render(nb=nb_copy, resources=resources)
+        if resources.get('strip_preceding_newlines', True):
+            output = output.lstrip('\r\n')
         return output, resources
 
     def _register_filter(self, environ, name, jinja_filter):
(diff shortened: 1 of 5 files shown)


t-makaro:
I honestly can't envision ```strip_preceding_newlines``` ever needing to be false, but I thought that I'd add some way to disable it in case someone needs the old behaviour. In the interest of not adding excess config, it would probably be better to remove that check. If anyone ever needs it, we can add it later (it would also be possible to re-add the space in an inherited exporter with some work).

meeseeksmachine:

PR #1077, opened 2019-07-31
fix ansi2latex filtering

### LINKED ISSUE #1071: Colored Text Crashes PDF Export
Printing colored text results in latex conversion when exporting to pdf that fail to process by xelatex.

Executing and then exporting via `jupyter nbconvet --to=pdf ColorExportFail`
```
from colorama import Fore, Back, Style 
print(Fore.RED + 'red')
```

Raises an exception:
```
! FancyVerb Error:
  Empty verbatim environment
.
\FV@Error ...ncyVerb Error:^^J\space \space #1^^J}
                                                  
l.381 }\end{Verbatim}
```

Just exporting the latex via `jupyter nbconvet --to=latex ColorExportFail` produces
```
    \begin{Verbatim}[commandchars=\\\{\}]
\textcolor{ansi-red}{red
}\end{Verbatim}
```

Instead of
```
    \begin{Verbatim}[commandchars=\\\{\}]
\textcolor{ansi-red}{red}
\end{Verbatim}
```
The extra newline breaks exports.

[ColorExportFail.txt](https://github.com/jupyter/nbconvert/files/3439599/ColorExportFail.txt)

This is on current master (5.5.0)

t-makaro:
Alternative to #1073. Fixes #1071

This produces:
image

As opposed to what #1073 produces:
image

@MSeal I'll leave it to you to decide which option to choose.

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -58,6 +58,7 @@ default_filters = {
         'get_metadata': filters.get_metadata,
         'convert_pandoc': filters.convert_pandoc,
         'json_dumps': json.dumps,
+        'strip_trailing_newline': filters.strip_trailing_newline,
 }
 
 class ExtensionTolerantLoader(BaseLoader):
--- a/nbconvert/filters/ansi.py
+++ b/nbconvert/filters/ansi.py
@@ -71,7 +71,13 @@ def ansi2latex(text):
         Text containing ANSI colors to convert to LaTeX
 
     """
-    return _ansi2anything(text, _latexconverter)
+    fix_ending_space = text.endswith('\n')
+    if fix_ending_space:
+        text = text[:-1]
+    text = _ansi2anything(text, _latexconverter)
+    if fix_ending_space:
+        text += '\n'
+    return text
 
 
 def _htmlconverter(fg, bg, bold, underline, inverse):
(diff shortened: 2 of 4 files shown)


mgeier:
FWIW, I've solved this problem in `nbsphinx` by stripping `'\n'` from the end of all `stream` outputs:

https://github.com/spatialaudio/nbsphinx/blob/f09f4490525300d5c45213aab3d50d4da68ec40a/src/nbsphinx.py#L1247-L1250

I assume this would still need #1073 (which is a good idea anyway).

I think there is no perfect solution though, because the Jupyter notebook format has some inconsistencies regarding trailing `'\n'`.

njapke:
The results here look much nicer than mine in #1073, but I feel like trimming all trailing `\n` might be a better solution. The verbatim environment should be produced correctly without needing to trim whitespace in the templates.

t-makaro:
Alright, I adjusted to just remove that whitespace. I agree this is a cleaner solution, and it does appear to render as expected with my test pdfs.

mgeier:
@t-makaro When you think about it, it doesn't really make sense to add this functionality to a function named `ansi2latex()`. I think the `'\n'` should be removed before passing the string to `ansi2latex()`.

> I could create a separate filter for removing this newline and only apply it too the stream block.

Yes, this sounds more reasonable.

This will not change the result, but I think in the long run it is better to have functions that do what their name suggests and not arbitrary other things on top of that.

t-makaro:
Now implemented using a filter called: ```strip_trailing_newline```.

meeseeksmachine:

PR #1090, opened 2019-08-19
adding cell tag data attributes to HTML exporter

choldgraf:
In Jupyter Book most of my custom templates exist basically to insert the contents of cell tags into the `<div>`s metadata. I was thinking maybe this is something that would be useful to the HTMLExporter in general.

This PR modifies the base HTML template to do the following:

If a cell has any tags defined, then it adds a `data-cell-tags` attribute to the code or markdown cell. this attribute has a comma-separated string of tags.

This could then be used with selectors etc to do various things to the output HTML based on the tags content.

I thought about doing this with _classes_ instead of `data-` attributes, but it seemed that the former might cause unexpected behavior. If folks think this PR is a good idea, but that using classes would be better, I'm happy to revisit that.

### DIFF
--- a/nbconvert/templates/html/basic.tpl
+++ b/nbconvert/templates/html/basic.tpl
@@ -1,8 +1,8 @@
 {%- extends 'display_priority.tpl' -%}
-
+{% from 'celltags.tpl' import celltags %}
 
 {% block codecell %}
-<div class="cell border-box-sizing code_cell rendered">
+<div class="cell border-box-sizing code_cell rendered"{{ celltags(cell) }}>
 {{ super() }}
 </div>
 {%- endblock codecell %}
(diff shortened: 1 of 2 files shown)


MSeal:
I also think this is a good idea. I'd be on-board after a test or two to cover the feature. I think I agree with your assessment of using `data-` attributes

choldgraf:
I added a test to make sure the tags make it into the HTML as expected. Let's see if travis is happy!

choldgraf:
A quick thought here: what do you think about instead of data tags, adding these as classes but with a short prefix to make sure they don't clash with anything else. Something like `celltag_{tagname}`?

The reason I'm rethinking this is because if people are writing CSS rules using these tags (spoiler, I might be one of these people 😄), it can be more complex to select subsets of data tags etc, while it is straightforward to use tags.

Do you have any thoughts on that @MSeal? Happy to discuss...

PR #1101, opened 2019-09-21
Allow HTML header anchor text to be HTML

choldgraf:
In many projects it's common for anchor links to be icons (e.g., SVGs) or images embedded with something like FontAwesome. I've tried to get this working with nbconvert, but it supplying HTML for the anchor link text (e.g. `<i class='far fa-link'></i>`) gets sanitized into raw text instead of HTML.

This PR does a quick check to see if the supplied anchor text is valid HTML and, if so, appends it to the link object with the assumption that it should be treated as HTML. This lets you do things like add fontawesome icons for anchor links.

### DIFF
--- a/nbconvert/filters/strings.py
+++ b/nbconvert/filters/strings.py
@@ -44,10 +44,10 @@ __all__ = [
 
 
 def wrap_text(text, width=100):
-    """ 
+    """
     Intelligently wrap text.
     Wrap text without breaking words if possible.
-    
+
     Parameters
     ----------
     text : str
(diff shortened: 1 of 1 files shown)


nbconvert/filters/strings.py
```diff
@@ -44,10 +44,10 @@
 
 
 def wrap_text(text, width=100):
-    """ 
+    """
```
choldgraf: All these whitespace fixes were auto-made by my editor, I can try to revert them if it's noise

choldgraf:
build error is on the parallel notebooks job, which I don't *think* is related to this...

nbconvert/filters/strings.py
MSeal: All good imo. I do the same when touching files.

PR #1107, opened 2019-10-01
fix pip extra install

00Kai0:
like https://github.com/jupyter/notebook/pull/4933


MSeal:
The quotes shouldn't be necessary to execute those commands. Was the instructions failing on a particular setup?

takluyver:
From the linked PR, it looks like zsh doesn't like that syntax.

Issue #1108, opened 2019-10-05
Current latex install command for ubuntu is not enough to export to PDF


fbidu:
Hello!

Currently, the docs offers the following command to install Latex for Ubuntu systems:

`sudo apt-get install texlive-xetex`

While this is on the right track, it misses some dependencies. Namely, it will first generate an error about missing the the `ecrm1095` file:

`I can't find file `ecrm1095'.`

This can be solved by installing the `texlive-fonts-recommended` package.

It will then comply about missing `ulem.sty`. This can be solved by installing `texlive-generic-recommended`.

Both of those packages are needed and I think they shold be referenced in the docs. I'll issue a PR adding them 

**Nbconvert version:**

fbidu:
#1109 closed this

PR #1109, opened 2019-10-05
Fix latex dependencies installation command for Ubuntu systems

fbidu:
As put in issue #1108, there are two additional packages needed
for Ubuntu systems in order to fully supply all the necessary
dependencies for PDF exportation.


MSeal:
Thanks!

PR #1112, opened 2019-10-09
add support for alt tags for jpeg and png images

hx2A:
Accessible web content needs to have proper 'alt' text for images. Currently, nbconvert will create html pages without any way for the user to add alt text if they wanted to. With this pull request, I added support for alt text for jpeg and png images.

The first way to use it is by adding to the output metadata, like so:

```
    from IPython.display import Image

    Image('images/some_image.jpg', metadata={'alt': 'description of image'})
```

This approach doesn't really work for matplotlib plots, however. Instead, the user must first create the plot:

```
    %matplotlib inline
    import matplotlib.pyplot as plt

    plt.plot([0, 2, 1])
```

And then set the cell's metadata to this:

````
    {
        "alt": "description of plot"
    }
````

Note: use View => Cell Toolbar => Edit Metadata to bring up a button to edit a cell's metadata.

Although writing proper alt text for complicated charts is challenging, there are resources explaining how it can and should be done:

http://diagramcenter.org/specific-guidelines-e.html

### DIFF
--- a/nbconvert/templates/html/basic.tpl
+++ b/nbconvert/templates/html/basic.tpl
@@ -155,6 +155,10 @@ height={{ height }}
 {%- if output | get_metadata('unconfined', 'image/png') %}
 class="unconfined"
 {%- endif %}
+{%- set alttext=(output | get_metadata('alt', 'image/png')) or (cell | get_metadata('alt')) -%}
+{%- if alttext is not none %}
+alt="{{ alttext }}"
+{%- endif %}
 >
 </div>
 {%- endblock data_png %}
@@ -176,6 +180,10 @@ height={{ height }}
 {%- endif %}
 {%- if output | get_metadata('unconfined', 'image/jpeg') %}
 class="unconfined"
+{%- endif %}
+{%- set alttext=(output | get_metadata('alt', 'image/jpeg')) or (cell | get_metadata('alt')) -%}
+{%- if alttext is not none %}
+alt="{{ alttext }}"
 {%- endif %}
 >
 </div>

### CODE REVIEW

File: nbconvert/templates/html/basic.tpl [hunk 1]
```diff
@@ -155,6 +155,10 @@ height={{ height }}
 {%- if output | get_metadata('unconfined', 'image/png') %}
 class="unconfined"
 {%- endif %}
+{%- set alttext=(output | get_metadata('alt', 'image/png')) or (cell | get_metadata('alt')) -%}
```
MSeal: What if the cell has multiple images? Are we ok with all images getting the `cell | get_metadata('alt')`?

hx2A: If a cell has multiple image outputs, the alt-text should be added to the output's metadata, not the cell's metadata. But there's no way to edit an output's metadata using Jupyter notebook or Jupyter lab. We can only edit the cell's metadata. We can use `IPython.display.Image` to programmatically set the output metadata, but that doesn't help people creating figures with `matplotlib` because you can't set the output metadata through `matplotlib`, and even if you could, not everyone would want to type out the alt-text in their code like that.

People who want or need to add alt-text to their notebooks for accessibility reasons can limit their cell outputs to one figure per cell, which isn't much of a limitation since that's what a lot of people do anyway. And without this change, all notebooks have zero alt-text added to figures and images, making Jupyter notebook blog posts less accessible to the blind or low vision community.

MSeal: Yeah I think that's a fair compromise. Just wanted to make sure it was thought out.

Thanks for the improvement!

hx2A: You are welcome!

I'd also like to update the documentation somewhere so people know about this feature. Can you advise how I should go about that?  Which github project should I update and where should the documentation go?

MSeal: It's this repo -- you can find to root of the docs here: https://github.com/jupyter/nbconvert/tree/master/docs

PR #1113, opened 2019-10-13
multiline search in RegexRemovePreprocessor

syoh:
Currently, `RegexRemovePreprocessor` does not match multiline code.

* Added `re.MULTILINE` to `re.compile`
* Changed `match` to `search` due to https://docs.python.org/3/library/re.html#search-vs-match

```python
import re
somepat = '(?:^hellothere)'
somestr = '\nhellothere'
print("match (master)      :", re.compile(somepat).match(somestr))
print("search w/o MULTILINE:", re.compile(somepat).search(somestr))
print("search w/  MULTILINE:", re.compile(somepat, re.MULTILINE).search(somestr))
```

### DIFF
--- a/nbconvert/preprocessors/regexremove.py
+++ b/nbconvert/preprocessors/regexremove.py
@@ -50,10 +50,11 @@ class RegexRemovePreprocessor(Preprocessor):
         # by a non-capturing group to ensure the correct order of precedence
         # and the patterns are joined with a logical or
         pattern = re.compile('|'.join('(?:%s)' % pattern
-                             for pattern in self.patterns))
+                             for pattern in self.patterns),
+                             re.MULTILINE)
 
         # Filter out cells that meet the pattern and have no outputs
-        return not pattern.match(cell.source)
+        return not pattern.search(cell.source)
 
     def preprocess(self, nb, resources):
         """


MSeal:
Hmm, the `TestRegexRemove.test_output` test seems to point out that current regex patterns around whitespace matches might be broken by this change. I'm ok with this change so long as the tests are updated and we merge / target the 6.0 release as I think the pattern change is more intuitive.

MSeal:
That's a good point. It might make sense to leave it as and just document how to use multiline patterns

syoh:
Thanks for pointing that out. I missed `(?m)` could be used in place of MULTILINE. (Removing re.MULTILINE may help to pass the tests)

How about `match` vs. `search`? Even when using the inline option, using `match` doesn't seem to work for me. Here is what I used:

```python
import re
somepat = '(?m:^hellothere)'
somestr = '\nhellothere'
print("match (master)      :", re.compile(somepat).match(somestr))
print("search w/o MULTILINE:", re.compile(somepat).search(somestr))
print("search w/  MULTILINE:", re.compile(somepat, re.MULTILINE).search(somestr))
```
The result is:

```python
match (master)      : None
search w/o MULTILINE: <re.Match object; span=(1, 11), match='hellothere'>
search w/  MULTILINE: <re.Match object; span=(1, 11), match='hellothere'>
```
Is there a different way to call `match` to make multiline search work?

> I don't know whether adding the flag `re.MULTILINE` is the right thing to do or not, but I just wanted to mention, in case people are not aware of it:
> 
> The "multiline" flag can also be set from within the regex string itself using `(?m)`, see https://docs.python.org/3/library/re.html#regular-expression-syntax.

mgeier:
> Is there a different way to call `match` to make multiline search work?

It *does* work as it's shown above, the given string is just not a match for the given pattern!

The first character of the pattern (`^`) is successful, because (by definition) the string starts with the beginning of a line. The problem is that the following character (`\n`) doesn't match the next part of the pattern (`h`).

Using `^` at the beginning of a pattern with `match()` is redundant, because `match()` only checks the beginning of the string anyway.
The idea of the "multiline" option is that you can use `^` in the *middle* of the pattern, e.g.

```
>>> re.match('(?m)\n^hellothere', somestr)
<re.Match object; span=(0, 11), match='\nhellothere'>
```

In this example, however, both the `^` character and the `(?m)` flag are redundant.

Coming back to the original idea of this PR, I have the feeling (but I'm not 100% sure) that people would expect the behavior of `search()` instead of `match()`, but hard-coding `re.MULTILINE` isn't necessary. Latter might even be harmful, because people might expect `^` to match only at the beginning of the string.

mgeier:
> Hmm, the `TestRegexRemove.test_output` test seems to point out that current regex patterns around whitespace matches might be broken by this change.

Yes, changing `match()` to `search()` is a breaking change.

syoh:
> > Is there a different way to call `match` to make multiline search work?
> 
> It _does_ work as it's shown above, the given string is just not a match for the given pattern!
> 
> The first character of the pattern (`^`) is successful, because (by definition) the string starts with the beginning of a line. The problem is that the following character (`\n`) doesn't match the next part of the pattern (`h`).
> 
> Using `^` at the beginning of a pattern with `match()` is redundant, because `match()` only checks the beginning of the string anyway.
> The idea of the "multiline" option is that you can use `^` in the _middle_ of the pattern, e.g.
> 
> ```
> >>> re.match('(?m)\n^hellothere', somestr)
> <re.Match object; span=(0, 11), match='\nhellothere'>
> ```
> 
> In this example, however, both the `^` character and the `(?m)` flag are redundant.
> 
> Coming back to the original idea of this PR, I have the feeling (but I'm not 100% sure) that people would expect the behavior of `search()` instead of `match()`, but hard-coding `re.MULTILINE` isn't necessary. Latter might even be harmful, because people might expect `^` to match only at the beginning of the string.

Ha! That is interesting.

So, if I understand correctly, two patterns are needed to match `hellothere` on the first line and subsequent lines. Is there another way to write this?

Sorry if the solution is obvious. I find regex challenging!

For example:
```python
import re
somestr1 = 'hellothere'
somestr2 = '\nhellothere'
print(re.match('(?m)^hellothere', somestr1))
print(re.match('(?m)^hellothere', somestr2))
print(re.match('(?m)\n^hellothere', somestr1))
print(re.match('(?m)\n^hellothere', somestr2))
```
gives the following output
```python
<re.Match object; span=(0, 10), match='hellothere'>
None
None
<re.Match object; span=(0, 11), match='\nhellothere'>
```

mgeier:
You don't need two patterns, but you need a different pattern:

```python
print(re.match('(?ms).*^hellothere', somestr1))
print(re.match('(?ms).*^hellothere', somestr2))
```

This uses the `re.S` flag in addition to the "multiline" flag: https://docs.python.org/3/library/re.html#re.S

There are of course many more patterns that match both your example strings, e.g.

```python
print(re.match('\n?hellothere', somestr1))
print(re.match('\n?hellothere', somestr2))
```

> Sorry if the solution is obvious. I find regex challenging!

They are indeed challenging!

Luckily, there are many tutorials available and there are nice online tools that let you experiment with regexes.

nbconvert/preprocessors/regexremove.py
```diff
@@ -50,10 +50,11 @@ def check_conditions(self, cell):
         # by a non-capturing group to ensure the correct order of precedence
         # and the patterns are joined with a logical or
         pattern = re.compile('|'.join('(?:%s)' % pattern
```
syoh: Users can only specify `%s` portion of a pattern and not the whole pattern

syoh:
@mgeier I came across another challenge in using a more sophisticated regexp pattern such as ones you suggested.
Currently, users can only specify a subpattern: https://github.com/jupyter/nbconvert/pull/1113/files#r349946204

Is there a user-specified way around this?

> You don't need two patterns, but you need a different pattern:
> 
> ```python
> print(re.match('(?ms).*^hellothere', somestr1))
> print(re.match('(?ms).*^hellothere', somestr2))
> ```
> 
> This uses the `re.S` flag in addition to the "multiline" flag: https://docs.python.org/3/library/re.html#re.S
> 
> There are of course many more patterns that match both your example strings, e.g.
> 
> ```python
> print(re.match('\n?hellothere', somestr1))
> print(re.match('\n?hellothere', somestr2))
> ```
> 
> > Sorry if the solution is obvious. I find regex challenging!
> 
> They are indeed challenging!
> 
> Luckily, there are many tutorials available and there are nice online tools that let you experiment with regexes.

mgeier:
@syoh Thanks for the information, I didn't know that `nbconvert` concatenates the given patterns into one large pattern.

I'm not sure, but this might not be a good idea.

It would probably be better to keep the patterns separate and iterate over them when running the preprocessor.

I didn't measure this, but I guess this wouldn't be any slower.
It might even be faster since there might be less back-tracking?

> Is there a user-specified way around this?

It might work but it looks like it will not work in the future.

When using a flag like `(?m)` I get this warning:

```pycon
>>> re.match('(?:(?m)something)', somestr)
__main__:1: DeprecationWarning: Flags not at the start of the expression '(?:(?m)something)'
```

Did you try using this with `nbconvert`?

MSeal:
I'm going to close this issue for now since there's been no activity for a while and no clear changes to be made.

PR #1114, opened 2019-10-14
Create docs on CircleCI

mgeier:
These are the artifacts (created on my CircleCI project https://app.circleci.com/github/mgeier/nbconvert/pipelines):

https://22-49829768-gh.circle-artifacts.com/0/html/index.html
https://22-49829768-gh.circle-artifacts.com/0/nbconvert.pdf

Somebody will have to switch on CircleCI for this to work in this repo.

PR #1120, opened 2019-10-19
CircleCI: Use conda root environment, no need to "activate"

mgeier:
I didn't like having to call "activate" before each step in my original setup (see #1114).

By installing everything in the "root" environment, we can avoid this.
We also have to save the `conda` paths etc. in `BASH_ENV`, for them to be made available in each new task.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,8 +12,8 @@ jobs:
     working_directory: ~/checkout
 
     environment:
+      BASH_ENV: /root/.bashrc
       MINICONDA_INSTALLER: https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh
-      ACTIVATE: source /root/miniconda/bin/activate nbconvert_docs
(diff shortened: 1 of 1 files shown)


mgeier:
That's the CircleCI log: https://app.circleci.com/jobs/github/mgeier/nbconvert/36

MSeal:
Running CI tooling as root is generally best avoided if possible, but since the conda env is already owned by root I don't see a problem changing it. Probably better long-term to use a localized environment for the user running the tests, but I'll merge it for now.

mgeier:
Thanks for merging!

TBH, I don't know whether and how I can change the user. I guess that depends on the used image, right?

I didn't think about changing the user and I didn't notice anyone changing the user in the examples I was consulting.

If you think pursuing this path further is worthwhile  please let me know.

MSeal:
Nah don't worry about it for now. If we get warnings on our builds then we should go take a look again later.

PR #1124, opened 2019-10-22
CircleCI: Don't use per-branch caching

mgeier:
I have the feeling that this would be unnecessary caching.

The cache only contains `miniconda` and the `conda` environment, and this only changes when `environment.yml` changes and it should be meaningful across branches.


MSeal:
Agreed,

PR #1142, opened 2019-11-13
Update nbconvert template

SylvainCorlay:
This upstreams a change that we made in the Voilà version of that template and is needed to filter out empty cells.


SylvainCorlay:
Will be self-merging soon if noone objects.

PR #1149, opened 2019-12-06
Correct the location of the javascript output element

### LINKED ISSUE #1148: Reveal.js does not render `Javascript` output correctly
When exporing a cell that contains an Ipython `Javascript` object, the result is:
```html
<div class="output_area">
  <div class="prompt"></div>
  <div id="299f148f-675b-44be-88c1-0da119ddd28b"></div>
  <div class="output_subarea output_javascript ">
    <script type="text/javascript">
    var element = $('#299f148f-675b-44be-88c1-0da119ddd28b');
    // script ends up here
    </script>
  </div>
</div>
```

It should have the `299f148f-675b-44be-88c1-0da119ddd28b` div within the output_subarea element
```html
<div class="output_area">
  <div class="prompt"></div>
  <div class="output_subarea output_javascript ">
    <div id="299f148f-675b-44be-88c1-0da119ddd28b"></div>  <!-- This moved -->
    <script type="text/javascript">
    var element = $('#299f148f-675b-44be-88c1-0da119ddd28b');
    // script ends up here
    </script>
  </div>
</div>
```

**Nbconvert version:** 5.6.1

eric-wieser:
Fixes #1148 

I assume this issue also applies to the jupyter widgets, so moved those too.

Without this change, my javascript element is rendered without padding, and compressed by an empty `output_javascript` flexbox:

image

cc @hugohadfield whose https://github.com/pygae/pyganja is outputing the `Javascript`

### DIFF
--- a/share/jupyter/nbconvert/templates/base/base.html.j2
+++ b/share/jupyter/nbconvert/templates/base/base.html.j2
@@ -217,8 +217,8 @@ alt="{{ alttext }}"
 
 {%- block data_javascript scoped %}
 {% set div_id = uuid4() %}
-<div id="{{ div_id }}"></div>
 <div class="output_subarea output_javascript {{ extra_class }}">
+<div id="{{ div_id }}"></div>
 <script type="text/javascript">
 var element = $('#{{ div_id }}');
 {{ output.data['application/javascript'] }}
@@ -230,8 +230,8 @@ var element = $('#{{ div_id }}');
 {% set div_id = uuid4() %}
 {% set datatype_list = output.data | filter_data_type %} 
 {% set datatype = datatype_list[0]%} 
-<div id="{{ div_id }}"></div>
 <div class="output_subarea output_widget_state {{ extra_class }}">
+<div id="{{ div_id }}"></div>
 <script type="text/javascript">
 var element = $('#{{ div_id }}');
 </script>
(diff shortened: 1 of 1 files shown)


eric-wieser:
To be clear @Mseal, this is moving the uuid container a level down - it's a leaf node, so nothing else moved.

Or are you suggesting an additional change?

maartenbreddels:
I'd like to take a look as well in on Monday. I find it odd this worked before, and I don't think I modified the structure, so maybe something went wrong in the template refactor.

eric-wieser:
If it helps, I needed to make a similar change in my local install of 5.6.x for this to work - so the 6.0.0 changes aren't relevant

MSeal:
Thanks @maartenbreddels and @eric-wieser. I'll let Maarten take a look and merge after he +1's then.

maartenbreddels:
I don't think this is the ideal solution either. To mimic the classical notebook, `element` should be the same DOM element as the one with the `output_subarea output_javascript` class. I think we should remove that extra div and put the id on *that* DOM element.

I'll open a seperate PR for this. Also, in master, we removed jquery in the classical notebook, this is a breaking change, and I'll revert that.
Note that in the lab template we do not have jquery (since jquery is past its end, and will not be available in jupyter lab either).

For the lab template, we now use `element_id`, but it would be easier to just have the same variable point to the DOM node instead, so runtime checks are easier.

PR's coming today.

maartenbreddels:
I prefer #1152 to this solution, so if we merge that one, this can be closed.

@eric-wieser Thanks for this report, it cleared up something that was bugging me for a while!

MSeal:
@eric-wieser Thanks for helping identify the issue. I merged maarten's which seemed to cover more things from the base issue on inspection. Do you think there's more that needs fixed / is this PR still needed for anything additional you saw?

PR #1152, opened 2019-12-13
Fix jquery and element var

### LINKED ISSUE #1148: Reveal.js does not render `Javascript` output correctly

maartenbreddels:
Fixes #1148 and alternative to #1149.

jQuery support was removed in #1056 based on what we've done with voila. However, since we now have the two template (classic and lab) it makes sense that classic stays as compatible as it is with the classical notebook.

Therefore I've restored the jQuery library, and made the template exposes the `element` variable again (used for js library/code).

However, we want to get rid of jQuery, so in the lab template, we simply point the `element` variable to the DOM Node, which has an API similar, meaning the example library mentioned in #1149 works with both the classical and lab template.

The original issue in #1148 was solved by removing the unnecessary div element, and point `element` to the same DOM element as what would happen in the classical notebook.

### DIFF
--- a/share/jupyter/nbconvert/templates/base/base.html.j2
+++ b/share/jupyter/nbconvert/templates/base/base.html.j2
@@ -248,7 +248,7 @@ var element = $('#{{ div_id }}');
 <div id="{{ div_id }}"></div>
 <div class="output_subarea output_widget_view {{ extra_class }}">
 <script type="text/javascript">
-var element_id = '#{{ div_id }}';
+var element = $('#{{ div_id }}');
 </script>
 <script type="{{ datatype }}">
 {{ output.data[datatype] | json_dumps }}
--- a/share/jupyter/nbconvert/templates/base/base.html.j2
+++ b/share/jupyter/nbconvert/templates/base/base.html.j2
@@ -217,8 +217,7 @@ alt="{{ alttext }}"
 
 {%- block data_javascript scoped %}
 {% set div_id = uuid4() %}
-<div id="{{ div_id }}"></div>
-<div class="output_subarea output_javascript {{ extra_class }}">
+<div id="{{ div_id }}" class="output_subarea output_javascript {{ extra_class }}">
 <script type="text/javascript">
 var element = $('#{{ div_id }}');
 {{ output.data['application/javascript'] }}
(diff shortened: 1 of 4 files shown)


share/jupyter/nbconvert/templates/base/base.html.j2
```diff
@@ -245,10 +243,9 @@ var element = $('#{{ div_id }}');
 {% set div_id = uuid4() %}
 {% set datatype_list = output.data | filter_data_type %} 
 {% set datatype = datatype_list[0]%} 
-<div id="{{ div_id }}"></div>
-<div class="output_subarea output_widget_view {{ extra_class }}">
+<div id="{{ div_id }}" class="output_subarea output_widget_view {{ extra_class }}">
 <script type="text/javascript">
-var element_id = '#{{ div_id }}';
+var element = $('#{{ div_id }}');
```
eric-wieser: I was under the impression that jupyter added this automatically as part of the script below. I'll try to confirm that by running this branch locally

maartenbreddels: that would be great if you could test it. But this part was only for widgets, for JS it was still correct.

eric-wieser: Ah, I sadly have no experience (yet) with widgets

PR #1153, opened 2019-12-16
Group \prompt command in jupyter latex template

### LINKED ISSUE #1151: LaTeX output cell breaks markdown rendering in LaTeX/PDF
When a code cell outputs LaTeX, all following markdown cells will use typewriter font.

An [example notebook](https://gist.github.com/mstimberg/d6087a0e7f0cd4611f309780d98a48f6) renders like this in PDF:

screenshot_nbconvert_latex_issue

Note that the markdown cell after the code output uses a typewriter font, even though the markdown code does not ask for it.

The example uses the `sympy` package to generate LaTeX output, but any other package generating LaTeX output should be fine (using `IPython.display.display_latex` directly will generate a cell of a different type and therefore not show the issue).

From looking at the LaTeX code, the issue seems that for non-latex output, the `\prompt` command displaying the `[1]` is wrapped in a `tcolorbox` environment, whereas for latex output it isn't. Since `\prompt` switches the font to `\ttfamily`, this affects all the following text not explicitly setting a font.

This was tested in a Python 3.8 conda environment, but with using `pip install -e '.[all]'` in the master branch of `nbconvert`.

**Nbconvert version:** 6.0.0a0.dev0

mstimberg:
Fixes #1151 by preventing `\ttfamily` to affect later markdown cells.

### DIFF
--- a/share/jupyter/nbconvert/templates/latex/style_jupyter.tex.j2
+++ b/share/jupyter/nbconvert/templates/latex/style_jupyter.tex.j2
@@ -103,7 +103,7 @@
     \makeatother
     ((*- block style_prompt *))
     \newcommand{\prompt}[4]{
-        \ttfamily\llap{{\color{#2}[#3]:\hspace{3pt}#4}}\vspace{-\baselineskip}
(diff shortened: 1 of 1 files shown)


t-makaro:
LGTM! Thank you!

Issue #1160, opened 2020-01-02
What to determine failure of cell execution: error iopub message or execute_reply with error status


BoPeng:
There has been some discussions in the context of papermill what should be used to determine the failure of cell execution (https://github.com/nteract/papermill/issues/337), which referred to the following trunk of code

https://github.com/jupyter/nbconvert/blob/b31a5af48ee159f579c021b897727374f28a0800/nbconvert/preprocessors/execute.py#L475-L479 

That is to say, nbconvert considers a cell to be failed if a kernel sends an `error` message through `ioPub`, **OR** if the kernel returns `execute_reply` with a `status` of `error`. In the meantime, papermill only uses `error` message, not `execute_reply` (https://github.com/nteract/papermill/blob/0ee0da7fe7a9e8a8334011e1b1606fa86bd0ed8e/papermill/preprocess.py#L58).

I believe `execute_reply` should be the only message used because this is how Jupyter behaves with "Run all cells". That is to say, papermill should stop if a kernel sends an `execute_reply` with `error` status (but without an `error` message, https://github.com/nteract/papermill/issues/337) , and neither papermill nor nbconvert should stop if a kernel sends an `error` message, and an `execute_reply` message with `ok` status. My argument for the latter is that a kernel might want to sends an `error` message to **demonstrates or document an error** and still returns an `execute_reply` with an `ok` status so that the document could be executed successfully in batch mode.

MSeal:
@mpacer @minrk Would like your input, but I believe @BoPeng's logic is correct and behavior should change in nbconvert 6.0 (nbexecute) and papermill should match that behavior.

mpacer:
I think that this is correct logic, however I don't know if we can say clearly one way or the other based on the protocol itself: https://jupyter-client.readthedocs.io/en/stable/messaging.html#execution-errors

> When an error occurs during code execution

> 
> Message type: error:
> 
> ```
> content = {
>    # Similar content to the execute_reply messages for the 'error' case,
>    # except the 'status' field is omitted.
> }
> 
> ```

> Changed in version 5.0: pyerr renamed to error

Given that the semantics are specifically pointed out as being similar I think there's an argument in both directions. 

I get that idea for a use case also imagine having a case where an error comes in from IOPub and it assumes that any error that is raised should immediately exit and by continuing computation state/data/&c. becomes corrupted. Of the two, I think nbconvert's behaviour is the more conservative of the two options.

[...]

PR #1163, opened 2020-01-08
Stop raising CellExecutionError with error ioPub message

BoPeng:
https://github.com/jupyter/nbconvert/issues/1160

Stop raising `CellExecutionError` if the cell output contains an `error` message but  returns `execute_reply` with an `ok` status. This makes nbconvert consistent with the batch-execution behavior (`Run all cells`) of classic Jupyter and Jupyter Lab.

@MSeal Is this good enough?

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -472,9 +472,6 @@ class ExecutePreprocessor(Preprocessor):
                               in cell.metadata.get("tags", []))
 
         if self.force_raise_errors or not cell_allows_errors:
-            for out in cell.outputs:
-                if out.output_type == 'error':
-                    raise CellExecutionError.from_cell_and_msg(cell, out)
             if (reply is not None) and reply['content']['status'] == 'error':
                 raise CellExecutionError.from_cell_and_msg(cell, reply['content'])
         return cell, resources


BoPeng:
It is not easy to create a test case because for example a `Python 3` kernel would always produce an `error` message in addition to `execute_reply` when we run `nbconvert` over a notebook with error. The fact that this PR does not break any existing test confirms this.

MSeal:
This would be more to test the run_cell function, not the end-to-end behavior. Clearly those tests have a gap around testing error processing since changing the logic didn't break any tests. If it's not obvious how to add such a test to https://github.com/jupyter/nbconvert/blob/master/nbconvert/preprocessors/tests/test_execute.py#L560 I can add the test to this PR in a few days.

MSeal:
Also thanks for raising and making the PR, really appreciate have people contribute changes!

BoPeng:
Thanks. I do not have much time in the next few days so I will leave the tests to you. You are more familiar with the test suite anyway.

MSeal:
I couldn't push the tests to your master branch, so I merged and pushed the tests directly in https://github.com/jupyter/nbconvert/commit/16d696b2eb9c1969cb0f3205931aaa779302042d

BoPeng:
Great. I will then see what can be done at the papermill side.

PR #1173, opened 2020-01-21
Make template system backward compatible

### LINKED ISSUE #1158: Extending `rst.tpl` fails in current master and 6.0.0a0
Extending `rst.tpl` with a DictLoader has worked up until the 6.0.0a0 alpha release, but now fails, in 6.0.0a0 as in current master (b31a5af4).

```python
import nbconvert
import nbformat
from jinja2 import DictLoader

DL = DictLoader({'rst_plus_stuff.tpl': """\
{%- extends 'rst.tpl' -%}
"""})


    template_file = 'rst_plus_stuff.tpl'

exporter = MyExporter(extra_loaders=[DL])

nb = nbformat.v4.new_notebook()
output, resources = exporter.from_notebook_node(nb)
```

This succeeds on current stable on pip (5.6.1), but fails for current master, with:

```
  File "/Users/mb312/tmp/nbconvert_check.py", line 17, in <module>
    output, resources = plots_exporter.from_notebook_node(nb)
  File "/Users/mb312/dev_trees/nbconvert/nbconvert/exporters/templateexporter.py", line 357, in from_notebook_node
    output = self.template.render(nb=nb_copy, resources=resources)
  File "/Users/mb312/.virtualenvs/test3/lib/python3.7/site-packages/jinja2/asyncsupport.py", line 76, in render
    return original_render(self, *args, **kwargs)
  File "/Users/mb312/.virtualenvs/test3/lib/python3.7/site-packages/jinja2/environment.py", line 1008, in render
    return self.environment.handle_exception(exc_info, True)
  File "/Users/mb312/.virtualenvs/test3/lib/python3.7/site-packages/jinja2/environment.py", line 780, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/mb312/.virtualenvs/test3/lib/python3.7/site-packages/jinja2/_compat.py", line 37, in reraise
  File "<template>", line 1, in top-level template code
jinja2.exceptions.TemplateNotFound: rst.tpl
```

**Nbconvert version:**
* 6.0.0a0
* Current master (b31a5af4)

maartenbreddels:
Fixes #1158

This also makes nbconvert 6.0.0 compatible with voila (0.1.20 as of writing), and I guess many more projects.

The basic idea is to add templates with the old names in the compatibility subdirectory, which will be added to the search path. I've done this for `rst.tpl` and `display_priority.tpl`, but if people like this I can do it for the other old template filenames as well.

I think it does not bring an extra dev burden for this project while staying backward compatible. 

Should we try to make it fully backward compatible, including running tests?

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -10,6 +10,7 @@ from __future__ import print_function, absolute_import
 import os
 import uuid
 import json
+import warnings
 
 from jupyter_core.paths import jupyter_path
 from traitlets import HasTraits, Unicode, List, Dict, Bool, default, observe
@@ -535,3 +536,11 @@ class TemplateExporter(Exporter):
         root_dirs.extend(self.template_path)
         root_dirs.extend(jupyter_path())
         return root_dirs
+
+    def _init_resources(self, resources):
+        resources = super()._init_resources(resources)
+        # inline function to avoid pickle errors
+        def deprecated(msg):
+            warnings.warn(msg, DeprecationWarning)
+        resources['deprecated'] = deprecated
+        return resources
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -482,6 +482,9 @@ class TemplateExporter(Exporter):
             base_dir = os.path.join(root_dir, 'nbconvert', 'templates')
             paths.append(base_dir)
 
+            compatibility_dir = os.path.join(root_dir, 'nbconvert', 'templates', 'compatibility')
+            paths.append(compatibility_dir)
+
         additional_paths = self.template_data_paths
         for path in additional_paths:
             try:
--- a/nbconvert/exporters/tests/test_rst.py
+++ b/nbconvert/exporters/tests/test_rst.py
@@ -4,6 +4,7 @@
 # Distributed under the terms of the Modified BSD License.
 
 import io
+import pytest
 import re
 
 import nbformat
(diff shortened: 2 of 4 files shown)


MSeal:
So I would like to not indefinitely support old patterns if we can avoid it. I think this is a good idea for the 6.0.0 release but it'd be nice if we could emit a deprecation warning stating the new path they should be using. Thoughts?

MSeal:
Pinging to refresh inboxes on PR conversation

maartenbreddels:
> So I would like to not indefinitely support old patterns if we can avoid it.

It would make many tools just work, it's 1 lines per template, I think it would make library owners happy. 

We could add the templates, indeed, with a warning, but only remove it in version 7?

Or we do the most popular only? html/rst?

MSeal:
> It would make many tools just work, it's 1 lines per template, I think it would make library owners happy.

I agree

> We could add the templates, indeed, with a warning, but only remove it in version 7?

If it's easy to add the warning and have an extra line or two per template that does sound ideal. Maybe we should pick a release sooner than 7 to remove it (who know how many years that is)? 6.2? 6.3? Or maybe we just say no sooner than 6.2 to leave it open to extending removal?

maartenbreddels:
I would say, we keep it to this, and we can add support for more old template names on a request/PR basis, using this as an example PR. This gives us some feedback on how people rely on it, and how they feel about migrating.

MSeal:
Also docs build needs a fix -- some linting error in sphinx that's unrelated to this

PR #1175, opened 2020-01-22
Fix memory leak in execute preprocessor

skizunov:
There is a memory leak in the `ExecutePreprocessor` when using a kernel
manager. `KernelClient` (from jupyter_client) does not have a `__del__`
that invokes `stop_channels()`. So if `stop_channels()` is not
invoked explicitly and the `self.kc` object goes away, a memory leak
will occur.

Signed-off-by: Sergey Kizunov <sergey.kizunov@ni.com>


MSeal:
Yep, good catch. Thanks for the fix!

PR #1176, opened 2020-01-28
Update Inkscape command line options in svg2pdf

18praveenb:
With Inkscape version "Inkscape 1.0beta2 (2b71d25, 2019-12-03)" and nbconvert version 5.6.1, svg2pdf failed with Unknown option --export-pdf

Changing to --export-file fixed this issue.


MSeal:
Ahh, thanks for finding that fix. I added some checks so it will work for both pre and post inkscape 1.0 versions.

MSeal:
Going to merge, if this gets reviewing during the 6.0 release prep I can make additional changes then

t-makaro:
LGTM!

PR #1179, opened 2020-02-02
line1 typo in latex/style_ipython.tex.j2

phaustin:


MSeal:
@maartenbreddels @SylvainCorlay Can you two review this? I'm less familiar with the expectations in tex files than you two are.

SylvainCorlay:
Hum, I really don't know about this. @Carreau may be able to give an educated review.

phaustin:
pretty sure it's just a stray editing error -- without this patch that fragment is injected at the start of the tex file and latex won't run.

phaustin:
Loosely related -- we're working on using paged.js and chrome headless to remove the need for latex (similar to what R is doing with bookdown).  See https://github.com/eoas-ubc/paged_trio/blob/master/Readme.md for some links

Carreau:
Don't remember seeing this file before, but the fix seem good to me. Thanks !

PR #1181, opened 2020-02-04
Fix error in docstring => latex/pdf export by converting ansi escapes

telamonian:
fixes jupyterlab/jupyterlab#4447

Currently, if you have a cell in a notebook with docstring output:

image

and you try to export it to pdf, you'll get an error:

image

Here's the actual error text

```
! Text line contains an invalid character.
l.403 ^^[
         [0;31mSignature:^^[[0m ^^[[0msum^^[[0m^^[[0;34m(^^[[0m^^[[0miterabl...

? 
! Emergency stop.
l.403 ^^[
         [0;31mSignature:^^[[0m ^^[[0msum^^[[0m^^[[0;34m(^^[[0m^^[[0miterabl...

No pages of output.
```

I've fixed this by handling `data_text` blocks (which are used for docstring output) in the latex template the same way as `stream` blocks (which are used for regular `stdout` and such): any potential latex command sequences are escaped, then the ansi codes are converted to latex.

If you want to reproduce the error, you'll have to use Jupyterlab, since notebook classic outputs docstrings to a separate pager area, rather than including them in cell output.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,4 +30,6 @@ docs/source/config_options.rst
 .settings
 
 # VSCode
+*.code-workspace
+.history
 .vscode
--- a/share/jupyter/nbconvert/templates/latex/document_contents.tex.j2
+++ b/share/jupyter/nbconvert/templates/latex/document_contents.tex.j2
@@ -6,9 +6,9 @@
 
 % Displaying simple data text
 ((* block data_text *))
-    \begin{verbatim}
-((( output.data['text/plain'] )))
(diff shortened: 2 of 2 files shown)


mgeier:
@MSeal FWIW, I've been using this combination of filters in `nbsphinx` since more than two years (https://github.com/spatialaudio/nbsphinx/pull/123) and I'm not aware of any problems.

MSeal:
Thanks @mgeier that helps to know.

PR #1182, opened 2020-02-05
Fix formatting of cell errors in exported latex/pdf

telamonian:
While investigating #1181, I noticed that cell errors are not formatted well in exported latex/pdf. Starting from a notebook with a cell that looks like this:

image

when you export to pdf, you end up with:

image

The pdf output in this case is significantly less legible than the initial error/trace from the live notebook.

This PR fixes the above by:

- correctly coloring error output
- correctly indenting error output lines
- ensuring that no extraneous newlines get inserted
- adding a colored background to the error output in the pdf that matches the output background from the live notebook

Here's what the pdf export looks like with the above fixes:

image

One thing to notice is that I only added the cell error background color to the edge of the background. This seems like a reasonable compromise between verisimilitude (w.r.t. the live notebook) and sparing expensive colored ink (since exporting a notebook to pdf is a common step before printing)

### DIFF
--- a/share/jupyter/nbconvert/templates/latex/style_jupyter.tex.j2
+++ b/share/jupyter/nbconvert/templates/latex/style_jupyter.tex.j2
@@ -95,6 +95,7 @@
     \definecolor{outcolor}{HTML}{D84315}
     \definecolor{cellborder}{HTML}{CFCFCF}
     \definecolor{cellbackground}{HTML}{F7F7F7}
+    \definecolor{outerrorbackground}{HTML}{FFDFDF}
     ((*- endblock style_colors *))
     
     % prompt
--- a/share/jupyter/nbconvert/templates/latex/document_contents.tex.j2
+++ b/share/jupyter/nbconvert/templates/latex/document_contents.tex.j2
@@ -11,15 +11,16 @@
     \end{Verbatim}
 ((* endblock data_text *))
 
-% Display python error text as-is
+% Display python error text with colored frame (saves printer ink vs bkgnd)
 ((* block error *))
-    \begin{Verbatim}[commandchars=\\\{\}]
-((( super() )))
+    \begin{Verbatim}[commandchars=\\\{\}, frame=single, framerule=2mm, rulecolor=\color{outerrorbackground}]
+(((- super() )))
     \end{Verbatim}
 ((* endblock error *))
-((* block traceback_line *))
-    ((( line | indent | strip_ansi | escape_latex )))
-((* endblock traceback_line *))
+% Display error lines with coloring
+((*- block traceback_line *))
+((( line | escape_latex | ansi2latex )))
+((*- endblock traceback_line *))
 
 % Display stream ouput with coloring
 ((* block stream *))


telamonian:
https://github.com/jupyter/nbconvert/pull/1181#pullrequestreview-352776460:

> The `| escape_latex | ansi2latex ` may have other odd rendering side effects for more obscure patterns

@MSeal This PR uses the same fix in order to correctly color the text in the error output, and your comment applies here too. I dug into the actual exported latex, and I tried to come up with some problematic edge cases. However, I couldn't think of any. Do you know of any good examples of problems that can arise from the successive application of the `escape_latex` and `ansi2latex` filters?

mgeier:
> I only added the cell error background color to the edge of the background. This seems like a reasonable compromise between verisimilitude (w.r.t. the live notebook) and sparing expensive colored ink

FWIW, the Classic Notebook doesn't use any background color at all. The (current version) of `nbviewer` doesn't, either (e.g. https://nbviewer.jupyter.org/github/spatialaudio/nbsphinx/blob/master/doc/pre-executed.ipynb), nor does `nbsphinx` (e.g. https://nbsphinx.readthedocs.io/_/downloads/en/0.5.1/pdf/#subsubsection.7.1.3).

Probably the color should be removed from JupyterLab's output?

MSeal:
> Do you know of any good examples of problems that can arise from the successive application of the escape_latex and ansi2latex filters

Not specifically, and it sounds like @mgeier has used it successfully for a while in his projects. Latex filters just tends to have a lot of odd side-effects, especially across different engines, that I always worry when making changes.

MSeal:
> Probably the color should be removed from JupyterLab's output?

Lab has it's own style. That's actually why there's a [lab template](https://github.com/jupyter/nbconvert/tree/master/share/jupyter/nbconvert/templates/lab) separate from the default [classic template](https://github.com/jupyter/nbconvert/tree/master/share/jupyter/nbconvert/templates/classic) for html conversions. Maybe we should do the same for other export formats (now or in the future)?

> I only added the cell error background color to the edge of the background. This seems like a reasonable compromise between verisimilitude (w.r.t. the live notebook) and sparing expensive colored ink

I am ok with this compromise for now to avoid making a larger PR that supports multiple styles. @mgeier @maartenbreddels @SylvainCorlay thoughts?

mgeier:
Well, since JupyterLab and the Classic Notebook disagree, there should first be a decision about who is right.

Did JupyterLab intentionally change the background color?
Was there a reason for the change?

If the JupyterLab color change gets reverted, `nbconvert` shouldn't use that background color either.

If not ... I guess we can see about that once we get there ...

MSeal:
> Well, since JupyterLab and the Classic Notebook disagree, there should first be a decision about who is right.

There's not a right or wrong here. Just different interfaces. Generally nbconvert defaults to classic look n' feel but for errors being visible in PDF format I'm ok with the compression mise here. If someone wants to split by lab / not lab later they can take that on in a future pr

MSeal:
Thanks @telamonian

share/jupyter/nbconvert/templates/latex/style_jupyter.tex.j2
```diff
@@ -95,6 +95,7 @@
     \definecolor{outcolor}{HTML}{D84315}
     \definecolor{cellborder}{HTML}{CFCFCF}
     \definecolor{cellbackground}{HTML}{F7F7F7}
+    \definecolor{outerrorbackground}{HTML}{FFDFDF}
```
t-makaro: This line should go in the document contents. Otherwise this will break other cell styles.

mgeier:
> There's not a right or wrong here.

Maybe there is?

I'm trying to find that out at https://github.com/jupyterlab/jupyterlab/issues/7892.

MSeal:
I'm pretty sure each UI is free to choose look n' feel: e.g. nteract components have their own rendering that's quite different for many aspects of notebooks. It's not a centrally managed pattern.

share/jupyter/nbconvert/templates/latex/style_jupyter.tex.j2
MSeal: Do you think you could help fix that up? Or should we make a new issue and tag it for the next alpha release?

mgeier:
I haven't used nteract before, but it seems like they don't use a pink background:

image

image

PR #1183, opened 2020-02-05
fix: avoid iopub losing messages by polling two iopub AND shell socket

maartenbreddels:
A POC/port to nbconvert of https://github.com/voila-dashboards/voila/pull/536
Solves: https://github.com/nteract/papermill/issues/426
Alternative to #994

I think eventually this should go into nbclient, but opening this just to show how it can be done.

#994 polls for 1 second on the shell channel before polling/reading from the iopub channel. In this 1 second, the iopub socket can hit the high water mark of 0mq (default of a 1000).
This causes some messages to be dropped, causing a IOPub timeout. Instead, in this PR, we poll both sockets, receiving them ASAP, no messages lost.

Fixing the unittests might be difficult, since it requires monkeypatching zmq.select I think.

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -24,6 +24,7 @@ except NameError:
 from traitlets import List, Unicode, Bool, Enum, Any, Type, Dict, Integer, default
 
 from nbformat.v4 import output_from_msg
+import zmq
 
 from .base import Preprocessor
 from ..utils.exceptions import ConversionException
@@ -569,8 +570,7 @@ class ExecutePreprocessor(Preprocessor):
         return False
 
     def run_cell(self, cell, cell_index=0, store_history=False):
-        parent_msg_id = self.kc.execute(cell.source,
-            store_history=store_history, stop_on_error=not self.allow_errors)
+        parent_msg_id = self.kc.execute(cell.source, store_history=store_history, stop_on_error=not self.allow_errors)
         self.log.debug("Executing cell:\n%s", cell.source)
         exec_timeout = self._get_timeout(cell)
         deadline = None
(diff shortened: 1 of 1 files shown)


nbconvert/preprocessors/execute.py
```diff
@@ -580,67 +580,45 @@ def run_cell(self, cell, cell_index=0, store_history=False):
...
+            if timeout >= (-1 if self.interrupt_on_timeout else 0):
+                # we include 0, which simply is a poll to see if we have messages left
+                rlist, wlist, xlist = zmq.select([self.kc.iopub_channel.socket, self.kc.shell_channel.socket], [], [], min(0, timeout))
+                # print("lists", rlist, wlist, xlist)
+                if not rlist and not wlist and not xlist:
+                    self._check_alive()
+                    if monotonic() > deadline:
+                        self._handle_timeout(exec_timeout, cell)
+                if xlist:
+                    raise RuntimeError("Oops, unexpected rror")
```
jasongrout:
```suggestion
                    raise RuntimeError("Oops, unexpected error")
```

jasongrout: (actually, the whole error message isn't that helpful :)

maartenbreddels: Yeah, this is partly the reason this is a draft, we don't have a test that could trigger this. My guess is that maybe closing the socket from the kernel side might do it, or segfaulting the kernel.

MSeal:
Thanks for putting this together. I'll want to dedicate some time to comparing and testing it as this section of the code is hard to fully reason about on inspection alone. The good news is this code is still the same in nbclient so it should translate over cleanly. I can help with the tests if we get to approval and there's still gaps.

echuber2:
Hi, was this implemented in another PR? I still see IOPub timeouts sometimes.

MSeal:
@echuber2 the work here is not longer applicable as it was rewritten and moved into [nbclient](https://github.com/jupyter/nbclient) with the 6.0 release. There is always an internal buffer somewhere with the zmq communication. Timeouts usually occur when there's 1000's or 10s of 1000's of messages in a very short time (a second). I'd look at if your notebook is trying to print way too much information at once as it's also likely a mess in output formatting as well when there's that many messages. If there's a smaller unit of messages but still hitting timeouts I'd post to nbclient with more details on what you're executing.

PR #1185, opened 2020-02-06
Update CI to test on 3.8 + upstate setup.py

Carreau:
Let's now test on 3.8 which is out;
remove unnecessary dependencies as we don't support 3.4;
remove Trove Classifiers as now Python_requires cover this


MSeal:
Btw that failing test for python 3.8 I fixed in nbclient. If you want I can just merge this after we remove that code with the first release of that library.

Issue #1045, opened 2019-06-09
NBConvert 6.0 Planning


MSeal:
In light of several conversations around larger changes, I wanted to start the thread on 6.0 targets and outline / discuss what changes should or shouldn't be in the next major release. We should make issues for each of the following that we want addressed. Any of the topics below can be questioned and challenged or pushed until later releases. Champions for each one would be appreciated to help move key subjects forward. I'll keep this top ticket updated with links and additional requested items from the thread below.

- [x] Move execute preprocessor code into separate thin project
@mpacer first brought up the idea, I think it would greatly help the separation of concerns and let nbconvert focus on conversion rather than execution. Other projects (i.g. papermill) that just want execution ability could import this new module rather than the much larger dependency chain of nbconvert.
- [x] Python 3 only
Remove, or at a minimum not support, python 2 moving forward. Seems straight forward enough for the major version upgrade.
- [x] Directory based template loading, following viola and other jupyter project patterns.
First part for adding additional paths will be in #1028 . It's intended to make templates a directory with resources and multiple asset files rather than a single template file with the jump to 6.0.
- [ ] Eliminating the majority of usability concerns with latex, pdf, and html conversion.
This is a tall order, and we've been trying to get improvements made for the major latex / pdf issues, but there's a lot of these and I'm not sure we'll have them closed for the last 5.x release. Will probably require some significant improvements to how templates are loaded, executed, and tested to get improvements to more production quality levels.
~- [ ] Deprecate / remove unnecessary or complicated configuration options
We have a lot of configuration options, some of them are old or could be unnecessary now. I'd like to prune options and figure out how templating improvements could help where flags were used as crutches in the past. Not sure what the full scope could be here for this.~
~- [ ] Consider lightening or removing traitlets
Not going to be a popular topic, but traitlets causes a lot of support burden and debugging issues. It's also been cited as the part that's kept many strong developers I know from contributing because it's complicated and difficult to understand how it operates without a lot of digging. I don't expect this topic to make it in for 6.0 but I thought I might at least spark a conversation on the topic to see how we could make the repo more contributor friendly.~

There's probably some more I'm not thinking of atm. I'll add items from the thread that seem relevant.

t-makaro:
> Move execute preprocessor code into separate thin project

I agree. It's weird that a conversion utility deals with execution, but we should keep the ```--execute``` flag in some fashion even if deprecated. We could reference the new package. It would be nice to have them developed independently. 

> Python 3 only

[...]

PR #1192, opened 2020-02-18
change default output format to notebook

t-makaro:
As discussed in #1045, this default makes notebook-to-notebook preprocessor transformations nicer and does not favour html arbitrarily. 

I know @takluyver suggested adding a warning if no preprocessors are configured, but I'm not sure how to detect if there has been no other changes in configuration. We could just add a warning that says: ```Default output has been changed to notebook. For the old behaviour add: --to html``` if the ```--to``` flag has not been used. We could then remove the warning in 6.1

### DIFF
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -263,7 +263,7 @@ class NbConvertApp(JupyterApp):
 
 
     export_format = Unicode(
-        'html',
+        'notebook',
         allow_none=False,
         help="""The export format to be used, either one of the built-in formats
         {formats}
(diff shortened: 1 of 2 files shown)


MSeal:
> Default output has been changed to notebook. For the old behaviour add: --to html

Seems like a perfectly fine warning. Another option is to just remove the default processor entirely and just always ask -- this is a bit safer of a default change as it doesn't do an expected execution, just halt progress. Either way changing default behavior will be a little annoying from people having changed expectations.

> I'm not sure how to detect if there has been no other changes in configuration

It's probably ok to just emit the warning from CLI calls as to the default change imo

MSeal:
Having thought more about it, I think removing the default processor in general and requiring a `--to` input is probably best. It won't silently do something different for people who were passing no args, it gives a clear message about how to fix, and it achieves the goal of not biasing the CLI towards one format or another.

t-makaro:
Default has been removed. It took a little be of re-arranging to let ```jupyter nbconvert" with nothing else to still display the help.

Issue #1203, opened 2020-02-26
Preserve Markdown hyperlink text in Latex output


flutefreak7:
I have notebooks with internal links that look something like this:

```markdown
# Data Analysis Methods

See the section on [Benchmarking](#Benchmarking) for more examples.

...

# Benchmarking

The methods discussed in [Data Analysis Methods](#Data-Analysis-Methods) are benchmarked in this section.

```

When I convert that to a PDF, the text I provided for the link text is omitted and replaced by "Section 1" or "Section 2.1.3" for example:

# Data Analysis Methods

See the section on [Section 2](#Benchmarking) for more examples.

...

# Benchmarking

The methods discussed in [Section 1](#Data-Analysis-Methods) are benchmarked in this section.

For my real example I have this markdown:

![image](https://user-images.githubusercontent.com/2966015/75390844-a2f8f580-58ae-11ea-8672-a5bbd4aadec7.png)

I see this in the HTML output:

![image](https://user-images.githubusercontent.com/2966015/75390794-8492fa00-58ae-11ea-9b33-176159e6f89d.png)

And this in the PDF output:

![image](https://user-images.githubusercontent.com/2966015/75390816-907ebc00-58ae-11ea-8f1a-bc75366a7018.png)

It looks like [this code](https://github.com/jupyter/nbconvert/blob/9dcc8435c789954727b189bc4622e59d34f7fc07/nbconvert/filters/filter_links.py#L35
) is adding the "Section" part explicitly with a RawInline.

I wonder if there isn't a better solution that preserves the Markdown behavior so the the resulting PDF looks similar to the HTML output.

flutefreak7:
I think I figured out a fix to [this function](https://github.com/jupyter/nbconvert/blob/9dcc8435c789954727b189bc4622e59d34f7fc07/nbconvert/filters/filter_links.py#L18) that works:

```python
def resolve_one_reference(key, val, fmt, meta):
    """
    This takes a tuple of arguments that are compatible with ``pandocfilters.walk()`` that
    allows identifying hyperlinks in the document and transforms them into valid LaTeX 
    \\ref{} calls so that linking to headers between cells is possible.

    See the documentation in ``pandocfilters.walk()`` for further information on the meaning
    and specification of ``key``, ``val``, ``fmt``, and ``meta``. 
    """
    
    if key == 'Link':
        text = val[1][0]['c']
        target = val[2][0]
        m = re.match(r'#(.+)$', target)
        if m:
[...]

PR #1204, opened 2020-02-26
Changed from \ref to \hyperref to preserve markdown link text

flutefreak7:
As discussed in #1203, I think it makes sense for the links in PDF's to preserve the behavior seen in Markdown and HTML. Using `\hyperref` instead of `\ref` enables custom link text as discussed in [this stackoverflow post](https://tex.stackexchange.com/questions/70143/cross-reference-with-custom-text).

I also checked to ensure that `\hyperref` would be available and the [pandoc documentation](https://pandoc.org/MANUAL.html#creating-a-pdf) says that it's a required package, so presumably it's in the pandoc latex template's preamble.

### DIFF
--- a/nbconvert/filters/filter_links.py
+++ b/nbconvert/filters/filter_links.py
@@ -4,7 +4,7 @@ Converts links between notebooks to Latex cross-references.
 """
 import re
 
-from pandocfilters import RawInline, applyJSONFilters
+from pandocfilters import RawInline, applyJSONFilters, stringify
 
 def resolve_references(source):
     """
@@ -26,7 +26,7 @@ def resolve_one_reference(key, val, fmt, meta):
     """
     
     if key == 'Link':
-        text = val[1][0]['c']
+        text = stringify(val[1])
         target = val[2][0]
         m = re.match(r'#(.+)$', target)
         if m:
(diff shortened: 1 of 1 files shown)


t-makaro:
Thank you very much for fixing this one. LGTM but I'm not familiar with this section of code. Could you please post a before/after picture of the rendering?

flutefreak7:
I'll try to provide a quick response...

The "before" PDF looks like the PDF screenshot in my original post, with
"Section 1.0" or "Section 2.0" style links. The "after" PDF looks more like
the HTML screenshot in my original post with linked text that matches the
markdown like "Introduction" or "Data Processing."

I can provide a better response when I'm back at work tomorrow.

flutefreak7:
@t-makaro 
In the process of making before and after shots I discovered a bug. Apparently pandocfilters splits strings up into different nodes if there is whitespace and has a `stringify` function to handle converting a chunk of content back into a string. When I added that I get the following before / after

Before:
image

After:
image

...now I just need to figure out how to add the next commit of mine to this PR...

t-makaro:
LGTM! Thank you! Re-ran docs build. Merging.

Issue #821, opened 2018-05-27
Consider factoring out the "execute" module into its own package


choldgraf:
Curious what folks think about this:

I've had a lot of confusion from people who want to run jupyter notebooks from the command line (e.g. as a part of batch jobs). They usually think they can do something like `python notebook.ipynb` or `jupyter notebook.ipynb`.

I usually correct them with the `nbconvert` solution (e.g. `jupyter nbconvert --to notebook --execute --inplace`) though I've found that this seems really unintuitive for most people (e.g., "why do I have to use a notebook conversion tool just to execute it?"). Moreover if you want to make other modifications (like define the kernel to be used), you need to start adding code like `----ExecutePreprocessor.kernel_name=python3` which feels a bit clunky.

In short, IMO the `execute` module is orthogonal enough and useful enough that it could be separated from `nbconvert` into its own repository. Then you could design an API around it that would be more intuitive for folks (e.g. `jupyter nbexecute mynotebook.ipynb --kernel_name=python3`) and design more complexity around this specific task (e.g. find ways to interface with projects such as [nbparameterize](https://github.com/takluyver/nbparameterise) or [papermill](https://github.com/nteract/papermill)). `nbconvert` could then use this API under the hood in conjunction with the `--execute` flag.

So I guess my questions are:

1. Do other folks agree that the "execute" part of nbconvert is useful enough to be its own separate repo?
2. How entangled is the execute preprocessor with `nbconvert`? Would it be extremely difficult to spin it off into its own thing?

choldgraf:
cc @rgbkrk and @takluyver since they've both built on top of the "execute" functionality here. curious what you two think!

takluyver:
Yeah, it probably makes sense, and I don't think it's particularly entangled with nbconvert.

Maybe the hardest decision is picking what to call the new package and the new command. One option is that it could go into `jupyter_client` (or its replacement, `jupyter_kernel_mgmt`) as part of the `jupyter run` command.

rgbkrk:
I do think there's a core of nbconvert that should be extracted out, especially when it's replacing the notebook itself, rather than converting to another format.

choldgraf:
> especially when it's replacing the notebook itself, rather than converting to another format

Totally. I opened this issue because I thought to myself "I'm not converting anything if I want to just run it. I'm just converting it into...itself..."

Something like `jupyter run XXX` makes sense to me...

rgbkrk:
[...]

PR #1211, opened 2020-03-09
Replaced execute preprocessor internals with nbclient calls

MSeal:
Finishes https://github.com/jupyter/nbconvert/issues/821. Took me a while to get around to it, but here's the PR with all the stuff moved into nbclient deleted.

I found that given how light the wrapper is now only a few very basic tests should be present. I didn't like how I had to subprocess NBClient's class to account for all the traitlet assignment junk, but it works and leave the interface backwards compatible. If nbclient ever needs to move it's interface forward or off trailets or whatever, we can capture it here for translation against future versions.


MSeal:
I think after this merges we can do 6.0.0a1 release

choldgraf:
The amount red in this PR makes me happy

MSeal:
Hmm need to look into the 3.5 failure -- might be related to the async wrappers since 3.5 has less support therein.

MSeal:
Not sure why 3.5 is hanging on allow-errors -- will debug that later

davidbrochart:
I tested this PR locally with python3.5, it seems to *randomly* hang on `test_allow_errors` (it doesn't happen with python3.6).
Asyncio was still missing some important features in 3.5, and nbclient is implemented using asyncio. We work around the limitations of 3.5 with e.g. the [async_generator](https://github.com/python-trio/async_generator) library, but there might be other things that are problematic.
Is there a plan to drop python3.5 in nbconvert, nbclient, and in the Jupyter ecosystem in general?
In Voila, switching to async means dropping python3.5 anyway, since we need async generators in Jinja, which is not going to happen in python3.5: https://github.com/pallets/jinja/issues/1058

choldgraf:
I am +1 on deprecating 3.5 after a deprecation cycle, now that we are already nearing 3.9

MSeal:
Yeah the general guidance I've been following is to support 3.5 until it cycles out of the supported path for python proper. The reason I've tried to keep maintaining it is there's some larger orgs whose internal default python version is still 3.5. Dropping support for it before python does would disrupt updates for those orgs. Still I think later this year 3.5 can drop. For now I'd be ok with marking those tests as 3.6+ only and putting a big warning in the release notes that 3.5 may have stability issues with this change given async support. Thoughts?

davidbrochart:
Sounds good to me.

MSeal:
Ok based on conversations across a few companies and in https://github.com/jupyter/nbclient/pull/34 I'm going to update this PR and drop support for Python 3.5. The async code in upstream libraries has caused issues in each repo for python 3.5.

MSeal:
Going to merge this to get some things moving

PR #1217, opened 2020-03-18
MRG: use Python 3 super syntax throughout

matthew-brett:
Are y'all in the market for Python 3 refactorings?

Here I did a perl pie conversion of all instances of `super(MyClass,
self).something` to `super().something`.


MSeal:
Thanks so much for submitting a PR -- and yes python 3 refactors are welcome additions for the 6.0 release!

Issue #1221, opened 2020-03-26
Python 3.9: ModuleNotFoundError: No module named 'xml.etree.cElementTree'


hroncok:
**Nbconvert version:** 5.6.0 but also master

When we rebuild the nbconver package in Fedora with Python 3.9.0a5, the following tracebacks is shown:

```
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/sphinx/config.py", line 361, in eval_config_file
    execfile_(filename, namespace)
  File "/usr/lib/python3.9/site-packages/sphinx/util/pycompat.py", line 81, in execfile_
    exec(code, _globals)
  File "/builddir/build/BUILD/nbconvert-5.6.0/docs/source/conf.py", line 25, in <module>
    exec(compile(f.read(), 'autogen_config.py', 'exec'), {})
  File "autogen_config.py", line 11, in <module>
  File "/builddir/build/BUILD/nbconvert-5.6.0/nbconvert/__init__.py", line 4, in <module>
    from .exporters import *
  File "/builddir/build/BUILD/nbconvert-5.6.0/nbconvert/exporters/__init__.py", line 3, in <module>
    from .html import HTMLExporter
  File "/builddir/build/BUILD/nbconvert-5.6.0/nbconvert/exporters/html.py", line 14, in <module>
    from nbconvert.filters.highlight import Highlight2HTML
  File "/builddir/build/BUILD/nbconvert-5.6.0/nbconvert/filters/__init__.py", line 7, in <module>
    from .strings import *
  File "/builddir/build/BUILD/nbconvert-5.6.0/nbconvert/filters/strings.py", line 22, in <module>
    from defusedxml import cElementTree as ElementTree
  File "/usr/lib/python3.9/site-packages/defusedxml/cElementTree.py", line 10, in <module>
    from xml.etree.cElementTree import TreeBuilder as _TreeBuilder
ModuleNotFoundError: No module named 'xml.etree.cElementTree'
```

Dependent packages also fail.

See https://docs.python.org/3.9/whatsnew/3.9.html#removed

> The xml.etree.cElementTree module has been removed.

I believe this:

https://github.com/jupyter/nbconvert/blob/275f7f909a803560e2459642a539874113fba959/nbconvert/filters/strings.py#L22-L23

Should be replaced to use ElementTree instead of cElementTree. I will also contact @tiran about the problem in defusedxml.

PR #1222, opened 2020-03-27
cElementTree no longer exists in Python 3.9

frenzymadness:
cElementTree has been just backward compatible import for a long time so this change has no impact on performance.

Fixes: https://github.com/jupyter/nbconvert/issues/1221


hroncok:
I confirm this works.

PR #1226, opened 2020-04-02
Enable configuring JQuery and RequireJS in HTML exporter

lresende:
Add --HTMLExporter.jquery_url and --HTMLExporter.require_js_url
to enable users to configure the URL to use for these required
dependencies. By default these will be pointing to the CDNJS
hosted by Cloudfare.

Note that these configuration options were already available in
the slide exporter, and based on inheritance, they were just
moved to the HTML exporter and are now available for both exporters.


lresende:
@MSeal could you please take a look at this pr.

MSeal:
Yes, sorry I missed it before. This seems like a reasonable fix, thanks for making it!

Issue #1246, opened 2020-05-03
Inkscape 1.0rc1: Unknown option --without-gui


hroncok:
In Fedora, we have Inkscape 1.0rc1, thta doesn't recognize the `--without-gui` option:

```
$ inkscape --version
Inkscape 1.0rc1 (09960d6f05, 2020-04-09)
    Pango version: 1.44.7
$ inkscape --without-gui
Unknown option --without-gui
```

nbocnvert tries to use this option.

Convert a notebook with SVG to PDF. Will try to get back with full example.

**Nbconvert version:** master

hroncok:
```
$ pytest -k svg2pdf
============================= test session starts ==============================
platform linux -- Python 3.8.2, pytest-5.4.1, py-1.8.0, pluggy-0.13.1
rootdir: /home/churchyard/Dokumenty/jupyter/nbconvert
plugins: cov-2.6.1
collected 253 items / 248 deselected / 5 selected

nbconvert/preprocessors/tests/test_svg2pdf.py ....F                      [100%]

=================================== FAILURES ===================================
___________________________ Testsvg2pdf.test_output ____________________________

self = <nbconvert.preprocessors.tests.test_svg2pdf.Testsvg2pdf testMethod=test_output>

    @onlyif_cmds_exist('inkscape')
    def test_output(self):
        """Test the output of the SVG2PDFPreprocessor"""
        nb = self.build_notebook()
        res = self.build_resources()
        preprocessor = self.build_preprocessor()
>       nb, res = preprocessor(nb, res)

nbconvert/preprocessors/tests/test_svg2pdf.py:73: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
nbconvert/preprocessors/base.py:47: in __call__
    return self.preprocess(nb, resources)
nbconvert/preprocessors/base.py:69: in preprocess
    nb.cells[index], resources = self.preprocess_cell(cell, resources, index)
nbconvert/preprocessors/convertfigures.py:45: in preprocess_cell
    output.data[self.to_format] = self.convert_figure(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

[...]

PR #1247, opened 2020-05-03
Fix CLI options given to Inkscape 1.0rc1+

hroncok:
Fixes https://github.com/jupyter/nbconvert/issues/1246

### DIFF
--- a/nbconvert/preprocessors/svg2pdf.py
+++ b/nbconvert/preprocessors/svg2pdf.py
@@ -76,10 +76,11 @@ class SVG2PDFPreprocessor(ConvertFiguresPreprocessor):
     @default('command')
     def _command_default(self):
         major_verison = self.inkscape_version.split('.')[0]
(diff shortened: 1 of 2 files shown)


LukeLR:
For anyone else who is already using inkscape1.0 and waiting for this change to make it in the release: Passing
```
--SVG2PDFPreprocessor.command="/usr/bin/inkscape --export-filename=\"{to_filename}\" \"{from_filename}\""
```
on the command line should also fix the issue (as a workaround). E.g.
```
jupyter-nbconvert --to pdf notebook.ipynb --SVG2PDFPreprocessor.command="/usr/bin/inkscape --export-filename=\"{to_filename}\" \"{from_filename}\""
```
should do the job. In theory, it should also be possible to set this permanently in a configuration file, as documented [here](https://nbconvert.readthedocs.io/en/latest/config_options.html) but I didn't manage to get this working.

PR #1251, opened 2020-05-07
Support SQL syntax highlighting

jaypeedevlin:
The SQL lexer is included with the standard Pygments install.  This change enables syntax highlighting when using sql magic such as https://github.com/catherinedevlin/ipython-sql.

### DIFF
--- a/nbconvert/preprocessors/highlightmagics.py
+++ b/nbconvert/preprocessors/highlightmagics.py
@@ -32,6 +32,7 @@ class HighlightMagicsPreprocessor(Preprocessor):
             '%%perl': 'perl',
             '%%ruby': 'ruby',
             '%%sh': 'sh',
+            '%%sql': 'sql',
     })
 
     # user defined language extensions


t-makaro:
Build failures look unrelated. 

It looks to me like this will work for markdown, asciidoc, and rst exporting, but not html/latex. This is because this metadata is only used for formats that understand highlighting code out of the box. For HTML, and LaTeX, the iPython lexer is used. It will need to be updated like [here](https://github.com/ipython/ipython/pull/11137).

I see no reason not to merge this here, but you should open an issue on the [iPython repo](https://github.com/ipython/ipython) about the lexer.

jaypeedevlin:
Thanks for the feedback Tyler.  Interestingly, when I made this change locally and converted using the HTML template, I got SQL syntax highlighting working as expected:

image

I tested without my change (to make sure the change wasn't a noop) and highlighting went away.

SylvainCorlay:
This looks good to me. Merging!

PR #1256, opened 2020-05-19
Log template paths with the debug level

jtpio:
The list of template paths tends to be quite long, and it would be useful to have it displayed only in debug mode.


MSeal:
Build failure is unrelated (check-manifest changed recently)

PR #1257, opened 2020-05-20
Fix invalid HTML in lab template

SylvainCorlay:
The `body_footer` from the base template was closing other tags that are not in the lab template.


SylvainCorlay:
This is a trivial fix, and the check manifest CI error occurs on master.

Issue #1161, opened 2020-01-06
Lab css files line ending on install


MSeal:
The 6.0.0a on master keeps downloading or recreating lab/static/theme-dark.css and theme-light.css after I pip install on Ubuntu 18.04 in python 3.6. I think we need to change where the source is for that to use `\n` instead of `\r\n` to keep consistent with the rest of the project's files, wherever that's coming form. It might be specific to my setup but it's been bugging me and I wanted to track the issue.

```
$ pip install -e .
....
10:51:34 mseal {master} ~/Workspace/nbconvert (venv:.py3local)  
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   share/jupyter/nbconvert/templates/lab/static/theme-dark.css
	modified:   share/jupyter/nbconvert/templates/lab/static/theme-light.css

no changes added to commit (use "git add" and/or "git commit -a")
10:51:35 mseal {master} ~/Workspace/nbconvert (venv:.py3local)  
$ git stash
warning: CRLF will be replaced by LF in share/jupyter/nbconvert/templates/lab/static/theme-dark.css.
The file will have its original line endings in your working directory.
warning: CRLF will be replaced by LF in share/jupyter/nbconvert/templates/lab/static/theme-light.css.
The file will have its original line endings in your working directory.
```

PR #1259, opened 2020-05-20
Ignore downloaded CSS

SylvainCorlay:
These files are downloaded when building from a source dist, and should not be included in the repository.


SylvainCorlay:
Merging as it is a trivial fix, removing files that should not have been versioned.

MSeal:
I can close https://github.com/jupyter/nbconvert/issues/1161 now too -- ty

PR #1261, opened 2020-05-20
Use body data attribute from JupyterLab

SylvainCorlay:
The `theme-dark` CSS class is not in the JupyterLab DOM and is not used anywhere.

This was an artifact of old Voilà code.


SylvainCorlay:
cc @maartenbreddels

maartenbreddels:
Cc @mariobuikhuizen this may have implications for voila-vuetify

PR #1262, opened 2020-05-21
WIP - Validate generated HTML

SylvainCorlay:
Verifying that html generated by nbconvert templates is valid.

### DIFF
--- a/nbconvert/tests/test_nbconvertapp.py
+++ b/nbconvert/tests/test_nbconvertapp.py
@@ -16,6 +16,9 @@ from nbconvert.exporters import Exporter
 from traitlets.tests.utils import check_help_all_output
 from testpath import tempdir
 import pytest
+from html5validator import Validator
+
+validator = Validator()
 
 #-----------------------------------------------------------------------------
 # Classes and functions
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
html5validator is not great. It is really only a CLI tool with no clean programmatic interface, and depends on a java thing.

MSeal:
That's too bad. I agree that doesn't like like a good dependency to add.

SylvainCorlay:
Yeah we use it in the voila-gridstack nbconvert template and caught some errors. We may be able to do something with a pure python solution.

SylvainCorlay:
We can apparently invoke the validator at http://validator.w3.org/ as per the stackoverflow answer here: https://stackoverflow.com/a/10519634/2502544

https://pypi.org/project/py_w3c/

MSeal:
I'd be quite hesitant to ship local html to a service in a library call here. Especially if someone has anything sensitive as we'd be airing to the whole internet at that point without the user knowing. Plus it reduces reliable execution substantially.

SylvainCorlay:
> I'd be quite hesitant to ship local html to a service in a library call here. Especially if someone has anything sensitive as we'd be airing to the whole internet at that point without the user knowing. Plus it reduces reliable execution substantially.

I totally agree, that we cannot rely on a service. Just pointing at the existing solutions to see if there is something better we can do about it.

SylvainCorlay:
Closing as I don't think we have a good solution for that yet.

PR #1263, opened 2020-05-22
Move base.html.j2 to classic template

SylvainCorlay:
`base.html.j2` contains the classic notebook DOM structure and should not be in `base/` but in `classic/`.


SylvainCorlay:
Ping @maartenbreddels

maartenbreddels:
Yeah, that should be fine!

SylvainCorlay:
> I get complaints about jinja2.exceptions.TemplateNotFound: static/index.css, should those be relative to classic as well:
> https://github.com/jupyter/nbconvert/blob/master/share/jupyter/nbconvert/templates/lab/index.html.j2#L10

I am not reproducing the issue. How are you invoking nbconvert to see the error message?

SylvainCorlay:
Oh I get it: you are probably using a development installs.

nbconvert relies on templates being installed in PREFIX/share as data files, so you cannot use `pip install -e`.

SylvainCorlay:
> Heres a minimal repro:
> [...]

OK, I was able to reproduce.

What is missing in your test template is the `base_template` entry in conf.json.

PR #1264, opened 2020-05-22
Make "lab" the default template for html export

SylvainCorlay:
This 7 characters' change makes the "lab" template the default for exporting to html.

The reason is that:

 - Nbconvert 6.0 is a major release and good time to make that change.
 - The classic notebook has been in maintainance mode for a while and JupyterLab has reached 2.0. Even though there are many people using the classic notebook, the front-end that we are now pushing is JupyterLab.
 -  The lab template enables the use of JupyterLab CSS themes. It is also the default basebase template for Voilà.

- [x] @SylvainCorlay 
- [x] @MSeal
- [x] @blink1073 
- [ ] @mpacer 
- [X] @maartenbreddels
- [ ] @minrk
- [x] @Carreau

### DIFF
--- a/nbconvert/exporters/html.py
+++ b/nbconvert/exporters/html.py
@@ -60,7 +60,7 @@ class HTMLExporter(TemplateExporter):
 
     @default('template_name')
     def _template_name_default(self):
-        return 'classic'
+        return 'lab'
 
     @default('template_data_paths')
     def _template_data_paths_default(self):
--- a/nbconvert/exporters/tests/test_templateexporter.py
+++ b/nbconvert/exporters/tests/test_templateexporter.py
@@ -398,7 +398,7 @@ class TestExporter(ExportersTestsBase):
         nb_no_output_prompt, resources_no_output_prompt = exporter_no_output_prompt.from_filename(self._get_notebook())
 
         assert not resources_no_output_prompt['global_content_filter']['include_output_prompt']
-        assert "Out[" not in nb_no_output_prompt
+        assert "Out[1]" not in nb_no_output_prompt
 
     def test_remove_elements_with_tags(self):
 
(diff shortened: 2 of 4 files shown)


maartenbreddels:
I am ok with this, it will also make people more aware of that 6 has changed, and times are changing :).

SylvainCorlay:
@mseal @maartenbreddels @mpacer @maartenbreddels @minrk @Carreau

maybe this is impactful enough that we need to approve the PR before moving on.

(I have been merging minor fixes that I am doing but this is a real change)

MSeal:
> maybe this is impactful enough that we need to formally approve the PR before moving on.

I appreciate that. Let's give a little time for the other @ pinged people to weight in. If all is +1 or no response by next week I'll approve and merge it.

blink1073:
Sounds reasonable to me to be future-biased in a new major release.

t-makaro:
Seems reasonable to me.

Carreau:
I'm ok with changing the default. Changing the filename of templates (with j2) was way more impactful (as it was bw incompat with  custom templates). So I think this is fine. 

Thanks for pushing it !

SylvainCorlay:
Thanks @Carreau ! Btw, the new template system is much more extensible!

SylvainCorlay:
When we get this in, I would like to publish another alpha of 6.0.

MSeal:
Hmm seems like there's still some failing output tests. I think we have consensus enough to merge once that's fixed.

SylvainCorlay:
@mseal I just fixed the tests:

- we were checking that `Out[` wad not in the notebook to check that outputs were filtered, but it is in the css.
- the lab template was actually not behaving correctly with respect to cell tags.

PR #1265, opened 2020-05-22
Fix example widget notebook

SylvainCorlay:
The example notebook for widgets was using old widgets APIs that are not available in 7.x anymore (which was released over two years ago).


SylvainCorlay:
I am merely using the current Widget_List from ipywidgets 7.5, and using a local image instead of the ipywidgets example image.

PR #1266, opened 2020-05-22
The lab template should not inherit from classic

SylvainCorlay:
This PR makes the lab template inherit from base and be completely independent from classic.


SylvainCorlay:
Ping @maartenbreddels as discussed. The lab template and the classic template now both inherit from base. I will be putting the requirejs and widgets things into a macro in base in a next PR.

Issue #1065, opened 2019-07-20
Is there a way to disable an export option?


ieglonew01f:
Currently, I'm trying to disable pdf exports and latex exports is there a way to do so? Is there an option to blacklist certain exports?

MSeal:
Well, if latex isn't installed the exports will fail. There's not an ability to remove the export options from the `--to` list afaik.

rahulpshah:
I am trying to disable pdf exporter for nbconvert based so that it does not show as an option when someone uses Jupyterlab. I don't have latex packages on the host machine so it does not make sense to show pdf as an option.
Have a c.NbExporter.blacklist_exporters = [""] in jupyter_server_config.py to specify valid exporters.

PR #1273, opened 2020-05-27
Allow exporters to be enabled or disabled via config

ianabc:
This change allows people to configure which Exporters are enabled by adding a `.enabled` attribute/traitlet to the `Exporter` class. It allows users to control functionality such as the "Download as" entry in the notebook and addresses issues like #1065 or [this discourse post](https://discourse.jupyter.org/t/remove-download-as-options-in-jupyter-notebook/4374).

To disable a single exporter, configuration looks like
```python
c.ASCIIDocExporter.enabled = False
```
Or, since everything inherits from `Exporter`, you could disable all but one with
```python
c.Exporter.enabled = True
c.NotebookExporter.enabled = True
```

People wanting to tweak the notebook "Download as" options those could go in their `jupyter_notebook_config.py` but to get complete control over those entries you also need a change to the notebook (e.g.  [this hack](https://github.com/ianabc/notebook/commit/bd0715cb281954a3ff045112e63794cbc6d9446e)). In principal you could do everything there, but I felt that the disable/enable functionality was more general than just the notebook so I thought I'd ask for feedback here first.

### DIFF
--- a/nbconvert/exporters/exporter.py
+++ b/nbconvert/exporters/exporter.py
@@ -52,7 +52,9 @@ class Exporter(LoggingConfigurable):
     accompanying resources dict.
     """
 
-    enabled = Bool(True).tag(config=True)
+    enabled = Bool(True,
+        help = "Disable this exporter (and any exporters inherited from it)."
+    ).tag(config=True)
 
     file_extension = FilenameExtension(
         help="Extension of the file that should be written to disk"
--- a/nbconvert/exporters/exporter.py
+++ b/nbconvert/exporters/exporter.py
@@ -17,7 +17,7 @@ import nbformat
 
 from traitlets.config.configurable import LoggingConfigurable
 from traitlets.config import Config
-from traitlets import HasTraits, Unicode, List, TraitError
+from traitlets import Bool, HasTraits, Unicode, List, TraitError
 from traitlets.utils.importstring import import_item
 from ipython_genutils import text, py3compat
 
@@ -51,6 +51,8 @@ class Exporter(LoggingConfigurable):
     NotebookNode object and then return the modified NotebookNode object and
     accompanying resources dict.
     """
+
+    enabled = Bool(True).tag(config=True)
 
     file_extension = FilenameExtension(
         help="Extension of the file that should be written to disk"
--- a/nbconvert/exporters/base.py
+++ b/nbconvert/exporters/base.py
@@ -7,6 +7,7 @@ import warnings
 
 import entrypoints
 
+from traitlets.config import get_config
 from traitlets.log import get_logger
 from traitlets.utils.importstring import import_item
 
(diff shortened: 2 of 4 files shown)


meeseeksmachine:

https://discourse.jupyter.org/t/remove-download-as-options-in-jupyter-notebook/4374/5

nbconvert/exporters/exporter.py
```diff
@@ -52,6 +52,8 @@ class Exporter(LoggingConfigurable):
     accompanying resources dict.
     """
 
+    enabled = Bool(True).tag(config=True)
```
betatim, line 55: Could you add a short "doc string"/help text to this to explain how/why people would use this

betatim:
Looks nice!

Do we have any data on how many exporters exist that don't inherit from `Exporter` and would not have this attribute? I was wondering if the selection loop should have `or True` in it for classes that don't have the traitlet in an attempt to not break existing code.

I like it but not very familiar with the nbconvert code and conventions to know what else we should look at/think about before merging.

nbconvert/exporters/exporter.py
ianabc: Yep, thanks, I've added a one line help, if you think it needs more explanation I can add it.

ianabc:
All of the exporters in `nbconvert/exporters` seem to inherit from either `Exporter` or `TemplateExporter` (which inherits from `TemplateExporter`) so I _think_ they are all covered. If there are other mechanisms for adding exporters I think I would need to extend the `or True` to work when people have explicitly set `.enabled = False`, maybe 
```python
[e for e in names if getattr(get_exporter(e)(config=c), 'enabled', True)] 
```
I'll leave this out for now, but if there are other exporters that seems to work OK.


I've been testing the changes in the notebook but they seem to work in jupyterlab as well. I haven't figured out exactly how lab builds it's list of exporters but it does seem to "just work". I'll try to dig into lab and make sure this is true in general.

ianabc:
The doc build failure seems to be related to [nbconvert/docs/example.ipynb](https://github.com/jupyter/nbconvert/blob/master/docs/source/example.ipynb). Something goes wrong extracting the output as a png and then LaTeX chokes on the broken png file.

MSeal:
Yes the docs build failure I need to go fix. Don't worry about it for this PR.

ianabc:
I took a look at moving the logic to `get_exporter`, it's a bit rough and ready but seems to be working OK. I define a new exception (`ExporterDisabledError`) which is raised when `get_exporter` is called on an exporter which is disabled in config. `get_exporter_names` can then use that exception to weed out the list of enabled exporters. Is this the sort of thing you were thinking of?

I think most things (notebook in particular) check to see what exporters are available with `get_exporter_names` before calling `get_exporter` on a specific exporter. Those uses should continue to work fine but if someone calls `get_exporter` blindly on something that happens to be disabled they will need to react to the exception (as in `get_exporter_names`).

betatim:
Nice work and thank you for contributing!

PR #1277, opened 2020-05-27
Allow specifying the widgets renderer semver range

SylvainCorlay:


maartenbreddels:
YES! We needed this ages ago :)

SylvainCorlay:
I want a better name for `jupyter_widgets_base_url`.

It is the CDN URL for third-party widgets. What about `jupyter_widgets_cdn_url`?

maartenbreddels:
in principle it can also be a local URL, i someday hacked this for local dev. It could also just be 1 variable right?

`jupyter_widgets_script_url` ?

SylvainCorlay:
Moving forward with this. Will include the script URL in another PR. Still not sure about the naming...

SylvainCorlay:
Ping @jasongrout

PR #1278, opened 2020-05-28
Escape HTML characters in latex output strings

eric-wieser:
Without this change, latex like `\left<content\right>` renders as a `<content\right>` html tag, which mathjax can't handle.

See https://nbviewer.jupyter.org/urls/galgebra--417.org.readthedocs.build/en/417/tutorials/algebra.ipynb#More-examples for an example of this failing.

This is like #514, but for cell outputs.

### DIFF
--- a/share/jupyter/nbconvert/templates/classic/base.html.j2
+++ b/share/jupyter/nbconvert/templates/classic/base.html.j2
@@ -191,7 +191,7 @@ alt="{{ alttext }}"
 
 {% block data_latex scoped %}
 <div class="output_latex output_subarea {{ extra_class }}">
-{{ output.data['text/latex'] | e}}
+{{ output.data['text/latex'] | e }}
 </div>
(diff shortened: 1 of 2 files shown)


eric-wieser:
CI failure look unrelated.

eric-wieser:
Restarting CI just to be sure.

MSeal:
CI docs build I'm going to try fixing today, don't worry about that one

eric-wieser:
Coping the notebook is fine, but it depends on `galgebra` and `sympy`, which make no guarantees of latex ewuivalence between versions.

I guess if you don't rerun the notebook then it does not matter.

eric-wieser:
Since I'm not really familiar with the structure of this project, would it be ok if I left the lab notebook and other HTML pages to you?

MSeal:
Sure, I can push into the PR where else it's obvious the same change needs applying when I get a chance.

PR #1285, opened 2020-06-08
Add --to webpdf for PDF rendering via browser

davidbrochart:
Compared to `--to pdf`, this allows to have widgets in the PDF document. It also supports the same templates as `--to html`.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -12,16 +12,17 @@ Installing nbconvert
 Nbconvert is packaged for both pip and conda, so you can install it with::
 
     pip install nbconvert
+
     # OR
+
     conda install nbconvert
 
-If you're new to Python, we recommend installing `Anaconda <https://www.anaconda.com/distribution/>`_,
-a Python distribution which includes nbconvert and the other Jupyter components.
+The `Miniconda <https://docs.conda.io/en/latest/miniconda.html>`_ and `Miniforge <https://github.com/conda-forge/miniforge/>`_ distributions both provide a minimal conda installation.
 
 .. important::
 
     To unlock its full capabilities, nbconvert requires Pandoc, TeX
-    (specifically, XeLaTeX) and Chromium. These must be installed separately.
+    (specifically, XeLaTeX) and Pyppeteer. These must be installed separately.
 
 Installing Pandoc
 -----------------
@@ -39,11 +40,8 @@ On other platforms, you can get pandoc from
 Installing TeX
 --------------
 
-For converting to PDF, nbconvert can either use `Puppeteer <https://pptr.dev>`_
-(with ``--to webpdf``) or the TeX document preparation ecosystem (with
-``--to pdf``). In the latter case it produces an intermediate ``.tex`` file
-which is compiled by the XeTeX engine with the LaTeX2e format (via the ``xelatex``
-command) to produce PDF output.
+For converting notebooks to PDF (with ``--to pdf``), nbconvert makes use of LaTeX
+and the XeTeX as the rendering engine.
 
 .. versionadded:: 5.0
 
@@ -72,13 +70,14 @@ notebooks to PDF.
 Installing Chromium
 -------------------
 
-When converting to PDF with ``--to webpdf``, nbconvert uses `Puppeteer <https://pptr.dev>`_, and
-more precisely its Python bindings `pyppeteer <https://github.com/pyppeteer/pyppeteer>`_. pyppeteer
-needs a special version of Chromium, and can automatically download it. Since the download size is
-quite large (around 100 MB), nbconvert prevents it by default. If pyppeteer finds a suitable
-version of Chromium on your system, it will try and use it. Otherwise, please use the
-``--allow-chromium-download`` flag to allow Chromium's download. Note that installing with
-``nbconvert[webpdf]`` we install a compatible version of pyppeteer for this usecase.
+For converting notebooks to PDF with ``--to webpdf``, nbconvert requires the
+`Pyppeteer <https://github.com/pyppeteer/pyppeteer>`_ Chromium automation library.
+
+Pyppeteer makes use of a specific version of Chromium. If it does not find a suitable
+installation of the web browser, it can automatically download it if the ``--allow-chromium-download``
+flag is passed to the command line.
+
+To install a suitable version of pyppeteer, you can pip install ``nbconvert[webpdf]``.
 
 PDF conversion on a limited TeX environment
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -77,7 +77,8 @@ more precisely its Python bindings `pyppeteer <https://github.com/pyppeteer/pypp
 needs a special version of Chromium, and can automatically download it. Since the download size is
 quite large (around 100 MB), nbconvert prevents it by default. If pyppeteer finds a suitable
 version of Chromium on your system, it will try and use it. Otherwise, please use the
-``--allow-chromium-download`` flag to allow Chromium's download.
+``--allow-chromium-download`` flag to allow Chromium's download. Note that installing with
+``nbconvert[webpdf]`` we install a compatible version of pyppeteer for this usecase.
 
 PDF conversion on a limited TeX environment
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -19,8 +19,8 @@ If you're new to Python, we recommend installing `Anaconda <https://www.anaconda
 a Python distribution which includes nbconvert and the other Jupyter components.
 
 .. important::
-    
-    To unlock nbconvert's full capabilities requires Pandoc and TeX 
+
+    To unlock nbconvert's full capabilities requires Pandoc and TeX
     (specifically, XeLaTeX). These must be installed separately.
 
 Installing Pandoc
@@ -39,22 +39,23 @@ On other platforms, you can get pandoc from
 Installing TeX
 --------------
 
-For converting to PDF, nbconvert uses the TeX document preparation 
-ecosystem. It produces an intermediate ``.tex`` file which is 
-compiled by the XeTeX engine with the LaTeX2e format (via the 
-``xelatex`` command) to produce PDF output. 
+For converting to PDF, nbconvert can either use `Puppeteer <https://pptr.dev>`_
+(with ``--to webpdf``) or the TeX document preparation ecosystem (with
+``--to pdf``). In the latter case it produces an intermediate ``.tex`` file
+which is compiled by the XeTeX engine with the LaTeX2e format (via the ``xelatex``
+command) to produce PDF output.
 
 .. versionadded:: 5.0
-    
-    We use XeTeX as the rendering engine rather than pdfTeX (as 
-    in earlier versions). XeTeX can access fonts through native 
-    operating system libraries, it has better support for OpenType 
-    formatted fonts and Unicode characters. 
 
-To install a complete TeX environment (including XeLaTeX and 
-the necessary supporting packages) by hand can be tricky. 
-Fortunately, there are packages that make this much easier. These 
-packages are specific to different operating systems: 
+    We use XeTeX as the rendering engine rather than pdfTeX (as
+    in earlier versions). XeTeX can access fonts through native
+    operating system libraries, it has better support for OpenType
+    formatted fonts and Unicode characters.
+
+To install a complete TeX environment (including XeLaTeX and
+the necessary supporting packages) by hand can be tricky.
+Fortunately, there are packages that make this much easier. These
+packages are specific to different operating systems:
 
 * Linux: `TeX Live <http://tug.org/texlive/>`_
 
@@ -63,10 +64,10 @@ packages are specific to different operating systems:
 * macOS (OS X): `MacTeX <http://tug.org/mactex/>`_.
 * Windows: `MikTex <http://www.miktex.org/>`_
 
-Because nbconvert depends on packages and fonts included in standard 
-TeX distributions, if you do not have a complete installation, you 
-may not be able to use nbconvert's standard tooling to convert 
-notebooks to PDF. 
+Because nbconvert depends on packages and fonts included in standard
+TeX distributions, if you do not have a complete installation, you
+may not be able to use nbconvert's standard tooling to convert
+notebooks to PDF.
 
 PDF conversion on a limited TeX environment
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -76,7 +77,7 @@ If you are only able to install a limited TeX environment, there are two main ro
 1. Using TeX by hand
     a. You could convert to ``.tex`` directly; this requires Pandoc.
     b. edit the file to accord with your local environment
-    c. run ``xelatex`` directly. 
+    c. run ``xelatex`` directly.
 2. Custom exporter
-    a. You could write a :ref:`custom exporter <external_exporters>` 
-       that takes your system's limitations into account. 
+    a. You could write a :ref:`custom exporter <external_exporters>`
+       that takes your system's limitations into account.
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -20,8 +20,8 @@ a Python distribution which includes nbconvert and the other Jupyter components.
 
 .. important::
 
-    To unlock nbconvert's full capabilities requires Pandoc and TeX
-    (specifically, XeLaTeX). These must be installed separately.
+    To unlock its full capabilities, nbconvert requires Pandoc, TeX
+    (specifically, XeLaTeX) and Chromium. These must be installed separately.
 
 Installing Pandoc
 -----------------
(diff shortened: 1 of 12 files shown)


maartenbreddels:
Very cool! Just a note, the pyppeteer dependency does not include the full chrome browser, it will only download it when it will be used. Meaning it is a light dependency, so I have no trouble with this being standard dependency, in case someone is wondering about that.

davidbrochart:
Yes, should we mention that in `docs/source/install.rst`?

bollwyvl:
I would recommend waiting until (and setting a minimum pin to) the next
release of pyppeteer... The current release has ssl disabled, and a lot of
other strange things which will be addressed in their next version.

I'm also interested in pursuing a conda-forge package of said 80mb
downloaded binary, as "doesn't work at all the first time without the
internet at runtime" is a pretty big caveat that I'd like to see have at
least _a_ solution.

davidbrochart:
> I'm also interested in pursuing a conda-forge package of said 80mb downloaded binary, as "doesn't work at all the first time without the internet at runtime" is a pretty big caveat that I'd like to see have at least _a_ solution.

On the other hand you will need an internet connection when installing with conda, so the first time download can also be seen as the installation. And you will need the internet at runtime anyway when e.g. showing a Leaflet map.

bollwyvl:
Indeed, I take issue with any runtime use of unpkg, requirejs, rawgit,
cdnjs, etc. for the same reason! I'm saddled with not-entirely-unique use
cases like corporate hpc and field work where getting pip/conda/yum/docker
stuff is reasonable, and that's about it.

Longer term, for the web case, i've imagined a static jupyterlab (e.g
Deathbeds/jyve) got smart enough to be used, with otherwise-useful
dependencies, demonstrated on:

https://github.com/deathbeds/nbconvert-pdfqt

As the name suggests, it used qt(webengine), which is at least going to be
stable, if you pin it, and shippable, and may already be in your
environment. Heck, i'd take white label Firefox ESR, and I think there's
some effort going on in p*ppeteer to make that happen.

So anyhow... Maybe naming the flag "pyppeteer" would be more precise.

MSeal:
> Compared to --to pdf, this allows to have widgets in the PDF document. It also supports the same templates as --to html.

Glad some alternative patterns to xelatex conversion could be offered!

> Yes, should we mention that in docs/source/install.rst?

Yes similar to xelatex we want large downloads / extra install requirements to be visible.

> I would recommend waiting until (and setting a minimum pin to) the next
release of pyppeteer... The current release has ssl disabled, and a lot of
other strange things which will be addressed in their next version.

Good to know.

MSeal:
> Indeed, I take issue with any runtime use of unpkg, requirejs, rawgit,
cdnjs, etc. for the same reason! I'm saddled with not-entirely-unique use
cases like corporate hpc and field work where getting pip/conda/yum/docker
stuff is reasonable, and that's about it.

Unfortunately most the tooling available for doing the more complex conversions require (large) external packaging to work :/

Having an `extra_requires` install pattern for pypuppeteer runtime dependencies could be made to download the runtime dependencies during installation if that helps.

maartenbreddels:
How I'd like to see this is to have the current pyppeteer be as it is, and a normal (non-optional) dependency of nbconvert, so it works for most users.
However, if we can also provide a way to package the chromium package on pypi and conda-forge, but without having an explicitly dependency (maybe `pip install nbconvert[all]` should though...) we can have the best of two worlds:
 * Most users will have a quick download/install, and this new features works out of the box.
 * In cases where you want the bundled chromium to be pre-downloaded, you'd pre install it with pip or conda (`pip install chromium`?) and pyppeteer should pick that up.

SylvainCorlay:
We could use pyppeteer's `check_chromium` first, and if it returns `False`, stop there with an error message

Unless a specific command line option such as `--allow_chromium_download` is used with nbconvert.

At least it would not be implicitely downloaded.

SylvainCorlay:
> https://github.com/deathbeds/nbconvert-pdfqt

The webpdf exporter could totally use QtWebkit instead of Chromium/pyppeteer's.

However, one advantage of going with HTML export first rather than rendering the notebook with a JS front-end is that we could use the new template system of nbconvert 6.0 (the same as Voilà), and produce PDFs from such a template.

```
jupyter nbconvert foobar.ipynb --to webpdf --template=gridstack
```

bollwyvl:
Sure, there are many wrinkles on it.

Perhaps the webpdf entry point is more of a facade/strategy pattern. Just
like latex, none of them are going to "just work" out of the box, on every
platform, for every user.

Because of that, I'm wary of stuffing everything into hard nbconvert deps,
which so many tools depend on: I don't think every user/deployment (e.g
nbviewer) will want pyqt _or_ pyppeteer, if it's even available for their
platform. Well-specified pip [extras] (and derived conda meta packages) are
probably more sustainable.

MSeal:
> Well-specified pip [extras] (and derived conda meta packages) are
probably more sustainable.

I agree with that sentiment. A lot of organizations don't install the dependencies for latex / pdf conversion today because they're massive dependency chains that take a lot of space on an image. I think any solution that pull large dependency chains should be an optional install so teams can continue to choose what they can / want to support / install by default.

SylvainCorlay:
> > `--allow-chromium-download` flag
>
> The last commit implements this idea.
> 

I think that with this change, this is an acceptable situation. What do you think @bollwyvl @MSeal ?

MSeal:
I'm good with this change, though I'd like to have a basic sanity test on the interfaces added for the preprocessor if we could before merge.

Do we want to add an extra_requires install for chromium dependencies as well? I could add that fairly easily I think if we want it.

davidbrochart:
@MSeal I added two tests:
- `test_webpdf_without_chromium`, which is executed first so chromium is not downloaded yet. It checks that the PDF cannot be created.
- `test_webpdf_with_chromium`, which allows chromium's download and checks that a PDF file is created.

Let me know if it's fine.
For the `extras_require`, I'm not sure how to do that because we need `pyppeteer` to be already installed in order to call `download_chromium()`. If you know how to do it I think it would be nice.

maartenbreddels:
Using pyproject.toml it should be possible:
```
[build-system]
requires = ["setuptools >= 40.6.0", "pyppeteer"]
```

Not sure we'd want to go there though.

I'd say we get chromium on pypi and conda-forge.

davidbrochart:
> I'd say we get chromium on pypi and conda-forge.

Not an easy task :smile:

davidbrochart:
> Using pyproject.toml it should be possible:
> 
> ```
> [build-system]
> requires = ["setuptools >= 40.6.0", "pyppeteer"]
> ```

Thanks @maartenbreddels, it works. @MSeal is it what you had in mind? If so we should document it.

davidbrochart:
I reverted the last commit because it seemed to install chromium although the extra requirement was not specified in the pip install.

MSeal:
Yeah that's close to what's needed, but you need to not fun the function call ahead of registering the action. I can look later to help, I don't remember without looking into it deeper again how to get the function hook into an extra requires... might require come distutils command registration hackery.

MSeal:
Ok I dug through the setuptools code and I believe it's not possible to define in setuptools the right response to the pip install command because pip doesn't pass it along. The workaround is to make a new package that always downloads chromium and make that the extra_requires dependency. Let's not worry about the extra_requires for now since it looks like it's not possible to do correctly atm without an extra package.

bollwyvl:
> maybe we should stick to the versions bundled by pyppeteer.

With these kinda things (p*ppeteer, chromedriver, etc.), there's pretty much The Version you can tell people to use unless _you_ test Some Other Version with your code on every platform where you expect people to use it. Browser automation is just that nasty.

For that reason, and others, I'll again lodge my :-1: about making `pyppeteer` a hard dependency: the `pyppeteer` deps are _not_ particularly lightweight: here's a cleaned up diff of an install from conda-forge (where, with some gross patching, SSL is enforced to download a _non-insecure-by-default_ version of chromium, which I _also_ [take issue with](https://github.com/conda-forge/pyppeteer-feedstock/pull/4)):

``` 
+  appdirs            
+  brotlipy           
+  cffi               
+  cryptography       
+  idna               
+  pycparser          
+  pyee               
+  pyopenssl          
+  pyppeteer          
+  pysocks            
+  tqdm               
+  urllib3            
+  websockets         
```

But let's say we ignore that: I would be a bit more careful in the docs distinguishing between `puppeteer` (which [decktape](https://github.com/astefanutti/decktape) uses) and `pyppeteer`, which claims to be:

- _Unofficial Python port of puppeteer JavaScript (headless) chrome/chromium browser automation library._

... and, as mentioned, is undergoing a significant overhaul after being partially abandoned, and then forked, and now somewhat back on track, where the next release will almost certainly be a breaking change. So if it _is_ kept, it would need to be pinned to `0.2.2` or so (and we'd have to re-apply all of the _basic security_ patches on conda-forge to get that version out), or just hold on this until... whatever the next version is comes out.

> implementing a qt version

@maartenbreddels: :crying_cat_face: yeah, with a bit of work, I could get the one running for 5.9 again. 5.12 is more problematic, I thought, due to some OSX issues. for the lab-based one, I had to do a fair amount of tinkering to listen for various signals, and then I'd _still_ wait a couple seconds afterwards. But yeah, qt is qt, and a whole separate universe of things to deal with, but does have stability in its favor.

> I'd say we get chromium on pypi and conda-forge.

It has been attempted a few times on conda-forge (by me _and_ @maartenbreddels i think... stalebot has since closed them). No idea on `pypi`. Seems hard to `whl`. 

For this _sole purpose_ (being driven by pyppeteer) it could make sense to [package the downloaded installer](https://github.com/conda-forge/pyppeteer-feedstock/issues/3), either _as installed_ (which gives me the willies), or into a cache, and then executed, but that would probably _also_ require an upstream patch. But would work offline, once you had installed, and _could_ be mirrored, sanely.

We do a binary repackaging of firefox, and nobody's too excited about it, but it works under most conditions. But chrom* is a particularly nasty beast: I don't even know how they build qt, with its fork of chromium: I think there's special, secret azure boxes I don't know about.

MSeal:
> or that reason, and others, I'll again lodge my -1 about making pyppeteer a hard dependency

alright we can make it an extra_requires field and make the option raise an exception that explains what needs to be installed to use the extension. I can push that change to the PR. This however isn't used in conda... so not sure what to do for conda other than make an entirely separate recipe.

> So if it is kept, it would need to be pinned to 0.2.2

I'm ok with the extra_requires being a hard pin dependency

> For this sole purpose (being driven by pyppeteer) it could make sense to package the downloaded installer,

I'm also ok with just listing the explicit install instructions in our readthedocs like we do for latex dependencies. They're outside the scope of a python package tbh and describing over complicated engineering is likely the safest approach where it's problematic imo.

MSeal:
Yay, tests (and docs) pass now. Any concerns with merging what's there now with the proposed changes applied?

docs/source/install.rst
```diff
@@ -19,9 +19,9 @@ If you're new to Python, we recommend installing `Anaconda <https://www.anaconda
 a Python distribution which includes nbconvert and the other Jupyter components.
 
 .. important::
-    
-    To unlock nbconvert's full capabilities requires Pandoc and TeX 
-    (specifically, XeLaTeX). These must be installed separately.
+
+    To unlock its full capabilities, nbconvert requires Pandoc, TeX
+    (specifically, XeLaTeX) and Chromium. These must be installed separately.
```
bollwyvl, line 24: Not a blocker, but I feel like this could be made more binary, and a bit less dramatic:

```
Exporting to PDF files requires additional, separately installed dependencies.
```

bollwyvl: Also, is pandoc required for this PR's workflow?

docs/source/install.rst
```diff
@@ -39,22 +39,23 @@ On other platforms, you can get pandoc from
 Installing TeX
 --------------
 
-For converting to PDF, nbconvert uses the TeX document preparation 
-ecosystem. It produces an intermediate ``.tex`` file which is 
-compiled by the XeTeX engine with the LaTeX2e format (via the 
-``xelatex`` command) to produce PDF output. 
+For converting to PDF, nbconvert can either use `Puppeteer <https://pptr.dev>`_
```
bollwyvl, line 42: This section should probably not even mentionthis PR's contant, and certainly not `puppeteer`, as it is not used.

docs/source/install.rst
```diff
@@ -63,10 +64,21 @@ packages are specific to different operating systems:
...
+Because nbconvert depends on packages and fonts included in standard
+TeX distributions, if you do not have a complete installation, you
+may not be able to use nbconvert's standard tooling to convert
+notebooks to PDF.
+
+Installing Chromium
+-------------------
+
+When converting to PDF with ``--to webpdf``, nbconvert uses `Puppeteer <https://pptr.dev>`_, and
+more precisely its Python bindings `pyppeteer <https://github.com/pyppeteer/pyppeteer>`_. pyppeteer
```
bollwyvl, line 76: Again, puppeteer is not used. Since there's less going on, perhaps moving this section above the TeX section makes sense.

docs/source/install.rst
```diff
@@ -63,10 +64,21 @@ packages are specific to different operating systems:
...
+needs a special version of Chromium, and can automatically download it. Since the download size is
+quite large (around 100 MB), nbconvert prevents it by default. If pyppeteer finds a suitable
+version of Chromium on your system, it will try and use it. Otherwise, please use the
```
bollwyvl, line 79: I wouldn't even advertise it working with any existing chrom*.

docs/source/usage.rst
```diff
@@ -82,17 +83,27 @@ PDF
 
   Generates a PDF via latex. Supports the same templates as ``--to latex``.
 
+.. _convert_webpdf:
+
+WebPDF
+~~~~~~
+* ``--to webpdf``
+
+  Generates a PDF via puppeteer (chromium web browser). Supports the same
```
bollwyvl, line 92: would again remove reference to pupeteer. Chromium should probably be capitalized.

nbconvert/exporters/webpdf.py
```diff
@@ -0,0 +1,64 @@
...
+from .html import HTMLExporter
+
+class WebPDFExporter(HTMLExporter):
+    """Writer designed to write to PDF files.
+
+    This inherits from :class:`HTMLExporter`. It creates the HTML using the
+    template machinery, and then runs puppeteer (through pyppeteer) to create
+    a pdf.
+    """
+    export_from_notebook="PDF via Puppeteer"
```
bollwyvl, line 18: again, non puppeteer

nbconvert/exporters/webpdf.py
```diff
@@ -0,0 +1,64 @@
...
+
+    allow_chromium_download = Bool(False,
+        help="Whether to allow downloading chromium if no suitable version is found on the system."
```
bollwyvl, line 21: capitalize Chromium

nbconvert/exporters/webpdf.py
```diff
@@ -0,0 +1,64 @@
...
+            from pyppeteer.util import check_chromium
+        except ModuleNotFoundError as e:
+            raise RuntimeError("Pyppeteer is not installed to support Web PDF conversion. "
+                               "Please install `nbconvert[webpdf]` to enable.") from e
+        if not self.allow_chromium_download and not check_chromium():
+            raise RuntimeError("No suitable chromium executable found on the system. "
+                               "Please use '--allow-chromium-download' to allow downloading one.")
+        return launch
+
+    def run_puppeteer(self, html):
```
bollwyvl, line 36: again, non puppeteer

setup.py
```diff
@@ -229,15 +229,23 @@ def get_data_files():
...
 
 extra_requirements = {
-    'test': ['pytest', 'pytest-cov', 'ipykernel', jupyter_client_req, 'ipywidgets>=7'],
+    'test': ['pytest',
+             'pytest-cov',
+             'pytest-dependency',
+             'ipykernel',
+             jupyter_client_req,
+             'ipywidgets>=7',
+             'pyppeteer==0.2.2',
```
bollwyvl, line 238: perhaps hoist this hard pin, like `jupyter_client`

docs/source/install.rst
MSeal: Pandoc is used extensively by a range of preprocessors and exporters, so yes it's required in the general sense but not for the webpdf installation specifically. I can try rewording this to match xelatex vs web pdf conversion requirements.

docs/source/install.rst
MSeal: I didn't touch those docs in my pass, I can clean it up to reference pyppeteer everywhere that's relevant instead of puppeteer.

docs/source/install.rst
MSeal: Good point since that was changed since first discussed.

setup.py
MSeal: Sure, can do

SylvainCorlay:
Made a quick pass taking into account @bollwyvl's review comments.

docs/source/usage.rst
```diff
@@ -89,9 +89,11 @@ WebPDF
 ~~~~~~
 * ``--to webpdf``
 
-  Generates a PDF via puppeteer (chromium web browser). Supports the same
-  templates as ``--to html``. Requires ``pyppeteer`` to be installed via
-  ``nbconvert[webpdf]``.
+  Generates a by first rendering to HTML, rendering the HTML Chromium headless, and
```
davidbrochart, line 92:
```suggestion
  Generates a PDF by first rendering to HTML, rendering the HTML Chromium headless, and
```

MSeal:
Gonna merge this. If need be we can tweak in master before the release (hoping to get to the big 6.0 soon rather than another alpha)

SylvainCorlay:
:tada:

PR #1286, opened 2020-06-10
Fix reveal template and make it work without --slides

maartenbreddels:
This will make the reveal template more in line with [voila-reveal](https://github.com/voila-dashboards/voila-reveal/].

Instead of having to do:
```
python -m nbconvert reveal.ipynb --execute --to slides
```

One can also do:
```
python -m nbconvert reveal.ipynb --execute --to html --template reveal
```

The `--slides` option has the advantage of being able to provide configuration options. I hope long term we can get rid of the `--slides` altogether.

TODO: see if we can use the lab template, and have theming.

### DIFF
--- a/share/jupyter/nbconvert/templates/reveal/conf.json
+++ b/share/jupyter/nbconvert/templates/reveal/conf.json
@@ -1,5 +1,5 @@
 {
-    "base_template": "classic",
+    "base_template": "lab",
     "mimetypes": {
         "text/html": true
     }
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
> dark theme

This is great!

> jupyter-widget manager 

I am not sure this is in the scope of this PR to resolve that. I think that the HTML manager of ipywidgets 8.0 should drop the CSS...

maartenbreddels:
> I am not sure this is in the scope of this PR to resolve that.

I agree, out of scope.

SylvainCorlay:
As discussed, this is a net improvement because the reveal template is broken on master, and this fixes it. Merging and iterating in the next PR.

damianavila:
So, you ended up moving logic to the template. I was expecting something like this... although I prefer the logic outside from the template, I see some benefit on this approach so I am OK with the change.

> The --slides option has the advantage of being able to provide configuration options.

Yep, that's important in some contexts.

> I hope long term we can get rid of the --slides altogether.

With this PR fixed, is the --slides option still working as expected?

PR #1289, opened 2020-06-15
[WIP] Changed preprocessor to processor.

MSeal:
The naming of Preprocessor always confused me as there was not Processor.. so what was it preempting? This was brought up by others in conversations about nbconvert as well, so while we're doing the major change I wanted to make the naming more understandable for others. We could add some of the additional functionality like chaining to this, but for now the PR is just a rename with deprecation warnings on the old names.

Also rebuilt the docs with current code changes in master + this branch.

I still need to write some tests and cover backwards compatibility. Haven't checked if traitlet assignments get forwarded and that old import / inheritrance patterns will still work with a warning (did some preliminary testing for the easiest cases).


t-makaro:
I'm not really a fan of this rename. Preprocessor has always made sense to me since it is processing pre-export. If you use nbconvert as a python library, then this makes total sense. You preprocess the notebook, then you export it. The definition of preprocess is "to subject (data) to preliminary processing" which it exactly what our preprocessors do.

Also, we still have postprocessors for after export (though I think technically anything that can be done with a postprocessor could be done with a custom writer).

Also, this will break users' config files, which I think are more commonly used than custom templates and thus this will cause more issues for people.

MSeal:
Hmm well alright. I always found the naming confusing personally. But we can leave it and I'll close the PR for now.

choldgraf:
My 2 cents with naming is that many names will make sense to some and not to others, and it could be hard to have something that makes sense to all. I think I personally agree w/ @MSeal that the "pre" is a bit unnecessary. That said, I don't have strong opinions about it.

Perhaps another way to approach this is to improve the documentation a bit. Looking at the docs (https://nbconvert.readthedocs.io/en/latest/) I don't see a strong user-facing high lever description of nbconvert, the main steps in conversion, etc. Perhaps we can add that in and use this as an opportunity to explain nomenclature (we can re-use some of @t-makaro's description above)

PR #1302, opened 2020-06-24
Minor fixups in template exporter

SylvainCorlay:


nbconvert/exporters/templateexporter.py
```diff
@@ -213,25 +213,20 @@ def _raw_template_changed(self, change):
...
         if self.file_extension:
-            return self.file_extension + ".j2"
+            return self.file_extension + '.j2'
         else:
-            return self.file_extension
-
-    @default('template_file')
-    def _template_file_default(self):
-        if self.template_extension:
-            return 'index' + self.template_extension
```
SylvainCorlay, line 234: This method was duplicated in the class.

nbconvert/exporters/templateexporter.py
```diff
@@ -213,25 +213,20 @@ def _raw_template_changed(self, change):
...
+            return '.j2'
```
SylvainCorlay, line 229: I don't think we ever fall in that case in nbconvert, but this appears to be cleaner IMO.

MSeal:
Thanks for the cleanup

PR #1303, opened 2020-06-24
Change preparation stage of slides exporter a preprocessor

SylvainCorlay:
Remove the `prepare` function and make it a (private) preprocessor.

### DIFF
--- a/nbconvert/exporters/slides.py
+++ b/nbconvert/exporters/slides.py
@@ -96,12 +96,6 @@ class SlidesExporter(HTMLExporter):
     def _template_extension_default(self):
         return '.html.j2'
 
-    # Register metadata preprocessor
-    #################################
-    def __init__(self, **kw):
-        resources = super().__init__(**kw)
-        self.register_preprocessor(_RevealMetadataPreprocessor, enabled=True)
-
     # Extra resources
     #################################
     reveal_url_prefix = Unicode(
--- a/nbconvert/exporters/slides.py
+++ b/nbconvert/exporters/slides.py
@@ -9,82 +9,101 @@ from warnings import warn
 from traitlets import Bool, Unicode, default
 
 from .html import HTMLExporter
+from ..preprocessors.base import Preprocessor
 
-def prepare(nb):
-    """Add some convenience metadata on cells for the slide template."""
-    nb = deepcopy(nb)
 
-    for cell in nb.cells:
-        # Make sure every cell has a slide_type
-        cell.metadata.slide_type = cell.metadata.get('slideshow', {}).get('slide_type', '-')
+class _RevealMetadataPreprocessor(Preprocessor):
+    # A custom preprocessor adding convenience metadata to cells
 
-    # Find the first visible cell
-    for index, cell in enumerate(nb.cells):
-        if cell.metadata.slide_type not in {'notes', 'skip'}:
-            cell.metadata.slide_type = 'slide'
-            cell.metadata.slide_start = True
-            cell.metadata.subslide_start = True
-            first_slide_ix = index
-            break
-    else:
-        raise ValueError("All cells are hidden, cannot create slideshow")
+    def preprocess(self, nb, resources=None):
+        nb = deepcopy(nb)
 
-    in_fragment = False
+        for cell in nb.cells:
+            # Make sure every cell has a slide_type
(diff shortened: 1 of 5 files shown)


nbconvert/exporters/slides.py
```diff
@@ -9,82 +9,101 @@
 from traitlets import Bool, Unicode, default
 
 from .html import HTMLExporter
+from ..preprocessors.base import Preprocessor
+
+
+class _RevealMetadataPreprocessor(Preprocessor):
```
MSeal, line 15: How does this now get associated to run for the slides exporter? I didn't see a change there to run this preprocessor by default unless I missed something?

MSeal: Also, should this be in preprocessors even though it's a private class? Maybe instead of being private we could put it inside a subdirectory `internal` under preprocessors so it's discoverable but indicates it's not nbconvert specific functionality? This would make any config assignments read a little nicer imo and fit the organization scheme a little more. Not fixed on the idea but thought I'd bring it up

SylvainCorlay: It gets associated in the constructor of the exporter.

SylvainCorlay: What I would like to do is make the templates specify a list of preprocessors to run.

Ideally we should be able to drop the SlidesExporter completely.

SylvainCorlay: I.e. have a new key in `conf.json` for `"prepropessors"`...

SylvainCorlay:
Ping @damianavila I would love to have your feedback on this.

This removes the need from a slides exporter different from the html exporter by having 

 - the `prepare` stage done in a preprocessor
 - registering the preprocessors to register in the template configuration.

Ideally, I will move the resources as well to the preprocessor, completely removing the need for the `slides` exporter.

SylvainCorlay:
ping @mkcor FYI

SylvainCorlay:
@MSeal I am tempted to move forward with this.

I would like to iterate with a refactor of how we include the various javascript into the templates (require, jquery), which will likely conflict with this.

nbconvert/exporters/slides.py
SylvainCorlay: Done.

share/jupyter/nbconvert/templates/reveal/conf.json
```diff
@@ -2,5 +2,8 @@
     "base_template": "lab",
     "mimetypes": {
         "text/html": true
-    }
-}
\ No newline at end of file
+    },
+    "preprocessors": [
+        "nbconvert.exporters.slides._RevealMetadataPreprocessor"
```
MSeal, line 7: Great!

SylvainCorlay:
> I'm ok with a merge if you feel comfortable. I didn't do extensive local testing outside the unittests still working.

I tested on a few presentations locally. Functionally, it appears to be completely the same as before.

SylvainCorlay:
If we agree to go with this, I will follow up with a PR

 - cleaning up the reveal template
 - documenting this new field
 - somewhat improving how we invoke the registration of the preprocessors.

SylvainCorlay:
Ping @timkpaine I think this also corresponds to what you were requesting with respect to the prelaunch hook: registering custom nbconvert preprocessors.

timkpaine:
@sylvaincorlay partially, though for voilà you'd want access to the tornado handler so you can pull headers and authenticate if necessary (which is why we still need it on my fork).

SylvainCorlay:
> @SylvainCorlay partially, though for voilà you'd want access to the tornado handler so you can pull headers and authenticate if necessary (which is why we still need it on my fork).

At least the example that you provided is something that can be done with this.

Since the preprocessor is run in the backend, why would you want to access the tornado handler?

timkpaine:
@SylvainCorlay if you authenticate with an access token, then when you visit voila that token will be present in the request. You can then inject user impersonation into the first cell of the notebook itself. Its also helpful for other reasons, like logging who the user is, where they're access from, the notebook/template/parameters they're passing in, etc.

damianavila:
@SylvainCorlay, I little bit late to the party, but changes look OK to me. I like the whole idea about making the prepare stuff into a preprocessor.

PR #1309, opened 2020-06-29
Pass on structure of reveal template

SylvainCorlay:
This makes the reveal template closer to the other html templates, reusing the blocks from the base template.

Ping @maartenbreddels @damianavila.


damianavila:
You are also adding some widget support and moving out css stuff from the template into the proper css file, right? I like that as well.

SylvainCorlay:
That's right.

I will make a pass to have the result ISO with the stable nbconvert.

The initial template refactor broke the reveal template, so I felt it was OK to iterate.

damianavila:
> so I felt it was OK to iterate.

That's totally OK!

PR #1310, opened 2020-07-01
fix: get rid of both template_path and template_paths

maartenbreddels:
We had both `template_path` and `template_paths`, which was confusing, now it's only the plural.

PS: and I sneaked in 2 totally unrelated fixes
 * unittest.mock is available in Python3
 * I apparently had a package installed that used 'text/tex' as mime-type.

### DIFF
--- a/nbconvert/exporters/tests/test_webpdf.py
+++ b/nbconvert/exporters/tests/test_webpdf.py
@@ -6,7 +6,7 @@
 import io
 import pytest
 
-from mock import patch
+from unittest.mock import patch
 
 from .base import ExportersTestsBase
 from ..webpdf import WebPDFExporter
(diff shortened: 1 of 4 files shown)


maartenbreddels:
1 Travis failure probably needs a restart.

nbconvert/exporters/tests/test_webpdf.py
```diff
@@ -6,7 +6,7 @@
 import io
 import pytest
 
-from mock import patch
```
MSeal, line 9: If we got all of the `from mock imports` changed we could also remove it from the requirements

MSeal:
Ahh nvm I was looking at the viola tests. Yes the that test failure is a known issue with jupyter_client.

PR #1311, opened 2020-07-01
Feature: make preprocessor in conf.json a dict

maartenbreddels:
Needs rebase after #1310  (only last commit matters)

This makes it easier for an inheriting template, or voila to add preprocessors.

We could also move the `CSSHTMLHeaderPreprocessor` in the HTML exporter also to the conf.json

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -11,6 +11,7 @@ import os
 import uuid
 import json
 import warnings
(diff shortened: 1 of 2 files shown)


maartenbreddels:
This change triggered an early 'fill' of template_paths. In the tests, template_name is set afterward, causing the template_paths (which depend on template_name) to be out of sync. The 2nd commit fixes this.

PR #1313, opened 2020-07-02
Fixed test warnings

MSeal:
Noticed some deprecation warnings .. these were fixable, and I created https://github.com/ipython/ipython/issues/12419 for the other fixable warning as I wasn't clear what the expected change should be


SylvainCorlay:
Thanks!

Issue #637, opened 2017-08-02
Add ``--clean`` option to ``--to notebook`` to strip all metadata and output?


ceball:
On projects where multiple people contribute to "source code only" notebooks stored in revision control, system-specific or user-specific metadata in the notebooks are often unwanted sources of diffs and conflicts. I would love to be able to tell contributors how to remove such metadata before submitting changes, without them having to download and configure external software/scripts. Apart from the extra steps involved in downloading, installing, and configuring external scripts, such scripts may not be up to date with the changing notebook format.

I think http://timstaley.co.uk/posts/making-git-and-jupyter-notebooks-play-nice/ describes my problem in more detail (although that article also covers git integration, which I'm not looking for here). I understand that many users of notebooks do want to keep all the metadata and output, but I guess where the notebooks are used more like e.g. python source code (which doesn't normally specify the version of python, or store any of the output of the program, or any metadata about the system that last opened the file), the metadata is always annoying. Right now we have to explain to people how to remove widget metadata, scroll state, kernel spec, etc, or we have to look at diffs on github that include all that stuff and deal with conflicts about things that don't matter to us.

My first question is, would it be an acceptable feature for ``--to notebook`` to offer a ``--clean`` option to remove everything other than the inputs and the minimum required metadata? Something that does the equivalent of the following, from the article above:

```
jq --indent 1 \
    '
    (.cells[] | select(has("outputs")) | .outputs) = []
    | (.cells[] | select(has("execution_count")) | .execution_count) = null
    | .metadata = {"language_info": {"name":"python", "pygments_lexer": "ipython3"}}
    | .cells[].metadata = {}
    ' notebook.ipynb
```

My second question is, how much work would it be to implement such a feature (if that's easy to answer)?

Thanks!

takluyver:
Something like this already exists as a separate tool, called nbstripout:
https://pypi.python.org/pypi/nbstripout

Personally, I'd say that having it as a separate tool is neater than adding another option to nbconvert.

ceball:
> Something like this already exists as a separate tool, called nbstripout:
> https://pypi.python.org/pypi/nbstripout

[...]

PR #1314, opened 2020-07-02
Changed clearmetadata preprocessor to remove all metadata by default

MSeal:
I decided to just go ahead and implement the requested change for https://github.com/jupyter/nbconvert/issues/637. Technically it changes the default behavior to clear notebook and cell metadata, but I think that's what's usually intended on first-use and I added a flag to fall back to the old behavior if need-be. I also added a flag for preserving specific keys by name.

### DIFF
--- a/nbconvert/preprocessors/clearmetadata.py
+++ b/nbconvert/preprocessors/clearmetadata.py
@@ -3,13 +3,20 @@
 # Copyright (c) IPython Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-from traitlets import Set
+from traitlets import Bool, Set
 from .base import Preprocessor
 
 class ClearMetadataPreprocessor(Preprocessor):
     """
     Removes all the metadata from all code cells in a notebook.
     """
+
+    clear_notebook_metadata = Bool(True,
+        help=("Flag to choose if notebook metadata is to be cleared "
+              "in addition to cell metadata.")).tag(config=True)
+    preserve_metadata_keys = Set(
+        help=("Indicates the keys to preserve when deleting metadata "
+               "across both cells and notebook metadata fields.")).tag(config=True)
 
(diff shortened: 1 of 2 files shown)


MSeal:
Gonna merge this in another day or two if no one any comments fyi.

nbconvert/preprocessors/clearmetadata.py
```diff
@@ -3,20 +3,47 @@
...
-from traitlets import Set
+from traitlets import Bool, Set
 from .base import Preprocessor
 
 class ClearMetadataPreprocessor(Preprocessor):
     """
     Removes all the metadata from all code cells in a notebook.
     """
 
+    clear_notebook_metadata = Bool(True,
```
t-makaro, line 14: Do we perhaps want the ability to choose between notebook metadata, cell metadata or both? It might be easier than specifically setting every piece of cell metadata that we want to keep.

MSeal:
Thanks for the review. I think I can incorporate those suggested changes fairly easily

t-makaro:
I just realised that ```"language_info"``` is more nested than I thought. It's specifically, the ```name``` subfield of ```language_info``` that I want to keep which I guess makes it a bit harder. The rest can be tossed.

MSeal:
@SylvainCorlay or @t-makaro If you wanted to quickly review this for the release tomorrow I believe it's addressed the PR comments.

SylvainCorlay:
I am not too opinionated about this.

Leaving this to you judgement @t-makaro @jupyter/everybody 
@maartenbreddels let me know if you want to chime in on this before the release!

MSeal:
I'm going to merge for now since I think the comments are well addressed and tested to avoid blocking release stuff

MSeal:
If we need to adjust later I can make tweaks

PR #1324, opened 2020-07-15
DOC: Autoreformat all the docstring to match the numpydoc format.

Carreau:
This uses a prototype project of mine, so might be incomplete, and try
to be better at understanding (and reformatting, docstrings compliant
with numpydoc.

I understand that you would be opposed to large scale reformatting.

This is on top of #1323 which made numpydoc choke on the docstring and
prevented correct reformatting.


Carreau:
That's all the changes my script found, so I think we are good for now.

PR #1329, opened 2020-07-20
Fix AttributeError when getting slide_type

Happyholic1203:
This is an updated PR for #604


MSeal:
A little annoying that different tools set that attribute to different types and we have to guard for it here :/

PR #1334, opened 2020-07-29
Further simplify revealjs template

SylvainCorlay:
This is a simple change in the reveal template that makes it slightly more similar to other html templates.


SylvainCorlay:
This is a trivial change.

PR #1335, opened 2020-07-29
Remove dependencies attribute to Reveal.initialize

SylvainCorlay:


SylvainCorlay:
These do not exist anymore with the latest nbconvert.

PR #1351, opened 2020-08-25
DOC: use pygments_style provided by HTML theme

mgeier:
The sickly-green background for code blocks is quite ugly, and it will become even more so when https://github.com/spatialaudio/nbsphinx/pull/463 will be merged, which will use the same background color for notebook code cells.

I guess the current setting is a remnant from using `sphinx-quickstart`, which added the `'sphinx'` style up to some time ago, when this was removed. Now the default is to not change the `pygment_style`.


MSeal:
Thanks for the improvements to our docs!

PR #1354, opened 2020-08-26
Add documentation for the nbconvert 6.0 template system

SylvainCorlay:

PR #1364, opened 2020-09-09
DOC: remove unused files from source directory

mgeier:
If I'm not mistaken, those files are not used anymore since the example notebook that was using them has been removed in #1354.

I'm not sure whether the content of those files should be kept in the docs, but for now it doesn't seem to be visible in the rendered docs.


SylvainCorlay:
Thanks! Good catch!

Issue #1370, opened 2020-09-11
nbconvert 6 doesn't run `preprocess_cell` with ExecuteProcessor


jph00:
With the latest nbconvert from pypi, `preprocess_cell` in subclasses of `ExecuteProcessor` no longer run. Looking at the source, I can see that nothing is calling `preprocess_cell` any more.

I'm not sure if this is a bug, a missing feature, or a question... It at least appears to be some missing documentation.

Can you please advise how, in nbconvert v6 / nbclient, you recommend executing a notebook (without saving it) and preprocessing each cell before executing?

(More generally, I'm totally overwhelmed at all the things that have broken in nbdev with nbconvert v6, and I'm only finding the answers by looking at the source code for v6 and commit logs. Is there anywhere I could find an 'upgrade guide', and/or a list of API changes? I've read the release notes, but they don't show many of the changes.)

MSeal:
The missing `preprocess_cell` looks like an oversight on my part when we moved the code to nbclient a long while back. So I'd classify this as a bug and I think I see what we need to do to fix it. Nbclient doesn't directly support cell manipulation just-in-time on it's own, and papermill persists notebooks as it executes, so there's not simple way to run a notebook and preprocess in the same way without the bug fixed. If you know what you want to change before executing any cells, you could run a custom preprocessor over a notebook and then run it through nbclient directly as an alternative pattern.

I'm sorry you feel frustrated with the change. In general most things should be behaving as they were with a few exceptions. And to be fair this is a major version bump on a large repo that's been sitting in alpha/beta releases for 6 months. We tried to test all usage patterns but it's a bit of a leaky boat with all the inheritance involved downstream, and a lot of expected behavior from nbconvert was/still is undocumented conventions. So apologies again that it's causing you friction, it wasn't intended to do so. There isn't a direct upgrade guide because of this intent, but I we can try to be responsive and update docs around things we missed or any unintentional changes in behavior.

MSeal:
Issue is resolved. There's 1 more patch needed for an unrelated issue before release. I have a working branch of nbdev that uses nbconvert 6.0 with minimal changes I'll PR once nbconvert 6.0.2 is released.

willingc:
Thanks @MSeal for your contributions to turn this around quickly. :sunny:

@jph00 FYI 6.0.2 has been released.

jph00:
Gosh that was speedy.

And @MSeal even jumped into nbdev and fixed our code for us too! 🎉😍

[...]

PR #1380, opened 2020-09-12
Execute preprocess cells

MSeal:
Fixes inheritance models where subclasses are used to implement preprocess_cells on the execute preprocessor.

There's one hack in this process, but the alternatives are all worse for maintenance pov (including some ones I didn't mention like not dual inheriting classes) so it is what it is to maintain backwards and forward compatibility.

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -78,7 +78,7 @@ class ExecutePreprocessor(Preprocessor, NotebookClient):
             cell: NotebookNode,
             cell_index: int,
             execution_count: Optional[int] = None,
-            store_history: bool = True) -> NotebookNode:
+            store_history: bool = False) -> NotebookNode:
         """
         Executes a single code cell.
 
@@ -113,7 +113,11 @@ class ExecutePreprocessor(Preprocessor, NotebookClient):
             The cell which was just processed.
         """
         # Copied and intercepted to allow for custom preprocess_cell contracts to be fullfilled
-        return self.preprocess_cell(cell, self.resources, cell_index, execution_count=execution_count, store_history=store_history)
+        self.store_history = store_history
+        cell, resources = self.preprocess_cell(cell, self.resources, cell_index)
+        if execution_count:
+            cell['execution_count'] = execution_count
+        return cell, resources
 
     def preprocess_cell(self, cell, resources, index, **kwargs):
         """
@@ -131,5 +135,5 @@ class ExecutePreprocessor(Preprocessor, NotebookClient):
             Index of the cell being processed
         """
         self._check_assign_resources(resources)
-        cell = run_sync(NotebookClient.async_execute_cell)(self, cell, index, **kwargs)
+        cell = run_sync(NotebookClient.async_execute_cell)(self, cell, index, store_history=self.store_history)
         return cell, self.resources
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -64,7 +64,7 @@ class ExecutePreprocessor(Preprocessor, NotebookClient):
             Additional resources used in the conversion process.
         """
         # Although this solution may look wrong, it considers that preprocess acts like an init on execution state.
-        # we need to capture it here again to properly reset (traitlet assignments are
+        # Therefore, we need to capture it here again to properly reset because traitlet assignments are
         # not passed). The risk is if traitlets apply any side effects for dual init,
         # The risk should be manageable, and this approach minimizes side-effects
         # relative to other alternatives. One alternative but rejected implementation would be to copy the client's init internals
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -44,6 +44,17 @@ class ExecutePreprocessor(Preprocessor, NotebookClient):
 
         The input argument *nb* is modified in-place.
 
+        Note that this function recalls NotebookClient.__init__, which may look wrong.
+        However since the preprocess call acts line an init on exeuction state it's expected.
+        Therefore, we need to capture it here again to properly reset because traitlet 
+        assignments are not passed. There is a risk if traitlets apply any side effects for
+        dual init.
+        The risk should be manageable, and this approach minimizes side-effects relative
+        to other alternatives.
+
+        One alternative but rejected implementation would be to copy the client's init internals
+        which has already gotten out of sync with nbclient 0.5 release before nbcovnert 6.0 released.
+
         Parameters
         ----------
         nb : NotebookNode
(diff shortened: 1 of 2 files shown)


nbconvert/preprocessors/execute.py
```diff
@@ -56,11 +63,77 @@ def preprocess(self, nb, resources=None, km=None):
         resources : dictionary
             Additional resources used in the conversion process.
         """
-        # Copied from NotebookClient init :/
-        self.nb = nb
-        self.km = km
-        if resources:
-            self.resources = resources
-        self.reset_execution_trackers()
+        # This may look wrong, but preprocess acts like an init on execution state and
```
willingc, line 66:
```suggestion
        # Although this solution may look wrong, it considers that preprocess acts like an init on execution state.
```

nbconvert/preprocessors/execute.py
```diff
@@ -56,11 +63,77 @@ def preprocess(self, nb, resources=None, km=None):
...
+        # we need to capture it here again to properly reset (traitlet assignments are
```
willingc, line 67:
```suggestion
        # Therefore, we need to capture it here again to properly reset because traitlet assignments are
```

nbconvert/preprocessors/execute.py
```diff
@@ -56,11 +63,77 @@ def preprocess(self, nb, resources=None, km=None):
...
+        # not passed). The risk is if traitlets apply any side effects for dual init,
```
willingc, line 68:
```suggestion
        not passed. There is a risk if traitlets apply any side effects for dual init.
```

MSeal:
Forgot to mention this fixes https://github.com/jupyter/nbconvert/issues/1370

nbconvert/preprocessors/execute.py
```diff
@@ -56,11 +63,77 @@ def preprocess(self, nb, resources=None, km=None):
...
+        # but it should be ok. The alternative is to copy the client's init internals
```
willingc, line 69:
```suggestion
        # The risk should be manageable, and this approach minimizes side-effects
        # relative to other alternatives. One alternative but rejected implementation would be to copy the client's init internals
```

nbconvert/preprocessors/execute.py
willingc, line 66: Let's move all of this into docstring with an important directive so that this populates through to the docs.

nbconvert/preprocessors/execute.py
```diff
@@ -56,11 +63,77 @@ def preprocess(self, nb, resources=None, km=None):
...
+        cell : NotebookNode cell
+            Notebook cell being processed
+        resources : dictionary
+            Additional resources used in the conversion process.  Allows
+            preprocessors to pass variables into the Jinja engine.
+        index : int
+            Index of the cell being processed
+        """
+        self._check_assign_resources(resources)
+        cell = run_sync(NotebookClient.async_execute_cell)(self, cell, index, store_history=self.store_history)
```
willingc, line 138: Let's comment this line to unpack what is happening here.

willingc (commented):
@MSeal Thanks for doing this. I've suggested some modifications for clarity. Consider this approved once the failing test is resolved.

There are a lot of deprecation warnings. Do we want to handle in docs or open an issue for future cleanup.

MSeal:
> There are a lot of deprecation warnings. Do we want to handle in docs or open an issue for future cleanup.

Future

nbconvert/preprocessors/execute.py
```diff
@@ -56,11 +63,78 @@ def preprocess(self, nb, resources=None, km=None):
...
         """
-        # Copied from NotebookClient init :/
-        self.nb = nb
-        self.km = km
-        if resources:
-            self.resources = resources
-        self.reset_execution_trackers()
+        # Although this solution may look wrong, it considers that preprocess acts like an init on execution state.
+        # Therefore, we need to capture it here again to properly reset because traitlet assignments are
+        not passed. There is a risk if traitlets apply any side effects for dual init.
```
willingc, line 68:
```suggestion
        # not passed. There is a risk if traitlets apply any side effects for dual init.
```

willingc (approved):

nbconvert/preprocessors/execute.py
jph00: Do subclasses need to call `super().preprocess_cell()` now? (IIRC they didn't need to before; it's fine either way, although if this has changed it would be great to document the breaking change in the release notes.)

willingc: @MSeal ^^ I'm happy to document the response.

MSeal: It's probably best to always call super for this preprocessor now.

Technically you don't have to so long as you call something equivilent to or exactly `NotebookClient.async_execute_cell`, but that's fairly complex to explain or expect given the subtlety of the multi-inheritance and overrides at play.

@willingc Documenting as such would be a good idea, happy to merge if you want to make the PR

SylvainCorlay:
cc @davidbrochart

nbconvert/preprocessors/execute.py
```diff
@@ -56,11 +74,73 @@ def preprocess(self, nb, resources=None, km=None):
...
+            Notebook cell being processed
+        resources : dictionary
+            Additional resources used in the conversion process.  Allows
+            preprocessors to pass variables into the Jinja engine.
+        index : int
+            Index of the cell being processed
+        """
+        self._check_assign_resources(resources)
+        # Because nbclient is an async library, we need to wrap the parent async call to generate a syncronous version.
+        cell = run_sync(NotebookClient.async_execute_cell)(self, cell, index, store_history=self.store_history)
```
davidbrochart, line 145: @MSeal You could have called `NotebookClient.execute_cell` here, right?
Shouldn't this cell execution be in `async_execute_cell`?

MSeal: Well, possibly. I was worried any refactors on the wrapping pattern for `NotebookClient.execute_cell` might make it lazy, causing in infinite loop here in the future since this overwrites async_execute_cell in this child class to allow for intercepting and running `preprocess_cell`... basically it's a hack to allow us to intercept the the execute_cell call from nbclient and run whatever the nbconvert inheritance tree demands.

davidbrochart:
Shouldn't `preprocess_cell` be in nbclient? I have the feeling that it would have led to a cleaner design.

MSeal:
> Shouldn't preprocess_cell be in nbclient? I have the feeling that it would have led to a cleaner design.

Should nbclient have pre/post execute hooks? Yes probably. I went with a solution here that could unblock the issue with dependencies as is. I think someone was working on adding these in a similar flavor to papermill's hooks but since it hasn't landed it's likely not an active effort. I would not copy the preprocess_cell pattern itself, but if we had the ability to register callbacks in nbclient it would help make this simpler for sure.

davidbrochart:
Maybe we could revive https://github.com/jupyter/nbclient/pull/79?

MSeal:
Maybe, or making a more concentrated PR for just hook operations?

Issue #1372, opened 2020-09-11
nbconvert raises NotImplementedError on Windows & Python 3.8


QuLogic:
See for example, [this matplotlib build](https://ci.appveyor.com/project/matplotlib/matplotlib/builds/35141301/job/97upt8o42a55wswk).

```
[NbConvertApp] Converting notebook C:\projects\matplotlib\lib\matplotlib\tests\test_nbagg_01.ipynb to notebook
Traceback (most recent call last):
  File "c:\miniconda37-x64\envs\test-environment\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "c:\miniconda37-x64\envs\test-environment\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Miniconda37-x64\envs\test-environment\Scripts\jupyter-nbconvert.EXE\__main__.py", line 7, in <module>
  File "c:\miniconda37-x64\envs\test-environment\lib\site-packages\jupyter_core\application.py", line 270, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "c:\miniconda37-x64\envs\test-environment\lib\site-packages\traitlets\config\application.py", line 837, in launch_instance
    app.start()
  File "c:\miniconda37-x64\envs\test-environment\lib\site-packages\nbconvert\nbconvertapp.py", line 345, in start
    self.convert_notebooks()
  File "c:\miniconda37-x64\envs\test-environment\lib\site-packages\nbconvert\nbconvertapp.py", line 519, in convert_notebooks
    self.convert_single_notebook(notebook_filename)
  File "c:\miniconda37-x64\envs\test-environment\lib\site-packages\nbconvert\nbconvertapp.py", line 484, in convert_single_notebook
    output, resources = self.export_single_notebook(notebook_filename, resources, input_buffer=input_buffer)
  File "c:\miniconda37-x64\envs\test-environment\lib\site-packages\nbconvert\nbconvertapp.py", line 413, in export_single_notebook
    output, resources = self.exporter.from_filename(notebook_filename, resources=resources)
  File "c:\miniconda37-x64\envs\test-environment\lib\site-packages\nbconvert\exporters\exporter.py", line 182, in from_filename
    return self.from_file(f, resources=resources, **kw)
  File "c:\miniconda37-x64\envs\test-environment\lib\site-packages\nbconvert\exporters\exporter.py", line 200, in from_file
[...]

PR #1383, opened 2020-09-13
Apply windows work-around fix for async applications

MSeal:
Mirrors https://github.com/nteract/papermill/pull/517 for patching the entrypoint. Should fix https://github.com/jupyter/nbconvert/issues/1372 if they're using command line. If they are importing as a library they might still need to apply the same work-around because it doesn't do anything if async calls have already begun so we only protect the application launch point.

No real way to write a unittest for this one

This should be the last issue for 6.0.2 release with a few fixes added

### DIFF
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -13,6 +13,7 @@ import logging
 import sys
 import os
 import glob
+import asyncio
 from textwrap import fill, dedent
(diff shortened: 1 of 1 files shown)


MSeal:
This is a pretty straight forward patch copy -- I'm going to merge and get a release going so the open issues can be addressed asap

Issue #1384, opened 2020-09-13
“Using as a command line tool” docs list outdated template names


Kwpolska:
The [Using as a command line tool](https://nbconvert.readthedocs.io/en/6.0.2/usage.html#convert-html) documentation page lists `full` and `basic` as valid options for `--template` with the HTML exporter. Those templates are no longer supported in v6.0.2:

```pytb
$ jupyter nbconvert --to html --template full hello.ipynb
Traceback (most recent call last):
  File "venv/lib/python3.8/site-packages/traitlets/traitlets.py", line 528, in get
    value = obj._trait_values[self.name]
KeyError: 'template_paths'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "venv/bin/jupyter-nbconvert", line 8, in <module>
    sys.exit(main())
  File "venv/lib/python3.8/site-packages/jupyter_core/application.py", line 270, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "venv/lib/python3.8/site-packages/traitlets/config/application.py", line 664, in launch_instance
    app.start()
  File "venv/lib/python3.8/site-packages/nbconvert/nbconvertapp.py", line 350, in start
    self.convert_notebooks()
  File "venv/lib/python3.8/site-packages/nbconvert/nbconvertapp.py", line 519, in convert_notebooks
    self.exporter = cls(config=self.config)
  File "venv/lib/python3.8/site-packages/nbconvert/exporters/templateexporter.py", line 323, in __init__
    super().__init__(config=config, **kw)
  File "venv/lib/python3.8/site-packages/nbconvert/exporters/exporter.py", line 115, in __init__
    self._init_preprocessors()
  File "venv/lib/python3.8/site-packages/nbconvert/exporters/templateexporter.py", line 489, in _init_preprocessors
    conf = self._get_conf()
  File "venv/lib/python3.8/site-packages/nbconvert/exporters/templateexporter.py", line 505, in _get_conf
    for path in map(Path, self.template_paths):
  File "venv/lib/python3.8/site-packages/traitlets/traitlets.py", line 556, in __get__
    return self.get(obj, cls)
  File "venv/lib/python3.8/site-packages/traitlets/traitlets.py", line 535, in get
    value = self._validate(obj, dynamic_default())
  File "venv/lib/python3.8/site-packages/nbconvert/exporters/templateexporter.py", line 516, in _template_paths
    template_names = self.get_template_names()
[...]

PR #1386, opened 2020-09-13
Fixed html doc usage docs to fix 6.x

MSeal:
Partially resolves https://github.com/jupyter/nbconvert/issues/1384

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -44,8 +44,8 @@ HTML
 ~~~~
 * ``--to html``
 
-  HTML Export. Be aware that if you were using custom copies of the old 5.x
-  template files you need to use ``--template-file path/to/old/file.tpl``
+  HTML Export.  Note on backward compatibility:  Be aware that if you were using custom copies of the old 5.x
+  template files (i.e. ``--template``), you will now need to use ``--template-file path/to/old/file.tpl``
   in order to use that file in compatibility mode as opposed to other options.
 
   - ``--template lab`` (default)
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -44,15 +44,18 @@ HTML
 ~~~~
 * ``--to html``
 
-  - ``--template full`` (default)
+  HTML Export. Be aware that if you were using custom copies of the old 5.x
+  template files you need to use ``--template-file path/to/old/file.tpl``
+  in order to use that file in compatibility mode as opposed to other options.
+
+  - ``--template lab`` (default)
 
     A full static HTML render of the notebook.
-    This looks very similar to the interactive view.
+    This looks very similar to the jupyter lab interactive view.
 
-  - ``--template basic``
+  - ``--template classic``
 
-    Simplified HTML, useful for embedding in webpages, blogs, etc.
-    This excludes HTML headers.
+    Simplified HTML, using the classic jupyter look and feel.
 
 .. _convert_latex:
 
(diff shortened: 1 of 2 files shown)


docs/source/usage.rst
```diff
@@ -44,15 +44,18 @@ HTML
 ~~~~
 * ``--to html``
 
-  - ``--template full`` (default)
+  HTML Export. Be aware that if you were using custom copies of the old 5.x
```
willingc, line 47:
```suggestion
  HTML Export.  Note on backward compatibility:  Be aware that if you were using custom copies of the old 5.x
```

docs/source/usage.rst
```diff
@@ -44,15 +44,18 @@ HTML
...
+  template files you need to use ``--template-file path/to/old/file.tpl``
```
willingc, line 48:
```suggestion
  template files (i.e. ``--template``), you will now need to use ``--template-file path/to/old/file.tpl``
```

willingc (approved):
Some small changes for clarity. Thanks @MSeal

docs/source/usage.rst
```diff
@@ -44,15 +44,18 @@ HTML
...
     A full static HTML render of the notebook.
-    This looks very similar to the interactive view.
+    This looks very similar to the jupyter lab interactive view.
 
-  - ``--template basic``
+  - ``--template classic``
 
-    Simplified HTML, useful for embedding in webpages, blogs, etc.
-    This excludes HTML headers.
+    Simplified HTML, using the classic jupyter look and feel.
```
Kwpolska, line 58: I think it might be useful to also document the “embeddable” version, which would be `--template-file=classic/base.html.j2` (or `lab/`)

MSeal: `--template=classic` is preferred over `--template-file=classic/base.html.j2`. It's possible in a future release that when the backwards compatibility is removed that the second pattern might not work that way anymore (`--template=classic --template-file=notbase.html.j2` might be required in some distance release, or no support template-file at all). https://nbconvert.readthedocs.io/en/6.0.2/changelog.html?highlight=--template-file#id3 outlines how to use template-file to refer to a specific `.tpl` 5.x file.

I'm hesitant to promote it more elsewhere for other uses because some of these always-supported patterns made the 6.0 upgrade harder. But I could be convinced it's a good idea to document it more if people feel strongly about it.

Kwpolska: `--template=classic` is not a replacement for `basic`/`classic/base.html.j2`, because it produces a full 13k line (?!) HTML document with styles and JS. This is not usable for embedding, at least not without using a HTML parser to get rid of the 13k lines of `<head>`. `classic/base.html.j2` produces just the notebook code and expects styles to be added in elsewhere (and we do it that way in Nikola).

MSeal: Hmm you're correct. I wasn't as involved with the html template changes from 5.x to 6.x. It looks like adding `--template-file=base.html.j2` along with `--template=classic` gets you back to close to the original `--template=base` pattern without all the styles and JS. Can you try that and see if it matches expectations? I can discuss with the other maintainers how to better preserve that capability in a clearer format with the new system if so.

Kwpolska: It seems to work, both from the command line and code. It would be good to have an officially recommended way to do this in the docs.

MSeal: I agree. Thanks for raising the concern and documentation gap. I'll talk with the other maintainers and see if we want to introduce an easier way to run this combination before I document it, but I don't believe we'll remove the pattern that worked for you here.

PR #1387, opened 2020-09-13
Fixed template assignments with old basic and full names with tpl extension

MSeal:
In the refactor that simplified the template_name fetching we broke using `template=basic` despite the path to the compatibility templates being added to the search paths. This fixes this by searching the compatibility directory last.

Second half of resolving https://github.com/jupyter/nbconvert/issues/1384

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -565,18 +565,37 @@ class TemplateExporter(Exporter):
                 template_dir = os.path.join(base_dir, base_template)
                 if os.path.exists(template_dir):
                     found_at_least_one = True
-                conf_file = os.path.join(template_dir, 'conf.json')
-                if os.path.exists(conf_file):
-                    with open(conf_file) as f:
-                        conf = recursive_update(json.load(f), conf)
-            for root_dir in root_dirs:
-                template_dir = os.path.join(root_dir, 'nbconvert', 'templates', base_template)
-                if os.path.exists(template_dir):
-                    found_at_least_one = True
-                conf_file = os.path.join(template_dir, 'conf.json')
-                if os.path.exists(conf_file):
-                    with open(conf_file) as f:
-                        conf = recursive_update(json.load(f), conf)
+                    conf_file = os.path.join(template_dir, 'conf.json')
+                    if os.path.exists(conf_file):
+                        with open(conf_file) as f:
+                            conf = recursive_update(json.load(f), conf)
+                    break
+            if not found_at_least_one:
+                for root_dir in root_dirs:
+                    template_dir = os.path.join(root_dir, 'nbconvert', 'templates', base_template)
+                    if os.path.exists(template_dir):
+                        found_at_least_one = True
+                        conf_file = os.path.join(template_dir, 'conf.json')
+                        if os.path.exists(conf_file):
+                            with open(conf_file) as f:
+                                conf = recursive_update(json.load(f), conf)
+                        break
+            if not found_at_least_one:
+                for root_dir in root_dirs:
+                    compatibility_name = base_template + '.tpl'
+                    compatibility_path = os.path.join(root_dir, 'nbconvert', 'templates', 'compatibility', compatibility_name)
+                    if os.path.exists(compatibility_path):
+                        found_at_least_one = True
+                        warnings.warn(
+                            f"5.x template name passed '{self.template_name}'. Use 'lab' or 'classic' for new template usage.",
+                            DeprecationWarning)
+                        self.template_file = compatibility_name
+                        # Without these added the templates can't find their transitive extensions
+                        if base_template == 'basic':
+                            template_names.extend(['classic', 'base'])
+                        if base_template == 'full':
+                            template_names.extend(['lab', 'base'])
+                        break
             if not found_at_least_one:
                 paths = "\n\t".join(root_dirs)
                 raise ValueError('No template sub-directory with name %r found in the following paths:\n\t%s' % (base_template, paths))
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -590,11 +590,6 @@ class TemplateExporter(Exporter):
                             f"5.x template name passed '{self.template_name}'. Use 'lab' or 'classic' for new template usage.",
                             DeprecationWarning)
                         self.template_file = compatibility_name
-                        # Without these added the templates can't find their transitive extensions
-                        if base_template == 'basic':
-                            template_names.extend(['classic', 'base'])
-                        if base_template == 'full':
-                            template_names.extend(['lab', 'base'])
                         break
             if not found_at_least_one:
                 paths = "\n\t".join(root_dirs)
--- a/share/jupyter/nbconvert/templates/classic/base.html.j2
+++ b/share/jupyter/nbconvert/templates/classic/base.html.j2
@@ -1,5 +1,5 @@
-{%- extends 'display_priority.j2' -%}
-{% from 'celltags.j2' import celltags %}
+{%- extends 'base/display_priority.j2' -%}
+{% from 'base/celltags.j2' import celltags %}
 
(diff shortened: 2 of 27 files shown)


nbconvert/exporters/templateexporter.py
```diff
@@ -560,18 +560,37 @@ def get_template_names(self):
...
-                    found_at_least_one = True
-                conf_file = os.path.join(template_dir, 'conf.json')
-                if os.path.exists(conf_file):
-                    with open(conf_file) as f:
-                        conf = recursive_update(json.load(f), conf)
+                    conf_file = os.path.join(template_dir, 'conf.json')
+                    if os.path.exists(conf_file):
+                        with open(conf_file) as f:
+                            conf = recursive_update(json.load(f), conf)
+                    break
```
MSeal, line 567: A minor optimization, this does change the algorithm to find the first instead of the last match. I _think_ this is fine but I'd like other eyes on it in case I need to reverse the iteration ordering.

SylvainCorlay:
thanks for the ping. i will look into this Monday morning paris time.

willingc (approved):
LGTM @MSeal. Will leave for @SylvainCorlay to review and comment/merge.

MSeal:
Hmm before merging the comment thread on https://github.com/jupyter/nbconvert/pull/1386 should be read. I think we (I?) eliminated the `base` template pattern and erroneously stated classic maps to that? I posted what I think works around it to get back to original functionality but I'd like some input and what we should do (because it impacts this backwards compatibility PR)

nbconvert/exporters/templateexporter.py
SylvainCorlay: I am not 100% sure about the `break` statement.
What if configuration is overriden in a higher precedence extra template dir?

Ping @maartenbreddels. Let me know what you think.

nbconvert/exporters/templateexporter.py
```diff
@@ -560,18 +560,37 @@ def get_template_names(self):
...
+            if not found_at_least_one:
+                for root_dir in root_dirs:
+                    template_dir = os.path.join(root_dir, 'nbconvert', 'templates', base_template)
+                    if os.path.exists(template_dir):
+                        found_at_least_one = True
+                        conf_file = os.path.join(template_dir, 'conf.json')
+                        if os.path.exists(conf_file):
+                            with open(conf_file) as f:
+                                conf = recursive_update(json.load(f), conf)
+                        break
```
SylvainCorlay, line 577: Same here. Not sure the recursion on root_dirs should be conditional on `found_at_least_one`.

nbconvert/exporters/templateexporter.py
```diff
@@ -560,18 +560,37 @@ def get_template_names(self):
...
+                        warnings.warn(
+                            f"5.x template name passed '{self.template_name}'. Use 'lab' or 'classic' for new template usage.",
+                            DeprecationWarning)
+                        self.template_file = compatibility_name
+                        # Without these added the templates can't find their transitive extensions
+                        if base_template == 'basic':
+                            template_names.extend(['classic', 'base'])
+                        if base_template == 'full':
+                            template_names.extend(['lab', 'base'])
+                        break
```
SylvainCorlay, line 593: This approach may collide with an actual template called `full` or `basic`.

Maybe we should condition on this at a higher level...

MSeal: I was thinking of changing the include paths inside template files to include the template name. I think that would resolve a few issues and not require these hacks for the backwards compatibility, as template extensions are similarly running into 2nd hop extension path failures.

MSeal: Updated PR with the change I discussed -- extending existing templates from custom templates no longer explodes when files are included across template boundaries. e.g. Fixes https://github.com/jupyter/nbconvert/issues/1394

MSeal:
I added the basic template back as well, since it has a need and we made using it more difficult and broke backwards compatibility.

nbconvert/exporters/templateexporter.py
MSeal: The recursion isn't conditional, it recurses to completion before returning back to this context

share/jupyter/nbconvert/templates/classic/base.html.j2
```diff
@@ -1,5 +1,5 @@
-{%- extends 'display_priority.j2' -%}
-{% from 'celltags.j2' import celltags %}
+{%- extends 'base/display_priority.j2' -%}
+{% from 'base/celltags.j2' import celltags %}
```
SylvainCorlay, line 2: Why is it necessary to refer to base here?

SylvainCorlay:
@MSeal it should not be required to add the path to the base template in jinja templates, unless you override them locally.

Does your change make it necessary?

share/jupyter/nbconvert/templates/classic/base.html.j2
SylvainCorlay: Actually, if these changes in the jinja templates are required, this means this is really backward incompatible.

SylvainCorlay:
@maartenbreddels @MSeal would you like to have a synchronous discussion about how to fix this the best early next week?

nbconvert/exporters/templateexporter.py
SylvainCorlay: But it will not merge the configuration with other matches?

nbconvert/exporters/templateexporter.py
SylvainCorlay: > changing the include paths inside template files to include the template name

I am not sure about this, because it is a backward incompatible change with respect to the current behavior.

MSeal: How is it backwards incompatible? The templates behave identically within nbconvert but now when extended from external templates will behave correctly (since they don't have the proper template paths set to include this template). It also protects from multiple inheritance of file extensions -- e.g. there's ambiguity when importing null.js is some places if you didn't include the template name in the extension path.

nbconvert/exporters/templateexporter.py
MSeal: It wasn't before because it overwrote the `conf` and `conf_file` whenever it found a template until the last one found. I don't think it changes the behavior beyond preferring the first exact match over the last exact match. If you're really worried I can remove the optimization.

MSeal:
> @MSeal it should not be required to add the path to the base template in jinja templates, unless you override them locally.
>
> Does your change make it necessary?

Only in the sense that the backwards compatibility templates were not working without it. Similarly there's several open issues pointing out that extending `lab/index.html.js` and other files within the templating system fail if those templates use relative extension paths within nbconvert.

> @maartenbreddels @MSeal would you like to have a synchronous discussion about how to fix this the best early next week?

Would have to be tomorrow (Sunday) or Tuesday morning as my Monday has no wiggle room on my end.

nbconvert/exporters/templateexporter.py
SylvainCorlay: It is backward incompatible for custom template authors who inherit from base.

maartenbreddels:
Tuesday works for me.

I think we should not prefix with `/lab` `/base` in the extends, as we'll lose the ability to override those files from other directories.

Going over the code, and trying to make the test pass without the prefixes, I don't think we can go with full 5.x compatibility as it is now. I think if we want compatibility with full.tpl and base.tpl, we should just copy the old content of nbconvert 5.x, and keep/add the deprecation warning.

MSeal:
> Tuesday works for me.

Invite sent.

> we'll lose the ability to override those files from other directories.

I see. Unfortunately since the template_paths don't dynamically reflect extensions made, supporting that pattern breaks the ability for folks to extend the `<template>/index.<format>.j2` files as the local extensions fail (not just for backwards compatibility). I didn't see a clean way to automatically add templates that were extended to the template_paths to prevent this issue.

MSeal:
Closing as alternative fixes were merged

Issue #1392, opened 2020-09-14
ExecutePreprocessor adds execution_count to markdown cells


hunse:
As of the newest release (6.0.2), ExecutePreprocessor adds an `execution_count` field to markdown cells, which causes problems when trying to load the notebook later (for example, when using `nbformat.read`).

```python
from nbconvert.preprocessors import ExecutePreprocessor
import nbformat

nb = nbformat.v4.new_notebook()
nb["cells"] = [
    nbformat.v4.new_markdown_cell("Title"),
]

# run notebook to generate output
ep = ExecutePreprocessor(timeout=600, kernel_name="python3")
ep.preprocess(nb, {"metadata": {"path": "."}})

print(nb["cells"][0])
```

Running the above script with `nbconvert==6.0.1` results in a cell with no `execution_count`, as expected, but with 6.0.2, `execution_count` is now there.

The error given by `nbformat.read` is:
```
Failed validating 'additionalProperties' in markdown_cell:

On instance['cells'][0]:
{'cell_type': 'markdown',
 'execution_count': 1,
 'metadata': {},
 'source': 'Title   \n\nserach\n\n'}
ERROR    traitlets:__init__.py:109 Notebook JSON is invalid: Additional properties are not allowed ('execution_count' was unexpected)
```

To see the above error, add the following to the above script
```python
import logging
logging.basicConfig(level=logging.ERROR)

path = "./tmp.ipynb"

with open(path, "w", encoding="utf-8") as f:
    nbformat.write(nb, f)

with open(path, "r", encoding="utf-8") as f:
    nbformat.read(f, as_version=4)
```

MSeal:
Thanks for raising -- I think I see the issue. Let me test and see if I can reproduce / patch it tonight

MSeal:
Will have to tackle this later this week. Been a busy couple days

MSeal:
6.0.3 release is out with fix on pypi (conda to follow when it catches up)

PR #1396, opened 2020-09-16
Fixed erroneous markdown counts for execute preprocessor breaking nbformat rules

MSeal:
Resolves https://github.com/jupyter/nbconvert/issues/1392. This was an issue in nbclient at one point as well and I forgot to guard against it in making the work-around to enable the preprocess cell calls in nbconvert.

This warrants a patch release when it's fixed,


willingc (approved):
Thanks @MSeal

PR #1406, opened 2020-09-21
Async execute cells

davidbrochart:
Fixes https://github.com/jupyter/nbclient/issues/108

### DIFF
--- a/nbconvert/preprocessors/execute.py
+++ b/nbconvert/preprocessors/execute.py
@@ -3,10 +3,10 @@ and updates outputs"""
 
 # Copyright (c) IPython Development Team.
 # Distributed under the terms of the Modified BSD License.
+import asyncio
 from typing import Optional
 from nbformat import NotebookNode
 from nbclient import NotebookClient, execute as _execute
-from nbclient.util import run_sync
 # Backwards compatability for imported name
 from nbclient.exceptions import CellExecutionError
 
@@ -76,7 +76,12 @@ class ExecutePreprocessor(Preprocessor, NotebookClient):
         """
         NotebookClient.__init__(self, nb, km)
         self._check_assign_resources(resources)
-        self.execute()
+        try:
+            loop = asyncio.get_event_loop()
+        except RuntimeError:
+            loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(loop)
+        loop.run_until_complete(self.async_execute())
         return self.nb, self.resources
 
     async def async_execute_cell(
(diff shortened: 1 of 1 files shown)


nbconvert/preprocessors/execute.py
```diff
@@ -120,13 +125,13 @@ async def async_execute_cell(
...
         self.store_history = store_history
-        cell, resources = self.preprocess_cell(cell, self.resources, cell_index)
+        cell, resources = await self.preprocess_cell(cell, self.resources, cell_index)
         # Apply rules from nbclient for where to apply execution counts
         if execution_count and cell.cell_type == 'code' and cell.source.strip():
             cell['execution_count'] = execution_count
         return cell, resources
 
-    def preprocess_cell(self, cell, resources, index, **kwargs):
+    async def preprocess_cell(self, cell, resources, index, **kwargs):
```
SylvainCorlay, line 134: other preprocessors'  `preprocess_cell` are not async. Should we make an async version of that method and call it from `async_execute_cell`?

davidbrochart: No, I think we just need to ensure it is async, I'm pushing another commit.

SylvainCorlay: it will most likely have consequences, because anyone overriding `preprocess_cell` in a non-async way will have surprises.

nbconvert/preprocessors/execute.py
```diff
@@ -76,7 +77,12 @@ def preprocess(self, nb, resources=None, km=None):
         """
         NotebookClient.__init__(self, nb, km)
         self._check_assign_resources(resources)
-        self.execute()
+        try:
+            loop = asyncio.get_event_loop()
+        except RuntimeError:
+            loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(loop)
+        loop.run_until_complete(self.async_execute())
```
SylvainCorlay, line 85: We may be just moving the issue here. `run_until_complete` will fail if the loop is already running.

davidbrochart: I agree, if nbconvert is used in a notebook for instance. But not if it is used from the CLI.

nbconvert/preprocessors/execute.py
davidbrochart: I don't think so, since we now call `ensure_async(preprocess_cell())`, it can be async or not.

SylvainCorlay: as discussed in person, for people who override `preprocess_cell` and then call `super.preprocess_cell` in their implementation without `await` this should be an issue, but I need to check.

MSeal: It is an issue and it will break downstream projects like `nbdev` if `preprocess_cell` is made async in nbconvert

davidbrochart: We should try because I don't think it is a problem with the `ensure_async`.

MSeal: I believe it will be an issue to mix:

```
In [8]: class A:
   ...:     async def foo(self):
   ...:         print("bar")
   ...: 

In [9]: class B(A):
   ...:     def foo(self):
   ...:         print("foo")
   ...:         A.foo(self)
   ...:         print("!")
   ...: 

In [10]: B().foo()
foo
<ipython-input-9-33f3ae08d87f>:4: RuntimeWarning: coroutine 'A.foo' was never awaited
  A.foo(self)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
!
```

Making the subclass aware of the parent being async but not being async itself on the override does weird things too:

```
In [12]: class B(A):
    ...:     def foo(self):
    ...:         print("foo")
    ...:         await A.foo(self)
    ...:         print("!")
    ...: 

In [13]: B().foo()
Out[13]: <coroutine object B.foo at 0x7f1d87179c40>
```

SylvainCorlay:
Note: it seems that the nest-asyncio author just pushed a fix: https://github.com/erdewit/nest_asyncio/issues/34

If a patch release is published, we should probably just use that and close this.

SylvainCorlay:
Closing for now, since this approach will probably not work.

However, I would like to make a pass on the interface between nbclient and the execute preprocessor.

PR #1411, opened 2020-09-22
Remove CSS preprocessor from default

SylvainCorlay:
This allows the preprocessor to be configured at a template level.

### DIFF
--- a/nbconvert/exporters/exporter.py
+++ b/nbconvert/exporters/exporter.py
@@ -84,7 +84,6 @@ class Exporter(LoggingConfigurable):
                                   'nbconvert.preprocessors.ExecutePreprocessor',
                                   'nbconvert.preprocessors.coalesce_streams',
                                   'nbconvert.preprocessors.SVG2PDFPreprocessor',
-                                  'nbconvert.preprocessors.CSSHTMLHeaderPreprocessor',
                                   'nbconvert.preprocessors.LatexPreprocessor',
                                   'nbconvert.preprocessors.HighlightMagicsPreprocessor',
                                   'nbconvert.preprocessors.ExtractOutputPreprocessor',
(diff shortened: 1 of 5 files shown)


SylvainCorlay:
@maartenbreddels what do you think of this?

nbconvert/exporters/exporter.py
```diff
@@ -84,7 +84,6 @@ class Exporter(LoggingConfigurable):
                                   'nbconvert.preprocessors.ExecutePreprocessor',
                                   'nbconvert.preprocessors.coalesce_streams',
                                   'nbconvert.preprocessors.SVG2PDFPreprocessor',
-                                  'nbconvert.preprocessors.CSSHTMLHeaderPreprocessor',
```
SylvainCorlay, line 87: Ultimately, we probably should remove all exporter-specific preprocessors from the defaults, and move them up to
 - `@default(preprocessor)` in exporters
 - or further up to the template configuration for template exporters.

maartenbreddels:
Yes, totally makes sense. Can you comment on the different configuration (style not specified) for the different configurations?

maartenbreddels (approved):

SylvainCorlay:
> Can you comment on the different configuration (style not specified) for the different configurations?

The default is `JupyterStyle`, which is the JupyterLab Pygment style...

We don't have the CSS for it in the case of classic.

SylvainCorlay:
@maartenbreddels this may require a 6.1.0 though.

maartenbreddels:
I don't consider this a feature, more like a chore or fix. It doesn't break anything right, and too light to be called a feature? Possibly..

SylvainCorlay:
> It doesn't break anything right

This removes highlighting for people who are doing a custom HTML template that does not inherit from `lab`, `classic`, or `reveal`, but for example only from `base`.

Currently, highlighting is broken in classic.

MSeal (approved):
I'm ok with a patch version for this change. It's a slight behavioral change but it's not a breaking change.

PR #1425, opened 2020-09-24
comment typo fix

akhmerov:
Nothing of importance, but made me think what is meant for a second. (Unless I really misunderstood, in which case sorry!)


MSeal (approved):
Thanks, would you mind also applying the change to the `latex/null.j2` template file?

akhmerov:
I was about to do so, but it says that it shouldn't be edited manually; I assumed it's regenerated automaticaly. Is manual edit OK?

akhmerov:
In case a manual edit is OK, it's done now.

MSeal:
Yeah should be fine I think

MSeal (approved):

PR #1292, opened 2020-06-15
DOC: Remove outputs from example notebooks

mgeier:
I think it sets a bad example when notebooks are committed with their outputs.

Also, when the outputs are committed, the notebooks are not executed (on RTD and during CI), therefore bugs are not noticed.

And that's the problem with this PR ... it exposes bugs in the notebooks, mostly related to changes in the templating system.

I don't know how to fix them, can somebody please help?

Feel free to ignore this PR and fix the problems in a separate PR.

PR #1429, opened 2020-09-30
fix: allow custom template files outside of the template system

### LINKED ISSUE #1394: TemplateNotFound error when extending built-in template
I have the following code:

```python
import nbformat
from nbconvert import HTMLExporter

html_exporter = HTMLExporter(template_file="foo.tpl")
with open("foo.ipynb") as infile:
    nb = nbformat.read(infile, as_version=4)
    html = html_exporter.from_notebook_node(nb)
    print(html)
```

and a template (which does nothing but still fails if it does do something) in the cwd, `foo.tpl`:

```jinja
{% extends 'lab/index.html.j2' %}
```

I get the error (with the venv path replaced with `<venv>` for brevity):

```python
  File "convert_test/__init__.py", line 7, in <module>
    html = html_exporter.from_notebook_node(nb)
  File "<venv>/lib64/python3.8/site-packages/nbconvert/exporters/html.py", line 122, in from_notebook_node
    return super().from_notebook_node(nb, resources, **kw)
  File "<venv>/lib64/python3.8/site-packages/nbconvert/exporters/templateexporter.py", line 382, in from_notebook_node
    output = self.template.render(nb=nb_copy, resources=resources)
  File "<venv>/lib64/python3.8/site-packages/jinja2/environment.py", line 1090, in render
    self.environment.handle_exception()
  File "<venv>/lib64/python3.8/site-packages/jinja2/environment.py", line 832, in handle_exception
    reraise(*rewrite_traceback_stack(source=source))
  File "<venv>/lib64/python3.8/site-packages/jinja2/_compat.py", line 28, in reraise
  File "/home/matt/temp/convert_test/foo.tpl", line 1, in top-level template code
    {% extends 'lab/index.html.j2'%}
  File "<venv>/share/jupyter/nbconvert/templates/lab/index.html.j2", line 1, in top-level template code
    {%- extends 'base.html.j2' -%}
jinja2.exceptions.TemplateNotFound: base.html.j2
```

It seems to be finding my `foo.tpl` file correctly, seeing the `extends 'lab/index.html.j2'` line and then correctly finding `<venv>/share/jupyter/nbconvert/templates/lab/index.html.j2`. This file, inside the install directory says it `extends 'base.html.j2'` but nbconvert fails to find *that* file.

If I do `ls <venv>/share/jupyter/nbconvert/templates/lab/` then I see:

```
static  base.html.j2  conf.json  index.html.j2
```

so `base.html.j2` is sitting there right next to `index.html.j2`.

This process works fine with nbconvert 5. I realise that templates have changed and I read https://nbconvert.readthedocs.io/en/latest/customizing.html but it doesn't seem to tell me how to create a custom template, only how the built-in templates are organised.

maartenbreddels:
As discussed over a videocall with @Mseal and @SylvainCorlay, this PR:
 * Fixes #1394 
 * is an alternative to a part of #1387 
 * and is a followup of #1292

In #1292, we kind of 'tricked' the template system, by replacing the template_name by the parent directory name of the template_file. That caused an override of the template name. If we don't do that, but simply add that directory to the template search path, the whole template system machinery works with us. This allows .e.g:
```
$ python -m nbconvert --to html --execute --template material --TemplateExporter.template_file=nbconvert/exporters/tests/files/lablike.html.j2 ~/src/QuantStack/voila/notebooks/basics.ipynb
```

So we use a particular template (material in this case, but could be labl) but override the template_file.

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -213,16 +213,12 @@ class TemplateExporter(Exporter):
         full_path = os.path.abspath(new)
         if os.path.isfile(full_path):
             directory, self.template_file = os.path.split(full_path)
+            self.extra_template_paths = [directory] + self.extra_template_paths
             # While not strictly an invalid template file name, the extension hints that there isn't a template directory involved
             if self.template_file.endswith('.tpl'):
                 warnings.warn(
                     f"5.x style template file passed '{new}'. Use --template-name for the template directory with a index.<ext>.j2 file and/or --template-file to denote a different template.",
                     DeprecationWarning)
-            if directory:
-                directory, self.template_name = os.path.split(directory)
-            if directory:
-                if os.path.isabs(directory):
-                    self.extra_template_basedirs = [directory]
 
     @default('template_file')
     def _template_file_default(self):
@@ -237,6 +233,7 @@ class TemplateExporter(Exporter):
 
     template_paths = List(['.']).tag(config=True, affects_environment=True)
     extra_template_basedirs = List().tag(config=True, affects_environment=True)
+    extra_template_paths = List([]).tag(config=True, affects_environment=True)
 
     @default('extra_template_basedirs')
     def _default_extra_template_basedirs(self):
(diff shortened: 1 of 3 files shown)


MSeal (approved):
Thanks! I tested some of the reported issues. We should probably also do a PR to update documentation for backwards compatibility. I'll get the basic template PR up later today to wrap up the other half of the implementation from the larger PR.

PR #1431, opened 2020-10-01
Restored basic template from 5.x

MSeal:
Extracted the basic template addition from https://github.com/jupyter/nbconvert/pull/1387 into an isolated PR. The full.tpl was mapped back to `classic` instead of `lab` because it was incorrectly pointing to the newer template in lieu of basic pointing to `classic` before this.

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -57,6 +57,10 @@ HTML
 
     Simplified HTML, using the classic jupyter look and feel.
 
+  - ``--template basic``
+
+    Base HTML, rendering with minimal structure and styles.
+
 .. _convert_latex:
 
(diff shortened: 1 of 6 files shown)


MSeal:
Merging to avoid conflicts with next change as it was already reviewed in linked PR

maartenbreddels:
LGTM

MSeal:
I am going to spin up a release with the template fixes we made then

ThiefMaster:
Unlike in 5.6.1, using `.template_name = 'basic'` (in 5.6.1: `.template_file = 'basic'`) no longer seems to include any CSS. Is this an intentional change from the old behavior?

PR #1432, opened 2020-10-01
Added proper support for backwards compatibility templates

MSeal:
Added last remaining feature from https://github.com/jupyter/nbconvert/pull/1387 to support the two templates from 5.x that didn't have an equivalent pattern in 6.x


MSeal:
I am going to ask for forgiveness here and proceed with a merge so we can get the overall fit deployed. If there's an issue here that was missed I'll happily help with an additional patch release after

Issue #1428, opened 2020-09-30
Feature: Update documentation to show how to configure custom template paths


txoof:
**Problem:** I am trying to build a system for maintaining nbconvert templates over several platforms (Mac OS, Ubuntu, Debian) that use various paths for storing templates
*Raspbian:*
```
['/home/pi/.local/share/jupyter/nbconvert/templates', '/usr/local/share/jupyter/nbconvert/templates', '/usr/share/jupyter/nbconvert/templates']
```
*Mac OS with Pyenv:*
```
['/Users/ac/Library/Jupyter/nbconvert/templates', '/Users/ac/.pyenv/versions/3.8.5/Python.framework/Versions/3.8/share/jupyter/nbconvert/templates', '/usr/local/share/jupyter/nbconvert/templates', '/usr/share/jupyter/nbconvert/templates']
```
I understand that I can drop my templates into one of the paths above, but that requires a different solution for each platform making updates more challenging. 

**Proposed Solution**
Update the documentation to clearly show how to specify an alternative or additional path to templates.

The [current documentation](https://nbconvert.readthedocs.io/en/latest/config_options.html) does not indicate how to do this.

There is a long discussion in Issue #1028 around this, but I could not determine how it was resolved.

```
TemplateExporter.template_paths : List
   Default: ['.']

   No description
```
```
PythonExporter.template_paths : List
   Default: ['.']

   No description
```

I have successfully added a specific path to a single template as follows:
```
c.PythonExporter.template_path: ['.', '/path/to/.jupyter/templates/my_python_template']
```
This results in the error:
```
jinja2.exceptions.TemplateNotFound: null.j2
```

I suspect this is because my added path clobbers the other paths and results in `null.j2` not being found at all. 

my_python_template/index.py.j2
```
{%- extends 'null.j2' -%}

## set to python3
{%- block header -%}
#!/usr/bin/env python3
# coding: utf-8
{% endblock header %}

## remove cell counts entirely
{% block in_prompt %}
{% if resources.global_content_filter.include_input_prompt -%}
{% endif %}
{% endblock in_prompt %}

## remove markdown cells entirely
{% block markdowncell %}
{% endblock markdowncell %}

{% block input %}
{{ cell.source | ipython2python }}
{% endblock input %}


## remove magic statement completely
{% block codecell %}
{{'' if "get_ipython" in super() else super() }}
{% endblock codecell%}
```

txoof:
Is #1394 related to this?

[...]

PR #1433, opened 2020-10-01
#1428 add docstring

txoof:
Add a docstring for `TemplateExporter.extra_template_basedirs` in response to issue #1428 

I'm unsure how to handle the `.tag()` method -- this probably requires attention from someone more knowledgeable.

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -303,6 +303,9 @@ class TemplateExporter(Exporter):
         help="""formats of raw cells to be included in this Exporter's output."""
     ).tag(config=True)
 
+    extra_template_basedirs = List(
+        help="Specify extra directories to search for templates").tag()
+
     @default('raw_mimetypes')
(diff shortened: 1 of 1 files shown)


nbconvert/exporters/templateexporter.py
```diff
@@ -303,6 +303,9 @@ def _template_extension_default(self):
         help="""formats of raw cells to be included in this Exporter's output."""
     ).tag(config=True)
 
+    extra_template_basedirs = List(
+        help="Specify extra directories to search for templates").tag()
```
MSeal, line 307: Instead of recreating the option, you want to add the help string to the `extra_template_basedirs` assignment that's already present. By restating it here you're replacing the other assignment which is causing the errors in the tests.

MSeal (changes requested):

MSeal:
Going to close for now since we added the other documentation about using this field. We open if you want to add the traitlet docstring in the future

Issue #1398, opened 2020-09-18
Exporters fail to pickle in nbconvert >= 6


kycutler:
With nbconvert version 5 I was able to invoke exporters via a process pool. After upgrading to nbconvert >= 6.0.0 I get the error `AttributeError: Can't pickle local object 'TemplateExporter._init_resources.<locals>.deprecated`.

Repro (python 3.7):

```python
import nbconvert, nbformat
from concurrent.futures import ProcessPoolExecutor

if __name__ == "__main"":
  executor = ProcessPoolExecutor()
  exporter = nbconvert.exporters.PythonExporter()
  nb = nbformat.read('example.ipynb', as_version=4)

  executor.submit(exporter.from_notebook_node, nb).result()
```
```
concurrent.futures.process._RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/concurrent/futures/process.py", line 205, in _sendback_result
    exception=exception))
  File "/usr/local/lib/python3.7/multiprocessing/queues.py", line 358, in put
    obj = _ForkingPickler.dumps(obj)
  File "/usr/local/lib/python3.7/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
AttributeError: Can't pickle local object 'TemplateExporter._init_resources.<locals>.deprecated'
"""

The above exception was the direct cause of the following exception:    

Traceback (most recent call last):
  File "test.py", line 8, in <module>
    executor.submit(exporter.from_notebook_node, nb).result()
  File "/usr/local/lib/python3.7/concurrent/futures/_base.py", line 435, in result
    return self.__get_result()
  File "/usr/local/lib/python3.7/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
AttributeError: Can't pickle local object 'TemplateExporter._init_resources.<locals>.deprecated'
```

**Nbconvert version: 6.0.0**

MSeal:
Hmm can you list the full chain of dependencies you have? I'm not sure what change would have caused pickling to be angry at the init_resources traitlet. I can dig into this regardless but having the jupyter library versions you are using will help.

kycutler:
@MSeal It was caused by a nested function defined in nbconvert -- I opened PR #1399 to fix it

PR #1435, opened 2020-10-02
Fix error when pickling HTMLExporter

kycutler:
Argh. Turns out the HTML exporter had a similar issue to the TemplateExporter (in #1398) which causes it to fail to pickle. This PR fixes the HTML exporter as well.

I checked the other exporters manually and none appeared to have nested functions, but I didn't write tests for all of them.

### DIFF
--- a/nbconvert/exporters/html.py
+++ b/nbconvert/exporters/html.py
@@ -118,47 +118,49 @@ class HTMLExporter(TemplateExporter):
         self.register_filter('highlight_code', highlight_code)
         return super().from_notebook_node(nb, resources, **kw)
 
+
+    def _resources_include_css(self, name):
+        env = self.environment
+        code = """<style type="text/css">\n%s</style>""" % (env.loader.get_source(env, name)[0])
+        return jinja2.Markup(code)
+        
+    def _resources_include_js(self, name):
+        env = self.environment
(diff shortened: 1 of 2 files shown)


MSeal:
Thanks for helping dig into those pickling issues more. It does look like it's still not hope with pickling those functions in the exporter though based on that stack trace

kycutler:
Yeah my bad, I missed the failed test in Python3.7 on my local. I fixed those errors (including in Python3.6) but now it looks like the `cannot pickle 'weakref' object` errors may be caused by something in Jinja... I'm wondering if there is a way to refactor these three resources to avoid this, but I'm afraid I'm out of my depth there.

For my specific use case, I only need the `output`, which is feasible with these current changes... :/

MSeal:
Try moving those resource functions outside of the class entirely and accept an `env` argument. Then wrap the resource assignments with a lambda / partial function that supplies the `self.environment` field as the first argument. I think that would avoid having the full self weak reference present but have the environment and name still be dynamic.

kycutler:
Hmm, that didn't quite work. It looks like there are references to the exporter within the environment -- for example `environment.filters["filter_data_type"].parent`, or any of the other filters that inherit from `LoggingConfigurable` and keep a reference to the parent.

Issue #1189, opened 2020-02-12
Using nbconvert --no-input does not indent code output properly


Juan-132:
When the notebook is converted using `--no-input` the code output is not indented properly.

I converted the attached notebook.
Run it using `jupyter nbconvert notebook.ipynb`
And using `jupyter nbconvert --no-input notebook.ipynb`.

[notebook.txt](https://github.com/jupyter/nbconvert/files/4193136/notebook.txt)

With a standard convert, the output is indented properly:
![output](https://user-images.githubusercontent.com/59505296/74349540-acfdfd00-4db4-11ea-85b7-f972fe9c6863.PNG)

But using --no-input, the output is not indented properly:
![output --no-input](https://user-images.githubusercontent.com/59505296/74349536-ac656680-4db4-11ea-9467-2a54d0a27144.PNG)

**Nbconvert version: 5.6.1

Juan-132:
This is the workaround I found elsewhere
`jupyter nbconvert --TemplateExporter.exclude_input=True --no-prompt notebook.ipynb`

MSeal:
Thanks for reporting I'll try to tag this for 6.0 release and see if we can get some focus on it before that goes out of alpha.

sgranade:
Adding `'exclude_input_prompt': True` to the `no-input` value in `nbconvertapp.py` will fix it as per the previously-mentioned workaround:

> `jupyter nbconvert --TemplateExporter.exclude_input=True --no-prompt notebook.ipynb`

t-makaro:
> Adding 'exclude_input_prompt': True to the no-input value in nbconvertapp.py will fix it

Funny. When ```--no-input``` was added, this was discussed and deemed redundant/an anti-pattern that we didn't want to support. See https://github.com/jupyter/nbconvert/pull/825#issuecomment-402013138. However, since this is an easy fix to the problem, I'd be open to merge it if someone wants to submit a pull request.

joyceerhl:
Hi @t-makaro, I submitted a pull request for this: https://github.com/jupyter/nbconvert/pull/1444 😊

MSeal:
The work-around has been merged, I'm going to close for now but if other rendering issues related come up with no-input feel free to create a new issue.

PR #1444, opened 2020-10-09
Fix code output indentation when running nbconvert --no-input

joyceerhl:
This PR adds the suggested fix for https://github.com/jupyter/nbconvert/issues/1189 in https://github.com/jupyter/nbconvert/issues/1189#issuecomment-674278616. I also verified that the original problem persists in nbconvert 6.0.3 with the provided repro before opening this PR.


MSeal:
Hello @joyceerhl -- Thanks for submitting the fix. While we could do more for input suppression rendering in the project overall this does address the root issue reported. Apologies for the late review on the PR.

Issue #1167, opened 2020-01-14
nbconvert failed: [Errno 13] Permission denied


eburgueno:
I've been trying to figure out the root cause of this for a while, but I can't see what's wrong with our config. When trying to export a notebook as PDF from JupyterLab I am getting a permission denied error:

```
[userA@host1 ~]$ jupyter nbconvert --to pdf Untitled.ipynb
(...)
PermissionError: [Errno 13] Permission denied: '/software/jupyterhub/share/jupyter/nbconvert/templates/latex/article.tplx'
```

`/software/jupyterhub` is where we deployed JupyterLab and `share/jupyter` is in the data path, but I don't understand why `~/.local/share/jupyter` isn't tried first:

```
[userA@host1 ~]$ jupyter --path
config:
    /home/userA/.jupyter
    /software/jupyterhub/etc/jupyter
    /usr/local/etc/jupyter
    /etc/jupyter
data:
    /home/userA/.local/share/jupyter
    /software/jupyterhub/share/jupyter
    /usr/local/share/jupyter
    /usr/share/jupyter
runtime:
    /home/userA/.local/share/jupyter/runtime

[userA@host1 software]$ jupyter --data-dir
/home/userA/.local/share/jupyter
```

If `/software/jupyterhub/share/jupyter` exists and is writeable to `userA`, the directories `nbconvert/templates/latex` get created and the conversion works (but only for `userA`):

```
[userA@host1 ~]$ ls -l /software/jupyterhub/share/jupyter/
total 152
drwxr-sr-x. 13 swuser  common 301 Jan 13 12:48 kernels
drwxr-sr-x.  8 swuser  common 181 Jan 10 14:25 lab
drwxr-sr-x.  4 swuser  common  62 Jan 13 09:47 nbextensions

[userA@host1 ~]$ touch /software/jupyterhub/share/jupyter/test_rw

[userA@host1 ~]$ jupyter nbconvert --to pdf Untitled.ipynb
[NbConvertApp] Converting notebook Untitled.ipynb to pdf
(...)
[NbConvertApp] PDF successfully created
[NbConvertApp] Writing 28002 bytes to Untitled.pdf

[userA@host1 ~]$ ls -l /software/jupyterhub/share/jupyter/
total 152
drwxr-sr-x. 13 swuser  common 301 Jan 13 12:48 kernels
drwxr-sr-x.  8 swuser  common 181 Jan 10 14:25 lab
[...]

PR #1450, opened 2020-10-14
fix issue (i'd call it a BUG) #1167

stela2502:
Hi, The issue is that if the users jupyter is running under did not install nbconvert this tools is breaking. That is extremely unsatisfying and I would almost call it immature. 
This is simply encapsulating the problematic area into a try: except: construct. Please fix this! Thank you!

I have tested this and it does work.

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -505,10 +505,13 @@ class TemplateExporter(Exporter):
     def _get_conf(self):
         conf = {}  # the configuration once all conf files are merged
         for path in map(Path, self.template_paths):
-            conf_path = path / 'conf.json'
-            if conf_path.exists():
-                with conf_path.open() as f:
-                    conf = recursive_update(conf, json.load(f))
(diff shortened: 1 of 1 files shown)


MSeal:
> Hi, The issue is that if the users jupyter is running under did not install nbconvert this tools is breaking. That is extremely unsatisfying and I would almost call it immature.

I am not following what this means, since to reach said code one must have installed nbconvert... since the code being changed is inside nbconvert. I'm going to close this PR for now, as I don't think the change in behavior as written in the PR is something we want added. If you'd like to open an issue that explains the problem you ran into in more depth that would be preferred.

Also please refrain from name calling as well. We have a [code of conduct](https://github.com/jupyter/governance/blob/master/conduct/code_of_conduct.md) if you're not aware, and calling code you're unfamiliar with immature could be interpreted as insulting to some maintainers or contributors. I'm assuming you meant no insult, but I'd like to point out that it's ambiguous to me here what the comment's intent was.

PR #1454, opened 2020-10-17
Include output of stdout stream in lab template

atsuoishimoto:
The lab template does not include stdin stream generated by `input()`. I propose to add new config `exclude_output_stdin` to exclude the stdin stream in lab template.  The default value of `exclude_output_stdin` should be `True` for compatibility.

The CSS in this PR renders stdin stream as follows.

image


MSeal (approved):
Sorry this was missed when it was opened. The change looks good to me and I'll get this into the alpha release going out.

atsuoishimoto:
Thank you!

PR #1463, opened 2020-11-10
Fix dangling reference to get_template_paths()

iromero91:
Use the existing property interface instead. Must have been missed when it got refactored.


iromero91:
Just pinging this PR as i have seen some activity in the repo lately. It is a pretty straightforward change i think,

yuvipanda:
Thanks a lot, @iromero91!

Issue #1439, opened 2020-10-07
Performance regression from 5.6.1 to 6.0.7


rmoe:
Loading notebooks and running `jupyter nbconvert` commands became much slower after we upgraded from 5.6.1 to 6.x. Our notebook load times went from 400-500ms to 30s or more. Our user notebooks run on kubernetes and use s3fs for the home directory.

```
jovyan@jupyter-5f1f30d0679987001a6cfef6:~$ time jupyter nbconvert --help
...
To see all available configurables, use `--help-all`.

real    0m31.991s
user    0m1.925s
sys     0m0.361s
```

**Nbconvert version:** 6.0.7

This was introduced in #1273. Previously `get_export_names` just returned the list of exporters in the `nbconvert.exporters` group. Now it loads and instantiates each one to check its `enabled` attribute. This causes a lot of stat calls and listing of directories which is very slow on s3fs.

mriedem:
>Performance regression from 5.6.1 to 6.0.7

A nit but to be clear I think the regression was introduced in 6.0.

mriedem:
>This causes a lot of stat calls and listing of directories which is very slow on s3fs.

Is it the file system work or the python module importing with the extensions? I thought it was the latter. The stat calls have always been around because the notebook server does a lot of lstat calls to check file existence.

rmoe:
> Is it the file system work or the python module importing with the extensions? I thought it was the latter. The stat calls have always been around because the notebook server does a lot of lstat calls to check file existence.

Instantiating the extensions to check their `enabled` attribute is what causes the increased file system activity (importing python modules, searching for configs, etc.). Running the same test on another filesystem is around 15 times faster.

MSeal:
Ahh, yes `s3fs` is pretty slow and the `get_export_names` is fetching config files to confirm their enabled status.

I see that `get_export_names` is now called to build the help string for nbconvertapp... we could change that function to optionally check for enabled / disabled and not load all exporters by default on app load for help strings. Would one of you be willing to try this change out and see if performance improves back to pre 6.0 speeds for you? (I don't have a quick s3fs setup to check atm)

rmoe:
> Ahh, yes `s3fs` is pretty slow and the `get_export_names` is fetching config files to confirm their enabled status.
> 
[...]

PR #1471, opened 2020-11-20
Allow get_export_names to skip configuration check

rmoe:
This introduces an environment variable named
NBCONVERT_DISABLE_CONFIG_EXPORTERS that will cause get_export_names to
return all entrypoints instead of checking the "enabled" status of
each one.

Closes: #1439

### DIFF
--- a/nbconvert/exporters/base.py
+++ b/nbconvert/exporters/base.py
@@ -3,6 +3,7 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
+import os
 import warnings
 
 import entrypoints
@@ -135,6 +136,10 @@ def get_export_names(config=get_config()):
     them as an nbconvert.exporter entrypoint.
     """
     exporters = sorted(entrypoints.get_group_named('nbconvert.exporters'))
+    if os.environ.get("NBCONVERT_DISABLE_CONFIG_EXPORTERS"):
+        get_logger().info("Config exporter loading disabled, no additional exporters will be automatically included.")
+        return exporters
+
     enabled_exporters = []
     for exporter_name in exporters:
         try:
--- a/nbconvert/exporters/tests/test_exporter.py
+++ b/nbconvert/exporters/tests/test_exporter.py
@@ -97,7 +97,7 @@ class TestExporter(ExportersTestsBase):
             # enabled in the config
             self.assertNotEqual(export_names, ['notebook'])
 
-        # In the abscence of this variable we should revert to
+        # In the absence of this variable we should revert to
         # the normal behavior.
         del os.environ["NBCONVERT_DISABLE_CONFIG_EXPORTERS"]
         export_names = get_export_names(config=config)
--- a/nbconvert/exporters/tests/test_exporter.py
+++ b/nbconvert/exporters/tests/test_exporter.py
@@ -13,8 +13,10 @@ Module with tests for exporter.py
 #-----------------------------------------------------------------------------
 # Imports
 #-----------------------------------------------------------------------------
-
+import os
 from traitlets.config import Config
+
+from unittest.mock import patch
 
 from .base import ExportersTestsBase
 from ...preprocessors.base import Preprocessor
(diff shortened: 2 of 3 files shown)


nbconvert/exporters/base.py
```diff
@@ -135,6 +136,9 @@ def get_export_names(config=get_config()):
     them as an nbconvert.exporter entrypoint.
     """
     exporters = sorted(entrypoints.get_group_named('nbconvert.exporters'))
+    if os.environ.get("NBCONVERT_DISABLE_CONFIG_EXPORTERS"):
```
mriedem, line 139: nit: might be good to have a comment here so people know what this is for, i.e. bypass the (potentially slow) enable/disable check per extension if you know that you're supporting all exporters.

I also wonder if there is somewhere in the docs or config options that we could mention this environment variable otherwise people would only ever know about it if they were familiar with this code. Like maybe somewhere in here:

https://github.com/jupyter/nbconvert/blob/master/docs/source/api/exporters.rst

Though that's just auto-generated module doc.

Looking at https://nbconvert.readthedocs.io/en/latest/config_options.html#exporter-options maybe `Exporter.enabled` should mention this env var? Or https://nbconvert.readthedocs.io/en/latest/external_exporters.html?

A CODEOWNER might have better ideas, but I'd think mentioning it with the `Exporter.enabled` config option would be the most _discoverable_ place for mentioning this variable.

rmoe:
@MSeal is there anything I need to do to help move this forward?

MSeal:
@rmoe Sorry I originally missed this PR (was on leave). Let me leave a comment in the implementation, but I think a quick log message and adding a note to the documentation is all that's needed.

nbconvert/exporters/base.py
MSeal: To @mriedem 's point, https://nbconvert.readthedocs.io/en/latest/config_options.html#exporter-options would be the best place to document the environment variable action. You can add this note by editing the `CLI Flags and Aliases` section in `autogen_config.py`.

Additionally we should probably log out a quick message when we're disabling config exporters so it's visible that the flag was set when helping future nbconvert issues. Use `from traitlets.log import get_logger` and `get_logger().info("Config exporter loading disabled, no additional exporters will be automatically included.")`

rmoe:
> @rmoe Sorry I originally missed this PR (was on leave). Let me leave a comment in the implementation, but I think a quick log message and adding a note to the documentation is all that's needed.

No worries, thank you for taking a look at this.

meeseeksmachine:

https://discourse.jupyter.org/t/current-state-of-nbconvert-drop-support-for-python-3-6/8882/1

mriedem:
@rmoe CI should be fixed on `master` now if you want to rebase.

nbconvert/exporters/tests/test_exporter.py
```diff
@@ -73,3 +75,30 @@ def test_get_export_names_disable(self):
...
+            export_names = get_export_names(config=config)
+            # get_export_names should not call get_exporter for
+            # any of the entry points because we return before then.
+            exp.assert_not_called()
+
+            # We should have all exporters, not just the ones
+            # enabled in the config
+            self.assertNotEqual(export_names, ['notebook'])
+
+        # In the abscence of this variable we should revert to
```
mriedem, line 100:
```suggestion
        # In the absence of this variable we should revert to
```

nbconvert/exporters/tests/test_exporter.py
```diff
@@ -73,3 +75,30 @@ def test_get_export_names_disable(self):
...
+        """
+        Does get_export_names behave correctly with respect to
+        NBCONVERT_DISABLE_CONFIG_EXPORTERS being set in the
+        environment?
+        """
+        config = Config({
+            'Exporter': {'enabled': False},
+            'NotebookExporter': {'enabled': True}
+        })
+        os.environ["NBCONVERT_DISABLE_CONFIG_EXPORTERS"] = "1"
```
mriedem, line 89: Fun fact, you can use `patch.dict` to temporarily set values in `os.environ`, there is an example in the docs:

https://docs.python.org/3.7/library/unittest.mock.html#unittest.mock.patch.dict

mriedem (approved):
Code looks good. A couple of small nits inline but the test looks good and there is a mention of this in the docs now (so comments have been resolved) and CI is passing. 👍

mriedem (approved):

MSeal (approved):
Thanks for getting this updated! LGTM

MSeal:
doc build issue is unrelated, I'll look into that before the next release.

PR #1474, opened 2020-11-26
JupyterLab export as HTML with widgets fails to load widgets

### LINKED ISSUE #1473: JupyterLab export as HTML with widgets fails to load widgets
Those resources should be defined in HTMLExporter:

https://github.com/jupyter/nbconvert/blob/f66ec517275f7c61bb5a36a5d3496ba73005d659/nbconvert/nbconvertapp.py#L267-L270

This will result in a similar behavior than for requirejs and jquery:

https://github.com/jupyter/nbconvert/blob/681729a1827fbb14c0642021f9e75b2632dfbcd2/nbconvert/exporters/html.py#L162-L163

Moreover this will ensure they are defined when exporting from jupyter server that is not using nbconvertapp but directly the exporter:

https://github.com/jupyter-server/jupyter_server/blob/b2e089506baf0723fe63112cb9df7ef8b493d093/jupyter_server/nbconvert/handlers.py#L122

fcollonval:
Fixes #1473

### DIFF
--- a/nbconvert/exporters/html.py
+++ b/nbconvert/exporters/html.py
@@ -54,6 +54,16 @@ class HTMLExporter(TemplateExporter):
         """
     ).tag(config=True)
 
+    jupyter_widgets_base_url = Unicode(
+        "https://unpkg.com/",
+        help="URL base for Jupyter widgets"
+    ).tag(config=True)
+
+    html_manager_semver_range = Unicode(
+        "*",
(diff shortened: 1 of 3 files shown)


MSeal (approved):

MSeal:
Sorry no one looked at this sooner. Merging now to have in the alpha release

PR #1489, opened 2020-12-14
Webpdf: Use a temporary file instead of an URL 

### LINKED ISSUE #1468: Export Notebook to Webpdf 1MB limit
So I'm really enjoying the Webpdf feature, however I'm noticing that if I have a notebook with many subplots, and the file size of the resulting pdf creeps up past 1MB, I am met with
```
nbconvert failed: Navigation Timeout Exceeded: 30000 ms exceeded.
```
 At first I thought there simply wasn't enough time for the pdf build to complete, so I changed the parameter `self._defaultNavigationTimeout` in  pyppeteer/pages.py to allow thrice this time, but soon realised that no matter the time, I am still met with this error for files above 1MB.

Is there something I can change to override this, be it in Pyppeteer or nbconvert?
Cheers.

iromero91:
Note: this PR includes #1463, please merge that one first.

Fixes #1468

### DIFF
--- a/nbconvert/exporters/webpdf.py
+++ b/nbconvert/exporters/webpdf.py
@@ -5,6 +5,8 @@
 
 import asyncio
 
+import tempfile, os
+
 from traitlets import Bool
 import concurrent.futures
 
@@ -65,6 +67,11 @@ class WebPDFExporter(HTMLExporter):
     def run_pyppeteer(self, html):
         """Run pyppeteer."""
 
+        temp_file = tempfile.NamedTemporaryFile(suffix=".html", delete=False)
+        temp_filename = temp_file.name
+        with temp_file:
+            temp_file.write(html.encode('utf-8'))
+
         async def main():
             args = ['--no-sandbox'] if self.disable_sandbox else []
             browser = await self._check_launch_reqs()(
@@ -75,7 +82,7 @@ class WebPDFExporter(HTMLExporter):
             )
             page = await browser.newPage()
             await page.waitFor(100)
-            await page.goto('data:text/html,'+html, waitUntil='networkidle0')
+            await page.goto(f'file://{temp_file.name}', waitUntil='networkidle0')
             await page.waitFor(100)
 
             pdf_params = {}
@@ -112,6 +119,7 @@ class WebPDFExporter(HTMLExporter):
             asyncio.set_event_loop(loop)
             return loop.run_until_complete(coro)
         pdf_data = pool.submit(run_coroutine, main()).result()
+        os.unlink(temp_filename)
         return pdf_data
 
     def from_notebook_node(self, nb, resources=None, **kw):
--- a/nbconvert/exporters/webpdf.py
+++ b/nbconvert/exporters/webpdf.py
@@ -67,12 +67,7 @@ class WebPDFExporter(HTMLExporter):
     def run_pyppeteer(self, html):
         """Run pyppeteer."""
 
-        temp_file = tempfile.NamedTemporaryFile(suffix=".html", delete=False)
-        temp_filename = temp_file.name
-        with temp_file:
-            temp_file.write(html.encode('utf-8'))
-
-        async def main():
+        async def main(temp_file):
             args = ['--no-sandbox'] if self.disable_sandbox else []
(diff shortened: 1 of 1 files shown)


nbconvert/exporters/webpdf.py
```diff
@@ -38,6 +40,11 @@ def _check_launch_reqs(self):
     def run_pyppeteer(self, html):
         """Run pyppeteer."""
 
+        temp_file = tempfile.NamedTemporaryFile(suffix=".html", delete=False)
```
yuvipanda, line 43: We should probably rely on the contextmanager somehow to cleanup the temporary file. I *think* right now, it's possible for a crash to leave this HTML file around

yuvipanda (changes requested):
Can you also add a test for files over 1MB? that'd be awesome.

Thank you so much for working on this :)

nbconvert/exporters/webpdf.py
iromero91: Yeah that is a possibility, but temp files are stored in temporary storage, so dangling files should not be an issue long term. I did it this way because it is important to close the file to ensure the whole data is transfered to the browser process without any buffering issues.

yuvipanda: True! But you can call `temp_file.fluash()` to ensure that.

iromero91: Not really. According to the documentation if the file is not closed, there is no guarantee that the file can be opened and read by another process in some operating systems (NT derived I think)

iromero91:
@yuvipanda I added better guarding on the file deletion. This should guard about as good as the context manager would. It wont guard against process crashes, but neither would the CM either.

How would I go about implementing that 1MB test? the behavior before this fix is that the generation just times out.

nbconvert/exporters/webpdf.py
```diff
@@ -105,13 +107,19 @@ async def main():
...
-        # TODO: when dropping Python 3.6, use
-        # pdf_data = pool.submit(asyncio.run, main()).result()
-        def run_coroutine(coro):
-            loop = asyncio.new_event_loop()
-            asyncio.set_event_loop(loop)
-            return loop.run_until_complete(coro)
-        pdf_data = pool.submit(run_coroutine, main()).result()
+        temp_file = tempfile.NamedTemporaryFile(suffix=".html", delete=False)
+        with temp_file:
+            temp_file.write(html.encode('utf-8'))
```
yuvipanda, line 112: I think just a tempfile.flush() here will do the right thing, and we can remove the explicit os.unlink call. IMO these two are almost equivalent, but putting everything in the `with` block is clearer to understand.

iromero91: As I said in the other conversation. According to the documentation, Chrome cannot open the file inside the `with` block on NT. Windows will not allow other process to access an open temp file, it has to be closed first.

yuvipanda: Ooh, oops. I totally missed that - am really sorry! GitHub had collapsed that comment as 'stale' so I did not see it!!!

Since it's non-obvious, can you just write a comment explaining why this is the case? I think I'm happy to merge after that - don't think we need a specific test case for this

iromero91: Added comments

iromero91:
@yuvipanda I believe it should be ready to merge now.

yuvipanda:
Thanks, @iromero91! I think the test failures are unrelated, so going to merge this now.

denzchoe:
> Thanks, @iromero91! I think the test failures are unrelated, so going to merge this now.

Any idea when 6.1 is going to release?
I would like the fix for [#1468](https://github.com/jupyter/nbconvert/issues/1468)

Or is there a temporary solution before 6.1 is released?

MSeal:
@denzchoe there were some reported issues on Windows tests I haven't had time to go verify / fix so it's still locked in the rc for pip until those get tackled.

As a temporary solution you should be able to install the pip version of the rc to your conda env:

```yaml
dependencies:
  - python=3.6
  - pip>=9.0
  ...
  - pip:
    - nbconvert==6.1.0rc0
```

Issue #1497, opened 2020-12-25
'traitlets>=4.2' is not enough for the use of  --TemplateExporter.extra_template_basedirs


BoPeng:
I have a linux system with the latest version of `nbconvert` installed. However, when I tried to use

```sh
jupyter nbconvert  my.ipynb --to html --TemplateExporter.extra_template_basedirs /path/to/template --template template_name
```

I got the error

```
[NbConvertApp] CRITICAL | Bad config encountered during initialization:
[NbConvertApp] CRITICAL | Invalid argument: '--TemplateExporter.extra_template_basedirs'
```

When I tried to use
```sh
jupyter nbconvert my.ipynb --to html --TemplateExporter.extra_template_basedirs=/path/to/template --template template_name
```

I got 

```
traitlets.traitlets.TraitError: The 'extra_template_basedirs' trait of a HTMLExporter instance must be a list, but a value of class 'str' (i.e. '/path/to/templates') was specified.
```

The errors were quite puzzling to me because the same commands work well on other systems.

After a frustrating hour of debugging, it turns out I had `traitlets 4.3.3` and upgrading it to `5.0.5` solved the problem.

 The recommended [`traitlets>=4.2` in `setup.py`](https://github.com/jupyter/nbconvert/blob/2e41429def702dd561586303356fdf1f9be7e95c/setup.py#L219)  needs to be updated  to the minimal version of `traitlets` for options such as `--TemplateExporter.extra_template_basedirs` to work properly.

t-makaro:
Should be fixed by #1498

PR #1498, opened 2020-12-25
Update version dependency of traitlets

BoPeng:
#1497 

Just tested, version 5.0.0 is the minimal version for `--TemplateExporter.extra_template_basedirs` to work properly.


eracle:
Hello everybody, 
this fix #1497

MSeal:
Should be fine to move that pin forward. Thanks for reporting and bumping

PR #1469, opened 2020-11-19
Solved svg2pdf conversion error if Inkscape is installed into the default path on a windows machine

MichaelAdolph:
The " " in "Program Files" in the default path for Inkscape on a windows machine makes the conversion fail. I updated the file to enclose the command string into double quotes. This allows the system to run properly on windows.

I'm not able to test on a Mac or Linux/Unix machine, so I'm unsure if my solution will create problems for users on different operating systems.

PR #1512, opened 2021-02-03
Don't use a shell to call inkscape

yuvipanda:
Python generally recommends against using `shell=True`
when calling
subprocesses (https://docs.python.org/3/library/subprocess.html#security-considerations).
This also causes issues with shell metacharacters (see
https://github.com/jupyter/nbconvert/pull/1469). I'm also
not entirely sure that the shell command *is* fully trustable
- I don't know where {to_filename} and {from_filename} are
from. Rest of nbconvert also prefers using lists instead
of strings to call commands.

svg2pdf also uses a command instead of a string now. We leave
the old string implementation alone for backwards compatibility,
although I'd really prefer to remove it.

We don't need the quotes set up in
https://github.com/jupyter/nbconvert/pull/1469,
since using a list automatically deals with that.

### DIFF
--- a/nbconvert/preprocessors/svg2pdf.py
+++ b/nbconvert/preprocessors/svg2pdf.py
@@ -62,7 +62,7 @@ class SVG2PDFPreprocessor(ConvertFiguresPreprocessor):
             raise RuntimeError("Unable to find inkscape executable --version")
         return output.decode('utf-8').split(' ')[1]
 
-    # FIXME: Deprecate passing a list here
+    # FIXME: Deprecate passing a string here
     command = Union(
         [Unicode(), List()],
         help="""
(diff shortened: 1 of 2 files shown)


nbconvert/preprocessors/svg2pdf.py
```diff
@@ -62,26 +62,37 @@ def _inkscape_version_default(self):
             raise RuntimeError("Unable to find inkscape executable --version")
         return output.decode('utf-8').split(' ')[1]
 
-    command = Unicode(
-        help="""The command to use for converting SVG to PDF
+    # FIXME: Deprecate passing a list here
```
agoose77, line 65: Shouldn't this be 
```suggestion
    # FIXME: Deprecate passing a string here
```

yuvipanda:
@agoose77 @technic fixed!

MSeal (approved):
Tested changed merged with main. Let's get this in and try to double check the alpha is working for folks on Windows more rigerously after that

Issue #1464, opened 2020-11-11
[RFC]: Print oriented webpdf output


iromero91:
I've been working on a webpdf exported designed for printing the resulting pdfs. Would this be something the project would be interested in having integrated with the webpdf exporter or should it remain a separate exporter?

To clarify the following features are already implemented in my code:

* Pagination (using css to set the page size and margins). Your webpdf exporter intentionally attempts to make a single page PDF out of the notebook (see PR #1402 ). If we integrate i would split that behavior off into an option.
* Templated HTML page headers and footers. Using chrome/puppeteer facilities to inject HTML at the header and footer of each printed page we can have information like page numbers change dynamically.

yuvipanda:
I would *really* like the option to turn pagination on / off. I think it should be in the default webpdf exporter, since that'll turn it into a full blown alternative to PDF via LaTeX

yuvipanda:
Also if you have a link to your code, I'd appreciate that too :)

yuvipanda:
Unfortunately, I had to get pagination working pretty soon for a class. So I opened https://github.com/jupyter/nbconvert/pull/1513. Very sorry about not waiting.

iromero91:
@yuvipanda I guess that pretty much settles this. I will look into submitting a PR for headers and footers once the tempfile PR gets in.

PR #1513, opened 2021-02-03
Make pagination configurable in webpdf

yuvipanda:
- Add option to paginate PDF generated via webpdf
- Default to pagination, to be a drop-in for the LaTeX
  PDF generator as much as possible.

Ref #1464

### DIFF
--- a/nbconvert/exporters/webpdf.py
+++ b/nbconvert/exporters/webpdf.py
@@ -24,7 +24,7 @@ class WebPDFExporter(HTMLExporter):
     ).tag(config=True)
 
     paginate = Bool(
-        False,
+        True,
         help="""
         Split generated notebook into multiple pages.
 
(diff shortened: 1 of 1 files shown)


yuvipanda:
@SylvainCorlay do you remember what the driving use case for turning off pagination in https://github.com/jupyter/nbconvert/pull/1402 was?

SylvainCorlay:
Hey @yuvipanda thanks for this!

When we did #1402, there was no release including webpdf yet, and I was having issues with pagination, so I probably decided to remove it together before the release.

SylvainCorlay:
It does not seem you are actually defaulting to pagination (in contrary to the PR description). Although I am not sure we should default to pagination (for backward compat). What do you think?

yuvipanda:
@SylvainCorlay ah, thanks for that catch. Made it the default now.

It looks like pagination was the default until 6.0.4, when single page became the default. IMO, it's ok to revert it in another patch release. webpdf is also such a new convertor, I think it's ok.

SylvainCorlay:
LGTM.

The issue in the tests looks like a problem with quotes in a string.

yuvipanda:
@SylvainCorlay yeah, that's unrelated - I've #1512 to try and fix that.

PR #1514, opened 2021-02-03
Set mimetype for webpdf correctly

yuvipanda:
- Browsers now detect the file as PDF and offer to
  open it in correct applications. Earlier, it was
  treated as a html file
- Fix the extension shown after the exporter in
  'Download as' menu. It was also .html before.

### DIFF
--- a/nbconvert/exporters/webpdf.py
+++ b/nbconvert/exporters/webpdf.py
@@ -5,14 +5,10 @@
 
 import asyncio
 
-<<<<<<< HEAD
-from traitlets import Bool, default
-from jupyter_core.paths import jupyter_path
-=======
 import tempfile, os
 
-from traitlets import Bool
->>>>>>> fa3ba7440cae2f2a7d3b5c0b4051e683b5576e82
+from traitlets import Bool, default
+from jupyter_core.paths import jupyter_path
 import concurrent.futures
 
 from .html import HTMLExporter
--- a/nbconvert/exporters/webpdf.py
+++ b/nbconvert/exporters/webpdf.py
@@ -5,7 +5,7 @@
 
 import asyncio
 
-from traitlets import Bool
+from traitlets import Bool, default
 import concurrent.futures
 
 from .html import HTMLExporter
@@ -33,6 +33,12 @@ class WebPDFExporter(HTMLExporter):
         Set to True to match behavior of LaTeX based PDF generator
         """
     ).tag(config=True)
+
+    output_mimetype = "application/pdf"
+
+    @default('file_extension')
+    def _file_extension_default(self):
+        return '.pdf'
 
     def _check_launch_reqs(self):
         try:
--- a/nbconvert/exporters/webpdf.py
+++ b/nbconvert/exporters/webpdf.py
@@ -6,6 +6,7 @@
 import asyncio
 
 from traitlets import Bool, default
+from jupyter_core.paths import jupyter_path
 import concurrent.futures
 
 from .html import HTMLExporter
@@ -39,6 +40,14 @@ class WebPDFExporter(HTMLExporter):
     @default('file_extension')
     def _file_extension_default(self):
         return '.pdf'
+
+    @default('template_name')
+    def _template_name_default(self):
+        return 'webpdf'
+
+    @default('template_data_paths')
+    def _template_data_paths_default(self):
+        return jupyter_path("nbconvert", "templates", "webpdf")
 
     def _check_launch_reqs(self):
         try:
--- a/nbconvert/preprocessors/tests/test_svg2pdf.py
+++ b/nbconvert/preprocessors/tests/test_svg2pdf.py
@@ -87,8 +87,8 @@ class Testsvg2pdf(PreprocessorTestsBase):
 
     def test_inkscape_pre_v1_command(self):
         preprocessor = self.build_preprocessor(inkscape='fake-inkscape', inkscape_version='0.92.3')
-        assert preprocessor.command == 'fake-inkscape --without-gui --export-pdf="{to_filename}" "{from_filename}"'
+        assert preprocessor.command == '"fake-inkscape" --without-gui --export-pdf="{to_filename}" "{from_filename}"'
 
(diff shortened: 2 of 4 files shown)


yuvipanda:
After a bunch of digging, this fails with:

```
[E 17:14:24.436 NotebookApp] Uncaught exception GET /notebooks/Untitled.ipynb (127.0.0.1)
    HTTPServerRequest(protocol='http', host='localhost:8888', method='GET', uri='/notebooks/Untitled.ipynb', version='HTTP/1.1', remote_ip='127.0.0.1')
      File "/Users/yuvipanda/.local/share/virtualenvs/nbconvert/lib/python3.9/site-packages/traitlets/traitlets.py", line 535, in get
        value = obj._trait_values[self.name]
    KeyError: 'template_paths'
    
    During handling of the above exception, another exception occurred:
    
      File "/Users/yuvipanda/.local/share/virtualenvs/nbconvert/lib/python3.9/site-packages/tornado/web.py", line 1704, in _execute
        result = await result
      File "/Users/yuvipanda/.local/share/virtualenvs/nbconvert/lib/python3.9/site-packages/tornado/gen.py", line 775, in run
        yielded = self.gen.send(value)
      File "/Users/yuvipanda/code/notebook/notebook/notebook/handlers.py", line 95, in get
        self.write(self.render_template('notebook.html',
      File "/Users/yuvipanda/code/notebook/notebook/base/handlers.py", line 516, in render_template
        return template.render(**ns)
      File "/Users/yuvipanda/.local/share/virtualenvs/nbconvert/lib/python3.9/site-packages/jinja2/environment.py", line 1090, in render
        self.environment.handle_exception()
      File "/Users/yuvipanda/.local/share/virtualenvs/nbconvert/lib/python3.9/site-packages/jinja2/environment.py", line 832, in handle_exception
        reraise(*rewrite_traceback_stack(source=source))
      File "/Users/yuvipanda/.local/share/virtualenvs/nbconvert/lib/python3.9/site-packages/jinja2/_compat.py", line 28, in reraise
      File "/Users/yuvipanda/code/notebook/notebook/templates/notebook.html", line 1, in top-level template code
        {% extends "page.html" %}
      File "/Users/yuvipanda/code/notebook/notebook/templates/page.html", line 154, in top-level template code
        {% block header %}
      File "/Users/yuvipanda/code/notebook/notebook/templates/notebook.html", line 115, in block "header"
        {% for exporter in get_frontend_exporters() %}
      File "/Users/yuvipanda/code/notebook/notebook/notebook/handlers.py", line 40, in get_frontend_exporters
        for name in get_export_names():
      File "/Users/yuvipanda/code/nbconvert/nbconvert/exporters/base.py", line 141, in get_export_names
        e = get_exporter(exporter_name)(config=config)
      File "/Users/yuvipanda/code/nbconvert/nbconvert/exporters/base.py", line 102, in get_exporter
        if getattr(exporter(config=config), 'enabled', True):
      File "/Users/yuvipanda/code/nbconvert/nbconvert/exporters/templateexporter.py", line 325, in __init__
        super().__init__(config=config, **kw)
      File "/Users/yuvipanda/code/nbconvert/nbconvert/exporters/exporter.py", line 114, in __init__
        self._init_preprocessors()
      File "/Users/yuvipanda/code/nbconvert/nbconvert/exporters/templateexporter.py", line 491, in _init_preprocessors
        conf = self._get_conf()
      File "/Users/yuvipanda/code/nbconvert/nbconvert/exporters/templateexporter.py", line 507, in _get_conf
        for path in map(Path, self.template_paths):
      File "/Users/yuvipanda/.local/share/virtualenvs/nbconvert/lib/python3.9/site-packages/traitlets/traitlets.py", line 575, in __get__
        return self.get(obj, cls)
      File "/Users/yuvipanda/.local/share/virtualenvs/nbconvert/lib/python3.9/site-packages/traitlets/traitlets.py", line 538, in get
        default = obj.trait_defaults(self.name)
      File "/Users/yuvipanda/.local/share/virtualenvs/nbconvert/lib/python3.9/site-packages/traitlets/traitlets.py", line 1578, in trait_defaults
        return self._get_trait_default_generator(names[0])(self)
      File "/Users/yuvipanda/.local/share/virtualenvs/nbconvert/lib/python3.9/site-packages/traitlets/traitlets.py", line 975, in __call__
        return self.func(*args, **kwargs)
      File "/Users/yuvipanda/code/nbconvert/nbconvert/exporters/templateexporter.py", line 518, in _template_paths
        template_names = self.get_template_names()
      File "/Users/yuvipanda/code/nbconvert/nbconvert/exporters/templateexporter.py", line 608, in get_template_names
        raise ValueError('Unsupported mimetype %r for template %r, mimetypes supported are: \n\t%s' %\
    ValueError: Unsupported mimetype 'application/pdf' for template 'lab', mimetypes supported are: 
        text/html
```

kycutler:
It seems like `WebPDFExporter` shouldn't actually inherit from `HTMLExporter`, since the clash between the two formats and their resulting mimetypes / extensions appears to be causing the error. It makes sense logically as well -- this is exporting a PDF, not HTML, so it doesn't make sense for it to "be" an `HTMLExporter` just because it uses HTML internally (and I suppose the same goes for the other `PDFExporter` and LaTeX too).

Can we refactor to just inherit from `Exporter`, and use an `HTMLExporter` internally?

MSeal:
> Can we refactor to just inherit from Exporter, and use an HTMLExporter internally?

I agree with that.

My time to look into these items to help them along is limited, but I believe can take a stab at helping this PR this weekend if no one else gets to fixing the build / implementing suggestions.

t-makaro:
Just a note: If you inherit from ```Exporter``` instead of ```TemplateExporter```, then the ```--template``` cli flag will not work.

MSeal:
So it turns out there's a lot of complication with not inheriting from the HTML templates. Those were used because the template first generates an HTML page and then sends it to pyppeteer. Instead what was needed was a unique config for the webpdf template to allow it to set the inherited HTML template and change the mimetype in it's nested config.

MSeal:
Docs build is unrelated ... looking into that separately

PR #1515, opened 2021-02-03
Change webpdf display name

yuvipanda:
The name has been confusing folks a fair bit. LaTeX
is at least better known.


yuvipanda:
This shouldn't affect any commandline or API users.

yuvipanda:
An alternative is 'PDF via Chrome' - more accurate, and better brand recognition.

MSeal (approved):
Thanks for the user feedback. I agree that this name is clearer

MSeal:
Probably means we should do a patch release in the near future -- I have not kept up with the repo over paternity leave though so I have little knowledge of current state of things

betatim:
What is the best way to deal with the fact that https://github.com/betatim/notebook-as-pdf uses the same text as display name? Will users get two entries in their menu and how can they tell the difference?

Issue #1434, opened 2020-10-01
Webpdf export doesn't seem to work in K8s/JupyterHub/JupyterLab setting


athornton:
I can't find a way to make Export As Webpdf work in my JH/JL/K8s setting.

Specifically: if I haven't installed `nbconvert[webpdf]` I still have the JL menu item, but it tells me to install `nbconvert[webpdf]`.  OK, fair, although I would prefer that I not have a menu item that doesn't work.

So I install it.  Then I get that I have to use `allow-chromium-download`.  Well, that's a problem, since I don't see how to add that flag to whatever the menu item invokes.  But, OK, first I tried installing chromium-headless into my container.  That didn't help but the container is Centos7-based so I'm totally willing to believe I need a newer Chromium.  So, fine, I run pyppeteer-install, and it downloads chromium into my .local.  But *then* when I try "export as webpdf", I get a pause for a while and then a new page that says `nbconvert failed: Browser closed unexpectedly:` .  My suspicion, although I have no evidence for this, is that the difference in the URL to the rendered PDF between what the user's browser sees and what the Jupyter environment sees, is what's breaking this.  Presumably it's been tested to death on local JL installations, and this wouldn't be the first time that I've been bitten by the assumption that the browser and the Jupyter server see the same URL.

I don't actually need this to work; if I can find out where the menu item is being inserted into the export menu, and just *not add it* I'm perfectly happy with my existing Export as PDF functionality.  But it bothers me that I have a menu item that doesn't work, and no way that I can find to not show it to my users.  So...where is that menu item injected into JupyterLab, and how do I make it not do that?

Pick a notebook.  "Export as webpdf".

**6.0.6**

ghuname:
Try to find out where pyppeteer is installed and than edit louncher.py.
In my case louncher.py is located at /opt/conda/lib/python3.8/site-packages/pyppeteer/launcher.py

In louncher.py at the very beginning you will see:

DEFAULT_ARGS = [
    '--disable-background-networking',
    '--disable-background-timer-throttling',
    '--disable-breakpad',
    '--disable-browser-side-navigation',
    '--disable-client-side-phishing-detection',
    '--disable-default-apps',
    '--disable-dev-shm-usage',
    '--disable-extensions',
    '--disable-features=site-per-process',
    '--disable-hang-monitor',
    '--disable-popup-blocking',
    '--disable-prompt-on-repost',
    '--disable-sync',
    '--disable-translate',
    '--metrics-recording-only',
[...]

PR #1516, opened 2021-02-03
Allow disabling pyppeteer sandbox

yuvipanda:
Needed when you're running in containers. This disables
downstream security features, so it needs to be explicitly
opted into.

Ref #1434


MSeal (approved):
Thanks @yuvipanda !

Issue #1526, opened 2021-02-25
Mark tests requiring network access


mcepl:
All builds (and running test suites as part of them) of OpenSUSE packages (the same for every other Linux or Mac distribution) happen in the chrooted environment without the network access, which makes running of the test suite for this package extremely awkward. Most of tests fail because they require network access, which is a bad idea even for other reasons (mocking is the way to go).

[skip_network_tests.patch](https://github.com/jupyter/nbconvert/files/6046285/skip_network_tests.patch.txt) is the solution of my problem, marking all tests requiring network access appropriately, and then skipping so marked ones.

MSeal:
Thanks for @mcepl ! And thanks for opening these across several projects. I agree that network requests during unittests is a bad idea. The chromium webpdf setup is a little annoying to both test the lazy download/install and not have network so we haven't had the time / energy to mock that interaction in a meaningful manner.

I made a PR with your patch applied. Were there other test you saw that needed this marking beyond this?

mcepl:
> I made a PR with your patch applied. Were there other test you saw that needed this marking beyond this?

Just with this (and setting `JUPYTER_PATH`) I had tests passing with this `%check` section (don't bother about details, it is basically bash syntax with some macros thrown in):
```rpm
%check
pushd docs
export LANG=en_US.UTF-8
export PYTHONDONTWRITEBYTECODE=1
export JUPYTER_PATH=%{buildroot}%{_datadir}/jupyter
%{python_expand export PYTHONPATH=%{buildroot}%{$python_sitelib}
$python -B -m ipykernel.kernelspec --user
pytest-%{$python_bin_suffix} -v -k 'not network' --pyargs nbconvert
}
popd
```

PR #1527, opened 2021-02-26
Applied patch for marking network changes

MSeal:
Co-authored-by: Matěj Cepl <mcepl@suse.com>

Closes https://github.com/jupyter/nbconvert/issues/1526


MSeal:
Test failure is unrelated -- that's from a different merge where the test update was missed.

PR #1534, opened 2021-03-22
Add correct output mimetype to WebPDF exporter

kycutler:
Similar to #970/#972 but for WebPDF, which currently inherits an HTML mimetype.


MSeal:
I think this is a duplicate of https://github.com/jupyter/nbconvert/pull/1514 -- which we need to go help review and determine the open issue on

MSeal:
Duplicate merged, should resolved this

Issue #1470, opened 2020-11-20
ExecutePreprocessor doesn't display images


Iuiu1234:
If I execute in console
```
jupyter nbconvert --to html --execute idea.ipynb
```
Everythings works as expected.
But if in python I execute
```
from nbconvert.preprocessors import ExecutePreprocessor

filename = 'idea.ipynb'
filename_out = 'idea2.html'

with open(filename) as f:
    nb = nbformat.read(f, as_version=4)

ep = ExecutePreprocessor()
ep.preprocess(nb)

with open(filename_out, 'w', encoding='utf-8') as f:
    nbformat.write(nb, f)
```

Html output doesn't display the plot 

idea.ipynb
```
# %%
import matplotlib.pyplot as p
p.plot([1,2,3])
```

System windows 10
nbclient==0.5.1
nbconvert==6.0.7
nbformat==5.0.8
nest-asyncio==1.4.3
notebook==6.1.5

Iuiu1234:
Same if do in py
```
from nbconvert.nbconvertapp import main

args = ['--to', 'html', '--execute', 'idea.ipynb']
main(args)
```

Iuiu1234:
It wasn't windows.
Was pycharm that scientific mode was redirecting my output

PR #1542, opened 2021-04-03
Drop Python 3.6 and test with Python 3.9

### LINKED ISSUE #1540: python 3.6 CI is broken because of traitlets
new traitlets>=5 version does not support python 3.6
this should be fixed in master, otherwise pull requests are kind of blocked.

SylvainCorlay:
Dropping Python 3.6

(Fixes #1540.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,9 @@ language: python
 
 matrix:
     include:
-        - python: 3.6
         - python: 3.7
         - python: 3.8
+        - python: 3.9
     allow_failures:
         - python: nightly
 env:


SylvainCorlay:
@MSeal what do you think about dropping Python 3.6?

agoose77:
@SylvainCorlay are we not forced to drop Python 3.6 support given that [we currently depend upon `traitlets >= 5.0`](https://github.com/jupyter/nbconvert/blob/e49b9a8a220c9f9f87c23764ea7b48eac4707937/setup.py#L219)?

mriedem (changes requested):
If this is really dropping support for python 3.6 then shouldn't the `python_requires` version in the setup.py be updated to be at least 3.7?

https://github.com/jupyter/nbconvert/blob/9689b2cb0658967e577585825cf8881a81feafa2/setup.py#L191

And if that's the case, the changelog should probably be updated as well right?

https://github.com/jupyter/nbconvert/blob/master/docs/source/changelog.rst

mriedem:
> If this is really dropping support for python 3.6 then shouldn't the `python_requires` version in the setup.py be updated to be at least 3.7?

I just noticed this in the README:

https://github.com/jupyter/nbconvert#python-requirement

So I guess 3.6 is supported but not all that strongly, and will be dropped in 6.1, so it seems this would be fine if it also triggers a 6.1 release.

agoose77:
@mriedem I don't think we can support 3.6 _at all_, because we have a hard dependency on `traitlets=5`. and [`traitlets` has `python_requires>=3.7`](https://github.com/ipython/traitlets/blob/5.0.0/setup.py#L90)

mriedem:
> However, I don't know what the protocol is when it's a dependency that is enforcing our version constraint. I would assume we drop support for 3.6?

Yeah I'd think so as well. I'm definitely not a regular contributor here, I'm just interested in getting #1470 merged since we need that downstream (we're capping nbconvert<6 as a result which is causing some other issues, i.e. jupyter/jupyter_client/issues/637).

CI has been broken for a few weeks and this PR has been up for about 3 weeks now so just trying to figure out how to proceed with getting CI passing again so things can start flowing for non-maintainers.

mriedem:
Also, I'm personally fine with dropping support for python 3.6, my -1 is just that I don't think we should probably drop CI testing for 3.6 without also officially dropping support for 3.6 which I think at least means:

- bumping the `python_requires` version in setup.py
- adding a changelog entry
- removing that section of the README about python 3.6
- release a 6.1 version

meeseeksmachine:

https://discourse.jupyter.org/t/current-state-of-nbconvert-drop-support-for-python-3-6/8882/1

agoose77:
OK, I have somehow made a mistake here - the current released versions of nbconvert use traitlets 4.x, so we're fine in terms of the release `install_requires`.

mriedem:
> The PR that introduced traitlets 5 is #1498

Yup and it failed CI but was merged (I can't see settings on the repo but seems failing CI should block a change unless an admin overrides it).

agoose77:
> > The PR that introduced traitlets 5 is #1498
> 
> Yup and it failed CI but was merged (I can't see settings on the repo but seems failing CI should block a change unless an admin overrides it).

Right, then I think it's just a case of reflecting the implied changes in the various metadata as you've outlined above :partying_face:

MSeal:
@SylvainCorlay Sorry I missed the notification on this originally. 3.6 technically has EOL from Python org until 12/21, but if it's desired to drop it given upstream deps dropping 3.6 it's probably fine, as 6.0.x is stable for 3.6 users anyway.

> bumping the python_requires version in setup.py
> adding a changelog entry
> removing that section of the README about python 3.6
> release a 6.1 version

We should probably get a couple other hanging PRs merged, e.g. extension for webpdf export fix, before the 6.1 release step. I've been a bit time constrained so haven't been able to put major effort into helping PRs get patched to a mergable state recently.

meeseeksmachine:

https://discourse.jupyter.org/t/nbconvert-6-1-0-release/9731/1

Issue #1423, opened 2020-09-24
Speaker notes not working for reveal.js slides


joelostblom:
I can't get speaker notes when exporting slides following [the instructions in the docs](https://nbconvert.readthedocs.io/en/latest/usage.html#convert-revealjs). Pressing `s` in the slide shows seems to not do anything while other keyboard shortcuts such as `o` works. Tested on recent versions of Firefox and Chromium, which work fine with speaker notes in [this demo reveal.js presentation](https://mvidal.net/slides/reveal.js/demo.html?transition=zoom#/) and with local reveal.js presentations generated from R Markdown by xaringan.

This is my directory layout:

```
.
├── reveal.js/
├── notes-test.ipynb
└── notes-test.slides.html
```

If I check out tag `3.5.0`, the slides don't format correctly and instead all code cells are visible on a single page. If I use `master` instead, the presentation works as expected except from the `s` key not bringing up the speaker view. I tried running both

```
jupyter nbconvert notes-test.ipynb --to slides --reveal-prefix reveal.js
```

and 

```
jupyter nbconvert notes-test.ipynb --to slides --reveal-prefix reveal.js --post serve
```

and

```
jupyter nbconvert notes-test.ipynb --to slides
```

All give seemingly the same result (when master is checkout in the reveal.js folder)

<details><summary>Contents of `notes-test.ipynb`</summary>
<pre>
{
 "cells": [
  {
   "cell_type": "code",
   "execution_count": 1,
   "metadata": {
    "slideshow": {
     "slide_type": "slide"
    }
   },
   "outputs": [
    {
     "data": {
      "text/plain": [
       "8"
      ]
     },
     "execution_count": 1,
     "metadata": {},
     "output_type": "execute_result"
    }
   ],
   "source": [
    "4 + 4"
   ]
  },
  {
   "cell_type": "markdown",
   "metadata": {
    "slideshow": {
     "slide_type": "slide"
    }
   },
   "source": [
    "# Main"
   ]
  },
  {
   "cell_type": "markdown",
   "metadata": {
    "slideshow": {
     "slide_type": "notes"
    }
   },
   "source": [
    "These are notes"
   ]
  },
  {
   "cell_type": "markdown",
   "metadata": {
    "slideshow": {
     "slide_type": "skip"
    }
   },
   "source": [
    "Skip"
   ]
  },
  {
   "cell_type": "markdown",
   "metadata": {
    "slideshow": {
     "slide_type": "slide"
    }
   },
[...]

PR #1543, opened 2021-04-06
Feature: add speaker notes to Reveal.js template

agoose77:
Reference in https://revealjs.com/plugins/#built-in-plugins


agoose77:
@SylvainCorlay here's the PR for #1423. I didn't run this locally but the changes are tiny, and mirror the template that I submitted to that issue :)

MSeal:
Seems like it's a reasonable change that was discussed. Tests pass locally with latest main pulled in. Merging for alpha release

PR #1545, opened 2021-04-08
Test jupyter_client

davidbrochart:


davidbrochart:
Not sure about the error for Python 3.6.7:
```
ERROR: Could not find a version that satisfies the requirement traitlets>=5.0 (from nbconvert)
ERROR: No matching distribution found for traitlets>=5.0
```
Other than that the next release of jupyter_client seems to be OK.

Issue #1507, opened 2021-01-27
CSSHTMLHeaderPreprocessor.style=<name> breaks with 6.x.x


jaymegordo:
### Description
I use `--CSSHTMLHeaderPreprocessor.style=solarized-dark` for example when exporting a notebook to modify the code syntax highlighting. In ` >6.0.0` this fails (silently) and only uses the default syntax. I've tested by both completely reinstalling everything in a new venv (still fails), then downgrading to `5.6.1` (works again).

Is there any way to confirm that `nbconvert` can find the correct `pygments` styles?

### Reporduce
- Create new conda env
- Install `ipython` and `nbconvert`
- `jupyter nbconvert --to html my_notebook.ipynb --CSSHTMLHeaderPreprocessor.style=solarized-dark` (doesnt use this theme's syntax)

**Nbconvert version:** 6.0.7

jupyter related packages:
Name | Version
-- | --
ipykernel | 5.3.4
ipython | 7.19.0
ipython_genutils | 0.2.0
jinja2 | 2.11.2
jupyter_client | 6.1.11
jupyter_core | 4.7.0
jupyter_server | 1.2.2
jupyterlab | 3.0.5
jupyterlab_pygments | 0.1.2
jupyterlab_server | 2.1.3
nbclassic | 0.2.6
nbclient | 0.5.1
nbconvert | 6.0.7
nbformat | 5.1.2
pygments | 2.7.4

JPinSPACE:
I am experiencing the same issue. When I run nbconvert with different styles and then diff the files there is no difference.

RagnosVA:
I got the solution for this after a lot of time
This solution contains changing Python Library Files
**Step 1 : ** Go to .\Python\Python39\Lib\site-packages\nbconvert\preprocessors
**Step 2: ** Open **csshtmlheader.py** with any text-editor
**Step 3: ** In class **CSSHTMLHeaderPreprocessor** there will be **style** variable, which will have **default_value=JupyterStyle**, just change value after '=' to any style you want but in qoutes( inverted commas ) 
for example : **default_value='vs'*
**Step 4:  Save** the file

![Jupyter Notebook](https://user-images.githubusercontent.com/76490924/112997558-daf5fa80-918a-11eb-9f53-48268bb5fe6c.jpg)

[...]

PR #1548, opened 2021-04-10
fix issue 1507: broken command line option --CSSHTMLHeaderPreprocessor.style=

finale80:
With the upgrade to version 6.x the pygment related command line option --CSSHTMLHeaderPreprocessor.style= is broken as reported by this open issue https://github.com/jupyter/nbconvert/issues/1507

I'm testing simply converting a notebook to html. For examples
```
jupyter nbconvert --HTMLExporter.theme=dark --to html notebook.ipynb --CSSHTMLHeaderPreprocessor.style=monokai
```

Looks like that exporters' config are not propagated to the preprocessors (?)
The function I'm changing does not appear in version 5.x so I assume there are design decisions underneath

The fix is simple, but there might be implications that I miss

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -499,6 +499,8 @@ class TemplateExporter(Exporter):
                 kwargs = preprocessor.copy()
                 preprocessor_cls = kwargs.pop('type')
                 preprocessor_cls = import_item(preprocessor_cls)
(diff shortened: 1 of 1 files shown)


MSeal:
Thanks for the improvement. I don't see an issue with the change and it seems helpful so I'll merge ahead of the alpha release here

MSeal:
Tests pass locally after merging in main, going to merge ignore CI

PR #1549, opened 2021-04-12
Test jupyter client

davidbrochart:


mriedem:
@davidbrochart I recreated the issue in #1548 

https://travis-ci.org/github/jupyter/nbconvert/jobs/766808690#L1370

```
self = <nbconvert.preprocessors.execute.ExecutePreprocessor object at 0x7f3e7ca1dd30>

msg_id = '238d6f9e-e6ea4495439b51dfd5175f5b_1'

cell = {'cell_type': 'code', 'metadata': {}, 'outputs': [], 'source': 'from __future__ import print_function\nfrom IPython.display import clear_output'}

timeout = 1

    def _poll_for_reply(self, msg_id, cell=None, timeout=None):

        try:

            # check with timeout if kernel is still alive

            msg = self.kc.shell_channel.get_msg(timeout=timeout)

>           if msg['parent_header'].get('msg_id') == msg_id:

E           TypeError: 'coroutine' object is not subscriptable
```

davidbrochart:
Thanks Matt, it is because we should use e.g. `self.kc.get_shell_msg()` instead of `self.kc.shell_channel.get_msg()` (see https://github.com/jupyter/jupyter_client/issues/640#issue-856223866).

jasongrout:
We are now testing master against jupyter_client 7.0.2: https://github.com/jupyter/nbconvert/runs/3592021910#step:7:35

Please let us know if we need to move forward with different tests of different versions of jupyter_client on master.

PR #1551, opened 2021-04-12
Allow child templates to override mathjax

yuvipanda:
Right now, mathjax is loaded synchronously in <head>, causing
render to block before moving forward. This is problematic,
especially since most notebooks don't actually have any
math to render.

This PR allows mathjax to be customized by child templates
that inherit from lab or classic. They can make it empty,
and load mathjax how they want - in the footer, only if
there is any math to be rendered, etc.

This should cause no difference in structure for the
default templates themselves.


SylvainCorlay:
This looks good to me. Merging!

PR #1554, opened 2021-04-16
fix if Inkscape path has spaces in it

MohsenEkhtiari:
fix if Inkscape path has spaces in it like on windows, resulting in error: 'C:\Program' is not recognized as an internal or external command

### DIFF
--- a/nbconvert/preprocessors/svg2pdf.py
+++ b/nbconvert/preprocessors/svg2pdf.py
@@ -79,7 +79,7 @@ class SVG2PDFPreprocessor(ConvertFiguresPreprocessor):
         export_option = ' --export-filename' if int(major_version) > 0 else ' --export-pdf'
         gui_option = '' if int(major_version) > 0 else ' --without-gui'
 
-        return '{inkscape}{gui_option}{export_option}='.format(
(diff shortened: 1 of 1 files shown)


nbconvert/preprocessors/svg2pdf.py
```diff
@@ -79,7 +79,7 @@ def _command_default(self):
         export_option = ' --export-filename' if int(major_version) > 0 else ' --export-pdf'
         gui_option = '' if int(major_version) > 0 else ' --without-gui'
 
-        return '{inkscape}{gui_option}{export_option}='.format(
+        return '"{inkscape}"{gui_option}{export_option}='.format(
```
ivanov, line 82: If you look at line 77 above, the quotes are already getting added there.

ivanov (commented):
Thanks for you contribution, @MohsenEkhtiari  - but I believe the fix you're proposing has already been applied a few lines higher in the code (line 77). I'm gonna go ahead and close this one.

MohsenEkhtiari:
Yes, thanks. It's fixed now. I had to fix it locally at that time!

PR #1567, opened 2021-05-12
small update to docs hide cell

lwasser:
closes #1194 
In the documentation, the code to hide cells currently returns an error where it can't find the `TagRemoveProcessor` object. A resolution was proposed in this issue to register the processor prior to calling it. 

Further the example is missing code to export the object to a file which many users may need to  do!  This pr  simply fixes the code sample.  @yuvipanda  I'm very open to making any changes and testing the locally! I am running the most current released version of nbconvert. i built the docs locally and ran into a small issue:

```bash
(nbconvert) CIRES-EL-LM-020:docs leahwasser$ make html
sphinx-build -b html -d build/doctrees   source build/html
Running Sphinx v3.5.4
loading intersphinx inventory from https://docs.python.org/3.6/objects.inv...
loading intersphinx inventory from http://jinja.pocoo.org/docs/dev/objects.inv...
loading intersphinx inventory from https://nbformat.readthedocs.io/en/latest/objects.inv...
intersphinx inventory has moved: http://jinja.pocoo.org/docs/dev/objects.inv -> https://jinja.palletsprojects.com/en/master/objects.inv
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 22 source files that are out of date
updating environment: [new config] 22 added, 0 changed, 0 removed
reading sources... [ 86%] nbconvert_library                                            
Notebook error:
NoSuchKernel in nbconvert_library.ipynb:
No such kernel named python3
```
i'm not sure if that is  related to  my  environment  install but  the docs that i modified should not actually cause any build issues. let's see what CI has to say.

UPDATE:  i was able to  fix this  by installing notebook in my nbconvert dev envt and rerunning the notebook figuring  perhaps it would then  find the right kernel and add it to the notebook  metadata. docs are now building nicely locally. i'll see  if that URL fix also  solves  the CI  issues.

PR #1571, opened 2021-05-13
add  pip to docs  envt

lwasser:
This  should  address the issue  where  CI  hangs when building docs due to not being able to find pip  correctly - note that the this builds upon issues discovered in #1567

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -12,5 +12,6 @@ dependencies:
 - tornado
 - entrypoints
 - ipykernel
+- pip
 - pip:
     - nbsphinx>=0.7.1


lwasser:
ok @yuvipanda so this pip fix makes the docs build fast! yay! 

```bash
Warning, treated as error:
/root/checkout/nbconvert/exporters/exporter.py:docstring of nbconvert.exporters.exporter.Exporter.register_preprocessor:8:more than one target found for 'any' cross-reference 'Preprocessor': could be :py:class:`nbconvert.preprocessors.Preprocessor` or :py:class:`nbconvert.preprocessors.base.Preprocessor`
```

But  this docstring issue is still  present.  i am not working tomorrow so i won't respond again until  next monday. Feel free to edit this  PR or  just  let me k now how to adjust that docstring  and i can  address it this  weekend  or monday! :)

yuvipanda:
This is already an improvement, @lwasser - without this there was hundreds of lines of output in the docs ci :) So I'm going to merge this now, and we can take another stab at the remaining issue when you come back :) Have a good weekend!

lwasser:
excellent! glad it was helpful. i'll dig a bit more into that docstring item one evening this week!

yuvipanda:
Thanks, @lwasser :)

PR #1576, opened 2021-05-17
Fix docstring issue and a broken  link

lwasser:
When building the docs a few issues come up. One is a bad reference link that no longer exists on the page.  Modified the link to be a jinja docs link for templates that works. The other is a docstring issue. will watch CI on this one and check back.

### DIFF
--- a/docs/source/architecture.rst
+++ b/docs/source/architecture.rst
@@ -151,7 +151,7 @@ Configurable classes.
 .. seealso::
 
     - :doc:`customizing`
-    - `More on Jinja Filters <https://jinja.palletsprojects.com/en/3.0.x/templates/#filters>`_
+    - `More on Jinja Filters`_
 
 Once it has passed through the template, an Exporter is done with the notebook,
 and returns the file data.
@@ -185,7 +185,7 @@ which is used for serving `reveal.js`_ HTML slideshows.
 
 .. links:
 
-.. _jinja: http://jinja.pocoo.org/
-.. _jinja_filter: https://jinja.palletsprojects.com/en/3.0.x/templates/#filters
+.. _jinja: https://jinja.palletsprojects.com/en/3.0.x/
+.. _More on Jinja Filters: https://jinja.palletsprojects.com/en/3.0.x/templates/#filters
 .. _reveal.js: http://lab.hakim.se/reveal-js
 .. _pandoc: https://pandoc.org/
--- a/docs/source/architecture.rst
+++ b/docs/source/architecture.rst
@@ -80,7 +80,7 @@ Once the notebook is loaded, it is preprocessed...
 Preprocessors
 -------------
 
-A `Preprocessor` is an object that transforms the content of the notebook to be exported.
+A ``nbconvert.preprocessors.Preprocessor`` is an object that transforms the content of the notebook to be exported.
 The result of a preprocessor being applied to a notebook is always a notebook.
 These operations include re-executing the cells, stripping output,
 removing bundled outputs to separate files, etc.
--- a/docs/source/architecture.rst
+++ b/docs/source/architecture.rst
@@ -180,7 +180,7 @@ Postprocessors
 --------------
 
 A *Postprocessor* is something that runs after everything is exported and written to the filesystem.
-The only postprocessor in nbconvert at this point is the `ServePostProcessor`,
+The only postprocessor in nbconvert at this point is the ``nbconvert.postprocessors.serve.ServePostProcessor``,
 which is used for serving `reveal.js`_ HTML slideshows.
 
 .. links:
--- a/docs/source/architecture.rst
+++ b/docs/source/architecture.rst
@@ -5,16 +5,16 @@ Architecture of nbconvert
 =========================
 
 This is a high-level outline of the basic workflow, structures and objects in nbconvert.
-Specifically, this exposition has a two-fold goal: 
-    
-    #. to alert you to the affordances available for customisation or direct contributions 
+Specifically, this exposition has a two-fold goal:
+
+    #. to alert you to the affordances available for customisation or direct contributions
     #. to provide a map of where and when different events occur, which should aid in tracking down bugs.
-    
+
 
 A detailed pipeline exploration
 ===============================
 
-Nbconvert takes in a notebook, which is a JSON object, and operates on that object. 
+Nbconvert takes in a notebook, which is a JSON object, and operates on that object.
 
 This can include operations that take a notebook and return a notebook.
 For example, that operation could be to execute the notebook as though it were a continuous script; if it were executed ``--in-place`` then it would overwrite the current notebook.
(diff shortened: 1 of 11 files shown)


docs/source/architecture.rst
```diff
@@ -186,6 +186,6 @@ which is used for serving `reveal.js`_ HTML slideshows.
 .. links:
 
 .. _jinja: http://jinja.pocoo.org/
-.. _filter: http://jinja.pocoo.org/docs/dev/templates/#filters
```
MSeal, line 189: Hmm still seems to be complaining about pocoo.org fetches -- maybe the line before or somewhere else in the project we're fetching objects that don't exist anymore?

MSeal (approved):
Thanks for helping improve the docs. I'm not sure what the change was causing the 404 jinja requests in the past. Might be worth looking at build history to see when that shifted / broke as well.

docs/source/architecture.rst
lwasser: yes @MSeal ! i'm really confused about the CI issue. Here is what i know. 

that url - jinja.pocoo.org/docs resolves to https://jinja.palletsprojects.com/en/3.0.x/

i've definitely seen builds have a hard time with redirects. i just fixed that but i'm still seeing that CI issue. i suspect somewhere there are other pocoo.org references perhaps? i can search for that next. 

http://jinja.pocoo.org/docs/dev/objects.inv 

i think i fixed it. the link for intersphinx_mapping was the old url. give me a few minutes to push it up. fingers crossed :)

docs/environment.yml
```diff
@@ -12,5 +12,6 @@ dependencies:
 - tornado
 - entrypoints
 - ipykernel
+- pip
```
lwasser, line 15: this was merged in my last pr. speeds up the build and simplifies the uutput

docs/source/architecture.rst
```diff
@@ -151,7 +151,7 @@ Configurable classes.
 .. seealso::
 
     - :doc:`customizing`
-    - :ref:`jinja:filters`
+    - `More on Jinja Filters`_
```
lwasser, line 154: updated link. this used `:ref:` before which is a local page link. I also fixed the URL to the jinja template page

docs/source/architecture.rst
lwasser: fixed this URL to be the new jinja docs link. palletsprojects.com

docs/source/conf.py
```diff
@@ -321,6 +321,6 @@
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {
     'python': ('https://docs.python.org/3.6', None),
-    'jinja': ('http://jinja.pocoo.org/docs/dev', None),
+    'jinja': ('https://jinja.palletsprojects.com/en/latest/', None),
```
lwasser, line 324: I suspect this was failing because the URL was a redirect and builds don't like redirects. maybe it was just a front end redirect which can cause problems. the url above is correct.

nbconvert/exporters/exporter.py
```diff
@@ -203,13 +203,13 @@ def register_preprocessor(self, preprocessor, enabled=False):
...
         Preprocessors are classes that act upon the notebook before it is
-        passed into the Jinja templating engine.  preprocessors are also
+        passed into the Jinja templating engine. Preprocessors are also
         capable of passing additional information to the Jinja
         templating engine.
 
         Parameters
         ----------
-        preprocessor : `Preprocessor`
+        preprocessor : `nbconvert.preprocessors.Preprocessor`
```
lwasser, line 212: i am not sure if this is the correct preprocessor object. It is either this or `.base.Preprocessor`. but this was causing the docs to fail too.

nbconvert/exporters/exporter.py
```diff
@@ -316,7 +316,7 @@ def _preprocess(self, nb, resources):
         # to each preprocessor
         for preprocessor in self._preprocessors:
             nbc, resc = preprocessor(nbc, resc)
-            try: 
+            try:
```
lwasser, line 319: my text editor Atom must do something with spaces. i hope this isn't problematic. if it is i can try to fix it.

lwasser:
ok the new point of failure is similar to the original issue i fixed. this is a simple docstring fix if we can identify which exporter object to use 
`exporters.TemplateExporter` OR `templateexporter.TemplateExporter`

```
/root/checkout/docs/source/api/filters.rst:6:more than one target found for 'any' cross-reference 'TemplateExporter': could be :py:class:`nbconvert.exporters.TemplateExporter` or :py:class:`nbconvert.exporters.templateexporter.TemplateExporter`
```
Which one is the right one?

lwasser:
ok i may need input from you @MSeal or @yuvipanda this seems to be a circular problem. once i fix a reference in a doc i get a new one. The nature of each issue is a potentially duplicative reference to an  object that has a base object with the same name and then maybe on that inherits or something - like the warning (treated as an error) below. i'm not familiar with the extension that is testing your docstrings for objects. and each time i've guessed which object is the correct one. the base one or the inherited one. i'm open to any input and also open to undoing  anything ive done so far. i am confident that the jinja fixes are good. but i cant replicate the other issues (like the one pasted below here) locally to troubleshoot.

`/root/checkout/docs/source/architecture.rst:64:more than one target found for 'any' cross-reference 'Exporter': could be :py:class:`nbconvert.exporters.Exporter` or :py:class:`nbconvert.exporters.exporter.Exporter``

MSeal:
Sorry it took a while for me to get back to this thread. I was able to get the builds working with all but the fitlers.rst `nbconvert.exporters.templateexporter.TemplateExporter`. I can't figure out why it dislikes that type reference after trying a few approaches, so I changed that one to a non-linked code reference and it appears to be the last issue in the docs locally.

MSeal:
Can't seem to get the errors to occur locally, so I have to keep pushing fixes :/

lwasser:
@MSeal I had the same issue - i think the errors [are related to this sphinx issue](https://github.com/sphinx-doc/sphinx/issues/4961) but i noticed the docs are not updating  so my guess i s until this is fixed they wont  build on readthedocs

lwasser:
one thing i  find odd is it's flagging files that are just code not even  references in  docstrings. i stopped when  i noticed that because i wasn't sure if/how  we  wanted to fix that! it seems like sphinx should ignore that type of issue or have an option to ignore it.

MSeal:
Yeah, we have all warnings as errors causing the build annoyance. Most of the references are in the changelog, but for now until that reference issue is resolved in sphinx I'm just going to change them to non-references for now to unblock.

MSeal:
Uhg, 'CellExecutionError' is being pulled in from nbclient which is outside the repo's ability to directly control

MSeal:
My local sphinx was on 3.x, upgrading to 4.x explicitly gets the errors to occur. I fixed all the ones not brought in but didn't know a way to get around the imported reference issues so I turned off warnings as errors for now

lwasser:
oh wow   -  docs built?!  i'll check read the docs !! this is great @MSeal thanks! what a pain to fix all of those references!

lwasser:
yay! docs updated finally!! 🎆

lwasser:
oh... good call. Yes that is really an annoying bug in sphinx right now! glad you were able to figure out how to just turn that off for the time being!

PR #1496, opened 2020-12-20
Fix order of template paths

m2-farzan:
Fixes https://github.com/jupyter/nbconvert/issues/1495.

I just played around and found this fix. I hope other devs can check this with more insight, so that nothing breaks by this patch.

Thanks!

P.S. After this patch, the repro steps described in https://github.com/jupyter/nbconvert/issues/1495 gives:
```
mostafa🐌phi ~ 
$ $VENV/jupyter nbconvert --debug --to=latex --template=my-latex-template my-notebook.ipynb
[NbConvertApp] Searching ['/repro/user-home', '/repro/user-home/.jupyter', '/repro/venv/etc/jupyter', '/usr/local/etc/jupyter', '/etc/jupyter'] for config files
[NbConvertApp] Looking for jupyter_config in /etc/jupyter
[NbConvertApp] Looking for jupyter_config in /usr/local/etc/jupyter
[NbConvertApp] Looking for jupyter_config in /repro/venv/etc/jupyter
[NbConvertApp] Looking for jupyter_config in /repro/user-home/.jupyter
[NbConvertApp] Looking for jupyter_config in /repro/user-home
[NbConvertApp] Looking for jupyter_nbconvert_config in /etc/jupyter
[NbConvertApp] Looking for jupyter_nbconvert_config in /usr/local/etc/jupyter
[NbConvertApp] Looking for jupyter_nbconvert_config in /repro/venv/etc/jupyter
[NbConvertApp] Looking for jupyter_nbconvert_config in /repro/user-home/.jupyter
[NbConvertApp] Looking for jupyter_nbconvert_config in /repro/user-home
[NbConvertApp] Converting notebook my-notebook.ipynb to latex
[NbConvertApp] Notebook name is 'my-notebook'
[NbConvertApp] Template paths:
	/repro/user-home/.local/share/jupyter/nbconvert/templates/my-latex-template 👈 Good.
	/repro/user-home/.local/share/jupyter
	/repro/user-home/.local/share/jupyter/nbconvert/templates
	/repro/user-home/.local/share/jupyter/nbconvert/templates/compatibility
	/repro/venv/share/jupyter
	/repro/venv/share/jupyter/nbconvert/templates
	/repro/venv/share/jupyter/nbconvert/templates/compatibility
	/usr/local/share/jupyter
	/usr/local/share/jupyter/nbconvert/templates
	/usr/local/share/jupyter/nbconvert/templates/compatibility
	/usr/share/jupyter
[...]

PR #1577, opened 2021-05-17
Fix template precedence when using a custom template (#1558)

### LINKED ISSUE #1558: article.tplx not found
Hi, I have seen many blogs saying users can modify article.tplx to create their own template and customize the pdf nbconvert generated. 
But with python =3.7 and nbconvert= 6.0.7, I can't find article.tplx or any file endswith tplx in these directories. 
```
from jupyter_core.paths import jupyter_path
print(jupyter_path('nbconvert','templates'))
['C:\\Users\\sssimonyang\\AppData\\Roaming\\jupyter\\nbconvert\\templates', 'D:\\ProgramData\\Miniconda3\\share\\jupyter\\nbconvert\\templates', 'C:\\ProgramData\\jupyter\\nbconvert\\templates']
```
I can only find files endswith .j2 in `D:\ProgramData\Miniconda3\share\jupyter\nbconvert\templates\latex`.

I wonder How can I generate an article.tplx or where can I find it ?
Thank you very much!

tobinjones:
The order of the template_paths matters. The additional and extra paths currently shadow the `index.*` template. This means custom templates are never called if these paths are set by the exporter (e.g. LatexExporter does) or by a configuration file.

This fixes #1558 but if there is a better way to do it, I'm happy to listen.

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -545,8 +545,8 @@ class TemplateExporter(Exporter):
                 ensure_dir_exists(path, mode=0o700)
             except OSError:
                 pass
-
-        return self.extra_template_paths + additional_paths + paths
+        
+        return paths + additional_paths + self.extra_template_paths 
 
     @classmethod
     def get_compatibility_base_template_conf(cls, name):


tobinjones:
This is a duplicate of #1496

PR #1590, opened 2021-06-09
Fix styles in WebPDF

### LINKED ISSUE #1589: WebPDF outputs missing styles in 6.1.0rc0
In nbconvert 6.1.0rc0, WebPDF outputs are missing any CSS styles. I was able to track down #1514 as the cause of the regression.

kycutler:
Fixes #1589


MSeal (approved):
Whoops! Ty for the fix!

Issue #1580, opened 2021-05-26
nbconvert generates URL fragments with invalid characters


charmasaur:
When headers in notebooks contain backslashes (for math, say), the generated URL fragments in the anchor links contain backslashes, which are invalid (https://datatracker.ietf.org/doc/html/rfc3986#section-3.5).

To reproduce, create a notebook with header: `# Pi $\pi$`.

This generates HTML code: `<h1 id="Pi-$\pi$">Pi $\pi$<a class="anchor-link" href="#Pi-$\pi$">&#182;</a></h1>`, which has in invalid fragment `#Pi-$\pi$` (due to the backslash).

See [pi.txt](https://github.com/jupyter/nbconvert/files/6543903/pi.txt).

**Nbconvert version:** 5.6.1 (edit: and 6.0.7)

(If the outcome here is just to tell me not to put math in headings, that's OK with me.)

Edit: note that this is similar, but not identical, to https://github.com/jupyter/nbconvert/issues/1233

MSeal:
Can you try with nbconvert 6 instead of 5.6.1? There aren't any more patches going to the older major version.

In general better parsing for math elements would be appreciated if people have time to help -- 6.x had improvements to this but there's still edge cases present.

charmasaur:
Same story with version 6. I can put it on my todo list to look into a fix here, not sure when I'll actually get to it but it's useful to know that contributions would be helpful (I'm never sure whether PRs from randoms like me are a help or a hinderance...).

MSeal:
Nah, new contributor PRs aren't hindrance and almost always welcomed :)

Sometimes it's not clear how to resolve follow-up concerns right away but I wouldn't call that a hindrance as everyone would hit those anyway.

PR #1591, opened 2021-06-10
Ensure generated URL fragments are valid

charmasaur:
Approach we take is to replace spaces with dashes (same as what was done
before), and then URL encode any remaining characters that aren't
allowed in URL fragments.

This shouldn't break any existing links, because any links that get changed by
this were invalid already.

Fixes https://github.com/jupyter/nbconvert/issues/1580.

### DIFF
--- a/nbconvert/filters/strings.py
+++ b/nbconvert/filters/strings.py
@@ -83,7 +83,10 @@ def _convert_header_id(header_contents):
 
     For use on markdown headings.
     """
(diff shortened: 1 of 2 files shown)


MSeal (approved):
Thanks @charmasaur ! I appreciate the test update as well :D

charmasaur:
No problem, thanks for the quick review!

Issue #129, opened 2015-10-06
Respect isolated metadata flag when converting to HTML


takluyver:
The live notebook looks for an `"isolated": true` flag in the metadata to put SVG/HTML in an iframe. However, conversion to static HTML appears to ignore this flag.

I suspect that this is the cause of IRkernel/IRkernel#198.

flying-sheep:
maybe this can work with static documents by doing `<iframe src="data:...">`

takluyver:
It looks like there's a srcdoc attribute, so you can do `<iframe srcdoc="<h1>Boo</h1>"/>`. Ugly, but better than a base-64 data URL. Not supported in IE, though.

flying-sheep:
neither does it support `src`, so IE users are out of luck anyway.

`srcdoc` sounds good. we only need to escape `"` and put it into `srcdoc` then?

flying-sheep:
maybe we should special-case SVG and use `<img src="data:...">`?

takluyver:
I think the srcdoc approach we described before is nicer than base64-ing text data.

flying-sheep:
i meant using imgs generally for SVG. but the most important thing is to get it in :wink:

michaelpacer:
Is this still open? 

Also, given the concerns elsewhere (https://github.com/jupyter/nbconvert/issues/328) that seem pro data URIs why is there anti data URI sentiment here? Is that not a valid solution?

flying-sheep:
the other thread doesn’t mention srcdoc, so that should work.

HTML/XML attribute values (strings) can span multiple lines, so we just need to think about quoting. would it be enough to

1. replace (in the HTML code) `&` with `&amp;` (so no entities will be expanded in the string)
2. replace `"` with `&quot;` (so the string isn’t ended prematurely)

this would result in this doc

```html
<div title="baz&quot;">
  &hellip;
</div>
```

being emedded like so:

```html
<iframe srcdoc="<span title=&quot;baz&amp;quot;&quot;>
    &amp;hellip;
</span>"></iframe>
```

which makes it still more readable than base64 encoded data

takluyver:
@michaelpacer data URIs are an OK if somewhat ugly way of embedding binary data like images into a text-based format like HTML. Encoding HTML, which is already text, into unreadable base64 data seems rather inelegant. That's why I prefer the srcdoc approach.

mpacer:
I can revisit this if desired after the 5.2 release.

[...]

PR #1593, opened 2021-06-21
Supports isolated iframe when converting to HTML

Nihiue:
Related issue:
  Respect isolated metadata flag when converting to HTML #129


MSeal (approved):
LGTM, thanks for submitting a PR for the fix! We're about to ship off a release tonight so it should be live by tomorrow.

PR #1596, opened 2021-06-24
Get TemporaryDirectory from stdlib tempfile, not from testpath.

anntzer:
They are aliases of one another on all supported Python versions.

I would like to get rid of the testpath dependency, but in any case,
this specific change should hopefully be uncontroversial.


MSeal (approved):
ty!

Issue #1401, opened 2020-09-19
Since 6.0 Markdown in HTML divs render as LaTeX instead of being skipped


joelostblom:
In 5.6.1, markdown inside HTML `<div>` tags was skipped in the nbconvert HTML output, which is not ideal, but was still readable. Since upgrading to 6.0, I have noticed that they are rendered as LaTeX math. In many cases, this makes the text ineligible, so this is quite a big issue (but possible to workaround by using HTML inside the div instead of markdown, although this means the same notebook can't be exported to both PDF and HTML without modifications).

As an example, the following code

````html
<div>

Testing HTML with `inline code` tags.<br>

```
I am a code block
```
    
</div>
````

Would show up like this (and [still does on GH](https://github.com/joelostblom/nbc-html-test/blob/master/Untitled.ipynb), which I guess has not update to 6.0)

![image](https://user-images.githubusercontent.com/4560057/93689200-08961480-fa81-11ea-9267-d1685bfcd127.png)

But since 6.0, it shows up like this

![image](https://user-images.githubusercontent.com/4560057/93689170-ccfb4a80-fa80-11ea-93ff-e708076e6612.png)

```
jupyter core     : 4.6.3
jupyter-notebook : 6.1.4
qtconsole        : not installed
ipython          : 7.18.1
ipykernel        : 5.3.4
jupyter client   : 6.1.7
jupyter lab      : 2.2.6
nbconvert        : 6.0.3
ipywidgets       : not installed
nbformat         : 5.0.5
traitlets        : 5.0.4
```

For reference to the old skipping behavior, these are the currently open HTML markdown related issues:

- https://github.com/jupyter/nbconvert/issues/1037
- https://github.com/jupyter/nbconvert/issues/1165
- https://github.com/jupyter/nbconvert/issues/1146
- https://github.com/jupyter/nbconvert/issues/1125

MSeal:
I'm not sure I will get time to dig into the HTML format issues referenced before I'm unavailable with a newborn... @t-makaro any chance you have some spare cycles to look into these? Also any new contributors that want to track this down we'd love to have more eyes on these open issues.

nh273:
I can help with this as a first issue! Would love to get any pointers!

MSeal:
Thanks for offering to help @nh273 !

Somewhere we're the code is not properly escaping characters or incorrectly passing them through a filter it shouldn't be using.

[...]

PR #1599, opened 2021-06-29
Adds dejavu

marimeireles:
This is the first PR to introduce the Dejavu feature.
Dejavu intends to be a tool to facilitate for the Jupyter user to generate static pdf outputs for their notebooks mimicking the behavior of voilà. So, with one command line you get all the presets configured for you, generating a pdf presentation:

```
jupyter dejavu notebook.ipynb
```

I'm trying to mimic `voila`'s results with this. For future PRs we'll probably have to change or add a new template configuration so we can have a correct pagination.
Please let me know what you think of it!

cc: @SylvainCorlay

### DIFF
--- a/docs/source/dejavu.rst
+++ b/docs/source/dejavu.rst
@@ -1,0 +1,39 @@
+.. highlight:: none
+
+Dejavu
+======
+
+Dejavu intends to be a tool to facilitate for Jupyter users to generate static pdf outputs from their notebooks, mimicking the behavior of `voilà <https://github.com/voila-dashboards/voila>`_.
+
+Configuring the Notebook
+------------------------
+
+In case the user intends to do a slide presentation out of their Jupyter notebook it's recommended to use the ``reveal`` template. In orders to obtain a better result from it's advised to use the slides metadatas available in the cells:
+
+
+* In the notebook, select a cell and click on the "Property Inspector menu"
+
+.. tip::
+
+   The "Property Inspector menu" can be located in the right side bar, its symbol contains two gears.
+
+* Select a cell in the notebook
+
+* In the Property Inspector menu select the cell's slide type:
+
+    * Slide
+    * Sub-Slide
+    * Fragment
+    * Skip
+    * Notes
+
+* Repeat the process for all cells
+
+Running dejavu
+--------------
+
+Dejavu works exactly the same as nbconvert and you can use all command line options that you would with nbconvert. To run a default instance::
+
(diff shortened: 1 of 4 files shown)


marimeireles:
I also have an example notebook in case you're interested in starting an example session. =)

MSeal (commented):
Interesting! So the intend is to mimic voila default behavior? Should we make this a template extension of reveal rather than a new App to avoid command confusion maybe?

@SylvainCorlay @maartenbreddels would like your input

SylvainCorlay:
Voilà is really just **nbconvert + live kernel and widgets** except that by default, inputs and prompt numbers are hidden.

The goal of déjàvu is to provide a shortcut for nbconvert that has the same defaults as Voilà.

```python
TemplateExporter.exclude_input = True            # Hide inputs
TemplateExporter.exclude_output_prompt = True    # Hide prompts
TemplateExporter.exclude_input_prompt = True     # Hide prompts
ExecutePreprocessor.enabled = True               # Execute notebook
WebPDFExporter.paginate = False                  # Prevent WebPDF pagination
```

We could not make it an alias because one should be able to overwrite these options (like in the case of Voilà).

I am not sure if we should default to HTML or WebPDF.

marimeireles:
If we're default to WebPDF we have to merge https://github.com/jupyter/nbconvert/pull/1601 too. I think if it was HTML we wouldn't have to make this change.
But anyways, I think https://github.com/jupyter/nbconvert/pull/1601 makes sense.

marimeireles:
IMO makes sense to default the `reveal` template as the default one, since it's the only officially supported template for slides visualization.

marimeireles:
>I am not sure if we should default to HTML or WebPDF.

So Sylvain and I were talking about this and we got in an agreement that's better to generate an html file so it exhibits the exact same result as voila. Even though pdf seems like a more useful format it produces a different less elegant result than when converted to html.

MSeal (approved):
Alright sounds reasonable to me. There's a doc build issue seems unrelated to this PR. There's been some issues with the new version of Sphinx that were unresolved last time I checked that we were working around.

marimeireles:
Thanks @MSeal, soon I'll also be publishing a blog post that talks a bit about it + nbconvert! :)

MSeal:
Great! I think if we can also address https://github.com/jupyter/nbconvert/issues/1401 this week we should do a patch release with the changes this weekend so they're available to others.

Issue #1568, opened 2021-05-13
DeprecationWarning: 'contextfilter' is renamed to 'pass_context'


slayoo:
https://github.com/jupyter/nbconvert/blob/d7c286b9bd07a8670d8ab08469c2e266140522c1/nbconvert/exporters/html.py#L14

As discussed here: https://github.com/pallets/jinja/issues/1381
``contextfilter`` is now deprecated.

This causes the following warning to show up whenever using the exporters:
```
...
/opt/hostedtoolcache/Python/3.9.5/x64/lib/python3.9/site-packages/nbconvert/exporters/html.py:100: in HTMLExporter
    def markdown2html(self, context, source):
/opt/hostedtoolcache/Python/3.9.5/x64/lib/python3.9/site-packages/jinja2/filters.py:54: in contextfilter
    warnings.warn(
E   DeprecationWarning: 'contextfilter' is renamed to 'pass_context', the old name will be removed in Jinja 3.1.
```

Thanks

sjdemartini:
This should presumably have been fixed in https://github.com/jupyter/nbconvert/pull/1624

slayoo:
thanks!

twhughes:
Hi, I am still getting this warning with nbconvert.  Version: 6.1.0

Is there a way to resolve?

dimpase:
`contextfilter` is removed in the current `jinja2` version, 3.1.1, on PyPI.
So fixing this is getting urgent.

PR #1604, opened 2021-07-16
Fix Deprecation warning for jinja2 contextfilter

sjdemartini:
Resolves https://github.com/jupyter/nbconvert/issues/1568.

As noted in the Jinja docs here https://jinja.palletsprojects.com/en/3.0.x/api/#jinja2.contextfilter and in the Changelog here https://jinja.palletsprojects.com/en/3.0.x/changes/#version-3-0-0, the `contextfilter` decorator has been replaced with `pass_context`, which provides the same functionality. `contextfilter` will be removed in Jinja 3.1.

Side note: I'm not really familiar with the internals of the html exporter and the markdown filter, so I wasn't sure how to test this (and didn't know whether existing coverage would ensure this still works).

### DIFF
--- a/nbconvert/exporters/html.py
+++ b/nbconvert/exporters/html.py
@@ -11,7 +11,7 @@ import base64
 from traitlets import default, Unicode, Bool
 from traitlets.config import Config
 from jupyter_core.paths import jupyter_path
-from jinja2 import contextfilter
+from jinja2 import pass_context
 from jinja2.loaders import split_template_path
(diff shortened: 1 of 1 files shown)


jasongrout:
Since we support jinja2>=2.4, can you put this in a try/except testing to see if pass_context is there?

jasongrout:
It looks like this is fixed in #1624 with a try/except to test for the import.

PR #1609, opened 2021-07-29
Allow users to customize mathjax URLs

bsyouness:
We're running into an issue where we'd like to customize the mathjax URL used by nbconvert but we're unable to do so easily. This PR uses the same methodology to make the mathjax customizable as is used to load the js scripts in the require.js jinja templates. This allows users to circumvent the use of CDNs to pull the mathjax dependencies.


bsyouness:
@bollwyvl @MSeal can I ask for a review? Not sure who to reach out to? Also the CI broke with this PR it looks like: https://github.com/jupyter/nbconvert/pull/1599

jasongrout:
@bsyouness - we just fixed the CI checks. Can you rebase or merge in main like you did above to get the CI checks working again?

share/jupyter/nbconvert/templates/base/mathjax.html.j2
```diff
@@ -1,7 +1,8 @@
 
-{%- macro mathjax(url="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.7/latest.js?config=TeX-MML-AM_CHTML-full,Safe") -%}
+{%- macro mathjax(mathjax_tex_url, mathjax_safe_url) -%}
```
jasongrout, line 2: Thanks for making this configurable. A couple of questions about these changes:

1. I'm curious why you separated out the config into two separate script tags. My understanding is that mathjax should be loaded once, and that config argument can give multiple configurations (like `config=TeX-MML-AM_CHTML-full,Safe`), but that would be different from loading mathjax two times, each time with a different config.
2. Is this macro considered part of the public api? In other words, will user templates use this macro expecting the same signature as currently exists? Maybe we should give some default argument so users can continue using `mathjax()` in their templates and have it still function. For our templates changed in this PR, it would be loading the traitlets config values still.

SylvainCorlay:
@bsyouness I noticed that your git author email does not match your github account email, which causes github to not credit them to you. You may want to either change the author email for these commits or add that address to your github account to be properly credited before the PR gets in.

jasongrout (approved):
This looks great to me now. Thanks!

jasongrout:
Since Youness and I both work at the same place, so I'd prefer if someone else (@MSeal?) looks at this too and does the merge. I'm also happy to merge it if someone else approves.

MSeal (approved):
LGTM as well

PR #1614, opened 2021-08-11
Add the ability to fully customize `widgetRendererSrc`

bsyouness:
Currently the `widgetRendererSrc` variable inside of the jinja template is only partially customizable. This would allow users to pass in a traitlet to fully customize it.


share/jupyter/nbconvert/templates/lab/index.html.j2
```diff
@@ -94,5 +94,4 @@ a.anchor-link {
 {% endblock footer_js %}
 {{ super() }}
 </html>
-{% endblock footer %}
-
+{% endblock footer %}
```
jasongrout, line 97: Add back in the newline.

```suggestion
{% endblock footer %}

```

share/jupyter/nbconvert/templates/classic/index.html.j2
```diff
@@ -74,7 +74,7 @@ div#notebook-container{
 {% endblock notebook_css %}
 
 {%- block html_head_js_mathjax -%}
-{{ mathjax() }}
+{{ mathjax(resources.mathjax_tex_url, resources.mathjax_safe_url) }}
```
jasongrout, line 77: Looks like this accidentally made it in from the other PR?

```suggestion
{{ mathjax() }}
```

SylvainCorlay:
Kicked CI.

SylvainCorlay:
Quick thought from looking at it: since we are adding an argument to the `jupyter_widgets` macro, we should check that it is not breacking Voilà templates which are also making use of this macro.

jasongrout:
> Quick thought from looking at it: since we are adding an argument to the `jupyter_widgets` macro, we should check that it is not breacking Voilà templates which are also making use of this macro.

It's not clear to me from the [Jinja2 docs](https://jinja.palletsprojects.com/en/latest/templates/#macros) nor from the [source](https://github.com/pallets/jinja/blob/35c69e9ec3597a88977ddd27628ecd9318e04737/src/jinja2/runtime.py#L740) how Jinja handles a positional argument without a value. This PR seems to indicate that it should work like Javascript, i.e., if you don't pass in a value for a positional argument, it is assigned a falsey value. If Jinja doesn't work like this, perhaps just giving that third argument a default value that is falsey is enough to cover this concern and the compatibility concern @SylvainCorlay raised.

bsyouness:
Thank you for fixing the CI and for the review! When you don't pass anything to `widgets_url` it's falsey, and like Jason is saying, it behaves like javascript. So if nothing is passed, the former behavior is unchanged.

SylvainCorlay:
> Also, I'm guessing the tests should be/would catch that?

They may not catch failures downstream (with e.g. Voilà).

bsyouness:
I added a default for `widgets_base_url` so it's explicitly set to `''` when nothing is passed.

share/jupyter/nbconvert/templates/base/jupyter_widgets.html.j2
```diff
@@ -1,19 +1,27 @@
-{%- macro jupyter_widgets(widgets_cdn_url, html_manager_semver_range) -%}
+{%- macro jupyter_widgets(widgets_cdn_url, html_manager_semver_range, widgets_url='') -%}
 
 <script>
 (function() {
   function addWidgetsRenderer() {
     var mimeElement = document.querySelector('script[type="application/vnd.jupyter.widget-view+json"]');
     var scriptElement = document.createElement('script');
+    {% if widgets_url %}   
+    var widgetRendererSrc = '{{ widgets_url }}embed-amd.js';
```
jasongrout, line 9: Since it said the widgets_url was the "full URL for Jupyter widgets", I did the following: `jupyter nbconvert  NB.ipynb --to html --HTMLExporter.widgets_url='https://cdn.jsdelivr.net/npm/@jupyter-widgets/html-manager/dist/embed-amd.js'`

However, that attempted to get `https://cdn.jsdelivr.net/npm/@jupyter-widgets/html-manager/dist/embed-amd.jsembed-amd.js`, which you can see has an extra `embed-amd.js` tacked onto the end.

@bsyouness, do you think it would be more clear to make this url the actual full url, and not tack on the embed-amd.js part?

I see that below, we have another call that changes `embed-amd.js` for `embed.js` for widget state below version 2, but I think that:

1. the widget state was upgraded to version 2 on the ipywidgets 7.0 release, which was Aug 18, 2017, so we probably don't need to automatically switch the fetched file anymore on new changes.
2. If a person is converting an old file, they can manually specify the `.../embed.js` file (since they will need to give an old version of the html manager anyway, I think).

So I'd suggest changing the above line to:

```suggestion
    var widgetRendererSrc = '{{ widgets_url }}';
```

and making a similar change to line 21 below (deleting `embed.js`.

SylvainCorlay: I agree with jason on this.

Also widgets_url really concerns the URL for the manager URL, and not the URL where custom widgets are fetched. Maybe the name should be a bit more specific.

SylvainCorlay: maybe `widget_renderer_url`...

bsyouness:
@SylvainCorlay @jasongrout are we ok to merge this then?

SylvainCorlay:
Thanks @bsyouness this looks good to me. Merging.

SylvainCorlay (approved):

Issue #1605, opened 2021-07-16
HTML exporter DeprecationWarning: 'jinja2.Markup' is deprecated and will be removed in Jinja 3.1


sjdemartini:
When using the HTML exporter, the following warning will show up:
```
   /.../venv/lib/python3.8/site-packages/nbconvert/exporters/html.py:125: DeprecationWarning: 'jinja2.Markup' is deprecated and will be removed in Jinja 3.1. Import 'markupsafe.Markup' instead.
     return jinja2.Markup(code)
```

This was updated in jinja2 here https://github.com/pallets/jinja/pull/1391. Semi-related to https://github.com/jupyter/nbconvert/issues/1568 (another Jinja deprecation warning).

`jinja2.Markup` is used in a few places in the HTML exporter:

https://github.com/jupyter/nbconvert/blob/90b3911ed9b155e8ca3680ca4581d9344aa10be0/nbconvert/exporters/html.py#L135

https://github.com/jupyter/nbconvert/blob/90b3911ed9b155e8ca3680ca4581d9344aa10be0/nbconvert/exporters/html.py#L140

https://github.com/jupyter/nbconvert/blob/90b3911ed9b155e8ca3680ca4581d9344aa10be0/nbconvert/exporters/html.py#L165

It seems it needs to be replaced with an import from the `markupsafe` package instead of `jinja2`.

**Nbconvert version:** 6.0.7 (deprecated class usage appears to be present on master as shown above, so will presumably show up with 6.1.0 as well)

minrk:
closed by #1737

PR #1624, opened 2021-08-30
Jinja 2 deprecated functionality

Carreau:
This make traitlets test suit fails as it tests downstream projects.


MSeal (approved):
Ty

sjdemartini:
Thanks! Would you be able to handle https://github.com/jupyter/nbconvert/issues/1605 as well? Another jinja2 deprecation problem

Issue #1490, opened 2020-12-14
What is the official way for handling "--no-sandbox" situation?


ghuname:
I get an error if I execute:

```
(base) jovyan@7876407ed131:~$ jupyter nbconvert --to webpdf bg_gradient.ipynb 
[NbConvertApp] Converting notebook bg_gradient.ipynb to webpdf

[NbConvertApp] Building PDF
Traceback (most recent call last):
  File "/opt/conda/bin/jupyter-nbconvert", line 11, in <module>
    sys.exit(main())
  File "/opt/conda/lib/python3.8/site-packages/jupyter_core/application.py", line 270, in launch_instance
    return super(JupyterApp, cls).launch_instance(argv=argv, **kwargs)
  File "/opt/conda/lib/python3.8/site-packages/traitlets/config/application.py", line 845, in launch_instance
    app.start()
  File "/opt/conda/lib/python3.8/site-packages/nbconvert/nbconvertapp.py", line 350, in start
    self.convert_notebooks()
  File "/opt/conda/lib/python3.8/site-packages/nbconvert/nbconvertapp.py", line 524, in convert_notebooks
    self.convert_single_notebook(notebook_filename)
  File "/opt/conda/lib/python3.8/site-packages/nbconvert/nbconvertapp.py", line 489, in convert_single_notebook
    output, resources = self.export_single_notebook(notebook_filename, resources, input_buffer=input_buffer)
  File "/opt/conda/lib/python3.8/site-packages/nbconvert/nbconvertapp.py", line 418, in export_single_notebook
    output, resources = self.exporter.from_filename(notebook_filename, resources=resources)
  File "/opt/conda/lib/python3.8/site-packages/nbconvert/exporters/exporter.py", line 181, in from_filename
    return self.from_file(f, resources=resources, **kw)
  File "/opt/conda/lib/python3.8/site-packages/nbconvert/exporters/exporter.py", line 199, in from_file
    return self.from_notebook_node(nbformat.read(file_stream, as_version=4), resources=resources, **kw)
  File "/opt/conda/lib/python3.8/site-packages/nbconvert/exporters/webpdf.py", line 93, in from_notebook_node
    pdf_data = self.run_pyppeteer(html)
  File "/opt/conda/lib/python3.8/site-packages/nbconvert/exporters/webpdf.py", line 83, in run_pyppeteer
    pdf_data = pool.submit(run_coroutine, main()).result()
[...]

PR #1625, opened 2021-08-31
Add CLI configuration for disable-chromium-sandbox

skovhus:
This add CLI configuration parameter for the feature introduced in https://github.com/jupyter/nbconvert/pull/1516

Closes https://github.com/jupyter/nbconvert/issues/1490

### DIFF
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -136,6 +136,13 @@ nbconvert_flags.update({
         },
         """Whether to allow downloading chromium if no suitable version is found on the system."""
         ),
+    'disable-chromium-sandbox': (
+        {'WebPDFExporter': {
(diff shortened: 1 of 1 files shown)


MSeal:
I think you'll still need to pass that trailer assignment to the pyppeteer call as well

skovhus:
> I think you'll still need to pass that trailer assignment to the pyppeteer call as well

I'm not sure I follow, can you elaborate? I tested this out and the new CLI flag is being forwarded correctly.

MSeal:
Ahh nvm the traitlet within the class was already set that you're setting at the app level. Thanks for testing that it's working!

MSeal:
Test failures are related to changing test patterns over from circleci and we can ignore those for this PR

Issue #1518, opened 2021-02-09
Doc: tex installation recommendation out dated


we-taper:
The package `texlive-generic-recommended` is no longer available in Ubuntu 20.04:

```bash
➜  ~ sudo apt-get install texlive-generic-recommended
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Package texlive-generic-recommended is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'texlive-generic-recommended' has no installation candidate

```

But it is mentioned in [docs/source/install.rst](https://github.com/jupyter/nbconvert/blob/master/docs/source/install.rst).
What is the alternative package? Is it `texlive-latex-recommended`?

mgeier:
According to the Debian package description (https://packages.debian.org/buster/texlive-generic-recommended), this is just a dummy package that makes sure that `	texlive-plain-generic` is installed.

PR #1617, opened 2021-08-24
Replace CircleCI and Travis tests with Github actions

jasongrout:
This moves us over to Github Actions for the CI, as we've been doing with many Jupyter projects.

Fixes #1615

PR #1630, opened 2021-09-04
Recommend texlive-plain-generic instead of texlive-latex-recommended

mathbunnyru:
Close: https://github.com/jupyter/nbconvert/pull/1530
Fix: https://github.com/jupyter/nbconvert/issues/1518

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -66,7 +66,7 @@ packages are specific to different operating systems:
 
   * E.g. on Debian or Ubuntu::
 
-        sudo apt-get install texlive-xetex texlive-fonts-recommended texlive-latex-recommended
+        sudo apt-get install texlive-xetex texlive-fonts-recommended texlive-plain-generic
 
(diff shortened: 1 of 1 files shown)


mathbunnyru:
I tried to use current docs for ubuntu image - but it doesn't work for me.
It works well with `texlive-plain-generic`.

https://github.com/jupyter/docker-stacks/pull/1458/files

jasongrout:
FYI, we also ran into this when updating the tests over in https://github.com/jupyter/nbconvert/pull/1617

We updated the line to https://github.com/jupyter/nbconvert/blob/53c46d2785bfbaebdcb1ab7a959f7c09e9209477/.github/workflows/docs.yml#L21
but I didn't try really hard to get a minimal package set, I just got to a package set that worked and was reasonably minimal.

jasongrout:
I tried this out over on #1636 and it seems to work great. Thanks!

PR #1637, opened 2021-09-13
Turn on deprecation warnings in tests

jasongrout:
I'm trying to duplicate the warning from #1604 before we fix it so we can catch things like this in the future.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -38,16 +38,25 @@ jobs:
             ${{ runner.os }}-pip-${{ matrix.python-version }}
       - name: Install
         run: pip install -v ".[all]"
+      - name: List installed packages
+        run: |
+          pip freeze
+          pip check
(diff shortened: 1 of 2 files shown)


jasongrout:
Interestingly, I do get some deprecation warnings in the logs, but I don't get the warning from #1604, even though I do get that contextfilter deprecation warning when converting a notebook to html on my desktop.

Are we not testing an end-to-end simple conversion to html?

MSeal:
There's end-to-end tests for conversions: https://github.com/jupyter/nbconvert/blob/main/nbconvert/exporters/tests/test_html.py

MSeal (approved):
I think we turned some of the warning monitoring off when Sphinx was on a version that would incorrectly spew a ton of errors from it's source code in our builds. Though I don't recall if that was a `-W` flag here or somewhere else

PR #1642, opened 2021-09-15
Update the pyppeteer dependency to resolve security issues.

jasongrout:
pyppeteer 0.2.6 includes a fix for https://github.com/advisories/GHSA-8ch4-58qp-g3mp by upgrading its dependency on the websocket library. See 	https://github.com/pyppeteer/pyppeteer/pull/252 and https://github.com/pyppeteer/pyppeteer/issues/275

CC @mnowacki-b

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -227,7 +227,7 @@ setup_args['install_requires'] = [
     'nbclient>=0.5.0,<0.6.0'
 ]
 
-pyppeteer_req = 'pyppeteer==0.2.2'
+pyppeteer_req = 'pyppeteer==0.2.6'
 
 extra_requirements = {
     'test': [


jasongrout:
@bollwyvl - do we still need to hard-pin this dependency? You mentioned in https://github.com/jupyter/nbconvert/pull/1285#issuecomment-643047401:

> So if it is kept, it would need to be pinned to 0.2.2 or so (and we'd have to re-apply all of the basic security patches on conda-forge to get that version out), or just hold on this until... whatever the next version is comes out.

I'm not sure if we still need an exact pin, or if it is fine for us to pin to `~0.2.6`.

jasongrout:
@SylvainCorlay - maybe you could also help answer if we need an exact version pin on pyppeteer anymore, or if we could do ~0.2.6?

SylvainCorlay:
I don't know why  we should pin down the pyppeteer dependency.

jasongrout:
> do we still need to hard-pin pyppeteer to an exact version (upgraded here to 0.2.6)? You mentioned in [#1285 (comment)](https://github.com/jupyter/nbconvert/pull/1285#issuecomment-643047401):

Reading @bollwyvl's comment again, it seems that he was saying that because pyppeteer was undergoing lots of evolution, we could either wait until 0.2.3 or we could pin to 0.2.2, but we shouldn't upgrade to 0.2.3 automatically. Perhaps that concern is resolved now?

jasongrout:
Looking at the [pyppeteer changelog](https://github.com/pyppeteer/pyppeteer/blob/dev/CHANGELOG.md), it looks like there are still lots of things happening on each patch release, so unless I get positive confirmation that it is okay to do `~=0.2.6`, I'll be conservative and pin to `==0.2.6`.

jasongrout:
> positive confirmation that it is okay to do `~=0.2.6`

(and we can do that in a separate PR...)

PR #1643, opened 2021-09-16
HTMLExporter: Add a new output filter that excludes widgets if there is no state

martinRenou:
This allows showing a static mimetype if the widget state is not available.

Example 1: Widget state not available, we render a static plot
nbconvert_static

Example 2: Widget state available, we render the widget (blank because the plot data is not part of the state (yet))
nbconvert_widget

This has no implication on Voila, voila continues rendering the widgets properly (this is because it will simply not use this filter):
voila

This is related to the work done in https://github.com/matplotlib/ipympl/pull/343


nbconvert/exporters/html.py
```diff
@@ -98,22 +99,22 @@ def _template_name_default(self):
     def default_config(self):
         c = Config({
             'NbConvertBase': {
-                'display_data_priority' : ['application/vnd.jupyter.widget-state+json',
```
martinRenou, line 101: I removed this as I don't see why this would be in cell outputs? I can revert this change if it can actually happens

SylvainCorlay: +1 for removing this. It does not make sense.

SylvainCorlay:
@jasongrout do you think we could consider this for 6.2?

We just released `ipympl 0.8` which now includes a PNG representation of figure widgets. This fallback mechanism will allow nbconvert to render static Matplotlib figures when the widget state is missing, like [GitHub does](https://github.com/matplotlib/ipympl/blob/master/examples/ipympl.ipynb).

SylvainCorlay (approved):

maartenbreddels:
Very nice! I think this makes total sense since if the state is not there, nothing can be done with the mime bundle entry. A potential future issue could be when the widget state is stored outside of the notebook, not something to worry about now.

jasongrout:
> @jasongrout do you think we could consider this for 6.2?

I'm planning on releasing 6.2 tomorrow since there haven't been bug reports on the RC. I think this is a big enough change I'd rather do the release (rather than reset the RC timeline) and get it out the door, and immediately start the process for 6.3.

SylvainCorlay:
Version 6.2 has been released a few days ago, and we haven't seen major breakage.
I think it is fine to start merging things in preparation for 6.3.

Issue #1652, opened 2021-10-04
"None" appears when converting to HTML with "full" template


alimanfoo:
When I use ``jupyter nbconvert --to html --template full`` I get an HTML file which looks great, apart from the fact that the text "None" appears at the top of the page. Looking at the HTML source, the text "None" is right at the beginning, before the doctype declaration.

To reproduce, run:

``jupyter nbconvert --to html --template full any.ipynb``

If instead of ``--template full`` I use ``--template classic`` this problem goes away. I gather that ``--template classic`` is now preferred, but the help docs still mention using ``--template full``.

Seems to happen with pretty much any notebook.

**Nbconvert version: 6.2.0**

PR #1653, opened 2021-10-04
Fix HTML templates mentioned in help docs

alimanfoo:
This PR updates the list of HTML templates available in the examples documentation.


alimanfoo:
Related to #1652. Updating the docs here would encourage use of ``--template classic`` instead of ``--template full``.

PR #1665, opened 2021-10-27
fix: avoid closing the script tag early by escaping a forward slash

### LINKED ISSUE #802: raw json in nbconvert output
I converted the attached notebook as part of a sphinx build and got the attached html. As you can see there is a problem: the end of the html has a load of raw stuff from the nb printed out.

When I created the notebook, ran it and saved it I did not see this problem. Then I shut the notebook, reloaded it, ran it and saved again. At that point the error appeared. The part of the notebook at which the raw stuff begins starts after the string "Export to plot.ly" and around that point tere are a lot of escaped "s. But that is as far as I got diagnosing the problem.  this issue is separate from the plot.ly output not appearing -- I'd consider it solved if there was no plot.ly output but the mess at the bottom of the page was gone.
[error.zip](https://github.com/jupyter/nbconvert/files/1946215/error.zip)

nbconvert (5.3.1)
Sphinx (1.7.0)

### LINKED ISSUE #1562: Widgets not rendered when notebook is output/printed, garbled when widget state is saved
I am running Jupyter with Jupyterlab and Sparkmagic to use Apache Spark through Livy. When I output results as a table or plot, the widgets are either not shown at all or the output is garbled.

If "Save widget state automatically" is enabled, the output is garbled, if not, it is missing

Run the attached notebook and open the print preview

[Widget.ipynb.txt](https://github.com/jupyter/nbconvert/files/6404375/Widget.ipynb.txt)

Without saving widget state
Without saving widget state

With saving widget state
With saving widget state

**Nbconvert version: 6.0.7**

maartenbreddels:
Closes #1562
Closes #802
Related #804
Related https://github.com/spatialaudio/nbsphinx/issues/183

Having a widget like this:
```
import ipywidgets as widgets
widgets.Label("Let me bug you with a closing tag </script>")
```

left an invalid HTML document, since there was a closing script tag that the browser directly parses, escaping the forward slash works around this, see:
 * https://stackoverflow.com/questions/1659749/script-tag-in-javascript-string
 * https://stackoverflow.com/questions/1580647/json-why-are-forward-slashes-escaped

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -64,6 +64,9 @@ default_filters = {
         'get_metadata': filters.get_metadata,
         'convert_pandoc': filters.convert_pandoc,
         'json_dumps': json.dumps,
+        # browsers will parse </script>, closing a script tag early
+        # Since JSON allows escaping forward slash, this will still be parsed by JSON
+        'escape_html_script': lambda x: x.replace('</script>', '<\\/script>'),
         'strip_trailing_newline': filters.strip_trailing_newline,
         'text_base64': filters.text_base64,
 }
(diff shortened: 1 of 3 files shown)


mgeier:
It turns out that the tag name is case-insensitive and that browses also have a problem when the tag name is followed by a space (in this case it doesn't matter if a `>` is following or not.

Here's an example that makes problems:

```python
widgets.Label('using </scRIPt can be problematic')
```

See also https://github.com/spatialaudio/nbsphinx/pull/611.

Issue #1663, opened 2021-10-21
Performance: skip repeated validator runs between preprocessors


sgoedecke:
Hello! I've been digging into some slow nbconvert runs with flamegraphs, and I noticed a lot of time (~60% of the overall run) being spent on validation. That's because it runs validation between every preprocessor: https://github.com/jupyter/nbconvert/blob/main/nbconvert/exporters/exporter.py#L320. The HTML exporter has about ten preprocessors, which means we do a lot of revalidation.

For my use case (building a nbconvert rendering server), I'm not interested in which preprocessor failed validation, so I'd prefer to only run validation at the beginning and end of `_preprocess`. Patching this out locally speeds up conversion pretty significantly (by 1.5 seconds on a ~400kb notebook).

I'd like to add a env var or config option to change the behaviour of `_preprocess` to what I described above.

I can write my own exporter class that overrides `_preprocess`, but as the method changes over time it might be tough to keep it up to date.

PR #1672, opened 2021-11-12
Optionally speed up valdiation

gwincr11:
Motivation:
  - We are looking to build out a notebook viewer with an eye toward performance.
    In our flamegraph research we found that as much as 90% of the time spent in
    Notebook creation can occur in the preprocessor valdiation logic. This pr
    introduces the optional idea of optimistic validation, in which we care not
    about which preprocessor introduces a validation error but that the notebook
    ends in a valid state, since we care more about speed.

Related Issues:
  - https://github.com/jupyter/nbconvert/issues/1663

Changes:
  - Introduces a config option to remove some of the validation checks.

Questions:
  - I am new to the project and am not sure where a test should live for
    this logic, or the best way to structure such a test. If a reference
    is available I will happily added a test.

### DIFF
--- a/nbconvert/exporters/exporter.py
+++ b/nbconvert/exporters/exporter.py
@@ -61,6 +61,10 @@ class Exporter(LoggingConfigurable):
         help="Extension of the file that should be written to disk"
     ).tag(config=True)
 
+    optimistic_validation = Bool(False,
+        help = "Reduces the number of validation steps so that it only occurs after all preprocesors have run."
+    ).tag(config=True)
+
     # MIME type of the result file, for HTTP response headers.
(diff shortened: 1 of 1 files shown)


MSeal (approved):
Thanks for making a PR for the change. There isn't a great example to copy for writing a test here. I'm good with merging as is but if you would like to add a test after you'd need to mock the validate call, trigger a preprocess event with the flag on, and then count the number of calls made.

gwincr11:
Thanks for the merge @MSeal I was less wondering how to test it and more wonder where it should live and how to setup the test properly, how to call nbconvert and pass it a file for example and what file this test best fits into. Thanks again!

PR #1675, opened 2021-11-18
Add an invalid notebook error

gwincr11:
Motivation:
  - Our new service is serving a significant amount of notebooks a day. In this we are
    seeing a number of errors from Notebook rendering that really stem
    from notebooks not being properly formatted. These show up in a number
    of errors such as the error handled in this pr, an assertion error,
    we have also seen key errors and others. This pr is a bit of a proposal
    to instead introduced a common exception that indicates the notebook
    is missing some required data that is needed to render itself.

### DIFF
--- a/nbconvert/filters/markdown_mistune.py
+++ b/nbconvert/filters/markdown_mistune.py
@@ -160,7 +160,10 @@ class IPythonRenderer(mistune.Renderer):
         attachment_prefix = 'attachment:'
         if src.startswith(attachment_prefix):
             name = src[len(attachment_prefix):]
-            assert name in attachments, "missing attachment: {}".format(name)
+            
+            if not name in attachments:
+                nbconvert.InvalidNotebook.new("missing attachment: {}".format(name))
+            
             attachment = attachments[name]
             # we choose vector over raster, and lossless over lossy
             preferred_mime_types = ['image/svg+xml', 'image/png', 'image/jpeg']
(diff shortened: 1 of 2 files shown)


Carreau:
I would be happy to have a common exception with subclasses, that seem like a reasonable requests.

nbconvert/filters/markdown_mistune.py
```diff
@@ -160,7 +160,10 @@ def image(self, src, title, text):
         attachment_prefix = 'attachment:'
         if src.startswith(attachment_prefix):
             name = src[len(attachment_prefix):]
-            assert name in attachments, "missing attachment: {}".format(name)
+            
+            if not name in attachments:
+                nbconvert.InvalidNotebook.new("missing attachment: {}".format(name))
```
Carreau, line 165:
```suggestion
                raise InvalidNotebook("missing attachment: {}".format(name))
```

I think this is what you were trying to do.

Carreau:
> As we are now service a significant amount of notebooks a day,

For other readers "We" in this context is GitHub themselves.

Sorry for the delay in responding, the jupyter team is quite thin, and we don't always watch all the repos.

nbconvert/filters/markdown_mistune.py
gwincr11: Thank you ❤️

gwincr11:
Thank you @Carreau

PR #1676, opened 2021-11-19
Adding styling to prevent page brake inside output area

mnowacki-b:
Currently when a PDF is created the output content (including widgets) can be split in half across two pages. This causes the PDF report to look rather awkward. With this change the output area will be pushed to a new page whenever the split is avoidable.

cc: @SylvainCorlay

### DIFF
--- a/share/jupyter/nbconvert/templates/lab/index.html.j2
+++ b/share/jupyter/nbconvert/templates/lab/index.html.j2
@@ -57,9 +57,24 @@ a.anchor-link {
     overflow: hidden;
 }
 
+/* Output area styling */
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Thanks! This is solving a longstanding issue with webpdf export.

SylvainCorlay:
This should definitely get into nbconvert. I am wondering if these CSS rules should be in the notebook lab CSS eventually.

jtpio:
Yes I think this can be opened as a PR on the JupyterLab repo. This way we can get the CI check to run and make sure there is no major issue there.

PR #1679, opened 2021-11-23
Prevent page breaks in outputs when printing

SylvainCorlay:
Alternative to #1676, based on the move from flexbox to table display implemented in https://github.com/jupyterlab/jupyterlab/pull/11508, which passes the visual regression tests.

I think that using `table` is cleaner than using `inline-box` which has other side effects (like adding vertical space between lines unless setting the font size to 0).

Closes #1676.


SylvainCorlay:
This is ready to go, especially as https://github.com/jupyterlab/jupyterlab/pull/11508 is now merged.

Issue #1649, opened 2021-09-22
convertion to latex \longtable content has problem


bon-netizen:
when inside table has long text content, generated latex file has problem and cannot further use it.
finally have to change the table to bullet points then can pass.

! Undefined control sequence.
<argument> @{} >{\raggedright \arraybackslash
                                              }p{(\columnwidth - 2\tabcolsep...
l.417 ...width - 2\tabcolsep) * \real{0.5806}}@{}}

?

tovrstra:
I noticed a similar issue. I my case, `\usepackage{tabularx}`  was missing from the preamble. Fixing that manually resulted in another error, `Missing number, treated as zero`. Can you try this?

Issue #1566, opened 2021-05-11
Pandoc 2.11.2 requires calc LaTeX package


buergi:
During export of jupyter notebook, I got loads of errors of the following kind

    notebook.tex:798: Missing number, treated as zero
    notebook.tex:798: Illegal unit of measure (
    notebook.tex:799: Missing number, treated as zero
    notebook.tex:799: Illegal unit of measure (
    notebook.tex:803: Missing number, treated as zero
    notebook.tex:803: Illegal unit of measure (

The notebook needs to contain markdown table of a certain width such as

    | AAAAAAAAAAAAAAAAA | BBBBBBBBBBBBBBBBB | CCCCCCCCCCCCCC | DDDDDDDDDDDDDD | EEEEEEEEEEEEEEEE | 
    | --- | --- | --- |--- |--- |

when exporting this notebook as PDF (`File > Download as > PDF via LaTeX (.pdf)`) the described errors appear.

The problem was introduced with Pandoc 2.11.2 with this commit https://github.com/jgm/pandoc/commit/bd7c9eb32be7c5499f4a50cb4b5f138f9eb919e8
To fix it we need to add `\usepackage{calc}` to [base.tex.j2](https://github.com/jupyter/nbconvert/blob/master/share/jupyter/nbconvert/templates/latex/base.tex.j2), ideally after the `\usepackage{booktabs}` line.

**Nbconvert version:** 6.0.7

ivanov:
closed by #1686

Issue #1536, opened 2021-03-26
Conversion to PDF fails with particular Markdown table


sgdecker:
Attempting to generate a PDF from JupyterLab (or jupyter nbconvert at the command line) fails. I've narrowed it down to a cell containing a Markdown table.

To reproduce, open the attached notebook file in JupyterLab, and attempt to export to PDF.

This file demonstrates the error:
[Bug.txt](https://github.com/jupyter/nbconvert/files/6214220/Bug.txt)

**Nbconvert version:** 6.0.7

The error message is:
```

500 : Internal Server Error

The error was:

nbconvert failed: PDF creating failed, captured latex output:
Failed to run "['xelatex', 'notebook.tex', '-quiet']" command:
This is XeTeX, Version 3.14159265-2.6-0.999992 (TeX Live 2020) (preloaded format=xelatex)
 restricted \write18 enabled.
entering extended mode
(./notebook.tex
LaTeX2e <2020-02-02> patch level 5
L3 programming layer <2020-07-17>
(/home/decker/local/texlive/2020/texmf-dist/tex/latex/base/article.cls
Document Class: article 2019/12/20 v1.4l Standard LaTeX document class
(/home/decker/local/texlive/2020/texmf-dist/tex/latex/base/size11.clo))
(/home/decker/local/texlive/2020/texmf-dist/tex/latex/tcolorbox/tcolorbox.sty
(/home/decker/local/texlive/2020/texmf-dist/tex/latex/pgf/basiclayer/pgf.sty
(/home/decker/local/texlive/2020/texmf-dist/tex/latex/pgf/utilities/pgfrcs.sty
(/home/decker/local/texlive/2020/texmf-dist/tex/generic/pgf/utilities/pgfutil-c
ommon.tex
(/home/decker/local/texlive/2020/texmf-dist/tex/generic/pgf/utilities/pgfutil-c
ommon-lists.tex))
(/home/decker/local/texlive/2020/texmf-dist/tex/generic/pgf/utilities/pgfutil-l
atex.def (/home/decker/local/texlive/2020/texmf-dist/tex/latex/ms/everyshi.sty)
)
(/home/decker/local/texlive/2020/texmf-dist/tex/generic/pgf/utilities/pgfrcs.co
de.tex
(/home/decker/local/texlive/2020/texmf-dist/tex/generic/pgf/pgf.revision.tex)))

[...]

PR #1686, opened 2021-12-07
Fix recent pandoc latex tables by adding calc and array (#1536, #1566)

cgevans:
Pandoc >= 2.11.1 requires the calc package, and pandoc >= 2.11.3
requires the array package, for larger tables (taken from git blame
for pandoc's templates).  This commits adds the two packages in
an order similar to that in pandoc's templates.

This should fix both #1536 and #1566.

### DIFF
--- a/share/jupyter/nbconvert/templates/latex/base.tex.j2
+++ b/share/jupyter/nbconvert/templates/latex/base.tex.j2
@@ -77,6 +77,8 @@ override this.-=))
     \usepackage{titling}
     \usepackage{longtable} % longtable support required by pandoc >1.10
     \usepackage{booktabs}  % table support for pandoc > 1.12.2
+    \usepackage{array}     % table support for pandoc >= 2.11.3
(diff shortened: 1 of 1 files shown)


cgevans:
Also fixes #1649.

SylvainCorlay:
cc @ivanov you may be more familiar with that part of the code base - wrt pandoc.

ivanov (approved):
Thanks for this @cgevans (and @SylvainCorlay for the ping).

This change looks good, https://github.com/jgm/pandoc/commit/bd7c9eb32be7c5499f4a50cb4b5f138f9eb919e8 and https://github.com/jgm/pandoc/commit/7c4d7db9c7b582c7eb574c8ef9381f41cd419a6e are the relevant commits on the `pandoc` side.

I just kicked the tests because one of them failed due to some transient network issue.

SylvainCorlay:
Thanks @ivanov! (Btw @cgevans opened a couple of other pandoc-related PRs).

PR #1687, opened 2021-12-07
Modernize latex greek math handling (partially fixes #1673)

### LINKED ISSUE #1673: Convert to HTML is fine, but Convert to PDF ignores Greek letters
This has been discussed many time here and on other forums, but I still don't know how to get Jupyter to print Greek letters when I convert the notebook to PDF. Interestingly, if converted to HTML, everything is fine.

When converted to PDF:
image

When converted to HTML:
image

I've tried using CLI pandoc and tweaking the settings (e.g., changing the font or using xelatex instead of pdflatex), but nothing seems to work. And since these Greek letters are part of my code, I can't just replace them with something that's known to Latex.

cgevans:
The current latex base template uses the ucs package with mathletters,
but as it does not also set utf8x on inputenc, the ucs package does
not work as intended, and Greek letters do not appear in math.

As utf8x and ucs are somewhat outdated and problematic, this PR uses
a few conditions to try to enable Greek letters in both math and text.

For pdflatex, if the alphabeta package (part of greek-inputenc, and
not necessarily installed) is present, this is used, as it better
supports Greek math and text, including diacritics.  It it is not,
ucs with mathletters is used, which will allow math, and some text,
though diacritics will cause an error.

For xelatex and lualatex, this instead uses unicode-math, which
handles math well but does not support pdflatex.  While this
arrangement does support text, note that by default, the fonts used
are unlikely to include support for Greek text, and the glyphs will be
omitted (both for text and for monospace text), particularly since
Latin Modern does not support Greek text.  I have an additional change
which uses Computer Modern Unicode if available, but I'd like to keep
these separate, as it's unclear to me that setting the font in this
template would be a good idea.

Mathpazo for pdflatex is removed for consistency with xelatex/lualatex
output, and the position of the conditional is moved to avoid an order
conflict with the ams packages; this more closely matches the position
of unicode-math in pandoc's template.

### DIFF
--- a/share/jupyter/nbconvert/templates/latex/base.tex.j2
+++ b/share/jupyter/nbconvert/templates/latex/base.tex.j2
@@ -14,13 +14,6 @@ override this.-=))
     ((* block docclass *))\documentclass[11pt]{article}((* endblock docclass *))
 
     ((* block packages *))
-    \usepackage{iftex}
-    \ifPDFTeX
-    	\usepackage[T1]{fontenc}
-    	\usepackage{mathpazo}
-    \else
-    	\usepackage{fontspec}
-    \fi
 
     % Basic figure setup, for now with no caption control since it's done
     % automatically by Pandoc (which extracts ![](path) syntax from Markdown).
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
I have no objections to this.
cc @ivanov FYI

blink1073 (approved):
Thanks!

PR #1688, opened 2021-12-07
Set cm-unicode font for latex if available (fixes #1673)

### LINKED ISSUE #1673: Convert to HTML is fine, but Convert to PDF ignores Greek letters

cgevans:
This includes PR #1686 and depends on it, but makes potentially more problematic changes.

For LuaLaTeX and XeLaTeX, the default fonts usually don't support
Greek text, causing problems when Greek letters are used outside of
math mode (eg, as text or as symbols in code).  This change to the
base template uses the cm-unicode fonts, if they are available.
Fonts are left unchanged if cm-unicode is not available, in which
case Greek letters in math will still work, but text will be missing.

Loading is done here using cm-unicode's filenames, rather than the
fonts' symbolic names, because XeLaTeX cannot search for symbolic
names of fonts that are not available to the system (but may be
available to LaTeX).  LuaLaTeX does not have this limitation, but can
also load using filenames.

### DIFF
--- a/share/jupyter/nbconvert/templates/latex/base.tex.j2
+++ b/share/jupyter/nbconvert/templates/latex/base.tex.j2
@@ -54,6 +54,28 @@ override this.-=))
     \else
         \usepackage{fontspec}
         \usepackage{unicode-math}
+        % Try to load cm-unicode, using filenames
+        % to support both XeLaTeX and LuaLaTeX.
+        \IfFontExistsTF{cmunrm.otf}{
+        \setmainfont{cmun}[
+               Extension=.otf,
+               UprightFont=*rm,
+               ItalicFont=*ti,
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
cc @ivanov FYI

SylvainCorlay:
cc @ivanov since you reviewed other related changes lately.

PR #1691, opened 2021-12-16
Updating to MathJax 3

gwincr11:
Motivation:
  - This pr starts to update the nbconvert app to use MathJax 3.
    MathJax 3 does not use inline script execution to pull in its
    libraries which improves the security story for using this library.

Notes:

This needs some solid testing, I have yet to build a bunch of notebooks with it, although the config is based on the MathJax setup we are using at GitHub for Notebook viewing. We hope to upstream this so it can be tweaked and owned by the community and not us.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ docs/source/*.j2
 docs/source/example.html
 docs/source/mytemplate/*
 docs/source/config_options.rst
+.venv/
 
 # Eclipse pollutes the filesystem
 .project
--- a/nbconvert/exporters/html.py
+++ b/nbconvert/exporters/html.py
@@ -52,7 +52,7 @@ class HTMLExporter(TemplateExporter):
     ).tag(config=True)
 
     mathjax_url = Unicode(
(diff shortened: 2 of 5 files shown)


gwincr11:
I temporarily update our app to use this branch and everything looks good. But you probably want to kick the tires yourself a bit 😅

SylvainCorlay:
Thanks! Will try to review and get this in for 6.5.

bollwyvl:
@gwincr11 this might be a good match for the recently-initiated `7.x` line on #1771: would it be possible to rebase or otherwise merge the changes on top of `main`?

krassowski:
Of note, we are upgrading to MathJax 3 on JupyterLab side too: https://github.com/jupyterlab/jupyterlab/pull/13877. It would be nice to land both PRs at similar time for consistency.

JeppeKlitgaard:
Note that MathJax 4 will ship soon 🤞, thought this release should be much more backwards compatible than the 2 → 3 transition.

PR #1701, opened 2022-01-10
Run black on the code

martinRenou:
Ran `black nbconvert` on the codebase and added a Github action check


SylvainCorlay:
This seems to conflict with the most recent changes.

martinRenou:
Rebased :)

martinRenou:
We should probably use darker instead (like IPython does), which makes the transition smoother

Issue #1216, opened 2020-03-16
nbconvert doesn't use theme


beasteers:
Nbconvert doesn't export using JupyterLab custom themes.

- Enable a custom theme
- Run `Export Notebook as HTML`
- Open downloaded file and 😕

After a quick dig thru the master branch, I found where the template chooses the theme styling (I think) and it's hardcoded to only work with the default light/dark themes.

https://github.com/jupyter/nbconvert/blob/master/share/jupyter/nbconvert/templates/lab/index.html.j2

I don't know how hard it would be to extend this to handle other themes, but I'd imagine that the infrastructure is mostly there seeing as you can render the html fine inside of lab.

**Nbconvert version:** 5.6.1

Update: actually it doesn't work for (builtin) dark mode either. It just looks like old jupyter notebook styling. Hmm.

SylvainCorlay:
@beasteers we have not tagged a release supporting JupyterLab themes.

With nbconvert 6.0, it will be possible to produce a JupyterLab-style HTML export, with support for dark and light theme.

We have no specific plan for custom themes yet.

redNixon:
@SylvainCorlay 
The [docs](https://nbconvert.readthedocs.io/en/latest/config_options.html#exporter-options) seem to have suggested that it is already possible to configure a JupyterLab CSS theme by this commit(I think): https://github.com/jupyter/nbconvert/commit/f4c7daefe0e24055a92f185d6fd803abf909db2f 

>HTMLExporter.theme : Unicode
>Default: 'light'
>
>Template specific theme(e.g. the JupyterLab CSS theme for the lab template)

Or at least that is what I understood that config option's help text to mean when I read it before spending time trying to debug why it was not applying.  Would you say that this is a documentation bug since switching to "dark" is not supported yet or I am completely off the mark and this is documenting something complete different than what this issue is requesting?

SylvainCorlay:
@redNixon this link to "latest" documentation reflects the state kf the master branch but there is no release with that feature yet.

martinRenou:
First attempt to fixing this issue: https://github.com/jupyter/nbconvert/pull/1703

PR #1703, opened 2022-01-11
HTMLExporter: Add basic support for federated labextensions themes

martinRenou:
Allow to use labextension themes when using the lab template.

The API is the following:

```bash
jupyter nbconvert --to html --HTMLExporter.template_name=lab --HTMLExporter.theme=@oriolmirosa/jupyterlab_materialdarker Notebook.ipynb
```

The following also works (omitting npm package namespace):

```bash
jupyter nbconvert --to html --HTMLExporter.template_name=lab --HTMLExporter.theme=jupyterlab_materialdarker Notebook.ipynb
```

It will fail with a `ValueError` if the theme is not found:
```
ValueError: Could not find lab theme unknown_theme
```

References:
-----------

Fix https://github.com/jupyter/nbconvert/issues/1216

Examples:
---------

https://github.com/AllanChain/jupyterlab-theme-solarized-dark

solarized

https://github.com/oriolmirosa/jupyterlab_materialdarker

material

https://github.com/timkpaine/jupyterlab_miami_nights

miami

### DIFF
--- a/nbconvert/exporters/html.py
+++ b/nbconvert/exporters/html.py
@@ -6,7 +6,9 @@
 
 import os
 import mimetypes
+import json
 import base64
+from pathlib import Path
 
 from traitlets import default, Unicode, Bool
 from traitlets.config import Config
@@ -25,6 +27,58 @@ from nbconvert.filters.widgetsdatatypefilter import WidgetsDataTypeFilter
 from nbconvert.filters.markdown_mistune import IPythonRenderer, MarkdownWithMath
 
 from .templateexporter import TemplateExporter
+
+
+def find_lab_theme(theme_name):
+    """
+    Find a JupyterLab theme location by name.
+
+    Parameters
+    ----------
+    theme_name : str
+        The name of the labextension theme you want to find.
+
+    Raises
+    ------
+    ValueError
+        If the theme was not found, or if it was not specific enough.
+
+    Returns
+    -------
+    theme_name: str
+        Full theme name (with scope, if any)
+    labextension_path : Path
+        The path to the found labextension on the system.
+    """
+    paths = jupyter_path('labextensions')
+
+    matching_themes = []
+    theme_path = None
+    for path in paths:
+        for (dirpath, dirnames, filenames) in os.walk(path):
+            # If it's a federated labextension that contains themes
+            if 'package.json' in filenames and 'themes' in dirnames:
+                # TODO Find the theme name in the JS code instead?
+                # TODO Find if it's a light or dark theme?
+                with open(Path(dirpath) / 'package.json', 'r', encoding="utf-8") as fobj:
+                    labext_name = json.loads(fobj.read())['name']
+
+                if labext_name == theme_name or theme_name in labext_name.split('/'):
+                    matching_themes.append(labext_name)
+
+                    full_theme_name = labext_name
+                    theme_path = Path(dirpath) / 'themes' / labext_name
+
+    if len(matching_themes) == 0:
+        raise ValueError(f'Could not find lab theme "{theme_name}"')
+
+    if len(matching_themes) > 1:
+        raise ValueError(
+            f'Found multiple themes matching "{theme_name}": {matching_themes}. '
+            'Please be more specific about which theme you want to use.'
+        )
+
+    return full_theme_name, theme_path
 
 
 class HTMLExporter(TemplateExporter):
@@ -89,8 +143,9 @@ class HTMLExporter(TemplateExporter):
     def _template_name_default(self):
         return 'lab'
 
-    theme = Unicode('light',
-                    help='Template specific theme(e.g. the JupyterLab CSS theme for the lab template)'
+    theme = Unicode(
+        'light',
+        help="Template specific theme(e.g. the name of a JupyterLab CSS theme distributed as prebuilt extension for the lab template)"
     ).tag(config=True)
 
     output_mimetype = 'text/html'
@@ -150,6 +205,30 @@ class HTMLExporter(TemplateExporter):
             code = """<style type="text/css">\n%s</style>""" % (env.loader.get_source(env, name)[0])
             return jinja2.Markup(code)
 
+        def resources_include_lab_theme(name):
+            # Try to find the theme with the given name, looking through the labextensions
+            _, theme_path = find_lab_theme(name)
+
+            with open(theme_path / 'index.css', 'r') as file:
+                data = file.read()
+
+            # Embed assets (fonts, images...)
+            for asset in os.listdir(theme_path):
+                local_url = "url({})".format(Path(asset).as_posix())
+
+                if local_url in data:
+                    mime_type = mimetypes.guess_type(asset)[0]
+
+                    # Replace asset url by a base64 dataurl
+                    with open(theme_path / asset, 'rb') as assetfile:
+                        base64_data = base64.b64encode(assetfile.read())
+                        base64_data = base64_data.replace(b'\n', b'').decode('ascii')
+
+                        data = data.replace(local_url, 'url(data:{};base64,{})'.format(mime_type, base64_data))
+
+            code = """<style type="text/css">\n%s</style>""" % data
+            return jinja2.Markup(code)
+
         def resources_include_js(name):
             env = self.environment
             code = """<script>\n%s</script>""" % (env.loader.get_source(env, name)[0])
@@ -178,9 +257,11 @@ class HTMLExporter(TemplateExporter):
             data = data.replace(b'\n', b'').decode('ascii')
             src = 'data:{mime_type};base64,{data}'.format(mime_type=mime_type, data=data)
             return jinja2.Markup(src)
+
         resources = super()._init_resources(resources)
         resources['theme'] = self.theme
         resources['include_css'] = resources_include_css
+        resources['include_lab_theme'] = resources_include_lab_theme
         resources['include_js'] = resources_include_js
         resources['include_url'] = resources_include_url
         resources['require_js_url'] = self.require_js_url
--- a/share/jupyter/nbconvert/templates/lab/index.html.j2
+++ b/share/jupyter/nbconvert/templates/lab/index.html.j2
@@ -37,8 +37,10 @@
 {{ resources.include_css("static/index.css") }}
 {% if resources.theme == 'dark' %}
     {{ resources.include_css("static/theme-dark.css") }}
-{% else %}
+{% elif resources.theme == 'light'  %}
     {{ resources.include_css("static/theme-light.css") }}
+{% else %}
+    {{ resources.include_lab_theme(resources.theme) }}
 {% endif %}
 <style type="text/css">
 /* Misc */


martinRenou:
cc. @jtpio you might have ideas/comments on the implementation

martinRenou:
For reference, this works nicely with Voila as well (for free with this PR)

```bash
voila notebooks/basics.ipynb --theme=jupyterlab-theme-solarized-dark
```

voila

jtpio:
Thanks, this is looking great!

nbconvert/exporters/html.py
```diff
@@ -150,6 +152,28 @@ def resources_include_css(name):
...
             return jinja2.Markup(code)
 
+        def include_lab_theme(name):
+            # Try to find the theme with the given name, looking through the labextensions
+            paths = jupyter_path('labextensions')
+
+            for path in paths:
+                for (dirpath, dirnames, filenames) in os.walk(path):
+                    # If it's a federated labextension that contains themes
+                    if 'package.json' in filenames and 'themes' in dirnames:
```
jtpio, line 162: Not sure yet whether this would be better, but an alternative could be to look for the `themePath` field in the `package.json`: 

https://github.com/jupyterlab/jupyterlab/blob/6911da2af507a73add0ae79469d806ea85118662/builder/metadata_schema.json#L76-L81

martinRenou: Indeed, I wasn't sure if I needed to use that. Isn't the `themePath` a relative path in the development repository structure? Does it still make sense in the generated federated extension?

jtpio: Right, it's probably better to stick to the approach this current PR proposes.

Which should be similar to what `jupyterlab_server` does: https://github.com/jupyterlab/jupyterlab_server/blob/main/jupyterlab_server/themes_handler.py

martinRenou: Thanks for the link! I assumed `index.css` would always be the entry point in the PR, is that a good assumption?

martinRenou:
There are still some things to be done on this PR:

- How do we handle custom fonts/images that the CSS points to?
- Is the `index.css` always the entry point for the theme?
- How do we fill the `data-jp-theme-light` property on the `body` tag? For that we will need to evaluate the `isLight` property of the theme extension, but this information lays in the JavaScript code of the theme, that we do not use in this PR.

nbconvert/exporters/html.py
jtpio: Probably this could be a different name? Although in the theme cookiecutter it's true it defaults to `index.css`: https://github.com/jupyterlab/theme-cookiecutter/blob/e95554b339bec0b4d44ee265a9d861f27e7db276/%7B%7Bcookiecutter.python_name%7D%7D/src/index.ts#L17

martinRenou:
@jtpio I updated the PR so that it embeds the fonts/images as base64 in the CSS.

martinRenou:
This PR should be ready @SylvainCorlay @jtpio

bollwyvl:
> omitting npm package namespace

what happens if there is a conflict? Perhaps a sane priority might be:
- an un-prefixed theme just wins (even if there are others)
- if no un-prefixed theme is found, and there are two themes with the same name but different `@org`, raise a `ValueError`

martinRenou:
@bollwyvl thanks for your review!

Yes that makes sense, currently it will pick the first one it finds, but I agree raising a `ValueError` is better so that the user understands what's going on.

bollwyvl:
> embeds the fonts/images as base64 in the CSS.

so... does it somehow magically collapse `@import` as well? i've done some fairly evil things to get to portable, single-file solutions, but it _hurts_... perhaps this could be a follow-on, e.g. `--HTMLExporter.embed_theme=True`


>  thanks for your review!

goes without saying i want this real bad... i tried my hand at bootstrap themes back in the day, and i liked the relative simplicity, but fully bought into the lab toolchain this will be great.

Actually, perhaps it is worth pointing out more explicitly that this _doesn't_ work with bootstrap themes, e.g. `HTMLExporter.lab_theme`?

martinRenou:
> so... does it somehow magically collapse @import as well?

From what I understood, the JupyterLab builder would already have collapsed the `@import` statements and made it a single CSS bundle.

I am not touching them in this PR, if there are still some `@import` statements they might fail.

The only things I touch are the `url(*)` calls, by replacing the files urls with base64 dataurl:
```css
@font-face {
    src: url(whatever.ttf);
}
.css-class {
   background: url(whatever.png);
}
```

> i've done some fairly evil things to get to portable, single-file solutions, but it hurts

Yeah I didn't want to go too deep through that path of trying to make my own "CSS blender"... I guess turning the urls into base64 urls is fairly simple though and might not fail that often?

> perhaps this could be a follow-on, e.g. --HTMLExporter.embed_theme=True

We actually have to embed the theme if we want the resulting HTML file to be portable, don't we?
Are you suggesting we don't embed the theme by default?

> Actually, perhaps it is worth pointing out more explicitly that this doesn't work with bootstrap themes, e.g. HTMLExporter.lab_theme?

Do you mean that this option should be renamed from `theme` to `lab_theme`? 

nbconvert advertises the `theme` option as being "template specific" https://github.com/jupyter/nbconvert/blob/main/nbconvert/exporters/html.py#L93. Other templates could very well implement their custom support for this `theme` option.

The only template provided by `nbconvert` by default, that supports the `theme` option, is the `lab` template, but other templates could make use of this `theme` option (I'm thinking about [this retro template for example](https://github.com/martinRenou/voila-retro, although it's a bad example because it inherits from the `lab` one)).

bollwyvl:
super sorry for the uninformed drive-by review! still haven't even cracked open the code, and will do so now.

> nbconvert advertises the theme option as being "template specific"

great!

> We actually have to embed the theme

great!

> made it a single CSS bundle.

great!

nbconvert/exporters/html.py
```diff
@@ -150,6 +152,44 @@ def resources_include_css(name):
...
+                        # TODO Find the theme name in the JS code instead?
+                        # TODO Find if it's a light or dark theme?
+                        with open(os.path.join(dirpath, 'package.json'), 'r') as fobj:
```
bollwyvl, line 165: seems like we could go ahead and start using pathlib on new work...

nbconvert/exporters/html.py
```diff
@@ -150,6 +152,44 @@ def resources_include_css(name):
...
+
+                        if labext_name == name or name in labext_name.split('/'):
+                            theme_path = os.path.join(dirpath, 'themes', labext_name)
+
+                            with open(os.path.join(theme_path, 'index.css'), 'r') as file:
+                                data = file.read()
+
+                            # Embed assets (fonts, images...)
+                            for asset in os.listdir(theme_path):
+                                local_url = "url({})".format(asset)
```
bollwyvl, line 176: pathlib's `Path.as_posix` would definitely be helpful here on e.g. windows

martinRenou:
No worries at all. It's all good questions and remarks :)

bollwyvl:
> if there are still some @import statements they might fail.

Well, as a worst-case, degenerate theme, i present the overly-complicated [jupyterlab-gt-coar-theme](https://github.com/gt-coar/jupyterlab-gt-coar-theme/). Though looking at the built content, it doesn't seem to have any dangling ones... but it does a lot of crazy things.

martinRenou:
> Well, as a worst-case, degenerate theme

Exactly

Thanks for the link! I'll try it with this PR

martinRenou:
Looks super cool

coar

bollwyvl:
Thanks! Yeah, it's no longer on-brand, apparently (they went to all proprietary fonts :crying_cat_face: ) but it was fun to build, and I have shipped it a few times.

So I do wonder about some more switches and doo-dads for:

- presentation mode
- single-document mode

Both themes shipped in that package are [sensitive to](https://github.com/gt-coar/jupyterlab-gt-coar-theme/blob/master/packages/brand/style/presentation.css)... i'm not necessarily saying we needs traitlets for those two values (when would you stop?), but it actually seems pretty relevant in this case.

martinRenou:
> presentation mode

Maybe a custom template could try to replicate the DOM structure?

martinRenou:
I took your comments into account. It will now fail with:

```
ValueError: Found multiple themes matching "jupyterlab_miami_nights": ['@gt-coar/jupyterlab_miami_nights', '@timkpaine/jupyterlab_miami_nights']. Please be more specific about which theme you want to use.
```

In case of conflict.

nbconvert/exporters/html.py
```diff
@@ -89,8 +140,9 @@ def _file_extension_default(self):
     def _template_name_default(self):
         return 'lab'
 
-    theme = Unicode('light',
-                    help='Template specific theme(e.g. the JupyterLab CSS theme for the lab template)'
+    theme = Unicode(
+        'light',
+        help="the JupyterLab CSS theme for the lab template"
```
jtpio, line 145: Maybe we should also mention somewhere that only JupyterLab themes distributed as prebuilt extensions will be compatible?

martinRenou: Indeed! I shouldn't have changed this comment that way though, I will revert this change and specify that it supports prebuilt extensions

nbconvert/exporters/html.py
```diff
@@ -27,6 +29,58 @@
...
+
+    Raises
+    ------
+    ValueError
+        If the theme was not found, or if it was not specific enough.
+
+    Returns
+    -------
+    theme_name: str
+        Full theme name (with organization any, if any)
```
jtpio, line 49: Looks like a minor typo?

```suggestion
        Full theme name (with organization, if any)
```

jtpio: We could also say `scope` as an alternative to `organization`

jtpio:
Thanks Martin this looks great!

nbconvert/exporters/html.py
```diff
@@ -27,6 +29,58 @@
...
+    paths = jupyter_path('labextensions')
+
+    matching_themes = []
+    theme_path = None
+    for path in paths:
+        for (dirpath, dirnames, filenames) in os.walk(path):
+            # If it's a federated labextension that contains themes
+            if 'package.json' in filenames and 'themes' in dirnames:
+                # TODO Find the theme name in the JS code instead?
+                # TODO Find if it's a light or dark theme?
```
jtpio, line 62: I think it would be fine to leave these TODOs here for now, so this PR can be merged and iterated on later :+1:

nbconvert/exporters/html.py
martinRenou: Thanks!

nbconvert/exporters/html.py
martinRenou: Yes I guess that's the best we can do for now :)

martinRenou:
I'm still iterating a tiny bit on it (making sure it works nicely with Voila)

martinRenou:
Paired PR in Voila: https://github.com/voila-dashboards/voila/pull/1089

I guess this one is ready :) Thank you for your review!

jtpio (approved):
Thanks!

SylvainCorlay (approved):

SylvainCorlay:
Really happy with this getting in! Merging.

PR #1711, opened 2022-01-26
Raise pyppeteer requirement to >=1,<1.1

### LINKED ISSUE #1707: evaluate upgrading to pyppeteer 1.x
## Elevator pitch

Raise the pin on pyppeteer to `>=1,<1.1` from  '==0.2.6'.

## Motivation

A few [1.x releases](https://pypi.org/project/pyppeteer/#history) have now gone out, and some time has elapsed to shake down any additional bugs. While a `1.x` doesn't really mean anything, it seems like giving a package the benefit of the doubt that this will be a more stable API seems fair.

Especially if a major rev of `nbconvert` is on the horizon to support major revs of heavyweights like mistune (#1685) and mathjax (#1706), it would be good to get this one updated as well.

## Additional context
Among the important things: _at least_ its provisioning is no longer insecure-by-default, and requires up-to-date certificates. Its default chromium is still `588429`, with claimed compatibility with puppeteer `v1.6.0`... but those haven't been updated in a _long_ time. No telling what state it's actually in.

SylvainCorlay:
Fixes #1707.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -227,7 +227,7 @@ setup_args['install_requires'] = [
     'nbclient>=0.5.0,<0.6.0'
 ]
 
-pyppeteer_req = 'pyppeteer==0.2.6'
+pyppeteer_req = 'pyppeteer>=1,<1.1'
 
 extra_requirements = {
     'test': [


bollwyvl:
Cool. I think this is ready to go, too: https://github.com/conda-forge/pyppeteer-feedstock/pull/11

SylvainCorlay:
I was about to tag a patch release including the DOM fixes in the lab template. What do you think about that one?

bollwyvl:
...though i do think this would prompt at least a 6.5.0 release...

bollwyvl:
i mean, i guess it's an `extra`... so maybe patch ok?

SylvainCorlay:
> ...though i do think this would prompt at least a 6.5.0 release...
> i mean, i guess it's an extra... so maybe patch ok?

exactly what I was wondering.

SylvainCorlay:
I pushed 6.4.1 out with the minuor fixes, and merging this one now to be published with 6.5.0.

PR #1712, opened 2022-01-27
Always hide the collapser element

martinRenou:
This element is unlikely to be used in nbconvert anyway?

We could probably use it to allow collapsing cells just like in JupyterLab. But this is not implemented yet, so it's probably best to not show it for now.

# Before this PR:

collapser

# After this PR:

nocollapser


martinRenou:
I opened an issue tracking the idea of implementing the collapser behavior: https://github.com/jupyter/nbconvert/issues/1713

SylvainCorlay:
LGTM!

PR #1717, opened 2022-02-04
HTMLExporter: Add embed_images option for embedding images from Markdown cells as base64

martinRenou:


nbconvert/filters/markdown_mistune.py
```diff
@@ -28,6 +31,10 @@
 from nbconvert.filters.strings import add_anchor
 
 
+class InvalidNotebook(Exception):
```
martinRenou, line 34: This un unrelated to the PR itself, `InvalidNotebook` was used in this file but never imported. Importing it was resulting in a circular import so we need to define it here (it's only used in this file anyway)

nbconvert/nbconvertapp.py
```diff
@@ -56,6 +56,7 @@ def validate(self, obj, value):
     'template' : 'TemplateExporter.template_name',
     'template-file' : 'TemplateExporter.template_file',
     'theme' : 'HTMLExporter.theme',
+    'embed-images' : 'HTMLExporter.embed_images',
```
bollwyvl, line 59: perhaps make it a `flag`, if it defaults to `False`?

martinRenou: thanks for the review! Makes total sense, I did the change

maartenbreddels:
Related to https://github.com/voila-dashboards/voila/pull/101

PR #1718, opened 2022-02-04
Fix theming support for WebPDF exporter

martinRenou:
Fixing the WebPDFExporter, overwriting the default Chromium behavior that forces a white background for pdf exports.


martinRenou:
Opening as draft. Cells are rendered with the proper theme, but the background is still not right for some reason.

PR #1719, opened 2022-02-11
Add section to `customizing` showing how to use template inheritance

stefanv:
@SylvainCorlay @choldgraf An update to the docs to illustrate new-style template usage, as suggested :)

@carreau Since this is based on your original template from https://github.com/Carreau/jupytercontrib

Note: there are still entire sections in the docs that need reviewing for new-style templates, such as https://nbconvert.readthedocs.io/en/latest/external_exporters.html


choldgraf:
@Carreau can you add me as an admin on ReadTheDocs? Or can you enable PR previews?

choldgraf (approved):
I can't enable the readthedocs PR previews because I'm not an admin there, but the copy in general looks good to me!

SylvainCorlay:
Thanks!

PR #1721, opened 2022-02-23
HTMLExporter: Respect the embed_images flag for HTML blocks

martinRenou:
This is a follow-up of https://github.com/jupyter/nbconvert/pull/1717 which was only working for markdown images:
```
![](./my_image.jpg)
```

This PR adds support for embedding images for HTML `img` tags as well:

```
<img src='./my_image.jpg'></img>
```

It adds beautifulsoup4 as a dependency for parsing the HTML contained in Markdown cells.

---

cc. @maartenbreddels maybe you'd have an opinion on this as you've been working on embedding images in Voila?

cc. @SylvainCorlay

### DIFF
--- a/nbconvert/tests/files/notebook5_embed_images.ipynb
+++ b/nbconvert/tests/files/notebook5_embed_images.ipynb
@@ -1,0 +1,51 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "e684608d",
+   "metadata": {},
+   "source": [
+    "![](./containerized_deployments.jpeg)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "e609fcaf",
+   "metadata": {},
+   "source": [
+    "<img src='./containerized_deployments.jpeg'></img>"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a0043f9e",
+   "metadata": {},
+   "source": [
+    "<div>\n",
+    "    <img src='./containerized_deployments.jpeg'></img>\n",
+    "</div>"
+   ]
+  }
+ ],
+ "metadata": {
+  "kernelspec": {
+   "display_name": "Python 3 (ipykernel)",
+   "language": "python",
+   "name": "python3"
+  },
+  "language_info": {
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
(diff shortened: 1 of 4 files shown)


nbconvert/tests/files/notebook5_embed_images.ipynb
```diff
@@ -0,0 +1,51 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a0043f9e",
+   "metadata": {},
+   "source": [
+    "<div>\n",
+    "    <img src='./containerized_deployments.jpeg'></img>\n",
+    "</div>"
```
martinRenou, line 26: This test will trigger the `block_html` method from the `IPythonRenderer`

nbconvert/tests/files/notebook5_embed_images.ipynb
```diff
@@ -0,0 +1,51 @@
...
+   "source": [
+    "![](./containerized_deployments.jpeg)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "e609fcaf",
+   "metadata": {},
+   "source": [
+    "<img src='./containerized_deployments.jpeg'></img>"
```
martinRenou, line 16: This test will trigger the `inline_html` method from the `IPythonRenderer`

martinRenou:
Thanks for reviewing! 

> I think that 101 PR will only work for Markdown images, which this PR does not implement right?

Yes, the PR here is for HTML `img` tags, which I believe Voila does not embed but serve them? The Markdown images logic was added in a separate PR: https://github.com/jupyter/nbconvert/pull/1717

> Looks good, but maybe also an opportunity to upstream

I was actually looking at the Voila logic for making those PRs :) But I think the logic in Voila cannot be upstreamed:

The Voila logic is kind of different, as we have a content-manager there for finding the files. Also, I was wondering if Voila could overwrite this embedding logic here so that it properly serves the images (like we've done for theming, where nbconvert embeds the theme, and Voila serves it with the proper endpoints).

SylvainCorlay:
> Also, I was wondering if Voila could overwrite this embedding logic here so that it properly serves the images (like we've done for theming, where nbconvert embeds the theme, and Voila serves it with the proper endpoints).

Yes, I think this is an amazing idea.

PR #1723, opened 2022-03-07
Drop dependency on testpath.

anntzer:
Instead, backport the tiny chdir contextmanager from CPython to replace TemporaryWorkingDirectory.

Followup to #1596


blink1073:
I think it would be cleaner to add the `TemporaryWorkingDirectory` class directly to a local util rather than use a more complicated context manager.

blink1073:
Fair enough, no strong objection.  Thanks!

PR #1737, opened 2022-03-24
Import Markup and escape directly from MarkupsSafe

### LINKED ISSUE #1736: [BUG] module 'jinja2.utils' has no attribute 'escape'
Today released `jinja2` version has deprecated `jinja2.utils.escape`. 
See here: https://github.com/pallets/jinja/issues/1626

Not directly, it raises on `readthedocs` when building docs from notebooks:
https://readthedocs.org/projects/sktime/builds/16457301/

**Nbconvert version:**
latest

rwedge:
Fixes #1736

Jinja 3.1.0 removed `jinja.Markup` and `jinja.escape` and suggested importing them from MarkupSafe

### DIFF
--- a/nbconvert/exporters/html.py
+++ b/nbconvert/exporters/html.py
@@ -14,6 +14,7 @@ from traitlets import default, Unicode, Bool
 from traitlets.config import Config
 from jupyter_core.paths import jupyter_path
 import jinja2
+import markupsafe
 
 if tuple(int(x) for x in jinja2.__version__.split(".")[:3]) < (3, 0, 0):
     from jinja2 import contextfilter
@@ -213,7 +214,7 @@ class HTMLExporter(TemplateExporter):
         def resources_include_css(name):
             env = self.environment
             code = """<style type="text/css">\n%s</style>""" % (env.loader.get_source(env, name)[0])
-            return jinja2.Markup(code)
+            return markupsafe.Markup(code)
 
         def resources_include_lab_theme(name):
             # Try to find the theme with the given name, looking through the labextensions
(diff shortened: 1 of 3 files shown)


minrk (approved):

rwedge:
Test failure appears related to Jinja2 3.1.0 release.  Pin Jinja2 below 3.1.0 and the tests pass.

Possibly related to this change? https://github.com/pallets/jinja/pull/1621

trungleduc:
should we bump the lower bound and add an upper bound for `jinja2` version?

rwedge:
I set an upper bound for `jinja2`.  Perhaps a different PR could investigate the windows errors?

rwedge:
@davidism could there have been a change in jinja2 behavior between 3.0 and 3.1 that would account for the new failures in these tests?

```python-traceback
    def test_absolute_template_dir(self):
        with tempdir.TemporaryDirectory() as td:
            template = 'mytemplate'
            template_file = os.path.join(td, template, 'index.py.j2')
            template_dir = os.path.dirname(template_file)
            os.mkdir(template_dir)
            test_output = 'absolute!'
            with open(template_file, 'w') as f:
                f.write(test_output)
            config = Config()
            config.TemplateExporter.template_name = template
            config.TemplateExporter.extra_template_basedirs = [td]
            exporter = self._make_exporter(config=config)
>           assert exporter.template.filename == template_file
E           AssertionError: assert 'C:\\Users\\R...e/index.py.j2' == 'C:\\Users\\R...\\index.py.j2'
E             - C:\Users\RUNNER~1\AppData\Local\Temp\tmpd72mcthi\mytemplate\index.py.j2
E             ?                                                            ^
E             + C:\Users\RUNNER~1\AppData\Local\Temp\tmpd72mcthi\mytemplate/index.py.j2
E             ?     
```
https://github.com/jupyter/nbconvert/runs/5689650683?check_suite_focus=true

davidism:
@rwedge can you open an issue on Jinja about that?

We did fix some issues with unsafe Windows paths in template names. The two filenames you're testing are in fact equivalent on Windows, even though they're not textually equivalent, since `/` is a valid path separator. A quick workaround for you would be to use `pathlib.Path()` on both values, which will normalize separators, which is what I'll do in a bugfix for Jinja as well.

davidism:
Created https://github.com/pallets/jinja/issues/1637, I will submit a PR and then release 3.1.1.

davidism:
Released Jinja2 3.1.1 with the filename fix.

SylvainCorlay:
Thanks everyone.

SylvainCorlay:
I wonder if we should exclude 3.1.0 (while accepting 3.1.1) from the versions in setup.py.

PR #1738, opened 2022-03-25
Use escape from markupsafe

### LINKED ISSUE #1736: [BUG] module 'jinja2.utils' has no attribute 'escape'

davidbrochart:
Fixes #1736


davidbrochart:
Ah yes, I didn't see it.

Issue #1740, opened 2022-03-28
Add upper bound on the Jinja and MarkupSafe dependencies


martinRenou:
Related https://github.com/jupyter/nbconvert/pull/1737

Considering [this discussion](https://github.com/pallets/jinja/issues/1626) and the fact that Jinja (and I assume MarkupSafe) do not use semver, we should probably add an upper bound on the Jinja2 and MarkupSafe dependencies. Something like the following could probably be good in order to not get bad surprises?
```
'jinja2>=2.4,<3.2',
'MarkupSafe>=2.0,<2.2'
```

akchinSTC:
Just piggy backing on this topic, it seems possible ?...
if given an environment with Jinja2<3.0 installed,
...for pip to resolve to using MarkupSafe >= 2.1  when using the latest release of nbconvert. If that happens, I believe there will be an import error here: 
https://github.com/pallets/jinja/blob/cf215390d4a4d6f0a4de27e2687eed176878f13d/src/jinja2/filters.py#L13  due to
```'soft_unicode' has been renamed to 'soft_str'. The old name will be removed in MarkupSafe 2.1.```

bollwyvl:
As a library almost as much as a tool, if possible we should try to support _and test_ as simple/wide a range as we can, without offering _too_ much of a crutch/impediment to others.

We could start a `_jinja_compat.py` module which normalized the behavior, during the 6.x line: having some `try:... ; except ImportError:` clauses isn't very expensive, (vs #1685 which is a whole other world), and we can even emit some warnings of our own. 

On a future, sensible major version increase, we could then bump the bottom range once and for all: `jinja2<3` is no longer supported, and historically, template libraries can be rich sources of security issues.

PR #1741, opened 2022-03-28
Add upper bounds to dependencies

martinRenou:
Fix https://github.com/jupyter/nbconvert/issues/1740

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -213,20 +213,20 @@ setup_args = dict(
 
 setup_args['install_requires'] = [
     'mistune>=0.8.1,<2',
-    'jinja2>=2.4',
-    'pygments>=2.4.1',
+    'jinja2>=2.4,<3.2',
+    'pygments>=2.4.1,<3',
     'jupyterlab_pygments',
-    'traitlets>=5.0',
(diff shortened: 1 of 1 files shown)


blink1073:
We've been trying to avoid preemptive upper bounds, after reading this [blog post](https://iscinumpy.dev/post/bound-version-constraints/)

maartenbreddels:
I agree, i think it's the less harmful options, but require developers to
do as little unnecessary breaking changes as possible.

(from mobile phone)

Op za 2 apr. 2022 03:15 schreef Steven Silvester ***@***.***>:

blink1073:
I think breaking changes will happen either way, but if we have pins then the users have less options to solve their local environments.

martinRenou:
Please feel free to close my PR if you don't think that change is relevant.

Should we at least consider adding an upper bound on Jinja2 for the next minor release? They don't follow semver and they [recommend pinning it](https://github.com/pallets/jinja/issues/1626).

martinRenou:
Replaced by https://github.com/jupyter/nbconvert/pull/1751

PR #1753, opened 2022-04-07
update cli example

leahecole:
I realized I forgot to add the "enabled" flag and I also updated the syntax with the new syntax


blink1073 (approved):
Thanks!

Issue #1754, opened 2022-04-07
`bleach` unconstrained dependency causes import errors


ReubenFrankel:
The [`bleach`](https://pypi.org/project/bleach/) package was updated to version `5.0.0` earlier today. This broke importing modules from `nbconvert` on the latest version (`6.4.5`). This seems to be an issue with older `6.4` versions as well.

The `bleach` dependency is currently unconstrained: https://github.com/jupyter/nbconvert/blob/68b496b7fcf4cfbffe9e1656ac52400a24cacc45/setup.py#L248

We tested with the previous major version of `bleach` (`4.1.0`), which resolved the issue.

---

### Steps to reproduce

(tested with Python 3.7.13 and 3.9.5)

```
pip install nbconvert
```

```py
from nbconvert import MarkdownExporter
```

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/private/tmp/.venv/lib/python3.7/site-packages/nbconvert/__init__.py", line 4, in <module>
    from .exporters import *
  File "/private/tmp/.venv/lib/python3.7/site-packages/nbconvert/exporters/__init__.py", line 3, in <module>
    from .html import HTMLExporter
  File "/private/tmp/.venv/lib/python3.7/site-packages/nbconvert/exporters/html.py", line 26, in <module>
    from nbconvert.filters.highlight import Highlight2HTML
  File "/private/tmp/.venv/lib/python3.7/site-packages/nbconvert/filters/__init__.py", line 6, in <module>
    from .markdown import *
  File "/private/tmp/.venv/lib/python3.7/site-packages/nbconvert/filters/markdown.py", line 23, in <module>
    from .pandoc import convert_pandoc
  File "/private/tmp/.venv/lib/python3.7/site-packages/nbconvert/filters/pandoc.py", line 1, in <module>
    from nbconvert.utils.pandoc import pandoc
  File "/private/tmp/.venv/lib/python3.7/site-packages/nbconvert/utils/pandoc.py", line 12, in <module>
    from nbconvert.utils.version import check_version
  File "/private/tmp/.venv/lib/python3.7/site-packages/nbconvert/utils/version.py", line 11, in <module>
    from packaging.version import Version
ModuleNotFoundError: No module named 'packaging'
```

bollwyvl:
ideally, we'd be able to add a dependency on `packaging`, which #1755 is exploring. it's already a dependency of many downstreams.

bollwyvl:
[...]

PR #1755, opened 2022-04-07
support bleach 5, add packaging and tinycss2 dependencies

bollwyvl:
## References
- investigating #1754 

## Code Changes
- [x] add dependency on `packaging`, which was previously being provided by `bleach`
  - it's better if we _can_ support bleach 4 and 5
- [x] add dependency on `tinycss2`
  - it has a rather tight pin upstream, but is hidden behind an `[extra]`, sigh 
- [x] sort dependencies
- [x] add compatibility for `bleach` 5 with `tinycss2`
  - [x] fallback to `bleach` 4, with deprecation warning
    - [x] fallback with  to no CSS santizaton, with user warning

## Alternatives
- pin to `bleach<5` (not great, because security-related)
- pin to `bleach>=5` (not terrible)

### DIFF
--- a/nbconvert/preprocessors/sanitize.py
+++ b/nbconvert/preprocessors/sanitize.py
@@ -8,35 +8,34 @@ import bleach
 from bleach import ALLOWED_ATTRIBUTES, ALLOWED_TAGS, clean
 from traitlets import Any, Bool, List, Set, Unicode
 
-
 _USE_BLEACH_CSS_SANITIZER = False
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
Thanks!

s-weigand:
Any chance that there will be a release with this fix in it soon? ❤
The missing `packaging` dependency is breaking my `pre-commit` hook.

blink1073:
I plan to make a release on Monday.

PR #1763, opened 2022-04-20
Match markedjs behavior for parsing HTML linebreaks

martinRenou:
Marked JS (used in JupyterLab and Jupyter Notebook) had a small behavior difference when parsing line breaks:

```python
<p>
Line 1

Line2
</p>
```

linebreaks

Behavior in nbconvert prior to this PR:

before

Behavior in nbconvert after this PR:

after


nbconvert/filters/markdown_mistune.py
```diff
@@ -106,6 +106,11 @@ def output_latex_environment(self, m):
 
 class MarkdownWithMath(mistune.Markdown):
     def __init__(self, renderer, **kwargs):
+        # This makes mistune behave more like markedjs
+        # (markedjs behing used by Jupyter Notebook and JupyterLab)
```
fcollonval, line 110: Small typo in `behing` :wink: 

Very nice find.

martinRenou: Thanks!

martinRenou:
I've got a working solution but I am really not happy with the implementation

Issue #1685, opened 2021-12-06
New release of mistune is a major version release: please take it up


ViktorHaag:
[Mistune's latest release](https://github.com/lepture/mistune) is a major change and a re-design. Now that its version 2.0 is out of a preview release, please consider taking it up here to facilitate supporting downstream CI installs.

It'd be great if nbconvert ensured it worked with the new version of mistune and unpinned its dependency (or repinned to a later version window).

Downstream consumers of nbconvert can pin, but may run into situations where other projects depend on the newer version of mistune.

michaelaye:
Yeah:

```python
----> 1 from nbconvert.preprocessors import Preprocessor

/luna4/maye/miniconda3/envs/py38/lib/python3.8/site-packages/nbconvert/__init__.py in <module>
      2 
      3 from ._version import version_info, __version__
----> 4 from .exporters import *
      5 from . import filters
      6 from . import preprocessors

/luna4/maye/miniconda3/envs/py38/lib/python3.8/site-packages/nbconvert/exporters/__init__.py in <module>
      1 from .base import (export, get_exporter, 
      2                    ExporterNameError, get_export_names)
----> 3 from .html import HTMLExporter
      4 from .slides import SlidesExporter
      5 from .templateexporter import TemplateExporter

/luna4/maye/miniconda3/envs/py38/lib/python3.8/site-packages/nbconvert/exporters/html.py in <module>
     13 
     14 from nbconvert.filters.highlight import Highlight2HTML
---> 15 from nbconvert.filters.markdown_mistune import IPythonRenderer, MarkdownWithMath
     16 
     17 from .templateexporter import TemplateExporter

/luna4/maye/miniconda3/envs/py38/lib/python3.8/site-packages/nbconvert/filters/__init__.py in <module>
      4 from .highlight import *
      5 from .latex import *
----> 6 from .markdown import *
      7 from .strings import *
      8 from .metadata import *

/luna4/maye/miniconda3/envs/py38/lib/python3.8/site-packages/nbconvert/filters/markdown.py in <module>
     11 
     12 try:
---> 13     from .markdown_mistune import markdown2html_mistune
     14 except ImportError as e:
     15     # store in variable for Python 3

/luna4/maye/miniconda3/envs/py38/lib/python3.8/site-packages/nbconvert/filters/markdown_mistune.py in <module>
     29 
     30 
---> 31 class MathBlockGrammar(mistune.BlockGrammar):
[...]

PR #1764, opened 2022-04-23
Update to Mistune 2.0.2

TiagodePAlves:
[Mistune 2.0](https://github.com/lepture/mistune/tree/799cd118cc5e664b72e98410ce1b68645f1a38c0) was major overhaul, requiring a lot of changes on the `markdown2html_mistune` filter. I understand that the update may noy be desirable (see #1685), but I'm leaving this here as an option. Still, some more testing seems needed.

## Possible problems

- [ ] I'm using mistune 2.0.2 since 2.0.0 seems to have [some XSS vulnerability](https://github.com/jupyter/nbconvert/issues/1685#issuecomment-1090180891), but is it safe?
- [ ] The new `BlockParser` [doesn't consider `#TITLE` without space](https://github.com/lepture/mistune/blob/799cd118cc5e664b72e98410ce1b68645f1a38c0/mistune/block_parser.py#L54) a valid header. I've fixed it with a custom `AXT_HEADING` regex, but there may be other problems hidden in those regexes.
- [ ] Mistune now uses the [undocumented](https://github.com/python/cpython/issues/84440#issuecomment-1101758660) [`re.Scanner`](https://github.com/lepture/mistune/blob/799cd118cc5e664b72e98410ce1b68645f1a38c0/mistune/scanner.py#L3), but I couldn't make it remove the `$$` suffix in block math and had to strip the text manually. Can we trust `re.Scanner`? It seems to have other [unexpected shortcomings](https://www.reddit.com/r/Python/comments/3edk1g/rescanner_undocumented_class_perfect_for_a/ctec8kl/?utm_source=share&utm_medium=web2x&context=3) too.
- [ ] Mistune also [considers `javascript:...` links harmful](https://github.com/lepture/mistune/blob/799cd118cc5e664b72e98410ce1b68645f1a38c0/mistune/renderers.py#L105) and removes them. I've bypassed that, but maybe it should be adopted too.
- [x] I've used f-strings, is that okay?

### DIFF
--- a/nbconvert/filters/markdown_mistune.py
+++ b/nbconvert/filters/markdown_mistune.py
@@ -48,6 +48,9 @@ class MathBlockParser(BlockParser):
 
     RULE_NAMES = ("multiline_math",) + BlockParser.RULE_NAMES
 
+    # Regex for header that doesn't require space after '#'
+    AXT_HEADING = re.compile(r" {0,3}(#{1,6})(?!#+)\s*([^\n]*?)$")
+
     def parse_multiline_math(self, m, state):
         """Pass token through mutiline math."""
         return {"type": "multiline_math", "text": m.group(0)}
@@ -61,6 +64,19 @@ def _dotall(pattern):
     to the undocumented `re.Scanner`.
     """
     return f"(?s:{pattern})"
+
+
+def _strip(text, *, prefix, suffix):
+    """Remove prefix and suffix from text, if present.
+
+    `InlineParser` sometimes return these affixes, even though it shouldn't.
+    """
+    np, ns = len(prefix), len(suffix)
+    if text[:np] == prefix:
+        text = text[np:]
+    if text[-ns:] == suffix:
+        text = text[:-ns]
+    return text
 
 
 class MathInlineParser(InlineParser):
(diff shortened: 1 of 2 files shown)


blink1073:
Thanks for working on this @TiagodePAlves!  I think what you've done so far is a great approach to keeping backwards compatibility.  I'd say if we get the existing test suite to pass, we should merge this PR.  Pinning isn't a long term solution, and vendoring `mistune` 0.8.4 does not sound fun either.  f-strings are totally fine!

TiagodePAlves:
Ok, I think the regexes may be the problem. I'll push it as soon as I fix it.

blink1073:
No rush, thanks again for taking this up!

TiagodePAlves:
I rewrote the regexes and they were passing the `test_markdown2html*` tests now. I couldn't test everything because pyppeteer hangs indefinitely while [creating a new page](https://github.com/jupyter/nbconvert/blob/6876519df3988f385a01569c9af03a9f13149683/nbconvert/exporters/webpdf.py#L91).

blink1073:
All green in CI!

@bollwyvl @SylvainCorlay I propose we release a 7.0rc0 with this change.

blink1073 (approved):
Changes LGTM!

EwoutH:
Thanks a lot for this!

PR #1767, opened 2022-05-04
Switch to flit build backend

blink1073:
Uses an [in-tree build backend](https://peps.python.org/pep-0517/#in-tree-build-backends) to handle the creation of the data files.


blink1073:
Docs failure is unrelated, due to a recent release of `nbclient`

blink1073:
Docs are passing, moving forward.

PR #1782, opened 2022-06-01
switch from entrypoints to importlib-metadata

konstin:
entrypoints is deprecated and recommends switching to importlib-metadata. I've implemented this because I'm using a custom import mechanism that works with the documented importlib.metadata interface but not with entrypoints' custom logic

### DIFF
--- a/nbconvert/exporters/base.py
+++ b/nbconvert/exporters/base.py
@@ -5,7 +5,10 @@
 
 import os
 
-import entrypoints
+try:
+    import importlib.metadata as importlib_metadata
+except ModuleNotFoundError:
+    import importlib_metadata
 from nbformat import NotebookNode
 from traitlets.config import get_config
 from traitlets.log import get_logger
@@ -97,20 +100,14 @@ def get_exporter(name, config=get_config()):  # noqa
         name = "notebook"
 
     try:
-        exporter = entrypoints.get_single("nbconvert.exporters", name).load()
+        exporters = importlib_metadata.entry_points()["nbconvert.exporters"]
+        exporter = [e for e in exporters if e.name == name or e.name == name.lower()][0].load()
         if getattr(exporter(config=config), "enabled", True):
             return exporter
         else:
             raise ExporterDisabledError('Exporter "%s" disabled in configuration' % (name))
-    except entrypoints.NoSuchEntryPoint:
-        try:
-            exporter = entrypoints.get_single("nbconvert.exporters", name.lower()).load()
-            if getattr(exporter(config=config), "enabled", True):
-                return exporter
-            else:
-                raise ExporterDisabledError('Exporter "%s" disabled in configuration' % (name))
-        except entrypoints.NoSuchEntryPoint:
-            pass
+    except IndexError:
+        pass
 
     if "." in name:
         try:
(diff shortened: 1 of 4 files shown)


blink1073 (approved):
LGTM, thanks!

pyproject.toml
```diff
@@ -23,7 +23,7 @@ dependencies = [
     "beautifulsoup4",
     "bleach",
     "defusedxml",
-    "entrypoints>=0.2.2",
+    "importlib-metadata;python_version<3.8",
```
bollwyvl: One thing: we have been trying to avoid `importlib-metadata<3.6`, and the stdlib one in `python<3.10`: it loads _every installed entrypoint_, which can take _a couple seconds_.

- https://github.com/jupyterlab/jupyterlab_server/blob/main/pyproject.toml#L27
- https://github.com/jupyterlab/jupyterlab_server/blob/main/jupyterlab_server/translation_utils.py#L20

nbconvert/exporters/base.py
```diff
@@ -97,20 +100,14 @@ def get_exporter(name, config=get_config()):  # noqa
         name = "notebook"
 
     try:
-        exporter = entrypoints.get_single("nbconvert.exporters", name).load()
+        exporters = importlib_metadata.entry_points()["nbconvert.exporters"]
```
bollwyvl, line 103: this loads _every single entry_point_, and can take _seconds_: indeed, i think it throws warnings on newer pythons. 

Using `entry_points(group="nbconvert.exporters.script")` will 

- https://github.com/jupyterlab/jupyterlab_server/blob/main/jupyterlab_server/translation_utils.py#L94

nbconvert/exporters/base.py
```diff
@@ -5,7 +5,10 @@
 
 import os
 
-import entrypoints
+try:
```
bollwyvl, line 8: one last thing: because the stdlib version was broken, this is most accurate as a check of `sys.version`, e.g.:

```py
if sys.version_info < (3, 10):
    from importlib_metadata import entry_points
else:
    from importlib.metadata import entry_points
```

https://github.com/jupyterlab/jupyterlab_server/blob/main/jupyterlab_server/translation_utils.py#L20

konstin:
Thanks for the notes @bollwyvl, I switched it to the group syntax

nbconvert/exporters/base.py
```diff
@@ -136,7 +134,7 @@ def get_export_names(config=get_config()):  # noqa
     Exporters can be found in external packages by registering
     them as an nbconvert.exporter entrypoint.
     """
-    exporters = sorted(entrypoints.get_group_named("nbconvert.exporters"))
+    exporters = (e.name for e in entry_points(group="nbconvert.exporters"))
```
bollwyvl, line 137: this is a generator: should probably still be sorted, as well...

konstin: indeed, fixed

konstin:
in my [ci run](https://github.com/konstin/nbconvert/commit/8712e1d798ddf1f35fd3ab4df8fe2fe59a84b1aa) tests are passing, but docs aren't, i assume the docs failure is unrelated?

PR #1791, opened 2022-06-09
WebPDFExporter: Emulate media print

martinRenou:
This PR reverts a change I made in https://github.com/jupyter/nbconvert/pull/1718 where I wrongly thought that we needed to emulate the "screen" media in order for the background to properly be colored when using a custom theme, but it does not seem to be the case, only the "printBackground" pyppeteer option and the "webkit-print-color-adjust" CSS rules are needed.

The "print" media emulation is needed so that the following CSS rules are properly applied (fixing issues with page breaks on the PDF output): https://github.com/jupyter/nbconvert/blob/main/share/templates/lab/index.html.j2#L117

cc. @mnowacki-b


blink1073 (approved):
Thanks!

martinRenou:
Thanks you for merging!

PR #1795, opened 2022-06-09
Fix lab template output alignment

### LINKED ISSUE #1794: Lab template output content is misaligned with prompt
Bascially, the output content doesn't align properly with the prompt, causing a mismatch with JupyterLab and a noticeable misalignment (look at `[8]: 6300.0`). In the screenshots, the nbconvert output is shown first, then the JupyterLab rendering:

Notebook in nbconvert showing misalignment
Same notebook in JupyterLab with proper alignment

This is due to the `padding-top` not being added. The relevant css rule from JupyterLab is 
```css
.jp-OutputArea-executeResult .jp-RenderedText.jp-OutputArea-output {
  padding-top: var(--jp-code-padding);
  border-top: var(--jp-border-width) solid transparent;
}
```
which means an element classed as `.jp-RenderedText.jp-OutputArea-output` must be a descendant of one classed as `.jp-OutputArea-executeResult`. In `share/templates/lab/index.html.j2`, the relevant class is added here:

https://github.com/jupyter/nbconvert/blob/58b0ae2cebfc550d0042a29ca9c3e1b62da78c66/share/templates/lab/base.html.j2#L112-L118

but this is added on the `.jp-OutputArea-output` element instead of its parent element to trigger the rule.

Potentially related to #1145.

**Nbconvert version:** 6.5.0

dakoop:
This uses an if/else block to add the class. Note that it does not remove the `{%- set extra_class=... -%}` directive on the `execute_result` block.

Fixes #1794

### DIFF
--- a/share/templates/lab/base.html.j2
+++ b/share/templates/lab/base.html.j2
@@ -80,7 +80,11 @@
 {% endblock output_area_prompt %}
 
 {% block output %}
+{%- if output.output_type == 'execute_result' -%}
+<div class="jp-OutputArea-child jp-OutputArea-executeResult">
+{%- else -%}
 <div class="jp-OutputArea-child">
+{%- endif -%}
 {% if resources.global_content_filter.include_output_prompt %}
     {{ self.output_area_prompt() }}
 {% endif %}


blink1073 (approved):
Thank you!

Issue #1819, opened 2022-08-02
nbconvert broken distribution


wvxvw:
Trivial install with `setuptools` from the current newest Jupyter is broken:

using this `setup.py`:

```
#!/usr/bin/env python
# -*- coding: utf-8 -*-

from setuptools import setup


if __name__ == '__main__':
    setup(
        install_requires=(
            'notebook',
        ),
    )

```

```
Installing nbconvert-7.0.0rc3-py3-none-any.whl to /home/wvxvw/<secret>/.venv/lib/python3.8/site-packages
Traceback (most recent call last):
  File "./setup.py", line 8, in <module>
    setup(
  File "/home/wvxvw/<secret>/.venv/lib/python3.8/site-packages/setuptools/__init__.py", line 153, in setup
    return distutils.core.setup(**attrs)
  File "/usr/local/lib/python3.8/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/usr/local/lib/python3.8/distutils/dist.py", line 966, in run_commands
    self.run_command(cmd)
  File "/usr/local/lib/python3.8/distutils/dist.py", line 985, in run_command
    cmd_obj.run()
  File "/home/wvxvw/<secret>/.venv/lib/python3.8/site-packages/setuptools/command/install.py", line 67, in run
    self.do_egg_install()
  File "/home/wvxvw/<secret>/.venv/lib/python3.8/site-packages/setuptools/command/install.py", line 117, in do_egg_install
    cmd.run(show_deprecation=False)
  File "/home/wvxvw/<secret>/.venv/lib/python3.8/site-packages/setuptools/command/easy_install.py", line 408, in run
    self.easy_install(spec, not self.no_deps)
  File "/home/wvxvw/<secret>/.venv/lib/python3.8/site-packages/setuptools/command/easy_install.py", line 650, in easy_install
    return self.install_item(None, spec, tmpdir, deps, True)
  File "/home/wvxvw/<secret>/.venv/lib/python3.8/site-packages/setuptools/command/easy_install.py", line 697, in install_item
    self.process_distribution(spec, dist, deps)
  File "/home/wvxvw/<secret>/.venv/lib/python3.8/site-packages/setuptools/command/easy_install.py", line 744, in process_distribution
    distros = WorkingSet([]).resolve(
  File "/home/wvxvw/<secret>/.venv/lib/python3.8/site-packages/pkg_resources/__init__.py", line 766, in resolve
    dist = best[req.key] = env.best_match(
[...]

PR #1822, opened 2022-08-06
Remove tests from bdist

TiagodePAlves:
As @wvxvw pointed out in #1819, nbconvert shouldn't be packaging and distributing test files in binary wheels. This PR is a simple solution for that.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -105,6 +105,12 @@ path = "nbconvert/_version.py"
 [tool.hatch.build.targets.sdist]
 artifacts = ["share/templates/lab/static/*.*"]
 
+[tool.hatch.build.targets.wheel]
+exclude = [
+    "nbconvert/tests/",
+    "nbconvert/**/tests/",
+]
+
 [tool.hatch.build.targets.wheel.shared-data]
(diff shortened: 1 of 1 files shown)


TiagodePAlves:
So both failing tests assume the test files are installed with nbconvert. Pip always install a wheel though, even when installing from a sdist, in which case the bdist is built before installing. The only solution for this is installing in editable, using the source files directly for that package. This would require changing the Github action [test-sdist](https://github.com/jupyterlab/maintainer-tools/blob/main/.github/actions/test-sdist/action.yml#L30) into

```bash
python -m pip install -e ${{inputs.package_spec}}
```

TiagodePAlves:
Wheel got reduced by 52%.

```fish
> ls -h dist/ new/
dist/:
total 1412
564 nbconvert-7.0.0rc3-py3-none-any.whl  848 nbconvert-7.0.0rc3.tar.gz

new/:
total 1116
268 nbconvert-7.0.0rc3-py3-none-any.whl  848 nbconvert-7.0.0rc3.tar.gz
```

```fish
> math \(564 - 268\) / 564
0.524823
```

PR #1827, opened 2022-08-09
Remove downloaded CSS from repository

martinRenou:


mascguy:
MacPorts member/maintainer here, and this change has proved a bit painful.

While I understand the desire to decouple the CSS files from this repo, that also precludes an easy offline installation: Unless I'm missing something (?), it looks like we now have to re-implement downloading of the CSS artifacts, as done in `hatch_install.py`.

And given that all of the logic is essentially hard-coded in the installation script - including versions, the specific artifacts themselves (and where they need to be placed, etc) - this makes things far more painful than it would be if they were included in a source tarball.

Is there any way that a source release could be built, with the artifacts included?

mascguy:
> MacPorts member/maintainer here, and this change has proved a bit painful.
> 
> While I understand the desire to decouple the CSS files from this repo, that also precludes an easy offline installation: Unless I'm missing something (?), it looks like we now have to re-implement downloading of the CSS artifacts, as done in `hatch_build.py`.
> 
> And given that all of the logic is essentially hard-coded in the installation script - including versions, the specific artifacts themselves (and where they need to be placed, etc) - this makes things far more painful than it would be if they were included in a source tarball.
> 
> Is there any way that a source release could be built, with the artifacts included?

We're tracking the issue here: [65809 - py-nbconvert @7.0.0: install fails, due to build-time artifact fetch](https://trac.macports.org/ticket/65809)

Guidance/advice on how we can best accomplish an offline installation - with the key being that our builds preclude dynamic artifact fetches via the Python build script - would be appreciated.

maartenbreddels:
Curious why an offline build is required? You can build the sdist or bwheel when online, and install that offline right?

mascguy:
> Curious why an offline build is required? You can build the sdist or bwheel when online, and install that offline right?

To ensure truly reproducible builds - and not be affected by things like, say, offline CDN resource(s) - we pull (and mirror) distribution files.

That requires having some control and knowledge of artifact downloads. And a build script that arbitrarily pulls resources - without formally listing them somewhere, and simply hard-coding it all in the build script - is highly problematic.

In short, our builders limit such activity, specifically to ensure that anything suspicious isn't happening. Otherwise artifacts could be arbitrarily downloaded from any location - including questionable sources and such.

Of note, this isn't just theoretical: Given more recent news of bad actors introducing spyware and viruses in some Python repos, this is a legitimate concern!

But by fetching and mirroring artifacts - and preventing build scripts from arbitrarily downloading anything they want - that helps reduce the likelihood of building packages with such attacks.

Does that make sense?

mascguy:
> I think so. How is this different from a js package that uses package.json? The fact that it's declarative and it's not in code?
> 
> What do you think are good alternatives to doing this?

Well, I'd prefer to see the CSS artifacts added back into the repo, and eliminate this: Assuming that fetching them externally doesn't solve a licensing conflict - and since they're being packaged with this component at build time, I don't see any such benefit - what does fetching from a CDN resource accomplish?

The build file has to be changed every time you want to use a new version of the CCS files anyway, so that throws any maintenance simplification out the window. Either way, new commits have to be made to this repo. Why not just commit the CSS files, and be done with it?

Issue #1849, opened 2022-08-23
Incorrect conversion of matplotlib SVG plots


goyodiaz:
Steps to reproduce:
* Display a matplotlib SVG plot in a notebook.
* Export as HTML
* There is no plot in the output.

[Notebook showing the issue](https://github.com/jupyter/nbconvert/files/9404089/svg_plot.txt)
[nbconvert output](https://github.com/jupyter/nbconvert/files/9404095/svg_plot_html.txt)

Screenshot of the notebook:
![notebook](https://user-images.githubusercontent.com/1676326/186203471-e63d2f6a-dc8a-45fa-a413-5872f18936c6.png)

Screenshot of the HTML output:
![nbconvert_output](https://user-images.githubusercontent.com/1676326/186203467-0b77614c-1522-4f8f-92dc-2d311614f9bc.png)

**Nbconvert version:** 7.0.0

adl:
This is related to #1836

goyodiaz:
@adl Yes, I guess `clean_html` is returning `bytes` now and the output needs to be decoded again.

adl:
The encoding expains the `b'...'` in the output, but the more serious issue is that `clean_html` seems to strips all tags inside the `<svg>...</svg>` element, making `<svg>` useless.

goyodiaz:
I guess that can be fixed by using a custom `Cleaner` object, but I have no idea how to express that in the template.

adl:
An additional point. 

Looking at the templates, I find weird that HTML is conditionally passed through `clean_html` with:

```
{%- if resources.should_sanitize_html %}
{{ output.data['text/html'] | clean_html }}
{%- else %}
{{ output.data['text/html'] }}
{%- endif %}
```
where `should_sanitize_html` defaults to False, while SVG goes through `clean_html` unconditionally.

adl:
@martinRenou and @SylvainCorlay this `clean_html` was added by 4a5eeb42ce3f65799a38fee785a5c84c38ea4ad9.  Can you comment?

adl:
With #1854 the implementation of `clean_html` has changed to only allow a handful of tags, and escape everything else.  So instead of a blank SVG image, we can now see the SVG source in the rendered HTML.

luboshanus:
Any changens? 
nbconvert 7.2.2 still doing the SVG source in html. Thought at first, it is julia thing https://github.com/jheinen/GR.jl/issues/482

JBlaschke:
[...]

PR #1683, opened 2021-12-01
Allow passing extra args to code highlighter

yuvipanda:
I primarily want to allow for the possibility of line numbers
in HTML output. Pygments supports this natively, but we don't
have a way to pass arguments through to pygments.

A traitlet is added to allow passing arbitrary extra args
to the Pygments formatter, for Html and LaTeX. Should allow for
line number control and much more :)

This has been something folks have asked for for a while.

Ref https://github.com/ipython/ipython/issues/5847
Ref https://github.com/jupyter/nbconvert/issues/427
Ref https://github.com/notebook-sharing-space/nbss/issues/34

PR #1854, opened 2022-09-01
Replace lxml.html.clean_html with bleach; drop lxml dependency

akx:
There's no need to pull in all of lxml for just `clean_html`, especially since nbconvert already depends on bleach anyway.

This PR:

1. removes the unused, outdated, auto-generated `notebook1.html` file from `nbconvert/tests/files`
2. simplifies a test that would fail with the next commit
3. drops the `lxml` dependency in favor of `bleach`; see below.

So turns out `bleach` was a bit stricter by default than lxml's `clean_html`, so I had to tweak the allowed tags rules a little:

* The `div`, `pre`, `code`, `span` tags (used by highlighting) are explicitly allowed
* The `class` and `id` attributes are explicitly allowed.

Beyond that, I manually checked that all of the .ipynb files under `nbconvert/examples/tests` are equivalently generated aside from `"` being `&quot;` and `'` being `&#39;`.


blink1073:
I'm a bit concerned about whether it is a drop in replacement.  There is one test that is failing that looks to be a result of different filtering.

akx:
@blink1073 I'll take a look. Thing is the filter isn't quite documented in any way.

akx:
@blink1073 The single test failure was due to a different quoting; `bleach` doesn't let a bare `'` through but converts it into an entity reference.

blink1073:
I changed the setting so you won't need approval to run the CI workflows.  If there is a slight tweak that can be made to the expected output that doesn't affect security then that seems reasonable.

akx:
@blink1073 Thanks – that makes running CI a little easier. Please see the updated PR description, too.

blink1073 (approved):
Nice work, thank you!

nbconvert/filters/strings.py
```diff
@@ -75,6 +78,21 @@ def html2text(element):
...
 
 
+def clean_html(element):
+    if isinstance(element, bytes):
+        element = element.decode()
+    else:
+        element = str(element)
+    return bleach.clean(
+        element,
+        tags=[*bleach.ALLOWED_TAGS, "div", "pre", "code", "span"],
```
adl, line 88: This `clean_html` function is clearly unsuitable to be called on `image/svg+xml` elements, since it will escape all xml tags.   Unfortunately the `base.html.j2` templates call `clean_html` in this context.  See #1849.

Rather than listing tags that are allowed, would it be possible to list tags that are disallowed?  (Like I suspect `<script>` and `<iframe>` are unwelcome. )

akx: Ah, that's a bummer. No tests clearly covered this, so that's a regression. SVG tags are well-known though so they could be added to the allowlist...

yuvipanda:
It looks like `table` and friends is also something that bleach maybe filters out while lxml passes through. Since pygments supports linenumbers via tables (see https://github.com/jupyter/nbconvert/pull/1683), with a new enough version of nbconvert, enabling line numbers as described in the linked PR makes them display like this:

image

With the workaround specified in https://github.com/jupyter/nbconvert/issues/1892#issuecomment-1294475221, it looks ok again:

image

yuvipanda:
Opened https://github.com/jupyter/nbconvert/pull/2083 to add those :)

PR #1890, opened 2022-10-26
clean_html: allow SVG tags and SVG attributes 

akx:
Fixes https://github.com/jupyter/nbconvert/issues/1849
Refs https://github.com/jupyter/nbconvert/pull/1854


akx:
@blink1073 How can I make `enforce-label` pass? I can't set labels on the PR by hand...

nbconvert/filters/svg_constants.py
```diff
@@ -0,0 +1,185 @@
+# Via bleach/_vendor/html5lib/filters/sanitizer.py;
```
blink1073, line 1: Let's depend on `html5lib` directly instead of inlining.

akx: The sanitizer is still deprecated in upstream html5lib.

https://github.com/html5lib/html5lib-python/blob/f7cab6f019ce94a1ec0192b6ff29aaebaf10b50d/html5lib/filters/sanitizer.py#L23-L29

blink1073: For posterity, from the [migration guide](https://github.com/mozilla/bleach/blob/main/docs/migrating.rst): "If you want to stick to the html5lib sanitizer's allow lists, get them from the [sanitizer code](https://github.com/html5lib/html5lib-python/blob/master/html5lib/filters/sanitizer.py). It's probably best to copy them as static lists"

blink1073: Maybe add that comment to the file itself?

akx:
Ah, I'll still need to see that `xlink:...` gets correctly kept.

akx:
Ah, darnit. `bleach` apparently happily cleans out `xlink:href` to `href`, which will probably break embedded SVGs... https://github.com/mozilla/bleach/issues/362

akx:
@blink1073 Right, this will allow a certain subset of SVG embedded into HTML.

The templates for actual `image/svg+xml` cells (e.g. https://github.com/jupyter/nbconvert/blob/a8067447af4a90690b762f48ceb46d40207a344e/share/templates/lab/base.html.j2#L166) should use an XML-savvy cleaner since a `image/svg+xml` is _not_ HTML to begin with. 

This PR is nevertheless an improvement to the status quo... What do you think?

blink1073 (approved):
LGTM!

SylvainCorlay:
@martinRenou FYI.

Issue #1660, opened 2021-10-15
Setting paper size and margins in webpdf


callegar:
Hi,
there seems to be no documentation on how to set the paper format and the margins when using webpdf.

The default margins appear absolutely too tight for any printout (in fact there seems to be no margin at all on the top and the bottom of the page) and it is unclear how one is expected to adjust them. Similarly it is unclear how to set the paper size.

Do you need to make a custom template for that? Should that "import" somehow the default template after overriding some styling?

JWock82:
I have the same issue. The margins are really tight. On the second sheet of my pdf the top margin isn't even there.

PR #1907, opened 2022-11-09
Fix for webpdf print margins

JWock82:
This is a fix for the webpdf print margins. It adds 0.5 inch margins to each printed page. This fixes Issue #1660


blink1073 (approved):
Thank you!

blink1073:
I'll make a release early next week.

PR #1934, opened 2023-01-10
Remove HTML escaping JSON-encoded widget state

### LINKED ISSUE #1900: Widget with "<" or ">" characters in Unicode state embeds "&lt" and "&gt" in emitted HTML
Exporting a widget with Unicode traitlet state containing `<` or `>` incorrectly serializes the `vnd.jupyter.widget-state+json` with `&lt` and `&gt` characters. This is a regression since v6.

```python
# Untitled.ipynb
import ipywidgets
ipywidgets.Label("This sould be the less than sign: <")
```

```bash
jupyter nbconvert --to html --execute Untitled.ipynb
```

`nbconvert v6.4.4`, `jinja2 v3.0.3`:

image

`nbconvert v7.2.3`, `jinja2 v3.1.2`:

image

<details>
<summary>Emitted `widget-state+json`</summary>

```html
<script type="application/vnd.jupyter.widget-state+json">
{
  "state": {
    "6c1921abf2704428a370acef0e14c85f": {
      "model_name": "LayoutModel",
      "model_module": "@jupyter-widgets/base",
      "model_module_version": "2.0.0",
      "state": {
        "_model_module": "@jupyter-widgets/base",
        "_model_module_version": "2.0.0",
        "_model_name": "LayoutModel",
        "_view_count": null,
        "_view_module": "@jupyter-widgets/base",
        "_view_module_version": "2.0.0",
        "_view_name": "LayoutView",
        "align_content": null,
        "align_items": null,
        "align_self": null,
        "border_bottom": null,
        "border_left": null,
        "border_right": null,
        "border_top": null,
        "bottom": null,
        "display": null,
        "flex": null,
        "flex_flow": null,
        "grid_area": null,
        "grid_auto_columns": null,
        "grid_auto_flow": null,
        "grid_auto_rows": null,
        "grid_column": null,
        "grid_gap": null,
        "grid_row": null,
        "grid_template_areas": null,
        "grid_template_columns": null,
        "grid_template_rows": null,
        "height": null,
        "justify_content": null,
        "justify_items": null,
        "left": null,
        "margin": null,
        "max_height": null,
        "max_width": null,
        "min_height": null,
        "min_width": null,
        "object_fit": null,
        "object_position": null,
        "order": null,
        "overflow": null,
        "padding": null,
        "right": null,
        "top": null,
        "visibility": null,
        "width": null
      }
    },
    "df0d2de504a94ae78a62b63fdf9c6bed": {
      "model_name": "LabelStyleModel",
      "model_module": "@jupyter-widgets/controls",
      "model_module_version": "2.0.0",
      "state": {
        "_model_module": "@jupyter-widgets/controls",
        "_model_module_version": "2.0.0",
        "_model_name": "LabelStyleModel",
        "_view_count": null,
        "_view_module": "@jupyter-widgets/base",
        "_view_module_version": "2.0.0",
        "_view_name": "StyleView",
        "background": null,
        "description_width": "",
        "font_family": null,
        "font_size": null,
        "font_style": null,
        "font_variant": null,
        "font_weight": null,
        "text_color": null,
        "text_decoration": null
      }
    },
    "8b9a84d33a1a48e3b1250936069c1c54": {
      "model_name": "LabelModel",
      "model_module": "@jupyter-widgets/controls",
      "model_module_version": "2.0.0",
      "state": {
        "_dom_classes": [],
        "_model_module": "@jupyter-widgets/controls",
        "_model_module_version": "2.0.0",
        "_model_name": "LabelModel",
        "_view_count": null,
        "_view_module": "@jupyter-widgets/controls",
        "_view_module_version": "2.0.0",
        "_view_name": "LabelView",
        "description": "",
        "description_allow_html": false,
        "layout": "IPY_MODEL_6c1921abf2704428a370acef0e14c85f",
        "placeholder": "​",
        "style": "IPY_MODEL_df0d2de504a94ae78a62b63fdf9c6bed",
        "tabbable": null,
        "tooltip": null,
        "value": "This sould be the less than sign: &lt;"
      }
    }
  },
  "version_major": 2,
  "version_minor": 0
}
</script>
```

</details>

<details>
<summary>`pip freeze`</summary>

```
anyio==3.6.2
appnope==0.1.3
argon2-cffi==21.3.0
argon2-cffi-bindings==21.2.0
asttokens==2.1.0
attrs==22.1.0
backcall==0.2.0
beautifulsoup4==4.11.1
bleach==5.0.1
certifi @ file:///private/var/folders/sy/f16zz6x50xz3113nwtb9bvq00000gp/T/abs_0ek9yztvu3/croot/certifi_1665076692562/work/certifi
cffi==1.15.1
debugpy==1.6.3
decorator==5.1.1
defusedxml==0.7.1
entrypoints==0.4
executing==1.2.0
fastjsonschema==2.16.2
idna==3.4
ipykernel==6.17.0
ipython==8.6.0
ipython-genutils==0.2.0
ipywidgets==8.0.2
jedi==0.18.1
Jinja2==3.0.3
jsonschema==4.16.0
jupyter==1.0.0
jupyter-console==6.4.4
jupyter-server==1.21.0
jupyter_client==7.4.4
jupyter_core==4.11.2
jupyterlab-pygments==0.2.2
jupyterlab-widgets==3.0.3
MarkupSafe==2.1.1
matplotlib-inline==0.1.6
mistune==2.0.4
nbclassic==0.4.7
nbclient==0.5.13
nbconvert==7.2.3
nbformat==5.7.0
nest-asyncio==1.5.6
notebook==6.5.2
notebook_shim==0.2.0
packaging==21.3
pandocfilters==1.5.0
parso==0.8.3
pexpect==4.8.0
pickleshare==0.7.5
prometheus-client==0.15.0
prompt-toolkit==3.0.31
psutil==5.9.3
ptyprocess==0.7.0
pure-eval==0.2.2
pycparser==2.21
Pygments==2.13.0
pyparsing==3.0.9
pyrsistent==0.19.1
python-dateutil==2.8.2
pyzmq==24.0.1
qtconsole==5.3.2
QtPy==2.2.1
Send2Trash==1.8.0
six==1.16.0
sniffio==1.3.0
soupsieve==2.3.2.post1
stack-data==0.6.0
terminado==0.17.0
testpath==0.6.0
tinycss2==1.2.1
tornado==6.2
traitlets==5.5.0
wcwidth==0.2.5
webencodings==0.5.1
websocket-client==1.4.1
widgetsnbextension==4.0.3

```

</details>

manzt:
Fixes #1900

The widget data in the templates are either the special JSON mime-type or within a `<script>` tag (JS). Transforming the JSON-encoded widget data with HTML escapes breaks application widget code. The widget author should be responsible for escaping HTML strings if necessary _before_ encoding the data as JSON in the template - not a post-encoding mutation.


manzt:
cc: @keller-mark

maartenbreddels:
Overview of the PR's / commits that touched this line:
 * https://github.com/jupyter/nbconvert/pull/1665
 * https://github.com/jupyter/nbconvert/commit/7100b2b7b12d501290f3142a791db9bec1fa1b32 (can't find the PR)
 * https://github.com/jupyter/nbconvert/commit/c90b746cddc7a4503e92ac8534e12e3eef616c7f (can't find the PR)
 * https://github.com/jupyter/nbconvert/pull/1830

With this PR we are reverting #1665, which is not what we want I guess (since it fixed a lot of issues).

I don't know why the 2nd and 3rd commit come from (ping @SylvainCorlay and @martinRenou ) but it seems that that has a different purpose (can you find the PR for this?).

maartenbreddels:
Ok, google answered that: 
 * https://securitylab.github.com/advisories/GHSL-2021-1013_1028-nbconvert/

maartenbreddels:
I don't think GHSL-2021-1025 applied anymore after https://github.com/spatialaudio/nbsphinx/pull/611 (which is the improved version of https://github.com/jupyter/nbconvert/pull/1665 cc @mgeier)

manzt:
Thanks for digging into this and having a look. I understand why avoiding a closing a script tag early is desirable, but the current html escaping modifies much more than this specific case. Would a method which is more precise to cover this case be accepted?

 As a side note, I think I understand issue in GHSL-2021-1025 but Jupyter widgets generally expose a mechanism to run any JavaScript on the page (in the global scope, with or without the identified </script> vulnerability).

martinRenou:
> I don't think GHSL-2021-1025 applied anymore after https://github.com/spatialaudio/nbsphinx/pull/611

Unfortunately the security issue applies more broadly than the nbsphinx case, so the nbsphinx PR does not fix the nbconvert issue.

> but Jupyter widgets generally expose a mechanism to run any JavaScript on the page (in the global scope, with or without the identified </script> vulnerability).

That's true, but this comment also applies to the following code cell, and we would probably like to escape it:

```python
from IPython.display import Javascript
display(Javascript('alert("this code runs")'))
```

**I think those were the main reasons for introducing the `--sanitize_html=False` option** and you should use it if you need widgets working.

In the long term we could probably think of a way to trust some well-known widgets libraries like ipywidgets/ipyleaflet.

maartenbreddels:
Why is https://github.com/spatialaudio/nbsphinx/pull/611 not sufficient for the widget output? So only addressing GHSL-2021-1025 and the widget output, not things like the title etc.

martinRenou:
> Why is https://github.com/spatialaudio/nbsphinx/pull/611 not sufficient for the widget output?

Because nbsphinx is not nbconvert, nbconvert is used in other tools, unless I misunderstand your point?

maartenbreddels:
Yes, I mean https://github.com/jupyter/nbconvert/pull/1665, but with the improvement of making it case insensitive (like what was done in https://github.com/spatialaudio/nbsphinx/pull/611)

maartenbreddels:
> **I think those were the main reasons for introducing the `--sanitize_html=False` option** and you should use it if you need widgets working.

Does this mean that with `--sanitize_html=False`, we still have the issues that https://github.com/jupyter/nbconvert/pull/1665 tried to address? If so, we should still make sure we escape closing script tags right?

martinRenou:
I think your fix still applies after https://github.com/jupyter/nbconvert/commit/c90b746cddc7a4503e92ac8534e12e3eef616c7f because it should be escaped by `escape_html`.

I may have read that thread and the code too fast, setting `--sanitize_html=False` would not fix the initial PR purpose to not escape the HTML in the widget JSON.

Maybe reintroducing https://github.com/jupyter/nbconvert/pull/1665 instead of using `escape_html_keep_quotes` would make everyone happy here? Though I fear that it would not handle all the security breaches.

martinRenou:
> Though I fear that it would not handle all the security breaches.

But when `--sanitize_html=False` we should not sanitize the widget output, so let's change this PR and replace the `escape_html_keep_quotes` call with the `escape_html_script` from https://github.com/jupyter/nbconvert/pull/1665/files so that we don't escape anything except the closing script tag.

maartenbreddels:
Just for clarity, #1665 does not try to solve a security issue, but fixes a bug when a script end tag is included in a string for the JSON. So what #1665 does (or better, the way https://github.com/spatialaudio/nbsphinx/pull/611 does it) should always happen indeed.

maartenbreddels:
Just for my understanding, if `--sanitize_html=True`, why would we need to sanitize the widget output, what are possible security issues? GHSL-2021-1025 only mentions the closing script tag, which would already be solved by #1665/https://github.com/spatialaudio/nbsphinx/pull/611 . Maybe I'm slow this morning :)

martinRenou:
> So what https://github.com/jupyter/nbconvert/pull/1665 does (or better, the way https://github.com/spatialaudio/nbsphinx/pull/611 does it) should always happen indeed.

Agreed 👍🏽 

> Just for my understanding, if --sanitize_html=True, why would we need to sanitize the widget output, what are possible security issues? 

We don't need to sanitize in that case indeed 👍🏽 I'm the one being slow this morning, sorry. I thought this was changing code that occurs when actually sanitizing the HTML but it's not. That code should not sanitize the HTML and only apply the https://github.com/jupyter/nbconvert/pull/1665 bug fix.

martinRenou:
> So what https://github.com/jupyter/nbconvert/pull/1665 does (or better, the way https://github.com/spatialaudio/nbsphinx/pull/611 does it) should always happen indeed.

Agreed 👍🏽 

> Just for my understanding, if --sanitize_html=True, why would we need to sanitize the widget output, what are possible security issues? 

You could inject any HTML in the widget JSON repr (the same way you can close the script tag). There are other ways than script tags for injecting malicious code in the page (through image sources for example).

I'm the one being slow this morning sorry. That code that the PR touches is behind a `{%- if not resources.should_sanitize_html %}` check so it should not do any sanitizing (apart from your script tag closing bug fix?).

martinRenou:
The code should probably be the following:

```html
{%- block footer %}
{% set mimetype = 'application/vnd.jupyter.widget-state+json'%}
{%- if not resources.should_sanitize_html %}
  {% if mimetype in nb.metadata.get("widgets",{})%}
    <script type="{{ mimetype }}">
    {{ nb.metadata.widgets[mimetype] | json_dumps | escape_html_script }}
    </script>
  {% endif %}
{%- elif %}
  {% if mimetype in nb.metadata.get("widgets",{})%}
    <script type="{{ mimetype }}">
    {{ nb.metadata.widgets[mimetype] | json_dumps | escape_html_keep_quotes }}
    </script>
  {% endif %}
{% endif %}
{{ super() }}
{%- endblock footer-%}
```

With `escape_html_script` from https://github.com/jupyter/nbconvert/pull/1665 that only removes the closing scripts.

maartenbreddels:
> With `escape_html_script` from #1665 that only removes the closing scripts.

Better, the regex version from here: https://github.com/spatialaudio/nbsphinx/pull/611



> You could inject any HTML in the widget JSON repr

But we don't insert whatever is in nb.metadata.widgets[mimetype] verbatim, but we do a `json_dumps`. This means that everything is numbers, strings etc. Only the strings can include a closing script tag, and that's already solved by `escape_html_script`.

For comparison:
```
{% set nb_title = nb.metadata.get('title', resources['metadata']['name']) | clean_html %}
<title>{{nb_title}}</title>
```

Here, we just include the value of the `nb_title` variable in the HTML, without any quotes/json_dumps, so we indeed need to clean using `clean_html`.

But I think that after a `json_dumps | escape_html_script` there shouldn't be a security issue at all, and we can skip `clean_html`.

martinRenou:
I'm not an expert in security, though `escape_html_script` seems to do less than `clean_html` does? It might not handle some cases like 
```
<[%00]/script><script>alert('malicious')
```

maartenbreddels:
Good question. So, can we write a water-tight `escape_html_script` so we don't need clean_html?

maartenbreddels:
> [%00]

What is this called? I have never seen this, is this some kind of encoding? What should it do?

maartenbreddels:
I wonder if replacing `/` by `\/` is the safest option for `escape_html_script`. It basically never allows the browser to see a close tag `...</...` since there is always something between the `<` and the `/` (i.e. the browser will see `...<\/`...`

martinRenou:
> What is this called? I have never seen this, is this some kind of encoding? What should it do?

This is a NULL byte, from what I've seen online this notation is only valid on IE (I don't know which version).

manzt:
Thanks for the active discussion. I'm just catching up...

> That's true, but this comment also applies to the following code cell, and we would probably like to escape it.

I meant more generally that jupyter widgets provide a mechanism to run third-party JS. The security issue here is concerned with preventing widget authors from injecting malicious scripts into the widget data, but every widget has third-party JS which is loaded from a CDN by `@jupyter-widgets/html-manager` and executed in the global scope.

https://github.com/jupyter-widgets/ipywidgets/blob/9fd4c3052ca90591d635ff802c2487196025a721/packages/html-manager/src/embed.ts#L10-L14

For example, if I use a `ipyobservable` widget in my notebook, `https://cdn.jsdelivr.net/npm/ipyobservable@0.1.1/dist/index.js` is loaded and executed by the widgets manger.

Widget authors control both the widget data (which _may_ be exploited) **and** widget JS (which needs no exploit to run any JS).

> With `escape_html_script` from https://github.com/jupyter/nbconvert/pull/1665 that only removes the closing scripts.

I would be in support of this solution, although with the point above don't see why this shouldn't just be the default (re: above).

> This is a NULL byte, from what I've seen online this notation is only valid (valid, as in it would read `</script>`) on IE (I don't know which version).

I'm not sure how prevalent this edge case is since [IE support officially ended](https://www.google.com/search?q=internet+explorer+deprecated&rlz=1C5CHFA_enUS994US994&oq=internet+explorer+deprecated&aqs=chrome.0.69i59j0i22i30l9.8500j0j4&sourceid=chrome&ie=UTF-8) as of June 2022.

maartenbreddels:
> replacing `/` by `\/`

I think this is a water-tight solution. I can't think how this can be circumvented (famous last words?)

SylvainCorlay:
> but Jupyter widgets generally expose a mechanism to run any JavaScript on the page

Not quite any JavaScript though, only the installed widget extensions.

maartenbreddels (approved):

ankostis:
Please apply the same fixes also to [share/templates/lab/base.html.j2](https://github.com/jupyter/nbconvert/blob/main/share/templates/lab/base.html.j2) file.

flekschas:
Are there any plans to get this merged? I'd love to use v7 and this PR seems to fix the issue. Happy to help getting it merged if there's something I can do.

PR #1935, opened 2023-01-11
Update codecov link

blink1073:


blink1073:
unpkg appears to be having problems today, giving 500 errors.

PR #1936, opened 2023-01-12
always pass relax_add_props=True when validating

minrk:
avoids crashing on notebooks that have no practical issues.

I think the relevant deprecation was a mistake, and is reverted in https://github.com/jupyter/nbformat/pull/343 because it's not related to normalization, and https://github.com/jupyter/nbformat/pull/282 deprecated normalization args passed to validate.


blink1073 (approved):
Thanks!

PR #1947, opened 2023-02-11
Do not import pyppeteer for installation check (#1947)Co-authored-by: pre-commit-ci[bot] <66853113+pre-commit-ci[bot]@users.noreply.github.com> Co-authored-by: Matthias Bussonnier <bussonniermatthias@gmail.com>

### LINKED ISSUE #1944: Revert importing Pyppeteer outside of testing
Hi,

https://github.com/jupyter/nbconvert/pull/1924 forces an extra import of Pyppeteer outside of testing. This is problematic because that depends on `trio` which tries to own the `sys.excepthook`. Our usage, similar to ubuntu, wants to manage the sys.execepthook (see [trio's comments](https://github.com/python-trio/trio/blob/333f9afef67dd5f16e385753572122362693d237/trio/_core/_multierror.py#L429)).

Can we move this import to somewhere that is only used for test cases?

krassowski:
Should fix #1944 because `importlib.util.find_spec` does not import the package.

### DIFF
--- a/nbconvert/exporters/webpdf.py
+++ b/nbconvert/exporters/webpdf.py
@@ -67,7 +67,7 @@ class WebPDFExporter(HTMLExporter):
 
     def _check_launch_reqs(self):
         try:
-            from pyppeteer import launch  # type: ignore
+            from pyppeteer import launch  # type: ignore[import]
             from pyppeteer.util import check_chromium  # type:ignore
         except ModuleNotFoundError as e:
             msg = (
--- a/nbconvert/exporters/webpdf.py
+++ b/nbconvert/exporters/webpdf.py
@@ -13,7 +13,6 @@ from traitlets import Bool, default
 
 from .html import HTMLExporter
 
-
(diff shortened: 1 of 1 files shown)


krassowski:
@MeeseeksDev tag bug

nbconvert/exporters/webpdf.py
```diff
@@ -71,7 +67,7 @@ def _template_name_default(self):
 
     def _check_launch_reqs(self):
         try:
-            from pyppeteer import launch
+            from pyppeteer import launch  # type: ignore
```
ndmlny-qs, line 70: I like having the type ignore codes when ignoring types, but if it is not standard to do here then ignore my comment.

krassowski:
@blink1073 FYI something strange happened with the title of this PR and of https://github.com/jupyter/nbconvert/pull/1949 on merge I suspect this was a GitHub bug? Or was it intended?

blink1073:
I'm not sure what happened there, it was not intentional.

PR #1957, opened 2023-03-03
add mermaidjs 10.2.3

bollwyvl:
Adds [mermaidjs diagram support](https://mermaid.js.org/), which is the most commonly-available, open source ascii-based diagram tool, supporting a large number of broadly-useful diagrams.

The motivation is that this syntax <code>```mermaid</code> is supported natively by a large number of existing platforms under a consistent syntax, most notably GitHub, which theoretically we track. 

## References

- related [JupyterLab PR](https://github.com/jupyterlab/jupyterlab/pull/14102#issuecomment-1453134883)

## Changes

- [x] adds a CDN URL for mermaidjs
- [x] adds via the recommended `<script type="module"` 
  - > this supports the lazy loading without e.g. requirejs for [95% of browser market share](https://caniuse.com/es6-module)
- [x] only load/start mermaid assets when mermaid actually found

## User-facing changes
- example output with rich display and markdown cells
  - [gist](https://gist.github.com/bollwyvl/b250b5ad584819b68923c97a45d68b22)
- this would _not_ yet be supported in
  - notebook classic
  - bespoke renderers of notebooks, such as the GitLab notebook viewer
  - other jupyter clients (beyond any linked PRs above)

### DIFF
--- a/nbconvert/exporters/html.py
+++ b/nbconvert/exporters/html.py
@@ -122,7 +122,7 @@ class HTMLExporter(TemplateExporter):
     ).tag(config=True)
 
     mermaid_js_url = Unicode(
-        "https://cdnjs.cloudflare.com/ajax/libs/mermaid/10.0.2/mermaid.esm.min.mjs",
+        "https://cdnjs.cloudflare.com/ajax/libs/mermaid/10.2.3/mermaid.esm.min.mjs",
         help="""
         URL to load MermaidJS from.
 
--- a/nbconvert/exporters/html.py
+++ b/nbconvert/exporters/html.py
@@ -121,6 +121,15 @@ class HTMLExporter(TemplateExporter):
         """,
     ).tag(config=True)
 
+    mermaid_js_url = Unicode(
+        "https://cdnjs.cloudflare.com/ajax/libs/mermaid/10.0.2/mermaid.esm.min.mjs",
+        help="""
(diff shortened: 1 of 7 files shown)


bollwyvl:
Will look into the failing tests.

maartenbreddels:
Nice to see this happen, we also have support for mermaid in [solara](https://github.com/widgetti/solara/), so we will keep an eye on this and see if we can track the same version number.

blink1073 (approved):
Thanks!

bollwyvl:
Cool. Over on the Lab PR I'll be working on bringing across more of the [accessibility features](https://mermaid.js.org/config/accessibility.html#accessibility), and will update here as appropriate.

Perhaps we'll be able to eventually reuse some of the code/style which are already highly shared, and provide a clean, offline files-on-disk install/deployment with e.g. `etc/share/jupyter/static/mermaid` or something, making it easier for multiple packages to have consistent versions and be more respectful of privacy concerns.

franzhaas:
thanks for the mermaid suport. It did not work for me out of the box wit reveal, but after changing from "plain" mermaid to mermaid js plugin it does work for me.

please review my changes here.:
https://github.com/jupyter/nbconvert/pull/2120

Is this a sensible improvement? Or did i make a stupid mistake and the original solution would work if i made a slight change...

Best regards,
Franz

PR #1960, opened 2023-03-15
Update jupyterlab CSS

martinRenou:


maartenbreddels:
Do you know if this gets rid of the blueprint CSS? I did a quick experiment once and it shrank the resulting CSS quite a bit! But that might have been with lab 4.

martinRenou:
I don't know. I checked the size difference between https://unpkg.com/@jupyterlab/nbconvert-css@3.1.11/style/index.css and https://unpkg.com/@jupyterlab/nbconvert-css@3.6.1/style/index.css and it seems that the size increases slightly actually

blink1073 (approved):
Thanks!

PR #1963, opened 2023-03-23
Make date configurable in latex/PDF

### LINKED ISSUE #1962: Enhancement: Set LaTeX/PDF document date
Similar to the ability to setting the document's author and title, I'd like to set the date.

The document might be compiled/reproduced from the source later (e.g. as part of a [DVC pipeline](https://dvc.org/doc)), but the contents didn't change.

This can be achieved fairly easily by changing https://github.com/jupyter/nbconvert/blob/main/share/templates/latex/base.tex.j2#L182

It would be great to have the possibility to suppress the date as well. This can avoid any ambiguity or false dates.

achimgaedke:
That's a working example of implementing https://github.com/jupyter/nbconvert/issues/1962.

Fix #1962

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -89,6 +89,29 @@ LaTeX
 
     Latex report, providing a table of contents and chapters.
 
+  Optionally you can specify ``authors``, ``title`` and ``date`` in the notebook's
+  metadata. These will be used to render the header of the LaTeX document.
+
+  .. code-block:: json
+
+      {
+          "authors": [
+              {
+                  "name": "Jane Doe"
+              },
+              {
+                  "name": "John Doe"
+              }
+          ],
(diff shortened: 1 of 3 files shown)


blink1073:
Thanks for working on this!  The changelog gets auto-generated, can you please add a note in `docs/source/usage.rst` instead?

achimgaedke:
@blink1073 - done, also documented the other fields, `author` and `title`.

achimgaedke:
I've added some command line options, they round the feature off nicely.

I find notebook metadata viewing/editing not very obvious, I guess this is a common fate of features becoming more important than expected.

blink1073 (approved):
Fantastic, thank you!

blink1073:
I'll make a release with this on Monday.

Issue #1966, opened 2023-03-28
Add configurable notebook name pattern


JeppeKlitgaard:
I have a use-case where I need to distinguish between manually written markdown files and those generated by `nbconvert`, which I would like to do by adding a 'pre-extension` such that `notebook.ipynb` would be written to `notebook.generated.md`.

I suggest changing the `output_base` configuration option of  `NbConvertApp`  such that this may be used with multiple inputs and can contain pattern information using a handlebars format similar to `NbConvertApp.output_files_dir` (Default: `{notebook_name}_files`) as detailed in [docs](https://nbconvert.readthedocs.io/en/latest/config_options.html)

The following two handle variables should be defined:
- `{notebook_name}` which is the base name of the notebook file
- `{index}` which is the one-indexed position of the notebook in the `self.notebooks` attribute of the `NbConvertApp` instance

I believe this change could be made backwards-compatible for all configurations except those containing either of the two handlebar expressions in the `output_base` option, which I presume no real-world configurations have.

I would be willing to provide a PR adding this functionality

PR #1974, opened 2023-04-12
Revert unintended effects of #1966

### LINKED ISSUE #1970: nbconvert --to notebook --output out.ipynb filename incorrect
Outputs to a file called "out.nbconvert.ipynb" instead of "out.ipynb" as expected.

The expected behavior was the case prior to v7.3, likely related to #1967, though not super obvious how this behavior emerges, looking at the diff as an outsider to this codebase.
My guess is that it is related to stripping the extension off there.

A similar change does not seem to happen for other conversions, e.g. `md` or `pdf`

Steps to reproduce:

Any ipynb will do, I just made a new, empty, ipynb file and called it "test.ipynb"

```bash
$ ls
test.ipynb
$ jupyter nbconvert --to notebook --output out.ipynb test.ipynb 
[NbConvertApp] Converting notebook test.ipynb to notebook
[NbConvertApp] Writing 617 bytes to out.nbconvert.ipynb
```

This is currently breaking matplotlib's tests, which rely on the expected output path being valid.

### LINKED ISSUE #1979: Cannot convert equally named notebooks in different directories in-place
This command used to work before 7.3.0:

```bash
jupyter nbconvert --clear-output --inplace a/nb.ipynb b/nb.ipynb
```

but now reports the following error:

```
ValueError: Conversion would override an already generated output. This is probably due to --output or output_base configuration leading to non-unique output names. Output notebook names were: ['nb', 'nb']
```

The check that is failing was introduced in #1967. As a workaround running the command for each notebook individually works, but it would be nice if this could work as it did before.

tuncbkose:
Revert to the previous default filename behavior
```
$ ls
test.ipynb
$ jupyter nbconvert --to notebook --output out.ipynb test.ipynb 
[NbConvertApp] Converting notebook test.ipynb to notebook
[NbConvertApp] Writing 72 bytes to out.ipynb
$ jupyter nbconvert --to notebook test.ipynb 
[NbConvertApp] Converting notebook test.ipynb to notebook
[NbConvertApp] Writing 72 bytes to test.nbconvert.ipynb
```

Closes #1970.

Edit:
Additional things to fix:
- [x] #1979 same nb names in different directories 
  - Fixed by removing the incorrect check that was failing (and an associated test). As a result, if `output_base` doesn't include some format section (like `{notebook_name}`), some files might get overwritten. This should not happen unless the user specifically wants it.
- [x] I don't think using `{notebook_filename}` will always work the way it was intended. Even if it does, it doesn't have any tests, which I should add.
  - Just removed this functionality as it doesn't seem to work cohesively with existing `FilesWriter` configuration.

Now also closes #1979.

### DIFF
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -420,9 +420,7 @@ class NbConvertApp(JupyterApp):
         """
         basename = os.path.basename(notebook_filename)
         notebook_name = basename[: basename.rfind(".")]
-        notebook_name = self.output_base.format(
-            notebook_name=notebook_name
-        )
+        notebook_name = self.output_base.format(notebook_name=notebook_name)
 
         return notebook_name
 
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -218,7 +218,7 @@ class NbConvertApp(JupyterApp):
     output_base = Unicode(
         "{notebook_name}",
         help="""Overwrite base name use for output files.
-            Supports pattern replacements '{notebook_name}' and '{notebook_filename}'.
+            Supports pattern replacements '{notebook_name}'.
             """,
     ).tag(config=True)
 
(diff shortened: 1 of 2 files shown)


tuncbkose:
@meeseeksdev tag bug

tuncbkose:
Not sure if failures are related

krassowski:
I restarted the failing CI jobs, maybe these are just flaky.

tuncbkose:
Rebasing onto main

tuncbkose:
Failure seems unrelated to me

blink1073 (approved):
Thank you!

Issue #1977, opened 2023-04-18
Attempting to convert a notebook with pasted image to PDF fails with an error


andrii-i:
Attempting to convert a notebook with pasted image to PDF fails with an error in both Jupyterlab and Notebook head-of-master. 

## Reproduce

1. Create a notebook.
2. Copy an image and paste it into the markdown cell.
3. Attempt File->Save and Export Notebook As ...->PDF

- Created based on jupyter/jupyterlab#14365 by @frohro

![image](https://user-images.githubusercontent.com/26686070/232924319-4c5d33d2-046e-45ce-8db6-e296b94cdbfc.png)

tuncbkose:
As far as I can tell, the problem is that the pasted image is stored directly in the notebook, base64 encoded (according to [this](https://nbformat.readthedocs.io/en/latest/format_description.html#cell-attachments)). 

It shouldn't be too difficult for `nbconvert` to look for these attachments and convert them into files in the same directory, so that outputs like markdown and latex will work properly.

PR #1978, opened 2023-04-19
Add ExtractAttachmentsPreprocessor

tuncbkose:
Adds a new preprocessor to extract notebook attachments into individual files, so that conversion outputs (like `.md/.tex` files) work properly.

Fixes https://github.com/jupyter/nbconvert/issues/1977.

Todos

- [x] Finish implementation
- [x] Finish preprocessor tests
- [x] Add tests related to FilesWriter tweaks

### DIFF
--- a/nbconvert/exporters/exporter.py
+++ b/nbconvert/exporters/exporter.py
@@ -95,6 +95,7 @@ class Exporter(LoggingConfigurable):
             "nbconvert.preprocessors.LatexPreprocessor",
             "nbconvert.preprocessors.HighlightMagicsPreprocessor",
             "nbconvert.preprocessors.ExtractOutputPreprocessor",
+            "nbconvert.preprocessors.ExtractAttachmentsPreprocessor",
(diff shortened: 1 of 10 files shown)


tuncbkose:
The previous iteration was fine with the tests, but it looks like the way I associated attachment directories with specific notebooks broke something. Will look into it soon.

tuncbkose:
I am not sure what is the best way to enable this preprocessor. I initially thought of adding it to `TemplateExporter.default_config` but ran into #1980.

For now, I enabled it for markdown and latex exporters.

tuncbkose:
It might be worth noting that I am aware `ExtractOutput` is pretty similar, but I am under the impression that cell outputs and attachments are different enough mechanisms to warrant different preprocessors. Still, they should probably work relatively similarly, for example if a directory is created to save related files.

I'll take a look to see how it works and make appropriate changes later.

tuncbkose:
Updated with latest commits to main

blink1073 (approved):
Thank you!

PR #1981, opened 2023-04-21
Tweak exporter default_config merging behavior

### LINKED ISSUE #1980: Exporter default_config merging
In exporters `templateexporter, html, latex, markdown,rst` and `asciidoc`, the following is present
```python
@property
def default_config(self):
    c = Config( some stuff here )
    c.merge(super().default_config)
    return c
```
where in the merging, the configuration of the exporter can be overwritten by the parent configuration. In other words, if `self` and `super()` contain different values for the same setting, the resulting config will contain the `super()`s setting.

I am a bit confused about why this is the case. My expectation is that the reverse should take place: a subclass should inherit its parent's configuration and modify it according to its own needs.

As a quick check, I replaced the line that is merging with
```python
if super().default_config:
    c2 = super().default_config.copy()
    c2.merge(c)
    c = c2
```
and the tests are still passing (except for some skipped ones due to not installed programs).

Is there a reason why the former is preferable to the latter style of merging configs?

tuncbkose:
Re #1980, changes `Config` merging so that `child` configuration overwrites `parent` configuration as opposed to the other way around.

Closes #1980

### DIFF
--- a/nbconvert/exporters/asciidoc.py
+++ b/nbconvert/exporters/asciidoc.py
@@ -48,5 +48,8 @@ class ASCIIDocExporter(TemplateExporter):
                 "HighlightMagicsPreprocessor": {"enabled": True},
             }
         )
-        c.merge(super().default_config)
+        if super().default_config:
+            c2 = super().default_config.copy()
+            c2.merge(c)
+            c = c2
         return c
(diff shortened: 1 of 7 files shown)


krassowski:
I think this looks good. Should there be a unit test for the changed behaviour (when there are keys in both parent and child)?

tuncbkose:
Rebased with main

blink1073 (approved):
Thanks!

PR #1987, opened 2023-05-01
Moved ensure_dir_exists to FilesWriter

### LINKED ISSUE #1986: Redundant "Making directory" messages
If a notebook has multiple outputs to be written, the making directory message is repeated. This is misleading since the directory is created only once.

## Reproduce
notebook cells:
```python
import matplotlib.pyplot as plt
```
```python
plt.plot(range(3))
```
```python
plt.plot(range(3))
```

result:
```
$ jupyter nbconvert --to markdown test.ipynb
[NbConvertApp] Converting notebook test.ipynb to markdown
[NbConvertApp] Support files will be in test_files/
[NbConvertApp] Making directory test_files
[NbConvertApp] Making directory test_files
[NbConvertApp] Writing 241 bytes to test.md
```

## Why does this happen?

The outputs are written in 
https://github.com/jupyter/nbconvert/blob/d519784fe13c40311135f38967419a612c68f3b8/nbconvert/writers/files.py#L86-L95

where this function is called for each file
https://github.com/jupyter/nbconvert/blob/d519784fe13c40311135f38967419a612c68f3b8/nbconvert/writers/files.py#L48-L52

So every call to `self._makedir` logs a message independent of whether the directory exists or not.

## Potential fixes

- Do nothing: This is only a minor annoyance that doesn't affect functionality, and I didn't notice anyone else complain about it.
- In `self._makedir` (or wherever this method is called), check if `path` exists: I think this would lead to the logging behavior I would expect, but it raises another (albeit unimportant) issue. `ensure_dir_exists` checks if `path` exists. So there will either be a redundant check, or we can modify `ensure_dir_exists` and (rename it something like `makedir`). I think it is fine to modify `ensure_dir_exists` since it has only one other usage, also in `writer/files.py`, and that usage can probably be replaced by `self._makedir` as well.
- pass `self.log` to `ensure_dir_exists` as an optional argument so that one can log from there

tuncbkose:
Closes #1986 by moving `ensure_dir_exists` from `utils.io` to `FilesWriter`.

### DIFF
--- a/nbconvert/utils/io.py
+++ b/nbconvert/utils/io.py
@@ -117,21 +117,3 @@ def link_or_copy(src, dst):
         # Either link isn't supported, or the filesystem doesn't support
         # linking, or 'src' and 'dst' are on different filesystems.
         shutil.copy(src, dst)
-
-
-def ensure_dir_exists(path, mode=0o755):
-    """ensure that a directory exists
-
-    If it doesn't exist, try to create it and protect against a race condition
-    if another process is doing the same.
-
-    The default permissions are 755, which differ from os.makedirs default of 777.
-    """
-    if not os.path.exists(path):
-        try:
-            os.makedirs(path, mode=mode)
-        except OSError as e:
-            if e.errno != errno.EEXIST:
-                raise
-    elif not os.path.isdir(path):
-        raise OSError("%r exists but is not a directory" % path)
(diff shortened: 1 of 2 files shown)


tuncbkose:
Failures are unrelated? The tests should probably be rerun

blink1073 (approved):
Thanks again!  I'll make a release on Monday.

PR #1997, opened 2023-06-01
bump pandoc max version

### LINKED ISSUE #1996: support for pandoc > 3.0.0
pandoc 3.0.0 has been released early 2023:

https://pandoc.org/releases.html

I get a warning about a non-supported version, but for my use case (converting jupyter notebooks for sphinx via https://nbsphinx.readthedocs.io) I did not encounter any issues yet.

tuncbkose:
Closes #1996.

Changes:
- `soul` package is added to the template ([relevant](https://github.com/jgm/pandoc/issues/8411))
  - There is an additional change relevant to this [here](https://github.com/jgm/pandoc/issues/8707) made for `lualatex`, but it seems to cause additional issues. I am not sure what is the best way forward. Given that `lualatex` is not the default engine, I opted to not include said change.

The tests are passing and skimming through the [release notes](https://pandoc.org/releases.html#pandoc-3.0-2023-01-18) I didn't see anything concerning (though I am not the expert).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -42,7 +42,7 @@ jobs:
           sudo apt-get install xvfb x11-utils libxkbcommon-x11-0  libxcb-xinerama0 python3-pyqt5
 
           # pandoc is not up to date in the ubuntu repos, so we install directly
-          wget https://github.com/jgm/pandoc/releases/download/2.14.2/pandoc-2.14.2-1-amd64.deb && sudo dpkg -i pandoc-2.14.2-1-amd64.deb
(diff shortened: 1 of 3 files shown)


blink1073:
I pushed a commit to see what happens with older pandoc.  I think we'll want to download the lowest supported version in our "minimum versions" job.

blink1073 (approved):
Thank you!

tuncbkose:
> I pushed a commit to see what happens with older pandoc. I think we'll want to download the lowest supported version in our "minimum versions" job.

That makes sense, thank you for catching it.

PR #2002, opened 2023-06-08
Ensure TEXINPUTS is an absolute path

### LINKED ISSUE #2001: bug in TEXINPUTS
## Summary

Converting a notebook to pdf when the notebook is not in the current directory sets `TEXINPUTS` incorrectly, which can create problems if local files are referenced.

## Explanation

During notebook to pdf conversion, converting the intermediate latex file to pdf is done in a temporary directory. While doing so, `TEXINPUTS` environment variable is used to point back to the original directory of the notebook so that local files (like images that may be used with `\includegraphics`) will be visible to the TeX engine:
https://github.com/jupyter/nbconvert/blob/209d853f9162bdfceea3d17c9cc83ecb4c2cbe1f/nbconvert/exporters/pdf.py#L188-L191

Thus, the value depends on `resources["metadata"]["path"]`, which is set here:
https://github.com/jupyter/nbconvert/blob/209d853f9162bdfceea3d17c9cc83ecb4c2cbe1f/nbconvert/exporters/exporter.py#L185-L188

Consider the following setting:
```
-parent_dir
|-dir
||-notebook.ipynb
||-image.png
```
where the notebook contains something like ``

If one does
```
$ cd dir
$ jupyter nbconvert --to pdf notebook.ipynb
```
`path` will be set to `None`, and thus `TEXINPUTS` will use `os.getcwd()` and everything is fine. However, if one does
```
cd parent_dir
jupyter nbconvert --to pdf dir/notebook.ipynb
```
`path` will be set to `dir`, so will be `TEXINPUTS`. Consequently, the conversion to pdf will be requested effectively with
```
$ cd temp_dir
$ TEXINPUTS=dir xelatex ...
```
The problem is that since we are in `temp_dir` and TEXINPUTS doesn't have an absolute path as a value, it doesn't find the right location.

tuncbkose:
Fixes #2001. At the time of `self.texinputs` being set, we are still in the directory where `nbconvert` is called, so this fix should resolve correctly.

~To-do: Add tests~

### DIFF
--- a/nbconvert/exporters/pdf.py
+++ b/nbconvert/exporters/pdf.py
@@ -186,7 +186,7 @@ class PDFExporter(LatexExporter):
         latex, resources = super().from_notebook_node(nb, resources=resources, **kw)
         # set texinputs directory, so that local files will be found
         if resources and resources.get("metadata", {}).get("path"):
-            self.texinputs = resources["metadata"]["path"]
+            self.texinputs = os.path.abspath(resources["metadata"]["path"])
         else:
             self.texinputs = os.getcwd()
 
--- a/nbconvert/exporters/tests/test_pdf.py
+++ b/nbconvert/exporters/tests/test_pdf.py
@@ -8,6 +8,7 @@ import shutil
 from tempfile import TemporaryDirectory
 
 from ...tests.utils import onlyif_cmds_exist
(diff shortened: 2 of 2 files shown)


blink1073 (approved):
Thank you!

PR #2013, opened 2023-06-28
Adopt playwright

brichet:
Since [pyppeteer](https://github.com/pyppeteer/pyppeteer) is now unmaintained we should adopt playwright instead.


brichet:
I have no clue on this test failure too (occurs locally also):
https://github.com/jupyter/nbconvert/actions/runs/5402860958/jobs/9814806087?pr=2013#step:5:379

Don't know why it used to pass as `exporter.from_file(f, resources={})` is supposed to return a tuple https://github.com/jupyter/nbconvert/blob/50b86f8755fe45c5ad0473727fc102ce957e7c1b/nbconvert/exporters/exporter.py#L203-L205

but `exporter.from_notebook_node(nb)` expects a `NotebookNode` (it lead to `exporter.html.from_notebook_node(nb)`):
https://github.com/jupyter/nbconvert/blob/50b86f8755fe45c5ad0473727fc102ce957e7c1b/nbconvert/exporters/html.py#L227-L229

blink1073:
I'm removing Python 3.7 support in #2008.  I'm updating this branch to pick up those changes.

blink1073 (approved):
Thanks again!

brichet:
Thanks @blink1073 for fixing the tests

nbconvert/exporters/webpdf.py
```diff
@@ -65,40 +68,44 @@ def _template_name_default(self):
...
 
         async def main(temp_file):
-            """Run main pyppeteer script."""
+            """Run main playwright script."""
             args = ["--no-sandbox"] if self.disable_sandbox else []
-            browser = await self._check_launch_reqs()(
-                handleSIGINT=False, handleSIGTERM=False, handleSIGHUP=False, args=args
-            )
-            page = await browser.newPage()
-            await page.emulateMedia("print")
```
fcollonval, line 96: It seems that that option was lost in the upgrade process. As the option is [available in playwright](https://playwright.dev/python/docs/api/class-page#page-emulate-media), I guess it will be better to restore it to increase backward compatibility.

cc @brichet

brichet: Thanks @fcollonval. I opened https://github.com/jupyter/nbconvert/pull/2022

Issue #1993, opened 2023-05-13
Provide default alt text for code generated images


rgaiacs:
Consider a cell with

```python
from matplotlib import pyplot as plt

plt.plot([1, 2, 3], [1, 2, 3]);
```

The default rendering of the output is

```
<div class="output_wrapper">
<div class="output">


<div class="output_area">

    <div class="prompt"></div>




<div class="output_png output_subarea ">
<img src="data:image/png;base64," >
</div>

</div>

</div>
</div>

</div>
```

Should we have an default `alt` text?

```
<div class="output_wrapper">
<div class="output">


<div class="output_area">

    <div class="prompt"></div>




<div class="output_png output_subarea ">
<img src="data:image/png;base64," alt="Image generated using the above code." >
</div>

</div>

</div>
</div>

</div>
```

Related issues and pull request: https://github.com/jupyter/nbconvert/pull/1112

cc @Jenneh @isabela-pf

tonyfast:
I'll agree that adding alt text is meaningful. Not all images need alt, and the existence of `alt` text differentiates between  [decorative images](https://www.w3.org/WAI/tutorials/images/decorative/) and [informative images](https://www.w3.org/WAI/tutorials/images/informative/). When images or figures are used in outputs we can assume they are informative and a lack of alt text is a impending violation.

With that said, coming up with a meaningful default message is completely ill defined. So yes we should have `alt` text, but what that should be is hard to imagine. `alt` should describe the image itself and we need to avoid misusing alt text. We can't make assumptions about the compute context or infer the meaning of the figure. The best we can say is "there is an informative image here, it is not decorative". Ultimately, we are delivering bad news to visually impaired users that they can't participate in all of the information, but at least the existence of the figure is announced because it is not decorative anymore.

another practical implementation would be some guard rails that produce errors (strict more) or warnings that they are producing inaccessible images or content.

PR #2021, opened 2023-07-10
[Accessibility] some accessibility improvements

brichet:
This PR improves the accessibility of the produced HTML file, using the `HTMLExporter`:
- adds a main landmark for the Notebook content
- adds a lang attribute to the `html` tag, configurable with a trait
- adds missing alternative text on images produced by matplotlib
- adds `tabindex="0"` to the inputs and outputs, in order to be able to navigate through the Notebook using the tab key


mgeier:
I'm not a user of a screenreader, so I might be missing something, but how is `alt="Image"` helping anyone? Doesn't a screenreader already tell you it's an image? Wouldn't it say "image" twice? Assuming the screenreader is even used by an english-speaking user ...

I don't know a solution here, but I would like to point out previous discussions: #1993, https://github.com/matplotlib/matplotlib/pull/21328

brichet:
> I'm not a user of a screenreader, so I might be missing something, but how is `alt="Image"` helping anyone?

Thanks for pointing out some discussions.

I agree that it's not really useful as such.
Nevertheless, it seems that the screenreader needs it, at least the Deque University Axe Devtools complains about it. But for informative image, the alternative text should be informative: https://dequeuniversity.com/rules/axe/4.4/image-alt, which is not the case here.

I don't think we can even tell that it's an image from matplotlib when building HTML, so I don't know neither how we can have any information about the content.
The best solution is to have if fixed from matplotlib.

blink1073 (approved):
Thank you!

krassowski:
> I agree that it's not really useful as such.
Nevertheless, it seems that the screenreader needs it, at least the Deque University Axe Devtools complains about it.

I don't think that blindly satisfying the automated checks here is the right approach. This will provide a false sense of satisfying the requirements. If a notebook contains images without alt the check should error out.

brichet:
> If a notebook contains images without alt the check should error out.

The problem comes from images generated by Matplotlib (at least), and I don't know if it can be fixed by a user currently.
Maybe we could at least have a more consistent text, something like "No description is provided for this image", and print a warning when the HTML is generated.

blink1073:
I'm cutting the release, we can follow up with further improvements in a bug fix release.

PR #2022, opened 2023-07-17
Restore 'media=print' option

brichet:
Follow up the playwright adoption to restore a option that has been removed.

Ref: https://github.com/jupyter/nbconvert/pull/2013#discussion_r1265239380


blink1073 (approved):
Thanks!

PR #2024, opened 2023-07-19
Show a warning if an image has no alternative text

brichet:
This PR changes the default alternative text for images, when exporting to HTML, and it also adds a warning in logs.

It also adds test on the warning, and some tests for the language code (which were missing).

### DIFF
--- a/nbconvert/exporters/html.py
+++ b/nbconvert/exporters/html.py
@@ -264,7 +264,7 @@ class HTMLExporter(TemplateExporter):
             elem.attrs["alt"] = "No description has been provided for this image"
             missing_alt += 1
         if missing_alt:
-            self.log.warning(f"Alternative text is missing on {missing_alt} image(s)")
+            self.log.warning(f"Alternative text is missing on {missing_alt} image(s).")
         # Set input and output focusable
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
Thank you!

0x26res:
@brichet I've recently updated nbconvert and I started getting warnings.

```
[NbConvertApp] WARNING | Alternative text is missing on 2 image(s).
```

So, it makes sense to have these warnings, but I'm not entirely sure how to fix them. 

For context I use simple plots from pandas, using `df.plot(title='foo').

From what I can see in [pandas's doc](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.plot.html) and [matplotlib's](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.plot.html) I don't think there's a way to provide an alternative text.

I found an example that overwrite the alternative text, but it's a lot of work: https://gist.github.com/tonyfast/d01a0248979b21c9484d0827e04c0e68

I'm wondering if `nbconvert` could provide an option to disable that warning?

brichet:
@0x26res would you mind opening a separated issue for it.
It should probably be discussed as the issue mainly come from matplotlib and pandas.

PR #2028, opened 2023-08-03
env var for extra_template_basedirs

### LINKED ISSUE #2026: Env variable for extra_template_basedirs
Hi team,

I want to be able to set extra_template_basedirs via some env variable. I am using in an internal infrastructure that needs this and some external library seems to use PythonExporter to convert a notebook into python.

Would be helpful to set this globally so this gets automatically picked up.

ndmlny-qs:
This commit add the following three items.

- Updates the method `_default_extra_template_basedirs` in the module `templateexporter.py` to search for a new environment variable called `NBCONVERT_EXTRA_TEMPLATE_BASEDIRS`. This environment variable will be added to the basedirs list if it is found, and it exists.
- Updates the tests to ensure if the new environment variable exists then it is included in the `TemplateExporter.template_paths`.
- Adds a context manager method that modifies the `os.environ` dictionary.

Closes #2026

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -260,7 +260,13 @@ class TemplateExporter(Exporter):
 
     @default("extra_template_basedirs")
     def _default_extra_template_basedirs(self):
-        return [os.getcwd()]
+        basedirs = [os.getcwd()]
+        env = os.environ
+        if "NBCONVERT_EXTRA_TEMPLATE_BASEDIR" in env:
+            extra_template_path = env["NBCONVERT_EXTRA_TEMPLATE_BASEDIR"]
+            if os.path.exists(extra_template_path):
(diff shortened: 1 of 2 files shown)


ndmlny-qs:
@Carreau let me know if you'd like additional items added or any changes or more tests. Thanks!

Carreau:
I have two worries: 

1) `extra_template_basedirs` is already a convenience config on top of `extra_template_paths`
2) we're modifying the default value, so if someone also use CLI or config file there will likely be conficts/unexpected behavior. 

Let me think about it a bit.

Carreau:
I've posted a followup in #2026 as I'd like to better understand the use case.
I'd like to avoid one-off patches if we can have a general one, and I think the use case is already covered by how traitlets can be configured, so just checking why and seeing if we can offer a more generic solution.

PR #2036, opened 2023-08-26
Fix: Prevent error from all whitespace lang string

Adamtaranto:
Add a check to make sure lang string is not an empty list after stripping out whitespace characters. Prevents error when user has a space after code block docstring but no language name.

Fixes jupyter/nbgrader#1689


Adamtaranto:
Can someone please add a bug label to this?

blink1073 (approved):
Thank you!  I'll cut a release with this fix tomorrow.

PR #2037, opened 2023-08-28
update Python version support in docs

minrk:
got out of date

not sure how best to ensure this stays in sync, since it needs to change every year. Maybe a comment where these restrictions are to note that install.rst needs updating when versions change?


blink1073:
Or we could remove this part of the docs altogether and rely on tooling like pip and PyPI:

image

bollwyvl:
Perhaps this could be dogfooded by porting `install.rst` to `install.ipynb`, and generating this (and probably other lines) from an appropriate source of truth, e.g. `pyproject.toml` or the `python-version` factor in the CI matrix.

ivanov (approved):
sure, we can do other dog fooding stuff, but this is a step forward from the outdated information we had before. 

But I think we should keep this information in the docs, because that's what packagers will bundle and make available to users in Linux distributions, for example. People don't always have access to the internet when they have access to these docs.

PR #2055, opened 2023-10-05
Restore ResourcesDict to the public API

### LINKED ISSUE #2054: Change in namespace / API for 7.9.0?
Over at https://github.com/matthew-brett/nb2plots, we were relying on the previous `nbconvert` namespace having `nbconvert.exporter.ResourceDict`, but I think you've just removed that for the latest release, resulting in an error in our code: https://github.com/matthew-brett/nb2plots/issues/32

I guess you intended to remove it - but isn't that an API change?  Should I expect that in a point release?

blink1073:
Fixes #2054

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1317,6 +1317,7 @@ raw template
 {%- endblock in_prompt -%}
     """
 
+
 exporter_attr = AttrExporter()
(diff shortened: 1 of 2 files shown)


matthew-brett:
Thanks - that's very helpful.

PR #2058, opened 2023-10-27
Update to mermaid 10.6.0, docs keyboard navigation

bollwyvl:
## References

- for parity with https://github.com/jupyterlab/jupyterlab/pull/15302

## Changes

- [x] update to `mermaidjs 10.6.0` for bug fixes, new charts
- [x] add `navigation_with_keys` to `conf.py` to avoid new warning and restore browser-native navigation

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -21,3 +21,5 @@ build:
   os: ubuntu-22.04
   tools:
     python: "3.11"
+  apt_packages:
+    - pandoc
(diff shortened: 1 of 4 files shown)


bollwyvl:
It looks like the `pandoc` is no longer supported... will roll that back, but perhaps we should consider the `mamba` provisioner.

blink1073 (approved):
Thanks!

PR #2063, opened 2023-10-31
Support es modules in js includes

timkpaine:
Straightforward update, adds a second argument to `include_js` resource to add module type to script tag. Tested on a real example to confirm working, and added a simple test to ensure the right html is created (Though this test does not run in a browser).


blink1073 (approved):
LGTM, thank you!

timkpaine:
thanks @blink1073!

PR #2075, opened 2023-11-28
Allow to load config from env.

Carreau:
Latest version of traitlets allow to load config from env,

Env loading is opt-in for backward compatibility.

### DIFF
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -357,6 +357,9 @@ class NbConvertApp(JupyterApp):
 
         self.init_syspath()
         super().initialize(argv)
+        return self.visit_unknown(node)
+        if hasattr(self, "load_config_environ"):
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Thanks!  I'll cut a release tomorrow.

blink1073:
I've been following a continuous release policy, it's been easier to reason about across the different repos.

jmmonteiro:
Hello,

I'm not sure this PR solves my problem. I'm running nbconvert version `7.12.0`.
I have a jupyter notebook that loads  `.env` file, and `nbconvert` is not working with it.
Currently I'm running it like this:
```bash
jupyter nbconvert --execute --to notebook --inplace my_notebook.ipynb
```
What command line arguments would I have to run to load the env file?

I've tried things like `--load-config-environ .env` and it did not work.

PR #2079, opened 2023-12-10
Update ruff config

blink1073:


mgeier:
In `nbconvert/preprocessors/execute.py`, this removed the backwards-compatibility line:

```python
from nbclient.exceptions import CellExecutionError  # noqa
```

Was this intentional?

Interestingly, it did not remove the comment for that line:

```python
# Backwards compatibility for imported name
```

See also https://github.com/spatialaudio/nbsphinx/issues/466#issuecomment-1865968622.

Issue #1892, opened 2022-10-27
Bleach seems to be significantly slower than lxml in 7.1.x+


MgenGlder:
### Description
After updating `nbconvert` from 6.5.0 to 7.2.2 for use here at GitHub we discovered the notebook rendering service degraded in render time (almost twice as long to render). We looked into what might have caused this and it looks as though the switch to `bleach` for sanitization is the culprit here. I'm wondering if we could provide a configuration for users to pass which cleaning library they would prefer to use considering it could have a major performance impact.

### Screenshots
7.0.0 Profile
![Screen Shot 2022-10-27 at 2 32 22 PM](https://user-images.githubusercontent.com/8964151/198375073-761ff7f3-88ee-408e-b06b-03645568c16b.png)

7.0.0 Flamegraph
![Screen Shot 2022-10-27 at 2 31 18 PM](https://user-images.githubusercontent.com/8964151/198374995-51fdeaff-ea57-40f5-b5c0-a5f6b754df45.png)

7.2.2 Profile
![Screen Shot 2022-10-27 at 12 26 31 PM](https://user-images.githubusercontent.com/8964151/198375132-46a47b21-d0e6-487f-a446-08146c1475f6.png)
7.2.2 Flamegraph
![Screen Shot 2022-10-27 at 12 26 14 PM](https://user-images.githubusercontent.com/8964151/198375157-ebd095ed-741f-498b-82b4-70154eceaa6c.png)

gwincr11:
Interestingly we (GitHub) looked at using bleach as a second sanitization step on this project and decided against it because it caused too many notebooks to timeout and increased render times my more than 75%.

bollwyvl:
Along the same lines of `python-fast-jsonschema` on `nbformat`, it seems reasonable to have a hard dependency on `bleach`, and  `nbconvert[fast]` to force the install of `lxml` (or whatever ends up being the fastest, maybe some rust-based thing). This would prefer the faster library, or behind `NBCONVERT_FAST=0` and/or a traitlet.

The two could then be pitted against each other during test, probably after some further homogenization.

`lxml` isn't really that uncommon a dependency, and is rather widely supported (including `pyodide`).

akx:
[...]

PR #2083, opened 2023-12-18
Add table, td, tr to allowed list of tags

yuvipanda:
pygments emits line numbers via tables, with tr and td elements (https://pygments.org/docs/formatters/#HtmlFormatter). lxml's clean_html considers table, tr and td as safe elements, but with https://github.com/jupyter/nbconvert/pull/1854 they are now considered unsafe. So instead of displaying line numbers, the table, tr and td elements are escaped, and show up as literal HTML if trying to enable line numbers via the method introduced in https://github.com/jupyter/nbconvert/pull/1683.

This PR adds table, tr and td as safe elements so that line numbers can continue to work.

I know that there are probably plans to move away from bleach (https://github.com/jupyter/nbconvert/issues/1892), but this is a small and focused change so hopefully doesn't need to block on

Without this change:

image

With this change:

image

### DIFF
--- a/nbconvert/filters/strings.py
+++ b/nbconvert/filters/strings.py
@@ -89,7 +89,7 @@ def clean_html(element):
         kwargs["css_sanitizer"] = css_sanitizer
     return bleach.clean(
         element,
-        tags=[*bleach.ALLOWED_TAGS, "div", "pre", "code", "span"],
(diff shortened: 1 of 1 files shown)


yuvipanda:
I don't think any of the test failures are related to this PR.

I did open https://github.com/jupyter/nbconvert/pull/2084 to fix the link check failures.

blink1073 (approved):
Thank you!

PR #2084, opened 2023-12-18
Remove twitter links that cause linkcheck to fail

yuvipanda:
Twitter links now cause linkcheck to fail, and I can't access the contents of those links from my browser either (I believe they are behind a paywall / loginwall now). So just removing the links.

See https://github.com/jupyter/nbconvert/actions/runs/7244142071/job/19731995736?pr=2083 for a check_links failure


blink1073 (approved):
Thanks!

yuvipanda:
Thanks a lot @blink1073!

PR #2094, opened 2024-01-09
Fix broken image scaling in case a custom width or height is provided for the image

AndSte01:
With the current implementation, if a custom image width or height is specified 
``` markdown
![alt](./some/img.png){ width=80% }
```
the image gets distorted during pdf export.

This pull request fixes the issue. It doesn't interfere with the feature that prevents images from overflowing the page.

Before:
grafik

After:
grafik


blink1073 (approved):
Thank you!

PR #2095, opened 2024-01-10
Allow pre-fetch of css files without attempting download

### LINKED ISSUE #2093: Allow pre-fetching of css files?
As far as I can tell, nbconvert cannot be built on non-internet connected systems, such as in an automated build system on an airgapped node. In hatch_build.py, it tries to download each css file (on at least one system I'm working on, the `urlopen(url).read()` function hangs indefinitely) _before_ checking for an existing copy. Would it be possible to check for the destination file before trying to download it? In my use case (installing through Spack), it's fine to stage the files in advance so they're available at install time.

AlexanderRichert-NOAA:
This PR allows for pre-fetching of css files (outside of the build system) without first attempting a download (which on some non-internet connected systems may cause the installation to hang indefinitely). Specifically, it rearranges the logic in hatch_build.py, checking to see whether the destination file exists before attempting a download (if the destination exists, then the `_get_css_file` function returns). I have tested installing on internet-connected systems (i.e., the normal way), as well as on non-internet connected systems with the css files pre-fetched through Spack (as resources).

Fixes #2093

### DIFF
--- a/hatch_build.py
+++ b/hatch_build.py
@@ -43,6 +43,9 @@ def _get_css_file(template_name, url, filename):
     """Get a css file and download it to the templates dir"""
     directory = osp.join(templates_dir, template_name, "static")
     dest = osp.join(directory, filename)
+    if osp.exists(dest):
+        print("Already have CSS: %s, moving on." % dest)
+        return
     if not osp.exists(directory):
         os.makedirs(directory)
     print("Downloading CSS: %s" % url)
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
LGTM, thank you!

blink1073:
I'll make a patch release tomorrow.

AlexanderRichert-NOAA:
Great, thanks!

PR #2104, opened 2024-02-05
Support configureable width and height of reveal presentations

franzhaas:
This is important, as reveal keeps the ratio, wasting serious monitor space if this is not configured correctly...


franzhaas:
This is an enhancement, especially for the use of reveal presentations without a server... Unfortunately i did not manage to set the label...

blink1073 (approved):
Thank you!

PR #2106, opened 2024-02-07
Add a toggle to skip the svg encoding of images

gwincr11:
The svg being encoded to base64 introduced in https://github.com/jupyter/nbconvert/pull/2018 is not a secure option for GitHub's use case. This pr adds an option to bypass this encoding so we can continue to use the standard svg output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Enhancements made
 
-- Support configureable width and height of reveal presentations [#2104](https://github.com/jupyter/nbconvert/pull/2104) ([@franzhaas](https://github.com/franzhaas))
+- Support configurable width and height of reveal presentations [#2104](https://github.com/jupyter/nbconvert/pull/2104) ([@franzhaas](https://github.com/franzhaas))
 
 ### Maintenance and upkeep improvements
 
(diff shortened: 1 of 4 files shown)


blink1073 (approved):
Thank you!

gwincr11:
Thanks!

jstorrs:
Can you provide some information about how base64 encoded XML is "not a secure option"?

Everything I've read is the opposite--bare svg directly encorporated into the DOM is a major security problem since the SVG becomes indistinguishable from the DOM and has full access. Malicious SVG can include javascript, etc. Base64 encoding the svg specifically triggers isolation in browsers. That isolation does not exist when incorporated directly into the page as this flag enables.

I have no issue with the flag being made available, but I find the comment to be very misleading. This flag enables vulnerabilities and does not increase security at all. It should be clearly marked as enabling unsafe behavior.

gwincr11:
I am not on the security teams so I am not the expert here. What I can say is that in reviewing attack vectors in python notebooks when building the notebook rendering service, base64 encoded data and image attributes were called out as a particular concern, so we strip all base64 encoded data regardles of where it is on the page. https://subfn.net/2020/04/14/base64-tricks/

We can more easily detect svg images and clear out what we don't want, the obfuscation of base64 makes it impossible to know what is in the payload.

PR #2107, opened 2024-02-07
Update project description

ivanov:
This description is what gets show in results of PyPI search, and also gets picked up verbatim as the package description in some operating systems (found it in FreeBSD ports, also Arch)


blink1073 (approved):
Thanks!

PR #2120, opened 2024-03-02
- changing mermaid suport

franzhaas:
the already existing mermaid support, did not work for me.

It showed me "syntax error bomb displays", which where displayed just milliseconds after the correctly rendered graph was shown.

I added the reveal mermaid plugin, and removed the already existing mermaid initialisation.

While this does work for me, it is probably quite a change and i really would like to get some input on how to continue, or change my environment to fit the preexisting solution (which is still used for the hard parts, i basically swap pure mermaid for the plugin, as a result of trial and error)

### DIFF
--- a/nbconvert/exporters/slides.py
+++ b/nbconvert/exporters/slides.py
@@ -119,12 +119,6 @@ class SlidesExporter(HTMLExporter):
         """
     ).tag(config=True)
 
-    reveal_mermaid_js_url = Unicode(
-        help="specifies the mermaid reveal plugin url. defaults to https://unpkg.com/reveal.js-mermaid-plugin@2.2.0",
-    ).tag(
-        config=True,
-    )
-
     @default("reveal_url_prefix")
     def _reveal_url_prefix_default(self):
         if "RevealHelpPreprocessor.url_prefix" in self.config:
@@ -135,6 +129,16 @@ class SlidesExporter(HTMLExporter):
             )
             return self.config.RevealHelpPreprocessor.url_prefix
         return "https://unpkg.com/reveal.js@4.0.2"
+
+    reveal_mermaid_js_url = Unicode(
+        help="Specifies the mermaid reveal plugin url.",
+    ).tag(
+        config=True,
+    )
+
+    @default("reveal_mermaid_js_url")
+    def _reveal_mermaid_js_url(self):
+        return "https://unpkg.com/reveal.js-mermaid-plugin@2.2.0"
 
     reveal_theme = Unicode(
         "simple",
--- a/nbconvert/exporters/slides.py
+++ b/nbconvert/exporters/slides.py
@@ -138,7 +138,7 @@ class SlidesExporter(HTMLExporter):
 
     @default("reveal_mermaid_js_url")
     def _reveal_mermaid_js_url(self):
-        return "https://unpkg.com/reveal.js-mermaid-plugin@2.2.0"
+        return "https://cdn.jsdelivr.net/npm/reveal.js-mermaid-plugin@2.2.0"
 
     reveal_theme = Unicode(
         "simple",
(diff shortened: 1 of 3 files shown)


share/templates/reveal/index.html.j2
```diff
@@ -145,21 +141,21 @@ require(
     },
     [
       "{{ reveal_url_prefix }}/dist/reveal.js",
-      "{{ reveal_url_prefix }}/plugin/notes/notes.js"
+      "{{ reveal_url_prefix }}/plugin/notes/notes.js",
+      "https://cdn.jsdelivr.net/npm/reveal.js-mermaid-plugin@2.2.0/plugin/mermaid/mermaid.js"
```
bollwyvl, line 145: I would recommend hoisting this to a configurable on the python side:

```py
# nbconvert/exporters/slides.py
reveal_url_prefix = Unicode(
  help="..."
).tag(config=True)

reveal_mermaid_js_url = Unicode(
  default="https://cdn.jsdelivr.net/npm/reveal.js-mermaid-plugin@2.2.0/plugin/mermaid/mermaid.js",
  help="...",
).tag(config=True)
```

```js
[
      "{{ reveal_url_prefix }}/dist/reveal.js",
      "{{ reveal_url_prefix }}/plugin/notes/notes.js",
      {% if reveal_mermaid_js_url %}"{{ reveal_mermaid_js_url }}",{% endif %}
]
```

And then below:

```js
plugins: [
  RevealNotes,
  {% if reveal_mermaid_js_url %}RevealMermaid,{% endif %}
]
```

franzhaas: ok thanks, makes sense, I did it.

Any further things I can do or should know?

Thanks again,
Franz

share/templates/reveal/index.html.j2
```diff
@@ -145,21 +141,24 @@ require(
     },
     [
       "{{ reveal_url_prefix }}/dist/reveal.js",
-      "{{ reveal_url_prefix }}/plugin/notes/notes.js"
+      "{{ reveal_url_prefix }}/plugin/notes/notes.js",
+      {% if reveal_mermaid_js_url %}"{{ reveal_mermaid_js_url }}",{% endif %},
     ],
 
-    function(Reveal, RevealNotes){
+    function(Reveal, RevealNotes, RevealMermaid){
```
bollwyvl, line 148: sorry, missed this: the positional argument would need to be conditional as well:

```js
function(
    Reveal, 
    RevealNotes, 
    {% if reveal_mermaid_js_url %}RevealMermaid,{% endif %}
){
```

Also feel free to normalize the whitespace: the "floating indent" style isn't necessary, just consistency with what's around (in this case four spaces).

franzhaas: good idea, i also did that...

franzhaas: Can someone please assign the improvement label? I am sorry, i either don't have the right to set it, or i just don't find the feature...

share/templates/reveal/index.html.j2
```diff
@@ -145,21 +143,26 @@ require(
     },
     [
       "{{ reveal_url_prefix }}/dist/reveal.js",
-      "{{ reveal_url_prefix }}/plugin/notes/notes.js"
+      "{{ reveal_url_prefix }}/plugin/notes/notes.js",
+      {% if reveal_mermaid_js_url %}"{{ reveal_mermaid_js_url }}",{% endif %},
```
bollwyvl, line 147: there's an extra comma after the `endif`

nbconvert/exporters/slides.py
```diff
@@ -119,6 +119,12 @@ def _template_extension_default(self):
         """
     ).tag(config=True)
 
+    reveal_mermaid_js_url = Unicode(
+        help="specifies the mermaid reveal plugin url. defaults to https://unpkg.com/reveal.js-mermaid-plugin@2.2.0",
```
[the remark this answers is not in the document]
franzhaas: I am unsure if i understood you correctly... do you mean those @default()?

anyhow i tried..

franzhaas:
Is there still something I need to do? Something i overlooked?

Thanks,
Franz

franzhaas:
- i believe to much changes happened in the meantime...

PR #2089, opened 2023-12-26
Convert `coalescese_streams` function to `CoalesceStreamsPreprocessor`

ryan-williams:
`coalescese_streams` was the last remaining "decorated function Preprocessor", and I couldn't find an example of how to use it.

Here it is converted to be a `Preprocessor` subclass, like the others. A top-level `--coalesce-streams` flag is also added, so this now works:

```bash
jupyter nbconvert --coalesce-streams --inplace notebook.ipynb
```

to normalize a notebook's otherwise-nondeterministic stdout/stderr chunking.

For example, a cell might end up with two `stderr` lines split across two `outputs`:

```json
{
  "cells": [
    {
      "cell_type": "code",
      "outputs": [
        {
          "name": "stderr",
          "output_type": "stream",
          "text": [
            "before sleep\n"
          ]
        },
        {
          "name": "stderr",
          "output_type": "stream",
          "text": [
            "after sleep\n"
          ]
        }
      ],
      "source": [
        "import sys\n",
        "import time\n",
        "sys.stderr.write(\"before sleep\\n\")\n",
        "time.sleep(1)\n",
        "sys.stderr.write(\"after sleep\\n\");"
      ]
    }
  ]
}
```

(I usually see this when running notebooks via [Papermill](https://papermill.readthedocs.io/en/latest/))

`jupyter nbconvert --coalesce-streams --inplace` combines consecutive `outputs` from the same `stream`, within each cell:

```diff
## inserted before /cells/0/outputs/0:
+  output:
+    output_type: stream
+    name: stderr
+    text:
+      before sleep
+      after sleep

## deleted /cells/0/outputs/0-1:
-  output:
-    output_type: stream
-    name: stderr
-    text:
-      before sleep
-  output:
-    output_type: stream
-    name: stderr
-    text:
-      after sleep
```

I also implemented this functionality in [`juq merge-outputs`](https://github.com/runsascoded/juq?tab=readme-ov-file#juq-merge-outputs), before noticing this preprocessor!

PR #2142, opened 2024-04-28
rst exporter: Re-enable coalescing streams

mgeier:
This has been (I guess inadvertently) disabled in #2089.

Other exporters might be affected as well, but this has been reported for the reST exporter, see #2092.


blink1073 (approved):
Thank you!

PR #2145, opened 2024-05-03
Add eol --newline option for text content in nbconvert.writers.files

### LINKED ISSUE #1062: nbconvert to notebook doesn't respect line endings
On Windows, when I run

    jupyter nbconvert --to notebook --execute --inplace "Untitled25.ipynb"

it uses CRLF line endings, but when when I open and save it directly via Jupyter notebook or Jupyterlab it always saves with LF line endings.

I think that the best solution would be to match the behaviour of Jupyter notebook, i.e. replace the line endings with LF. Alternatively, it should maintain the existing line endings, rather than (presumably) always using "native".

nbconvert 5.5.0, notebook 5.7.8, pandoc 2.7.3, jupyerlab 1.0.1

thisiswhereitype:
closes #1062 

Now users can specify the newline string to be passed to [open](https://docs.python.org/3/library/functions.html#open) for string data.

This is fiddly requiring `--newline $'\r\n'` on Bash to force windows newline.

The default is to fall back onto `os.linesep` as currently used.

I didn't look closely at tests but wanted to see what the CI makes of this.

### DIFF
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -17,7 +17,16 @@ import typing as t
 from textwrap import dedent, fill
 
 from jupyter_core.application import JupyterApp, base_aliases, base_flags
-from traitlets import Bool, DottedObjectName, Instance, List, Type, Unicode, default, observe
+from traitlets import (
+    Bool,
+    DottedObjectName,
+    Instance,
+    List,
+    Type,
+    Unicode,
+    default,
+    observe,
+)
 from traitlets.config import Configurable, catch_config_error
 from traitlets.utils.importstring import import_item
 
@@ -60,6 +69,7 @@ nbconvert_aliases.update(
         "writer": "NbConvertApp.writer_class",
         "post": "NbConvertApp.postprocessor_class",
         "output": "NbConvertApp.output_base",
+        "newline": "NbConvertApp.newline",
         "output-dir": "FilesWriter.build_directory",
         "reveal-prefix": "SlidesExporter.reveal_url_prefix",
         "nbformat": "NotebookExporter.nbformat_version",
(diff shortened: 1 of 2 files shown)


bollwyvl:
This is a great feature.

However, as your example shows, it might require complex shell escaping (windows is truly trash at that). 

Perhaps in addition to the explicit alias which accepts any-old-thing, add two flags:

```
    --unix-line-endings
        Write text files with UNIX LF line endings:      \n
    --windows-line-endings  
        Write text files with Windows CRLF line endings: \r\n
```

bollwyvl:
In addition to _reading_, being able to force _writing_ with a newline variant would be _fantastic_.

thisiswhereitype:
> In addition to _reading_, being able to force _writing_ with a newline variant would be _fantastic_.

Thanks for the feedback - `open` as linked above handles to reading and writing depending on the file pointer `mode` arg. The docs provide a bullet point for how the treatment under r/w. This arugment is illegal in `mode='b'` (binary).

The `--newline` is only passed to output via the `Writer` class as of writing this PR only affects writing.

This is because reading text is universal mode where all line endings are normalised to `\n` by default. This [line](https://github.com/jupyter/nbconvert/blob/d6dc8a57adb1c2ac17090c6382f011eb682d1261/nbconvert/exporters/exporter.py#L200) appears to be reading is done.

If we expose an input flag users can work with systems which have alternative EOL characters which aren't by default normalised, but I don't see that requested anywhere.

NB: some [comments](https://github.com/jupyter/nbconvert/issues/1062#issuecomment-528140715) in the issue thread point out the current behaviour "use-system eol" differs from the ecosystem. 

This pr as is needed but the default behaviours need deprecating in favour of use `LF` to satisfy the discussion, this has been adopted by [VScode Jupyter](https://github.com/microsoft/vscode/pull/133762) .

thisiswhereitype:
The trailets config classes somerwhere are not using the repr as the docs are rendering newlines:
```
--use-crlf-newline
    Implies CLRF newline.
    Equivalent to: [--NbConvertApp.newline=
    ]
--use-lf-newline
    Implies LF newline.
    Equivalent to: [--NbConvertApp.newline=
    ]
```

thisiswhereitype:
Reset bad author on commits

thisiswhereitype:
Rebased

PR #2152, opened 2024-05-30
Fix markdown2asciidoc function for pandoc >= 3.0 (closes #2017)

### LINKED ISSUE #2017: nbconvert export omits markdown cells
I have noticed that newer _base_ and _minimal_ docker images on dockerhub do not export markdown cells to asciidoc format. It seems that this problem has been introduced at some point of version 7.5.0.

For further details, please have a look at this [stackoverflow question](https://stackoverflow.com/questions/76580643/jupyter-notebook-export-omits-markdown).

thomasjm:
Pass different arguments to `pandoc` depending on whether the version is >= 3.0, to fix #2017.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1586,6 +1586,7 @@ raw template
 {%- endblock in_prompt -%}
     """
 
+
 exporter_attr = AttrExporter()
 output_attr, _ = exporter_attr.from_notebook_node(nb)
 assert "raw template" in output_attr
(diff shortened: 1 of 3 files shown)


takluyver:
The fix looks good to me, but I think the [asciidoc tests](https://github.com/jupyter/nbconvert/blob/main/tests/exporters/test_asciidoc.py) should be extended to check for Markdown cells. The notebook used in `test_export` already has Markdown cells, but the test still passes if they're missing from the output.

(I'm also kind of surprised that nbconvert carries on if pandoc fails, but that's a bigger question)

thomasjm:
Okay, added a test. I manually confirmed that the new test fails without the fix in this branch.

takluyver:
Thanks!

PR #2160, opened 2024-06-28
Fix crash when running with optimization

QuLogic:
In that case, `__doc__` is None.

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -151,7 +151,11 @@ class TemplateExporter(Exporter):
     """
 
     # finish the docstring
-    __doc__ = __doc__.format(filters="- " + "\n    - ".join(sorted(default_filters.keys())))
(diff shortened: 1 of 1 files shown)


takluyver:
I think this is actually a good illustration of why we shouldn't care about the `PYTHONOPTIMIZE` or `-O` options. This line has been there with minor tweaks for over a decade (from 70be6cc40b3f94a94fbbe5f84421b6de23b10daf),  and would break as soon as you import nbconvert with optimize level 2. I can't find any issue filed about it before this one, and it looks like you've only noticed it because of a test failure on matplotlib, from an issue that went 5 years between comments.

The 'optimize' options make hardly any performance difference to normal code, anyway. If someone has a use case where they're actually useful, then they're easy enough to support, but I don't think everyone should spend time and thought on them just because they exist.

krassowski (approved):
If that helps someone else running tests, why not?

PR #2162, opened 2024-07-06
Don't die if template path cannot be read

stuaxo:
Ordinary users may not have read access to conf in system directories,

In those cases .exists() will trigger a PermissionError when enumerating the conf directories.

### DIFF
--- a/nbconvert/exporters/templateexporter.py
+++ b/nbconvert/exporters/templateexporter.py
@@ -565,13 +565,19 @@ class TemplateExporter(Exporter):
         for template_name in template_names:
             for base_dir in self.extra_template_basedirs:
                 path = os.path.join(base_dir, template_name)
-                if not prune or os.path.exists(path):
-                    paths.append(path)
+                try:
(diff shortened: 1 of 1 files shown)


stuaxo:
This may be https://github.com/jupyter/jupyter/issues/634

takluyver:
Is this sufficient to avoid the problem you encountered? It looks like there are other `exists()` checks in `get_template_names()` which could also run into permission errors.

(It also seems weird to me that standard system directories under `/usr` wouldn't be readable, but I agree Jupyter should ideally cope with that, even if it's unusual)

stuaxo:
It was sufficient for the bug I was seeing, it's certainly possible there could be other places with the same issue.

stuaxo:
(I also found the fact .exists() can throw a permission error suprising, but it seems to be a thing.

stuaxo:
Cheers, I think this will help others who do what I did and combine homebrew and virtualenvs.

PR #2166, opened 2024-07-26
Directly depend on bleach[css], instead of pulling in tinycss2.

xiacunshun:
Tinycss2 is not used in our code, so it may be better that let it in bleach requirements.


takluyver:
This makes sense to me. I'm just going to close & reopen the PR to re-run the tests with the fix from #2168.

PR #2167, opened 2024-07-30
Allow including `text/x-rst` outputs in rst conversion, transition away from `text/restructuredtext`

takluyver:
Outputs in a notebook can include arbitrary mimetypes. From IPython this can be done with `_repr_mimebundle_` method, for instance. Most of the export formats that nbconvert knows about have a corresponding template block for output already in that format, e.g. `data_latex` or `data_html`. However, there's no template block for rst output.

I've partially generalised this, by adding a `data_native` block to the display priority template rather than a specific `data_rst`. This is meant to hold output in the exporter's native format (its `output_mimetype`), iff there isn't already a dedicated block for that format. I admit that's not the tidiest definition, but it seemed the best way to avoid breaking anything.

The exporter class then has to customise `filter_data_type` to accept its additional mimetype. I've done this in `RSTExporter`. It could be generalised in `TemplateExporter`, but that runs more risk of surprising behaviour for subclasses, which are unlikely to have a `data_native` block in their templates.

### DIFF
--- a/share/templates/rst/index.rst.j2
+++ b/share/templates/rst/index.rst.j2
@@ -44,6 +44,10 @@
 {{ output.text | indent }}
 {% endblock stream %}
 
+{% block data_native %}
+{{ output.data['text/restructuredtext'] }}
+{% endblock data_native %}
+
 {% block data_svg %}
 .. image:: {{ output.metadata.filenames['image/svg+xml'] | urlencode }}
 {% endblock data_svg %}
--- a/share/templates/rst/index.rst.j2
+++ b/share/templates/rst/index.rst.j2
@@ -45,7 +45,7 @@
 {% endblock stream %}
 
 {% block data_native %}
-{{ output.data['text/restructuredtext'] }}
+{{ output.data['text/x-rst'] }}
 {% endblock data_native %}
 
 {% block data_svg %}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -122,6 +122,8 @@ jobs:
 
       - name: Run tests
         run: |
+          # Attempt to work around https://github.com/pypa/pip/issues/12781
+          PIP_CONSTRAINT= hatch env run -e test -- pip install 'pip>=24.2'
           xvfb-run --auto-servernum hatch run test:nowarn || xvfb-run --auto-servernum hatch run test:nowarn --lf
 
   test_prereleases:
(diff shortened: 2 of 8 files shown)


takluyver:
The remaining test failure is because [a pip bug](https://github.com/pypa/pip/issues/12781) broke installing pandocfilters 1.4.1. That's already been fixed in pip 24.2, but the environment that hatch creates still seeems to get pip 24.1. :face_exhaling:

takluyver:
I've got all the CI passing now.

takluyver:
I don't suppose I can interest anyone in reviewing this? :slightly_smiling_face:

share/templates/rst/index.rst.j2
```diff
@@ -44,6 +44,10 @@
 {{ output.text | indent }}
 {% endblock stream %}
 
+{% block data_native %}
+{{ output.data['text/restructuredtext'] }}
```
krassowski, line 48: Should it also support `text/x-rst` and `text/prs.fallenstein.rst`? `text/restructuredtext` seems pretty non-standard to me (although I acknowledge that this is what nbconverte is already using for some reason)

takluyver: Huh, good point. :thinking: 

I think it's best for users to have 1 MIME type for rst, rather than trying to wrangle several. So maybe we should change `RSTExporter.output_mimetype` as well - probably to `text/x-rst`, since that's what [docutils suggests](https://docutils.sourceforge.io/FAQ.html#what-s-the-official-mime-type-for-restructuredtext-data).

What would break if we do that?

- Raw cells in notebooks marked with the mimetype `text/restructuredtext` would stop showing up - but we could work around that by overriding `raw_mimetypes` to add back `text/restructuredtext` for compatibility.
- There's a mechanism for finding templates based on `output_mimetype`, so custom templates marked `text/restructuredtext` would no longer work. I can't imagine that many people have custom rst templates, and it's an easy change, so I'd be inclined to document that as a breaking change rather than adding extra complexity to keep them working.

Does that sound reasonable?

krassowski: I think it is reasonable. If I were doing the change I would probably try to make it as backward compatible as possible.

krassowski: Feel free to ping me for review once you think it is ready

krassowski: I trust your judgement. If your assessment is accurate, I would say it might be ok to release and see if anyone complains and only then add back the extra compatibility layer.

krassowski (approved):
Thank you @takluyver!

share/templates/rst/index.rst.j2
takluyver: Thanks!

PR #2168, opened 2024-07-30
Work around pip 24.1 bug which prevents installing pandocfilters 1.4.1

takluyver:
Originally debugged in #2167, but this fix is separate, and required for CI to pass on main.

pip 24.1 introduced a bug (https://github.com/pypa/pip/issues/12781 ) which breaks installing pandocfilters 1.4.1. pip 24.2 fixed that again.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -122,6 +122,8 @@ jobs:
 
       - name: Run tests
         run: |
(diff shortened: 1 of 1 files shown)


takluyver:
If there are no objections, I'll merge this tomorrow, because I think working CI is important, even if there might be better ways to make it work.

PR #2171, opened 2024-08-19
Fix font for code elements

nchiapol:
Without this change exported html notebooks use a proportional font for the code blocks. Really not what one usually wants.


krassowski:
Can you add before/after screenshots?

nchiapol:
sorry, I do not have the affected setup any more. 

Current versions from my OS (Debian) and pip did seem to work correctly in a minimal test. Feel free to close the request.

krassowski:
Closing as per above

Issue #917, opened 2018-11-18
Bullet list not getting properly generated in PDF


zHaytam:
Hello,

My jupyter notebook contains a markdown with a bullet list:
```Markdown
Some text here:
 - Item 1.
 - Item 2.
 - Item 3.
```

When I generate the PDF using `jupyter-nbconvert jn.ipynb --to pdf`, the result is:
```
Some text here: - Item 1. - Item 2. - Item 3.
```

Is there are way for nbconvert to correctly generate a nice bullet list?
Thank you!

t-makaro:
Trying adding an empty line before the list like:
```
Some text here:

 - Item 1.
 - Item 2.
 - Item 3.
```

ert485:
I experienced this as well. Adding the empty line made it format better. 

It also affects the > in markdown, and even with this fix: now it's on its own line, but is the same as any paragraph (markdown has a grey bar on the indent).

Its possible that's an inherent difference in normal markdown and latex's style, but any difference in the appearance (vs directly dowload as html->print to pdf) is undesirable to some degree

t-vi:
This isn't PDF-specific. I'm using [nbsphinx](https://github.com/spatialaudio/nbsphinx) and it seems like the conversion to HTML (though rst?) has the same effect. Similarly it seems that the Python markdown package doesn't like lists starting without an empty line before. So it seems that there is a discrepancy between "Jupyter Markdown" and expectations of other tools.

kraktus:
Note that this is an issue with pandoc, created an issue: https://github.com/jgm/pandoc/issues/9079

douglas-raillard-arm:
I also experienced that on HTML output. It seems that pandoc has a setting to recognize bullet lists without a preceding blank line, maybe nbconvert should make use of it ?
https://github.com/jgm/pandoc/issues/9079#issuecomment-1719850251

krassowski:
The fix is available in `nbconvert` v7.16.6. Thank you @douglas-raillard-arm!

PR #2177, opened 2024-09-11
Fix bullet list parsing in markdown

douglas-raillard-arm:
Jupyter markdown allows bullet list to start right after the previous element, without any blank line in between. The default settings of pandoc is to not recognize that as a bullet list.

Therefore, enable the pandoc markdown extension that makes the blank line optional.

Fixes https://github.com/jupyter/nbconvert/issues/917


t-makaro:
This looks great to me. This is the appropriate way to solve this (upstream to pandoc) and the tests pass. I would merge this, but I appear not to have maintainer access anymore (I assume due to inactivity). @takluyver You seem to be the one most active recently. Can we get this merged?

t-makaro (approved):

krassowski (approved):
This makes sense to me, aligning it with valid Markdown as used by Notebook and JupyterLab

PR #2183, opened 2024-09-28
Add overflow-x on pre tags

### LINKED ISSUE #2182: Enable overflow on pre tags
When exported to HTML the pre tags can't be scrolled left or right. Adding a `overflow-x: auto` will fix it.

akshaybabloo:
Fixes #2182


KAD4-ctrl:
https://github.com/jupyter/nbconvert/issues/2182

Issue #1586, opened 2021-06-08
nbconvert -Number Headings when exporting from jupyter ipython to html


oracle1987:
Im trying to export a ipynb notebook to html while maintaining the headers numbered with # in markdown cells, however, when i export using the following terminal command, the numbering of the sections is ignored.

Note: Headings auto-generated using toc nbextension.

terminal command:
jupyter nbconvert --TemplateExporter.exclude_input=True --no-prompt --to html sample.ipynb

To generate the numbered headings content of notebook is as follows(please ignore brackets below):

(#) SectionA

(##) SubsectionA1

(##) SubsectionA2

Please advise how can the numbering be maintained as per the defined hashtags when exporting to html.

gitcuz:
Any progress? I'm also interested in this feature.

oracle1987:
Try adding '_toc'as follows:
jupyter nbconvert --TemplateExporter.exclude_input=True --no-prompt --to html_toc filename.ipynb

DheemanthBhat:
+1

JingyiShi:
Following. Still have the same issue in 2023. The command above did not work for me.

guigoruiz1:
Hello,

I have the same problem

letian-w:
Same problem.

PR #2187, opened 2024-10-19
feat: add NumberedHeadingsPreprocessor

sapristi:
Add pre-processor that adds numbering to markdown headers


krassowski:
Thank you @sapristi for sending the PR. Was this requested in an issue before? Just trying to understand if there is sufficient community need for this to live in `nbconvert` given the maintenance costs for adding a new pre-processor.

sapristi:
Hello, I just checked, it seems that some people were interested in this as well, e.g. https://github.com/jupyter/nbconvert/issues/1586

IMO this is a very essential  feature for working with a document - it's also a feature provided in the Jupyter web UI, so I think I'm not the only one.

nbconvert/preprocessors/numbered_headings.py
```diff
@@ -0,0 +1,51 @@
...
+            self.current_numbering = self.current_numbering + [0] * (
+                level - len(self.current_numbering)
+            )
+        elif level < len(self.current_numbering):
+            self.current_numbering = self.current_numbering[:level]
+        self.current_numbering[level - 1] += 1
+
+    def _transform_markdown_line(self, line, resources):
+        """Rewrites one markdown line, if needed"""
+        if m := re.match(r"^(?P<level>#+) (?P<heading>.*)", line):
```
krassowski, line 34: What if I we have a code in markdown, e.g.:

~~~markdown
This is how you create a heading in Markdown
```markdown
# heading
```
~~~

It feels like a more robust solution would be to use mistune API for parsing, see https://mistune.lepture.com/en/latest/api.html (`markdown_mistune` is already in use here, but it should fail nicely).

It is also possible to instruct pandoc to run with `--number-sections` (see https://www.youtube.com/watch?v=gDV63UMJR_U and https://pandoc.org/chunkedhtml-demo/8.3-headings.html and https://pandoc.org/MANUAL.html) but pandoc is neither the only nor even default markdown parser.

sapristi: Hello,
sorry for the late reply, I'm a bit busy at the moment.

Indeed, I'll make that more robust with mistune parsing / treatment / dumping back to markdown.

About your pandoc remark, I'm not sure to follow you.

sapristi: Implemented changes to use mistune for parsing

- only compatible with mistune v3, do you think it's worth it keeping compatibility with v2 as well ?
- I'm using a logger, not sure if that's the preferred way to report warning / errors to users ?

nbconvert/preprocessors/numbered_headings.py
```diff
@@ -0,0 +1,63 @@
...
+from nbconvert.preprocessors.base import Preprocessor
+
+logger = get_logger()
+
+try:  # for Mistune >= 3.0
+    import mistune
+    from mistune.core import BlockState
+    from mistune.renderers.markdown import MarkdownRenderer
+except ImportError:  # for Mistune >= 2.0
+    logger.error("NumberedHeadingsPreprocessor requires mistune >= 3")
```
krassowski, line 16: I would set a variable here (and in the try block too), like:

```suggestion
    mistune_3_available = False
```

and only emit the error (throw?) if someone tries to initialize `NumberedHeadingsPreprocessor` (this is in `__init__`). This is because logging on import of preprocessors would affect these who do not wish to use `NumberedHeadingsPreprocessor`

nbconvert/preprocessors/numbered_headings.py
krassowski: > only compatible with mistune v3, do you think it's worth it keeping compatibility with v2 as well ?

I think that's fine

> I'm using a logger, not sure if that's the preferred way to report warning / errors to users ?

I left a comment on the code

nbconvert/preprocessors/numbered_headings.py
sapristi: Went for throwing an exception ( for some reason log output does not seem displayed)

nbconvert/exporters/exporter.py
```diff
@@ -97,6 +97,7 @@ class Exporter(LoggingConfigurable):
             "nbconvert.preprocessors.ExtractOutputPreprocessor",
             "nbconvert.preprocessors.ExtractAttachmentsPreprocessor",
             "nbconvert.preprocessors.ClearMetadataPreprocessor",
+            "nbconvert.preprocessors.NumberedHeadingsPreprocessor",
```
sapristi, line 100: Just added the processor here, otherwise it was not running `python -m nbconvert --to=html --NumberedHeadingsPreprocessor.enabled=true notebook.ipynb` (even though the option to enable it appeared in `python -m nbconvert --help-all `

paolovic:
nice work, thank you @sapristi !

PR #2188, opened 2024-11-01
Fixed issue with INLINE_MATH_TEX regex to accept spaces within inline

lokashrinav:
Updated the INLINE_MATH_TEX regex pattern to allow spaces within the inline math expression. Ensured proper escaping of dollar signs and backslashes. This change addresses the issue with spaces that users encounter.

This is a further fix for this issue from Jupyter Lab/Notebook since the original style may confuse the users with issues, such as in the following: https://github.com/jupyterlab/jupyterlab/issues/11062

Transformed oriignal INLINE_MATH_TEX to INLINE_MATH_TEX = _dotall(r"(?<![$\\])$(?!\s)(?P<math_inline_tex>[^$\\\s]+?)(?<!\s)(?<![$\\])$")

Feel free to decide what you think should be done, and if further testing should be done.

### DIFF
--- a/nbconvert/filters/markdown_mistune.py
+++ b/nbconvert/filters/markdown_mistune.py
@@ -119,7 +119,7 @@ if MISTUNE_V3:  # Parsers for Mistune >= 3.0.0
         # Display math mode, using newer LaTeX delimiter: \[ \pi \]
         BLOCK_MATH_LATEX = _dotall(r"(?<!\\)\\\\\[(?P<math_block_latex>.*?)(?<!\\)\\\\\]")
         # Inline math mode, using older TeX delimiter: $ \pi $  (cannot be empty!)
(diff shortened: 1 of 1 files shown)


krassowski:
I think this makes sense but before merging we need a unit test here

lokashrinav:
@krassowski Hey krassowski, could i ask for some help on how to add unit tests here/understand what next to do?

nicolas-joffre:
I just tested it on my device and unless I did something wrong, it doesn't work: `$ a$` in a notebook Markdown cell still gets converted to `\$ a\$` in LaTeX.

Edit: I think I misunderstood what it was aiming to do. I thought it was supposed to make math with spaces render as math in LaTeX, but by taking a closer look I see it's the opposite.

PR #2190, opened 2024-11-12
LaTeX and PDF export: Support SVG image files via the `svg` LaTeX package

aravindh-krishnamoorthy:
> [!NOTE]
> This is a draft pull request which is a WIP. Community contributions and suggestions are welcome.

# Summary
Jupyter notebooks allow inclusion of SVG images via `Caption`. However, for `LaTeX` and `PDF via LaTeX` exports, unlike other image formats, SVG images does not work. This is mainly because `\includegraphics` doesn't support SVGs yet.

In this PR, a solution is provided via `\usepackage{svg}`. The fix works by hooking onto `\includegraphics` to detect SVG images from their file extension and replace the `\includegraphics` call by `\includesvg`.

# Cautionary areas
The following are the reasons why we have to be cautious before merging this PR.

- The `svg` LaTeX package utilizes Inkscape to convert SVGs to PDFs. Hence, the experience for LaTeX export may not be seamless for novice LaTeX users.
- PDF export requires `-shell-escape` options to be passed to `XeTeX`.

# Pending issues
- While `LaTeX` export works well for SVG files if the requirements of the `svg` LaTeX package are met, **the `PDF via LaTeX` export is still broken.** Community inputs are welcome here.

    - Initial look at the issue reveals that the compilation of `notebook.tex` in a temporary directory is a likely cause of the issue. It seems that the `svg` package cannot handle paths from `TEXINPUTS` well yet. Or, maybe I've not setup the package or `pdf.py` exporter file properly.

### DIFF
--- a/nbconvert/exporters/pdf.py
+++ b/nbconvert/exporters/pdf.py
@@ -60,7 +60,7 @@ class PDFExporter(LatexExporter):
     latex_count = Integer(3, help="How many times latex will be called.").tag(config=True)
 
     latex_command = List(
-        ["xelatex", "{filename}", "-quiet"], help="Shell command used to compile latex."
+        ["xelatex", "-shell-escape", "{filename}", "-quiet"], help="Shell command used to compile latex."
     ).tag(config=True)
 
     bib_command = List(["bibtex", "{filename}"], help="Shell command used to run bibtex.").tag(
--- a/share/templates/latex/base.tex.j2
+++ b/share/templates/latex/base.tex.j2
@@ -20,8 +20,6 @@ override this.-=))
     \usepackage{graphicx}
     % Keep aspect ratio if custom image width or height is specified
     \setkeys{Gin}{keepaspectratio}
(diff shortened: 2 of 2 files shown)


t-makaro:
I looked into using the svg package for nbconvert latex once. It requires running latex with a shell escape which is a no-go. Given that we already have an svg2pdf preprocessor, I think that making that work with markdown cells would be better.

aravindh-krishnamoorthy:
> I looked into using the svg package for nbconvert latex once. It requires running latex with a shell escape which is a no-go. Given that we already have an svg2pdf preprocessor, I think that making that work with markdown cells would be better.

True! Lemme try to find a fix which uses the `svg2pdf` preprocessor... it currently preprocesses based on MIME headers of a cell and will also have to be activated for `xxx` instead of `pandoc`. I'll get back (hopefully) soon with a fix.

t-makaro:
I'm not super familiar with svg2pdf, but it would likely have to handle markdown cells very differently, since you'd need to parse the markdown to find the image first. Hence why this has been an open issue so long.

PR #2178, opened 2024-09-16
Add a toc for html exporter

HaudinFlorence:
…markdown contents.

PR #2191, opened 2024-11-14
Move the `check_release.yml` to its own workflow

jtpio:
- [x] Move the `check_release` step to a separate workflow, so it's easier to see if it passes
- [x] And add an extra step to upload the artifacts built by the releaser, so it's easier to test PRs locally by installing the built wheel from a branch.

### DIFF
--- a/.github/workflows/check-release.yml
+++ b/.github/workflows/check-release.yml
@@ -1,0 +1,33 @@
+name: Check Release
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  release:
+    types: [published]
+
+jobs:
+
+  check_release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Base Setup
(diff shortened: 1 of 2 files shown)


jtpio:
@martinRenou I was wondering if this was maybe ready to be merged?

Having the built artifacts could help test changes from other PRs more easily, for example https://github.com/jupyter/nbconvert/pull/2178.

martinRenou (approved):
Thanks!

krassowski:
`test_check` does not pass because it is waiting for `check_release`:

https://github.com/jupyter/nbconvert/blob/6e5fdb31b77314a85f6ac199a92d4d1dc0f67f1d/.github/workflows/tests.yml#L161-L170

It would need to be removed from it and added as a separate protection rule

jtpio:
> It would need to be removed from it and added as a separate protection rule

Done in https://github.com/jupyter/nbconvert/pull/2191/commits/5541bdbd91bdd5af89b65764dc78f864ef3e48bb.

However I don't have access to the repo settings to enable the protection rule in case we want to add one (not sure if this may have an impact on the releaser though).

PR #2196, opened 2024-12-03
 Update base.tex.j2 for macro compatibility with newer versions of Pandoc

### LINKED ISSUE #2173: Incompatibility issue with latest version of pandoc
When converting a notebook that has an image to a PDF via xelatex, you will now receive an error with version 3.2.1 and above on pandoc. Pandoc has added a new macro that is not defined in the jinja templates. See pandoc issue [10125](https://github.com/jgm/pandoc/issues/10125) for more details on reproducing the issue.

PetalAdrift:
Added the macro \pandocbounded that is used by newer versions of pandoc.

### DIFF
--- a/share/templates/latex/base.tex.j2
+++ b/share/templates/latex/base.tex.j2
@@ -164,7 +164,26 @@ override this.-=))
     \newcommand{\AttributeTok}[1]{\textcolor[rgb]{0.49,0.56,0.16}{{#1}}}
     \newcommand{\InformationTok}[1]{\textcolor[rgb]{0.38,0.63,0.69}{\textbf{\textit{{#1}}}}}
     \newcommand{\WarningTok}[1]{\textcolor[rgb]{0.38,0.63,0.69}{\textbf{\textit{{#1}}}}}
-
+    \makeatletter
+    \newsavebox\pandoc@box
+    \newcommand*\pandocbounded[1]{%
+      \sbox\pandoc@box{#1}%
+      % scaling factors for width and height
+      \Gscale@div\@tempa\textheight{\dimexpr\ht\pandoc@box+\dp\pandoc@box\relax}%
+      \Gscale@div\@tempb\linewidth{\wd\pandoc@box}%
+      % select the smaller of both
+      \ifdim\@tempb\p@<\@tempa\p@
+        \let\@tempa\@tempb
+      \fi
+      % scaling accordingly (\@tempa < 1)
+      \ifdim\@tempa\p@<\p@
+        \scalebox{\@tempa}{\usebox\pandoc@box}%
+      % scaling not needed, use as it is
+      \else
+        \usebox{\pandoc@box}%
+      \fi
+    }
+    \makeatother
 
     % Define a nice break command that doesn't care if a line doesn't already
     % exist.


PetalAdrift:
Sorry about closing the previous PR on accident. Opened a new one.

krassowski:
@PetalAdrift thank you! How can we test this PR? Is there a test that you could add to ensure this is tested on CI?

PetalAdrift:
@krassowski  Sorry, I am not too familiar with the terminology (or Git, actually). Is there a link to relevant documentation on tests used for nbconvert?

krassowski:
It may not be much but:
- https://github.com/jupyter/nbconvert/blob/main/CONTRIBUTING.md#testing
- https://github.com/jupyter/nbconvert/tree/main/tests

gabemorris12:
When is this going to be implemented? It's been months, and nbconvert for latex still only works with pandoc 3.2 or older.

PetalAdrift:
For now you can just make the change to your files locally.

gabemorris12:
> For now you can just make the change to your files locally.

That is fine for me, but I have about a hundred students each semester that have to setup jupyter to successfully convert a notebook to a pdf via latex. I'd hate to redo my setup process only for the solution to finally have been implemented later.

PetalAdrift:
I guess one solution is to just use pandoc. It doesn't look as nice (subjective opinion), but it gets the job done. (edited a typo)

ChristofKaufmann:
@PetalAdrift Maybe I can help a bit to write a simple test. I have not tried it, but you would have to add in `tests/exporters/test_latex.py` a function like:

``` python
    @onlyif_cmds_exist("pandoc")
    def test_template_contains_pandocbounded(self):
        """
        Does the template contain the pandocbounded command for compatibility with pandoc >= 3.2.1?
        """
        (output, resources) = LatexExporter().from_filename(self._get_notebook())
        assert r"\newcommand*\pandocbounded" in output
```

Either follow the [instructions](https://github.com/jupyter/nbconvert/blob/main/CONTRIBUTING.md#testing) to test it locally before committing or commit and let CI test it. In the CI test output, you should see about [here](https://github.com/jupyter/nbconvert/actions/runs/12581211157/job/35065952645?pr=2196#step:5:77) a new test named `test_template_contains_pandocbounded`.

@krassowski Would that simple test be enough for you? It is actually just testing that the code, which this PR adds, is there. So it does not make much sense. Maybe it would be better to add e.g. `\pandocbounded{check}` in [rawtest.ipynb](https://github.com/jupyter/nbconvert/blob/main/tests/exporters/files/rawtest.ipynb?short_path=f343d2e#L54) and try to compile it with a new test in [test_pdf.py ](https://github.com/jupyter/nbconvert/blob/main/tests/exporters/test_pdf.py). What is your opinion?

krassowski:
> Maybe it would be better to add e.g. \pandocbounded{check} in [rawtest.ipynb](https://github.com/jupyter/nbconvert/blob/main/tests/exporters/files/rawtest.ipynb?short_path=f343d2e#L54) and try to compile it with a new test in [test_pdf.py](https://github.com/jupyter/nbconvert/blob/main/tests/exporters/test_pdf.py)

Yes, that would be better. Basically if compilation errors out without this PR and test checks that compilation passes after, it is a fine test in my book, as long as it can be run on CI.

krassowski (approved):

PetalAdrift:
Thanks everyone! I have been busy with math since Jan. and have not had the time to implement the test. But we have it sorted out, yay!

PR #2199, opened 2024-12-30
Add support for mistune 3.1.0

fcollonval:
Fixes #2198

PR #2200, opened 2024-12-30
bugfix: AXT -> ATX

laurencap:
Mistune does not have an axt_heading.

https://github.com/lepture/mistune/blob/177a0ceefd23a8ebc7f7efb5c89c0be8023ae3bf/src/mistune/block_parser.py#L61

In some instances, this can cause the following error:
`AttributeError: 'MathBlockParser' object has no attribute 'parse_axt_heading'. Did you mean: 'parse_atx_heading'?`


laurencap:
Some linting errors which appear to be outside the scope of this fix

laurencap:
Closing because this will be fixed by https://github.com/jupyter/nbconvert/pull/2199

Issue #2156, opened 2024-06-14
Notebook to script does not respect leading whitespace


drblarg:
I am using nbconvert as a package vs command line to convert a notebook to a script using ScriptExporter class in python:

```
contents, meta = nbconvert.exporters.export(
    nbconvert.exporters.ScriptExporter,
    notebook
)
```

When the contents of a notebook cell are all indented by the same amount, this intention is removed in the output. This whitespace should not be removed. Is there already a way to enforce this that I am missing?

A use case for keeping the white space is to enclose multiple output cells within some scope, such as an if statement. For example: 

```
In[1]
if is_true:

In[2]
    do_stuff1()

In[3]
   do_stuff2()
```

This allows greater flexibility in using the notebook as a notebook (where you choose the cells to run), and the generated script as a script (e.g. AWS pipelines) without putting scoping statements in every cell.

aryansk:
I’m taking a look at this. I’ll keep the fix focused on preserving common leading indentation through the IPython-to-Python transform, with regression coverage for indented code cells.

PR #2203, opened 2025-01-06
Prevent leading whitespace in markdown code blocks from being stripped

peytondmurray:
This PR prevents leading whitespace from being stripped from markdown code blocks before they are parsed for highlighting. For certain languages, leading whitespace is syntactically significant. Fixes https://github.com/jupyter/nbviewer/issues/1021.

I tried to go as far back as I could through the git history to figure out why this was added, but it appears in the earliest tag, `4.0.0`. If there _is_ a reason to strip leading whitespaces, it's not explicitly tested for in the test suite. If you have any context around this choice, I'd be interested in hearing about it.

I also added a test to check that whitespaces aren't stripped from code blocks. I've also added a `RUF001` rule to ignore uses of ambiguous unicode characters in `tests/exporters/test_html.py`, because of the "ɩ" which is apparently a valid APL function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@
 
 ### Maintenance and upkeep improvements
 
-- enhancement dep-chain: directly depend on bleach[css], instead of pulling in tinycss2. [#2166](https://github.com/jupyter/nbconvert/pull/2166) ([@xiacunshun](https://github.com/xiacunshun))
+- enhancement dep-chain: directly depend on bleach\[css\], instead of pulling in tinycss2. [#2166](https://github.com/jupyter/nbconvert/pull/2166) ([@xiacunshun](https://github.com/xiacunshun))
 - chore: update pre-commit hooks [#2146](https://github.com/jupyter/nbconvert/pull/2146) ([@pre-commit-ci](https://github.com/pre-commit-ci))
 
 ### Contributors to this release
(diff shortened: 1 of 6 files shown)


krassowski:
Does it also fix https://github.com/jupyter/nbconvert/issues/2156?

krassowski (commented):
I think this change makes sense. Should we add an option to allow to opt-in into the old behaviour so that users who may have relied on it are not blocked from upgrading?

peytondmurray:
> Does it also fix https://github.com/jupyter/nbconvert/issues/2156?

I don't think so - this PR only addresses code blocks inside markdown. The issue above :point_up: is about being able to split statements across multiple code cells - for example, putting a single (empty) `if`, `with`, or `while` etc in one cell, and having indented code in subsequent cells further down be part of the scope of the first cell:

image

Maybe this is a discussion for that issue, but I'm not sure that makes sense to do here - at the moment, each cell needs to be a syntactically complete snippet of python code, which isn't true in the use case provided.

About your other suggestion: I like it, I'll add the option to pass arbitrary lexer options to the pygments lexers now!

peytondmurray:
I've added a test for the lexer options. The user can revert to the old behavior by setting

```python
HTMLExporter(lexer_options={"stripall": True})
```

The option should be documented by traitlets, so I haven't included anything extra about reverting back to the old behavior. But if some additional documentation would be helpful please let me know and I can add an explicit section about this to the docs.

divyansshhh:
I see the PR is being worked on, is there an ETA on when it would be released?

krassowski (approved):
Thank you @peytondmurray!

krassowski:
> I see the PR is being worked on, is there an ETA on when it would be released?

I released it just now:
- https://github.com/jupyter/nbconvert/releases/tag/v7.16.6
- https://pypi.org/project/nbconvert/7.16.6/

PR #2206, opened 2025-01-23
remove any twitter mention

Carreau:


Carreau:
Thanks, I though those would be controversial; but happy to see you agree.

PR #2220, opened 2025-06-04
Save temp file next to notebook for WebPDF export

shreve:
By putting the HTML file in the same location as the notebook, any
relative file paths correctly resolve. This change saves us from needing
to convert relative paths to absolute.

This slightly changes behavior by creating the temp file in the user's
working directory. We already make sure we delete this temp file, and by
adding the `".webpdf-"` prefix to make it hidden, less-technical users
should never see this file at all.


shreve:
Rebased from main

PR #2223, opened 2025-06-09
Add support for Chemical Reactions and more symbols

ksinkar:
- The additional symbols are very helpful in notebooks for Astrophyics

Supported symbols by mathjax: https://docs.mathjax.org/en/latest/input/tex/macros/index.html#environments

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -39,6 +39,7 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install texlive-plain-generic inkscape texlive-xetex latexmk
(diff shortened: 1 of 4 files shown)


shreve:
Would you be able to add or update a test to demonstrate the usage of these packages? I think a place that makes sense would be [`tests/files/notebook1.ipynb`](https://github.com/jupyter/nbconvert/blob/main/tests/files/notebook1.ipynb) because it's a demonstration of library that supports outputting LaTeX equations.

ksinkar:
@krassowski @shreve :wave: I have updated the PR as requested. Is there anything else that I need to do to get this PR merged?

PR #2228, opened 2025-07-16
Fix get_export_names and get_exporter default args

shreve:
These functions were annotated with `noqa: B008` to silence a warning from flake8-bugbear. It turns out bugbear was onto something.

`get_export_names` and `get_exporter` were defined with a default argument, `config=get_config()`. The problem is that this function call is executed once when the function is defined, not each time the function is called. I encountered this problem in a JupyterLab server where disabling exporters in the `jupyter_server_config.py` file wouldn't remove them from the `/api/nbgrader` endpoint because the default values had been set to `{}`.

This commit fixes the problem by using `None` as the default value and calling `get_config()` at function call time.

### DIFF
--- a/nbconvert/exporters/base.py
+++ b/nbconvert/exporters/base.py
@@ -91,11 +91,14 @@ def export(exporter, nb, **kw):
     return output, resources
 
 
-def get_exporter(name, config=get_config()):  # noqa: B008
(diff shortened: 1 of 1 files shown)


krassowski (approved):
Thank you!

PR #2229, opened 2025-07-24
Replace `@flaky.flaky` decorate with pytest marker

mgorny:
Use the `@pytest.mark.flaky` marker in place of the `@flaky.flaky` decorator, to modernize the code and improve compatibility with other plugins providing the feature such as `pytest-rerunfailures`.

### DIFF
--- a/tests/exporters/test_qtpng.py
+++ b/tests/exporters/test_qtpng.py
@@ -6,7 +6,6 @@
 import os
 
 import pytest
-from flaky import flaky
 
 from nbconvert.exporters.qt_screenshot import QT_INSTALLED
 from nbconvert.exporters.qtpng import QtPNGExporter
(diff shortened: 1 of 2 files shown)


mgorny:
> Sounds great; Thanks for this, I'm going to merge, it would be good if you can link to some docs listing the advantage of this and pytest-rerunfailures vs using flaky directly.

I'm afraid I don't know of anything specific. The rough idea is that if you `import flaky`, then you must use `flaky`, but if you use the marker you can use any plugin that implements that marker — so probably the most relevant docs are the docs of different plugins implementing the marker.

As for why `pytest-rerunfailures`, it's pretty arbitrary. It's just that we've happened to hit packages that requires `pytest-rerunfailures` specifically and failed when `flaky` was enabled (sorry, don't recall which exactly), but worked the other way around.

PR #805, opened 2018-05-03
Create ClearMetadata Preprocessor

henriqueribeiro:
Very simple preprocessor that cleans all the metadata on the ipynb code cells.
This is good for apply a git filter in order to be able to git diff more easily.

PR #2239, opened 2025-10-23
Make ClearMetadataPreprocessor work for cells of any type, and some related improvements

mskiptr:
Hello. I've been pleasantly surprised to discover that `nbconvert` can now clear metadata fields. That makes for much nicer diffs when you're trying to revision-control your notebooks.

But then I soon noticed that not _all_ of the metadata is actually being removed. More specifically, markdown cells seemed to retain their `id`s at least. So today, after looking through the source code and doing some Git archaeology, (I think) I've finally figured out why.

When the `ClearMetadataPreprocessor` was added in #805, it was based on the `ClearOutputPreprocessor` which too performs some metadata transformations. Removing cell output only makes sense for `code` cells, and so [there is a a check for that](https://github.com/jupyter/nbconvert/blob/5eec897ca319d2593029895ce75cfed22b1a2153/nbconvert/preprocessors/clearoutput.py#L22). But [the same check](https://github.com/jupyter/nbconvert/blob/5eec897ca319d2593029895ce75cfed22b1a2153/nbconvert/preprocessors/clearmetadata.py#L78) is now done for general metadata clearing, which leads to this unexpected behavior.

While trying to fix it, I also ended up making a patch that clarifies some of the documentation present in that file. And then I recalled that at first it wasn't really clear to me how I should go about enabling the `ClearMetadataPreprocessor`, so I also wrote a patch to add a `--clear-metadata` command line flag that works just like `--clear-output`

I marked this PR as a draft, because:

- ~~I still need to test this truly works correctly~~ _I have now built and tested it (in a `guix shell jupyter --with-patch=python-nbconvert=metadata.diff`)_
- I have not extended the unit tests to cover this (help appreciated)
- the way `ClearOutputPreprocessor.remove_metadata_fields` and `ClearMetadataPreprocessor.preserve_cell_metadata_mask` interact is rather confusing, so I would like to fix that
- maybe there's something I'm missing and we shouldn't actually clear metadata for certain cells

### DIFF
--- a/nbconvert/nbconvertapp.py
+++ b/nbconvert/nbconvertapp.py
@@ -118,6 +118,18 @@ nbconvert_flags.update(
             """Clear output of current file and save in place,
         overwriting the existing notebook. """,
         ),
+        "clear-metadata": (
+            {
+                "NbConvertApp": {
+                    "use_output_suffix": False,
+                    "export_format": "notebook",
+                },
+                "FilesWriter": {"build_directory": ""},
+                "ClearMetadataPreprocessor": {"enabled": True},
+            },
+            """Clear metadata of current file and save in place,
+        overwriting the existing notebook. """,
+        ),
         "coalesce-streams": (
             {
                 "NbConvertApp": {"use_output_suffix": False, "export_format": "notebook"},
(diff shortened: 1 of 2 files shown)


mskiptr:
I guess this is as ready-for-review as it will ever be.

edit: I'm happy to fill in any gaps (like missing tests) if there's a will to get this merged. Please comment if so.

PR #2216, opened 2025-04-21
- bump reveal.js

franzhaas:
reveal.js is currently at version 4.0.2, this is quite outdated.

This is an attempt to move reveal.js to the current 5.2.x releases.

This is an initial effort, but i expect more steps to be necessary. Lets see what CI finds and carry on from there.

Issue #1851, opened 2022-08-27
Is there a simple way to customize Reveal.initialize arguments?


paugier:
I would need to change values of `Reveal.initialize` arguments (see https://revealjs.com/presentation-size/). Is there a simple way to do that? I would like to avoid writing a custom exporter for that.

bouzidanas:
Inside reveal.js there is a chunk of code that exposes configuration options:
```
		// Compose our config object in order of increasing precedence:
		// 1. Default reveal.js options
		// 2. Options provided via Reveal.configure() prior to
		//    initialization
		// 3. Options passed to the Reveal constructor
		// 4. Options passed to Reveal.initialize
		// 5. Query params

		config = { ...defaultConfig, ...config, ...options, ...initOptions, ...Util.getQueryHash() };
```
So the [query string](https://en.wikipedia.org/wiki/Query_string) ( remove `#` at end of url and add ?config_option=value ) is used to set configuration options first and whatever is not set, gets set by initOptions ( `Reveal.initialize()` ) and whatever is not set by initOptions is set by the arguments of the Reveal constructor and so on.

So to answer your question, the only way that I think might be simpler is to put your settings in the url.

For example: ..yourslides.html?embedded=false&width=600&height=700&margin=0.04&minScale=0.4&maxScale=1.0#/

canyon289:
Thank you for this! This helped me tremendously

franzhaas:
Is there also a way to have this set in the standalone HTML file?

PR #849, opened 2018-07-29
Add additional configuration options to SideExporter

njallam:
Add SlidesExporter configuration options for controls, progress, slide number and history

Issue #612, opened 2017-06-27
Better support for reveal.js functionality


mattilyra:
I really like creating slide presentation using Jupyter Notebooks, but the workflow is currently fairly cumbersome. I often find myself wanting to control HTML `<section>` tag parameters like

- slide transitions (per individual slide)
-  individual slide backgrounds
- background transitions
- different fragment types (http://slides.com/hakim/reveal-js/embed?transition=none#/transitions)
    - grow
    - shrink
    - fade in/out
    - highlight colour
- footer 
...

It would be great if controls for these could be added to the slide toolbar so that one does not need to edit the HTML file manually each time after running `nbconvert --to slides`.

buffetboy2001:
+1 Definitely agree. I assumed I could use the slide toolbar in some way to edit the JSON and control slide transitions, but I don't think it's possible. Please open up some of these options, if possible.

bouzidanas:
#1861 

Give that a try. You can set the `data-` attributes for slides, subslides (vertical slides), and fragments by means of the cells metadata. 

Example:  Click `View > Cell Toolbar > Edit Metadata` . Then go to a cell (in notebook) thats going in a slide/subslide/fragment and click Edit Metadata in the cell's toolbar on the right. You will see a popup input with something like this:
```
{
  "id": "p9eicnYjOoBv",
  "slideshow": {
    "slide_type": "slide"
  }
}
```
Change that to something like this:
```
{
  "id": "p9eicnYjOoBv",
  "slideshow": {
    "slide_type": "slide",
    "data": {
      "background_video": "your-video-URL-here.mp4",
      "background_video_loop": "",
      "background_video_muted": ""
    }
  }
}
```
This should put `data-background-video="your-video-URL-here.mp4" data-background-video-loop="" data-background-video-muted=""` inside the `<section>` tag that corresponds to this slide. 

A few things to note.

1.  That `background_video` sets `data-background-video` so if you want to add say ` data-transition="zoom"`, you would put "transition": "zoom"` inside the Metadata instead.
2.  Setting the Metadata attribute to an empty string `""` is equivalent to adding a `data-` without a value.

Give it a shot. There is only so much testing I have time for.

MP-MaximilianLattka:
Sorry for bringing this up again, but is it possible as of now to provide reveal.js configuration parameters on the **presentation-level** like this?
https://revealjs.com/config/

bouzidanas:
[...]

PR #2257, opened 2025-12-14
Add support for full configuration of reveal slides in notebook metadata

bouzidanas:
(**Suggested labels:** enhancement, feature)

This pull request introduces a new feature that allows `reveal.js` presentation settings to be configured directly within a notebook's metadata.

Currently, `reveal.js` options for slide exports are primarily set via command-line arguments or configuration files. This separates the presentation's configuration from its content, making it less portable. For example, if a user wants to share a notebook that is designed to be presented with a specific theme or transition, they must also share the command-line flags needed to reproduce it or edit the html file themselves which requires familiarity with `reveal.js`. Furthermore, only a limited number of reveal.js options are accessible via the command line and in the configuration files.

This enhancement allows users to embed `reveal.js` settings within the `.ipynb` file itself, making presentations self-contained and easier to share and reproduce consistently.

## Implementation Details

The feature is implemented by overriding the `from_notebook_node` method in the `SlidesExporter`. The new implementation preserves the existing conversion logic by calling `super().from_notebook_node()`. It also inspects the notebook's metadata for a `reveal` key and merges the settings with the configuration already loaded from command-line options and defaults.

A clear order of precedence (override order) is maintained to ensure intuitive and predictable behavior:
1. Command-line arguments (highest)
2. Notebook metadata 
3. Config file
4. Hard-coded defaults (lowest)

This ensures that users can set project-wide defaults in a notebook but still easily override them from the command line for a specific export.

## How to Use

To use this feature, a user can edit the notebook's metadata (e.g., in Jupyter, via `Edit > Edit Notebook Metadata`) and add a `reveal` object.

**Example Notebook Metadata:**
```diff
{
  "kernelspec": {
    "display_name": "Python 3",
    "language": "python",
    "name": "python3"
  },
  "language_info": {
    "name": "python",
    "version": "3.9.7"
  },
+ "reveal": {
+   "theme": "night",
+   "transition": "zoom",
+   "scroll": true
+}
}
```

When a notebook with this metadata is exported using `jupyter nbconvert --to slides`, it will automatically use the "night" theme and "zoom" transition, unless a different value is provided on the command line.

[List of config options](https://revealjs.com/config/)

### Related Issues

1. #612 
2. #1851
3. #849
4. #2216 (new config options)

### DIFF
--- a/nbconvert/exporters/slides.py
+++ b/nbconvert/exporters/slides.py
@@ -6,7 +6,7 @@
 from copy import deepcopy
 from warnings import warn
 
-from traitlets import Bool, Unicode, default
+from traitlets import Bool, Dict, Unicode, default
 
 from nbconvert.preprocessors.base import Preprocessor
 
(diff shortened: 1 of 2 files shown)


petrelharp:
This would really help make jupyter a more functional platform for making presentations. Thanks for working on this! Where's the right way to document this - [here](https://ipython.readthedocs.io/en/3.x/notebook/nbformat.html)?

PR #2261, opened 2026-01-07
Fix CVE-2025-53000: Secure Inkscape Windows path (registry first + block CWD)

h3pdesign:
## Changes

| Problem | Fix | Impact |
|---------|-----|--------|
| `which("inkscape")` searches CWD first | Registry lookup **first** + `NoDefaultCurrentDirectoryInExePath=1` | Prevents arbitrary code execution via `inkscape.bat` [GHSA-xm59-rqc7-hhvf] | | Registry `QueryValueEx` returns `"C:\path\to\inkscape.exe,0"` | `.split(",")[0]` + `os.path.isfile()` verification | Valid executable path only | | Fallback `return "inkscape"` triggers CWD | **Raise `FileNotFoundError`** | No unsafe fallback | | No `.com` priority | `shutil.which("inkscape.com") or which("inkscape.exe")` | Native Windows console execution |

## Verification
- **PoC blocked**: Malicious `inkscape.bat` in CWD ignored
- **Legit Inkscape**: Registry/PATH paths work unchanged
- **macOS/Linux**: Identical behavior

Closes #CVE-2025-53000
Refs: https://github.com/advisories/GHSA-xm59-rqc7-hhvf

PR #2264, opened 2026-01-14
Fix QtPNGExporter returning empty bytes on macOS

h3pdesign:
Fix QtPNGExporter returning empty bytes on macOS. Fix for failed Test run for Fix CVE-2025-53000 (Secure Inkscape Windows path (registry first + block CWD) #2261): PR is failing because QtPNGExporter().from_filename(...) returns b'' on macos-latest (Python 3.10), and the test asserts the returned PNG bytes are non-empty

### DIFF
--- a/nbconvert/exporters/qt_exporter.py
+++ b/nbconvert/exporters/qt_exporter.py
@@ -75,7 +75,7 @@ class QtExporter(HTMLExporter):
         except OSError:
             pass
 
-        return data 
+        return data
 
     def from_notebook_node(self, nb, resources=None, **kw):
         """Convert from notebook node."""


Carreau (approved):
Many thanks for this. I'm approving and will come back to it soon and merge if no comment. I'm a bit overwhelmed at the moment.

h3pdesign:
Hi @krassowski @mberlanda — quick context on #2264:

This PR fixes the macOS CI regression where `QtPNGExporter().from_filename(...)` returns empty bytes on `macos-latest` (Py 3.10), causing the “PNG bytes non-empty” assertion to fail.

There are no behavioral changes beyond restoring the expected test invariant; the fix is narrowly scoped and unblocks #2261, which is otherwise merge-ready.

Once this lands, CI for the Inkscape CVE fix should go fully green again. Thanks so much

